PR #9, opened 2019-03-12
Segment specific scripts added

zsombal:
[WIP] In docs, segments section added with same sample data and sample scripts. Plus new section added with tutorials with different programs (python for now). The idea there is to have sample scripts that show at least how to open .h5 files using matlab, igor, origin, etc


dean0x7d:
Thanks Zsombor! I've merged this separately with some light restructuring (for consistency) in commit c6dd549d8c93ac36224c303ca6a753549b4f29f1.

PR #10, opened 2019-03-13
Improve performance of confocal image reconstruction

dean0x7d:
Should mostly resolve NG-1472.

Benchmarked with a 2 GiB HDF5 file containing a continuous scan with 212 frames:
```
|            | before (ms) | after (ms) | speedup |
|------------|-------------|------------|---------|
| load       |   1353.3    |     1.1    |    ∞    |
| rgb image  |   3906.6    |   863.9    |   4.5x  |
| num_frames |   0.003     |    19.7    |    ∞    |
|------------|-------------|------------|---------|
|  total     |   5259.9    |   884.7    |  5.95x  |
```

* `load` is just listing the scan using `file.scans`. This was previously very slow for continuous scans because the the number of frames needed to be calculated and this was being done in the constructor. This is now done lazily, but the calculation is also faster. See below.
* `rgb image` indicates the time needed to reconstruct the full RGB image with all the frames. This got a massive speedup thanks to a more efficient way of slicing continuous channels and a faster way of calculating the number of samples per pixel.
* `num_frames` is the time needed to look up the number of frames in a continuous scan. This was previously instant becaues it was calculated up front. This is now done lazily and it uses a dedicated algorithms which only looks at the infowave instead of reconstructing the entire image to get number of frames.

In total, the speedup is quite significant and should also reduce RAM usage.

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -3,6 +3,37 @@ import math
 import numpy as np
 
 
+class InfowaveCode(enum.IntEnum):
+    discard = 0  # this data sample does not contain useful information
+    use = 1  # useful data to be used to form a pixel
+    pixel_boundary = 2  # useful data and marks the last data sample in a pixel
+
+
+def reconstruct_num_frames(infowave, pixels_per_line, lines_per_frame):
+    """Reconstruct the number of frames in a continuous scan
+
+    Unfortunately, continuous scans do not store the number of frames as part of their
+    metadata (instead the number is zero) so we must reconstruct this number based on
+    the infowave.
+
+    Parameters
+    ----------
+    infowave : array_like
+        The famous infowave.
+    pixels_per_line : int
+        The number of pixels on the fast axis of the scan.
+    lines_per_frame : int
+        The number of pixels on the slow axis of the scan. Only needed for multi-frame scans.
+
+    Returns
+    -------
+    int
+    """
+    num_pixels = np.count_nonzero(infowave == InfowaveCode.pixel_boundary)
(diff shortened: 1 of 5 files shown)


onnodb:
Might be good to contact Filip in Delft directly with this change, and see whether it helps him. 

I'll have a closer look later, but I did not this is not going to support 3D scans. That's totally fine (Bluelake itself doesn't support it either), but would it make sense to make a note of that somewhere? Or is there already a check for this somewhere?

dean0x7d:
> I'll have a closer look later, but I did not this is not going to support 3D scans. That's totally fine (Bluelake itself doesn't support it either), but would it make sense to make a note of that somewhere? Or is there already a check for this somewhere?

I just added an explicit check in `__init__()`. See the last commit.

dean0x7d:
@onnodb Any change of reviewing this soonish? That way we can release v0.3.1. and update the relevant users.

onnodb:
Sorry Dean, this fell through the cracks. I'll look at it today!

lumicks/pylake/detail/image.py
```diff
@@ -3,6 +3,37 @@
 import numpy as np
 
 
+class InfowaveCode(enum.IntEnum):
+    discard = 0  # this data sample does not contain useful information
```
onnodb: I'd be in favor of having the naming here be consistent with Bluelake. Renaming the one in Bluelake is also fine with me, though, if that makes more sense.

dean0x7d: I'm not a fan of `combine` and `store` naming in Bluelake because it's very suggestive of serial data processing. I'll make a PR there to make it consistent.

PR #11, opened 2019-08-09
Bugfix: continuous channel bug when sliced within a time step.

JoepVanlier:
image

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -158,7 +158,7 @@ class Continuous:
             """Convert a timestamp into a continuous channel index (assumes t >= self.start)"""
             return (t - self.start + self.dt - 1) // self.dt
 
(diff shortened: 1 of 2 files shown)


lumicks/pylake/channel.py
```diff
@@ -158,7 +158,7 @@ def to_index(t):
             """Convert a timestamp into a continuous channel index (assumes t >= self.start)"""
             return (t - self.start + self.dt - 1) // self.dt
 
-        start = max(start, self.start)
+        start = max(start + np.mod(start, self.dt), self.start)
```
dean0x7d: You can use the builtin `%` operator here since it's faster for scalars.

JoepVanlier: Ah yes, makes sense. Thanks!

PR #12, opened 2019-08-13
Ng 954 - Read calibrations from HDF5

JoepVanlier:
* Adds calibration slicing for continuous and time series force channels.

* Adds slicing for calibration data.

* Adds tests for the new functionality.

* Add error message when start or stop of empty timeseries is requested.

### DIFF
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -1,23 +1,31 @@
+from collections import namedtuple
+
+
 class ForceCalibration:
-    """Parameters
+    """Calibration handling
+
+    Parameters
     ----------
-    calibration data : list of dictionaries
-        Calibration data with a field called "Stop time (ns)"
+    calibration data : named tuple containing
+        time_field - Name of the field used for time
+        list of dictionaries containing the raw calibration attribute data
     """
     def __init__(self, calibration):
         self._calibration = calibration
 
-    """Calibration data"""
-    """Filter calibration data based on time stamp range [ns]"""
+    """filter calibration data based on time stamp range [ns]"""
     @staticmethod
     def _filter_calibration(calibration, start, stop):
+        if len(calibration.items) == 0:
+            return {}
+
         def timestamp(x):
-            return x['Stop time (ns)']
+            return x[calibration.time_field]
 
         # Sort by time
-        calibration = sorted(calibration, key=timestamp)
-        calibration_items = [calibration[0]]
-        for i, v in enumerate(calibration):
+        calibration.items = sorted(calibration.items, key=timestamp)
+        calibration_items = [calibration.items[0]]
+        for i, v in enumerate(calibration.items):
             ts = timestamp(v)
             if ts <= start:
                 calibration_items[0] = v
@@ -29,27 +37,47 @@ class ForceCalibration:
 
         return calibration_items
 
+    """Filter calibration based on time stamp range
+
+        Parameters
+        ----------
+        start : time stamp at start [ns]
+        stop  : time stamp at stop [ns]
+        """
+    def filter_calibration(self, start, stop):
+        return self.__class__._filter_calibration(self._calibration, start, stop)
+
+    @property
     def calibration(self):
         """Calibration data for this channel"""
         """Calibration data slicing is deferred until calibration is requested to avoid"""
         """slicing values that may be needed."""
         if self._calibration:
-            return self.__class__._filter_calibration(self._calibration, self.start, self.stop)
+            return self.filter_calibration(self.start, self.stop)
         else:
             return AttributeError("No calibration data available")
 
-    def from_dataset(hdf5, n, xy):
+    def from_dataset(hdf5, n, xy, time_field='Stop time (ns)'):
         """Fetch the force calibration data from the HDF5 file"""
         def parse_force_calibration(cdata, force_idx, force_axis) -> list:
-            calibration_src = []
+            calibration = namedtuple("Calibration", {"time_field", "items"})
+            calibration.time_field = time_field
+            calibration.items = []
             for i, v in enumerate(cdata):
-                calibration_src.append(dict(cdata[v][f"Force {force_idx}{force_axis}"].attrs))
+                attrs = cdata[v][f"Force {force_idx}{force_axis}"].attrs
+                if len(attrs.keys()) > 0 and time_field in attrs.keys():
+                    calibration.items.append(dict(attrs))
 
-            return calibration_src
+            return calibration
 
         if "Calibration" in hdf5.keys():
-            calibration_data = parse_force_calibration(hdf5["Calibration"], n, xy)
+            if xy:
+                calibration_data = parse_force_calibration(hdf5["Calibration"], n, xy)
+            else:
+                raise NotImplementedError("Calibration is currently only implemented for single axis data")
         else:
-            calibration_data = {}
+            calibration_data = namedtuple("Calibration", {"time_field", "items"})
+            calibration_data.time_field = time_field
+            calibration_data.items = {}
 
         return calibration_data
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -22,9 +22,9 @@ def _filter_calibration(calibration, start, stop):
 
     calibration.items = sorted(calibration.items, key=timestamp)
 
-    calibration_items = [x for x in calibration.items if start <= timestamp(x) < stop]
+    calibration_items = [x for x in calibration.items if start < timestamp(x) < stop]
     pre = [x for x in calibration.items if timestamp(x) <= start]
-    if pre and (not calibration_items or start < timestamp(calibration_items[0])):
+    if pre:
         calibration_items.insert(0, pre[-1])
 
     return calibration_items
@@ -36,7 +36,6 @@ class ForceCalibration:
     Parameters
     ----------
     calibration: Calibration
-        Calibration information
     """
     def __init__(self, calibration):
         if not calibration:
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -38,9 +38,6 @@ class ForceCalibration:
     calibration: Calibration
     """
     def __init__(self, calibration):
-        if not calibration:
-            raise ValueError("Missing argument calibration")
-
         self._src = calibration
 
     def filter_calibration(self, start, stop):
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -1,26 +1,12 @@
-from collections import namedtuple
-
-"""A source of calibration data
-
-Parameters
-----------
-time_field : string
-    name of the field used for time
-items : list
-    list of dictionaries containing raw calibration attribute data
-"""
-Calibration = namedtuple("Calibration", {"time_field", "items"})
-
-
-def _filter_calibration(calibration, start, stop):
+def _filter_calibration(time_field, items, start, stop):
     """filter calibration data based on time stamp range [ns]"""
-    if len(calibration.items) == 0:
+    if len(items) == 0:
         return []
 
     def timestamp(x):
-        return x[calibration.time_field]
+        return x[time_field]
 
-    items = sorted(calibration.items, key=timestamp)
+    items = sorted(items, key=timestamp)
 
     calibration_items = [x for x in items if start < timestamp(x) < stop]
     pre = [x for x in items if timestamp(x) <= start]
@@ -35,10 +21,18 @@ class ForceCalibration:
 
     Parameters
     ----------
-    calibration: Calibration
+    A source of calibration data
+
+    Parameters
+    ----------
+    time_field : string
+        name of the field used for time
+    items : list
+        list of dictionaries containing raw calibration attribute data
     """
-    def __init__(self, calibration):
-        self._src = calibration
+    def __init__(self, time_field, items):
+        self._time_field = time_field
+        self._items = items
 
     def filter_calibration(self, start, stop):
         """Filter calibration based on time stamp range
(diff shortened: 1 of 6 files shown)


lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+        stop  : time stamp at stop [ns]
+        """
+    def filter_calibration(self, start, stop):
+        return self.__class__._filter_calibration(self._calibration, start, stop)
+
+    @property
+    def calibration(self):
+        """Calibration data for this channel"""
+        """Calibration data slicing is deferred until calibration is requested to avoid"""
+        """slicing values that may be needed."""
```
dean0x7d: Similar to the above, see the `numpydoc` style guide. This can be one single docstring with a title and body:
```
"""Calibration data for this channel
        
Calibration data slicing is deferred until calibration is requested to avoid
slicing values that may be needed.
"""
```

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+        if self._calibration:
```
dean0x7d: No need to do this check here. `__init__` should ensure we have a `self._calibration` and from then on it's an invariant of the class.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+            return AttributeError("No calibration data available")
+
+    @staticmethod
+    def from_dataset(hdf5, n, xy, time_field='Stop time (ns)'):
+        """Fetch the force calibration data from the HDF5 file"""
+        def parse_force_calibration(cdata, force_idx, force_axis) -> list:
+            calibration = namedtuple("Calibration", {"time_field", "items"})
+            calibration.time_field = time_field
+            calibration.items = []
+            for i, v in enumerate(cdata):
```
dean0x7d: `i` is never used.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+                attrs = cdata[v][f"Force {force_idx}{force_axis}"].attrs
+                if len(attrs.keys()) > 0 and time_field in attrs.keys():
```
dean0x7d: No need for the extra `len(keys) > 0` check. `key in keys` will naturally return `False` if `keys` is empty.

lumicks/pylake/channel.py
```diff
@@ -194,11 +203,17 @@ def from_dataset(dset, y_label="y"):
 
     @property
     def start(self):
-        return self.timestamps[0]
+        if len(self.timestamps) > 0:
```
dean0x7d: The Pythonic way to check for a non-empty sequence is a simple `if self.timestamps:`, because an empty sequence evaluates to `False`.

lumicks/pylake/channel.py
```diff
@@ -247,6 +262,91 @@ def downsampled_by(self, factor, reduce):
         raise NotImplementedError("Downsampling is not available for time tag data")
 
 
+class ContinuousCalibrated(Continuous, ForceCalibration):
```
dean0x7d: I would not reach for inheritance here. This implementation seems to double the number of classes and also duplicates a lot of code. It should be possible to just make `ForceCalibration` an attribute of the existing `Continuous` class.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+        calibration_items = [calibration.items[0]]
+        for i, v in enumerate(calibration.items):
+            ts = timestamp(v)
+            if ts <= start:
+                calibration_items[0] = v
+            elif ts < stop:
+                calibration_items.append(v)
+            else:
+                # Since the list is sorted, we can early out
+                break
```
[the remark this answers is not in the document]
JoepVanlier: The previous one is only needed if start doesn't equal the calibration timestamp.

So I'd have to add something like pre and start < timestamp(calibration_items[0]), but add another check that calibration items is not empty, which gives rise to 
if pre and (not calibration_items or start < timestamp(calibration_items[0])):

In addition, there's one more corner case, if we slice with [120,120] and a calibration point exists at 120, we'd probably want 120 rather than the one before it. This code would still gives you the calibration point before that since the filter uses <. This could be fixed by using <= for the stop condition in the second filter I suppose.

What I'd end up with is this:
```
    calibration_items = [x for x in calibration.items if start <= timestamp(x) < stop]

    pre = [x for x in calibration.items if timestamp(x) <= start]

    if pre and (not calibration_items or start < timestamp(calibration_items[0])):

        calibration_items.insert(0, pre[-1])
```

Then again, it may not be desirable, since [120,120] should already return an empty slice. Maybe it'd make more sense to return an empty calibration struct to signal that the slice will be empty. In the timeseries case that actually already happens since the slicing there is not lazily evaluated. What do you think?

Personally, the for loop is easier to read for me, but I should probably get used to writing more in terms of list comprehensions; since it's more in style with python. I will get used to this and then this'll probably feel more natural.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+        if len(calibration.items) == 0:
+            return []
+
+        def timestamp(x):
+            return x[calibration.time_field]
+
+        # Sort by time
+        calibration.items = sorted(calibration.items, key=timestamp)
+        calibration_items = [calibration.items[0]]
+        for i, v in enumerate(calibration.items):
```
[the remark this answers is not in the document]
JoepVanlier: Ah yeah. Enumerate isn't needed here.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+        assert calibration
+
+        if len(calibration.items) == 0:
+            return []
+
+        def timestamp(x):
+            return x[calibration.time_field]
+
+        # Sort by time
+        calibration.items = sorted(calibration.items, key=timestamp)
```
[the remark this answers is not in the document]
JoepVanlier: Fair point. Removed.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if not hasattr(calibration, "time_field"):
+            raise TypeError("Calibration structure is missing time_field field")
+
+        self._calibration = calibration
+
+    """filter calibration data based on time stamp range [ns]"""
+    @staticmethod
+    def _filter_calibration(calibration, start, stop):
+        assert calibration
```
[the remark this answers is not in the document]
JoepVanlier: Alright. Makes sense.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+            raise TypeError("Calibration structure is missing items field")
+
+        if not hasattr(calibration, "time_field"):
+            raise TypeError("Calibration structure is missing time_field field")
+
+        self._calibration = calibration
+
+    """filter calibration data based on time stamp range [ns]"""
+    @staticmethod
+    def _filter_calibration(calibration, start, stop):
```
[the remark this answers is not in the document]
JoepVanlier: Good point. Done.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+
+        if not hasattr(calibration, "items"):
+            raise TypeError("Calibration structure is missing items field")
+
+        if not hasattr(calibration, "time_field"):
+            raise TypeError("Calibration structure is missing time_field field")
+
+        self._calibration = calibration
+
+    """filter calibration data based on time stamp range [ns]"""
```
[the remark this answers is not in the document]
JoepVanlier: Changed.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+    """
+    def __init__(self, calibration):
+        if not calibration:
+            raise ValueError("Missing argument calibration")
+
+        if not hasattr(calibration, "items"):
+            raise TypeError("Calibration structure is missing items field")
+
+        if not hasattr(calibration, "time_field"):
+            raise TypeError("Calibration structure is missing time_field field")
```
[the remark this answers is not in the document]
JoepVanlier: Good point. Changed.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
+from collections import namedtuple
+
+
+class ForceCalibration:
+    """Calibration handling
+
+    Parameters
+    ----------
+    calibration: named tuple containing
```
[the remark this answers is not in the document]
JoepVanlier: Done. I modified it, but the text string after the type seems almost redunant now.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,95 @@
...
+
+            return calibration
+
+        if "Calibration" in hdf5.keys():
+            if xy:
+                calibration_data = parse_force_calibration(hdf5["Calibration"], n, xy)
+            else:
+                raise NotImplementedError("Calibration is currently only implemented for single axis data")
+        else:
+            calibration_data = namedtuple("Calibration", {"time_field", "items"})
```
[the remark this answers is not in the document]
JoepVanlier: Sounds good! Amended.

lumicks/pylake/calibration.py
JoepVanlier: Amended.

lumicks/pylake/calibration.py
JoepVanlier: True. Amended.

lumicks/pylake/calibration.py
JoepVanlier: Yeah, enumerate was superfluous.

lumicks/pylake/channel.py
JoepVanlier: The issue was that timestamps usually has more than one element, which raises an error that this is ambiguous. Would you prefer self.timestamps.any() over the length check? I thought the len here was more clear about the intent than self.timestamps.any() is.

lumicks/pylake/channel.py
JoepVanlier: Yeah. I was actually doubting here.

Considering that we'd just have ForceCalibration as an internal, I also renamed its class's data as _src, since _calibration._calibration looks pretty horrible and unclear.

Do you think it makes more sense to have the assignment of the calibration attribute in the constructor, or via a separate attachment function?

The new implementation makes the named tuple feel a little redundant, since I construct ForceCalibration anyway.

lumicks/pylake/calibration.py
JoepVanlier: Ah yeah. Fixed.

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
dean0x7d: It's perfectly fine to just keep the `name : type` without a description when it's obvious from the type.

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
dean0x7d: Indeed, you are right, my previous code does not quite do the right thing. Modifying the conditions a bit would resolve this:
```python
# before
results = [x for x in items if start <= timestamp(x) < stop]
pre = [x for x in items if timestamp(x) < start]

# after
results = [x for x in items if start < timestamp(x) < stop]
pre = [x for x in items if timestamp(x) <= start]
```
There should be no need for the extra conditions in this case. But anyway, I don't want to push for the comprehensions since they can be a bit of a foreign concept to start with and also not really more efficient in this case. But in the long run, I do really like them because, once used to them, they make the logic clearer and more composable.

lumicks/pylake/channel.py
dean0x7d: Ah, I forgot that these were `numpy` arrays which don't quite do the same thing with regard to `True`ness as regular sequences in Python. Never mind, this is fine.

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
JoepVanlier: Yep. That's a lot cleaner. I must have been sleepy when I wrote those ugly extra conditions.

I'll update it, thanks! 👍

lumicks/pylake/channel.py

lumicks/pylake/channel.py
```diff
@@ -122,21 +131,22 @@ class Continuous:
...
         Delta between two timestamps. Constant for the entire data range.
     """
-    def __init__(self, data, start, dt):
+    def __init__(self, data, start, dt, calibration=None):
         self._src_data = data
         self._cached_data = None
         self.start = start
         self.stop = start + len(data) * dt
         self.dt = dt
+        self._calibration = calibration
```
dean0x7d: Have you considered storing the calibrations in `Slice` instead of in the individual data sources? I see some small differences in the implementations of the calibrations for continuous and time series sources, but I think it would be a huge advantage for maintenance to make the behavior uniform and pull it up into `Slice`. 

If the reason for the different behavior of calibrations is the case of the empty slice, I'd be in favor of making the empty slice behavior more uniform rather than diverging the calibration behavior.

JoepVanlier: It was the latter, but makes sense. Moved calibration into slice now.

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,91 @@
...
+        Calibration data slicing is deferred until calibration is requested to avoid
+        slicing values that may be needed.
+        """
+        return self.filter_calibration(self.start, self.stop)
+
+    @staticmethod
+    def from_dataset(hdf5, n, xy, time_field='Stop time (ns)'):
+        """Fetch the force calibration data from the HDF5 file"""
+        def parse_force_calibration(cdata, force_idx, force_axis) -> list:
+            calibration = Calibration
```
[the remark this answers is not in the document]
JoepVanlier: Good catch! 👍

lumicks/pylake/calibration.py
```diff
@@ -0,0 +1,91 @@
...
+        if not calibration:
+            raise ValueError("Missing argument calibration")
+
+        self._src = calibration
+
+    def filter_calibration(self, start, stop):
+        """Filter calibration based on time stamp range
+
+            Parameters
+            ----------
```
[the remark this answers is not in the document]
JoepVanlier: Done.

lumicks/pylake/file.py
```diff
@@ -119,13 +120,20 @@ def print_group(group, name="", indent=-2):
         return print_attributes(self.h5) + "\n" + print_group(self.h5)
 
     def _get_force(self, n, xy):
-        return Continuous.from_dataset(self.h5["Force HF"][f"Force {n}{xy}"], "Force (pN)")
+        force_group = self.h5["Force HF"][f"Force {n}{xy}"]
+        calibration_data = ForceCalibration(ForceCalibration.from_dataset(self.h5, n, xy))
```
dean0x7d: `ForceCalibration(ForceCalibration.from_dataset(...))` is a bit redundant. You should make sure that `from_dataset()` already returns a `ForceCalibration`.

PR #14, opened 2019-08-20
[blocked] It should be possible to do arithmetic with two channels with the same units

JoepVanlier:
- Adds arithmetic operations to slice object (/, *, +, -, rdiv, rmul, radd, rsub).

- Adds tests for various combinations of arithmetic on slices.


lumicks/pylake/channel.py
```diff
@@ -21,14 +22,54 @@ class Slice:
...
+        unit = self.compare_unit(other, 'mul')
+        return Slice(self._src._with_data_source(self._src.data * self._unpack_other(other)), unit=unit)
+
+    def __rtruediv__(self, other):
+        unit = self.compare_unit(other, 'div')
+        return Slice(self._src._with_data_source(self._unpack_other(other)/self._src.data), unit=unit)
+
+    def __rsub__(self, other):
+        unit = self.compare_unit(other, 'sub')
+        return Slice(self._src._with_data_source(self._unpack_other(other)-self._src.data), unit=unit)
```
tiagobonetti: I'm not even sure if this is standard or not but `__rsub__` and `__rtruediv__` both lack space between operators:
`self._unpack_other(other)-self._src.data` -> `self._unpack_other(other) - self._src.data`

lumicks/pylake/tests/test_channels.py
```diff
@@ -106,6 +155,168 @@ def test_calibration_continuous_channels():
...
+    scc1 = channel.Slice(channel.Continuous([1, 2, 3, 4, 5], start=0, dt=1))
+    assert np.allclose((scc1 - scc2).data, [0, 1, 0, 3, 0])
+    assert np.allclose((scc1 + scc2).data, [2, 3, 6, 5, 10])
+    assert np.allclose((scc1 * scc2).data, [1, 2, 9, 4, 25])
+    assert np.allclose((scc1 / scc2).data, [1, 2, 1, 4, 1])
+    with pytest.raises(ValueError):
+        scc1 - scc3
+
+
+def test_arithmetic_operations():
```
tiagobonetti: A function so nice you wrote it twice?

I'm confused. :thinking:

JoepVanlier: Huh... 😕 You and me both. My best guess is that I must have done this accidentally when I rebased it. Removed the duplication now.

lumicks/pylake/channel.py
JoepVanlier: 👍 Added

lumicks/pylake/detail/units.py
```diff
@@ -0,0 +1,38 @@
...
+            raise NotImplementedError(f"Division not implemented between units {unit1} and {unit2}")
+    elif operation == "mul":
+        if unit2 == au:
+            output_unit = unit1
+        elif unit1 == au:
+            output_unit = unit2
+        else:
+            raise NotImplementedError(f"Multiplication not implemented between units {unit1} and {unit2}")
+
+    return output_unit
```
tiagobonetti: I would return form inside the `if`s directly and maybe just return `None` here.

But actually, that opens an interesting question, shouldn't we raise in the case the operation is unknown?

lumicks/pylake/channel.py
```diff
@@ -21,14 +22,54 @@ class Slice:
...
     def __len__(self):
         return len(self._src)
 
+    def _unpack_other(self, other):
+        if isinstance(other, self.__class__):
+            return other._src.data
+        else:
+            return other
+
+    def compare_unit(self, other, operation):
```
tiagobonetti: `compare_unit` sounds like we're getting a boolean as a result, can we opt for something like the underlying `determine_unit`?

lumicks/pylake/channel.py
```diff
@@ -21,14 +22,54 @@ class Slice:
...
+        other_unit = getattr(other, "unit", None)
+        if other_unit:
+            return determine_unit(self.unit, other_unit, operation)
+
+    def __add__(self, other):
+        unit = self.compare_unit(other, 'add')
+        return Slice(self._src._with_data_source(self._src.data + self._unpack_other(other)), unit=unit)
```
tiagobonetti: My python is old and crappy so I may be wrong.

Aren't we constructing `Slice` twice here? one inside `_with_data_source` and another explicitly?  
If so we can probably drop the explicit one.

lumicks/pylake/tests/test_channels.py
```diff
@@ -106,6 +155,87 @@ def test_calibration_continuous_channels():
     assert nested_slice.calibration[0]["Calibration Data"] == 120
 
 
+def test_arithmetic_operations():
```
tiagobonetti: There is a significant amount of repetition in this function.  
I'm not sure what's the best strategy, but I see a lot of it could be reused for different channel type and also between raw lists a _numpy_ arrays.

lumicks/pylake/detail/units.py
JoepVanlier: Thanks man. Good point on both accounts. Changed! 👍 

Also added a little test for the invalid operation as well.

lumicks/pylake/channel.py
JoepVanlier: Yep. Makes sense. Actually, it should probably be a private function, since we don't really intend for the user to mess with this directly. I changed it to _determine_unit now.

lumicks/pylake/channel.py
JoepVanlier: Nah, the function _with_data_source generates a new instance of the underlying datatype (Continuous, TimeSeries or TimeTags), not slice (note the _src).

lumicks/pylake/tests/test_channels.py
JoepVanlier: Yes, there is a significant amount of repetition, but I figured it'd be easier to interpret a failing test when each line can be attributed to a specific underlying datatype being tested.

Anyways, I changed it to avoid the repetition now. It's true that the repetition is a bit meh. :)

lumicks/pylake/detail/units.py
```diff
@@ -0,0 +1,37 @@
+
+au = "dimensionless"
+
+
+def determine_unit(unit1, unit2, operation):
+    """Compares units. Raises exception when units are not easily converted or incompatible.
+
+    Parameters
+    ----------
+    unit1 : string
```
dean0x7d: `str` is the canonical type name in Python.

lumicks/pylake/detail/units.py
```diff
@@ -0,0 +1,37 @@
+
+au = "dimensionless"
```
dean0x7d: `arbitrary_unit != dimensionless` -- AU has a unit, it's just unknown to us.

lumicks/pylake/file.py
```diff
@@ -148,14 +148,14 @@ def make(channel):
         fx = make(f"Force {n}x")
         fy = make(f"Force {n}y")
         return Slice(TimeSeries(np.sqrt(fx.data**2 + fy.data**2), fx.timestamps),
-                     labels={"title": f"Force LF/Force {n}", "y": "Force (pN)"})
+                     labels={"title": f"Force LF/Force {n}", "y": "Force (pN)"}, unit="pN")
 
     def _get_distance(self, n):
         return TimeSeries.from_dataset(self.h5["Distance"][f"Distance {n}"],
-                                       r"Distance ($\mu$m)")
+                                       r"Distance ($\mu$m)", unit="micrometer")
```
dean0x7d: `"µm"` would be preferable.

lumicks/pylake/channel.py
```diff
@@ -21,14 +22,54 @@ class Slice:
...
+    def _unpack_other(self, other):
+        if isinstance(other, self.__class__):
+            return other._src.data
+        else:
+            return other
+
+    def _determine_unit(self, other, operation):
+        other_unit = getattr(other, "unit", None)
+        if other_unit:
+            return determine_unit(self.unit, other_unit, operation)
```
dean0x7d: This function has a path without a `return`. Python does an implicit `return None` in that case, but we really should be explicit if at least one path has a `return`.

lumicks/pylake/channel.py
```diff
@@ -21,14 +22,54 @@ class Slice:
         Plot labels: "x", "y", "title".
     calibration: ForceCalibration
     """
-    def __init__(self, data_source, labels=None, calibration=None):
+    def __init__(self, data_source, labels=None, calibration=None, unit=None):
```
dean0x7d: If `unit == None` the same thing as `unit == "dimensionless"`? Should it be?

lumicks/pylake/detail/units.py
JoepVanlier: Replaced 👍

lumicks/pylake/detail/units.py
JoepVanlier: True. The variable name misrepresents the intent. I will change this.

lumicks/pylake/file.py
JoepVanlier: Alright. Modified.

lumicks/pylake/channel.py
JoepVanlier: Actually, this comment made me think.

It's probably wiser to be more explicit and move lack of unit handling into units. For addition and subtraction, adding a value results in the unit that is present; because this is the only operation that makes any sense. For the other operations, it results in an undefined unit (signified by None).

It does mean that when working with datatypes that have no explicit unit, strict unit checking goes out the window (but I don't think that can be avoided, really).

lumicks/pylake/channel.py
JoepVanlier: No unit (None) would be the true arbitrary unit I suppose, i.e. no information on the unit being available. Dimensionless was renamed to make this more clear.

dean0x7d:
Can you add unit tests specifically for `determine_unit`? And document exactly the meaning of `None` and `dimensionless`. I'm a bit confused now. I'm not entirely sure about the rules are here. Is ti that a `Slice` without an explicit unit set is `None` and also `np.ndarray` is `None`? I would actually expect those to be dimensionless.

I feel like it would be best to simplify to just dimensionless (represented by `None`) and a specific unit. Adding arbitrary units into the mix seems to complicate things a bit too much. Do we then consider AU like NaNs, i.e. they never compare equal?

JoepVanlier:
> Can you add unit tests specifically for `determine_unit`? And document exactly the meaning of `None` and `dimensionless`. I'm a bit confused now. I'm not entirely sure about the rules are here. Is ti that a `Slice` without an explicit unit set is `None` and also `np.ndarray` is `None`? I would actually expect those to be dimensionless.
> 
> I feel like it would be best to simplify to just dimensionless (represented by `None`) and a specific unit. Adding arbitrary units into the mix seems to complicate things a bit too much. Do we then consider AUs to be like NaNs, i.e. they never compare equal?

I have added a description of the reasoning in the determine_unit function and added a test which covers all the cases.

I was reasoning along the lines of what can be inferred about a unit when the datatype does not specify one. We know for instance that 5 + ? is valid iff both units are the same. Hence the unit of the unknown quantity can be inferred as being the same. For division and multiplication this is not the case. In this case units cannot be inferred and therefore the return "unit" is None (unknown). I was debating whether to raise an exception instead, but I'm afraid that this would be obnoxious and prohibitive in any sort of analysis where normalization is involved, and would likely result in users just extracting the raw data from the objects early on.

The special case of the dimensionless unit is only to allow some additional operations to be performed. If we ever switch to a proper unit library, this special case would become superfluous as the libraries I have looked at all support dimensionless quantities natively. In this case, the units can just be forwarded to that library to deal with.

Maybe it would make sense to have a short meeting with everyone about this topic, since it is a fairly early API decision on whether to err on the side of safety or on the side of trusting the user.

In terms of comparison, I would say that yes, any operation where two quantities that do not have units associated with them should return a quantity without a unit.

lumicks/pylake/detail/units.py
```diff
@@ -0,0 +1,98 @@
...
+
+        Returns
+        -------
+        array_like
+        """
+        if isinstance(other, WithUnit):
+            if hasattr(other._src, "data"):
+                return other._src.data
+            else:
+                return other._src
```
JoepVanlier: This bit I'm unhappy with, but the ability to use bare lists is nice.

Alternatively to this, I could do a check in the constructor of withUnit and convert the incoming list to an nparray before assigning it to _src if it doesn't have a data field. Neither solution seems elegant to me.

JoepVanlier:
rip

PR #15, opened 2019-08-27
Ng 987 - Export pixel size and dwell time to TIFF

JoepVanlier:
- Adds dwell time / pixel time as metadata to exported TIFF file.

- Adds resolution to exported TIFF file (resolution is verified to be correctly recognized in ImageJ).

- Add tests for resolution and pixel time export.

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -124,4 +124,5 @@ def save_tiff(image, filename, dtype, clip=False, pixel_size=1.0, pixel_time=1.0
     pixel_time = pixel_time * 1e-3              # ms => s
 
     tifffile.imsave(filename, image.astype(dtype), resolution=(1/pixel_size, 1/pixel_size, "CENTIMETER"),
-                    metadata={'PixelTime': pixel_time})
+                    metadata={'PixelTime': pixel_time, 'PixelTimeUnit': 's'})
+
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -3,7 +3,7 @@ import math
 import numpy as np
 
 
-class ImageMetaData:
+class ImageMetadata:
     """Image metadata
 
     Parameters
@@ -25,34 +25,31 @@ class ImageMetaData:
             self._pixel_size_y = pixel_size_x
 
     @staticmethod
-    def from_dataset(json=None):
+    def from_dataset(json):
         """
         Fetch metadata from json structure
 
         Parameters
         ----------
-        json : json structure containing kymograph metadata
+        json : dict
+            json structure containing kymograph metadata
         """
         if json:
             pixel_size = json["scan volume"]["scan axes"][0]["pixel size (nm)"]
             pixel_time = json["scan volume"]["pixel time (ms)"]
-            return ImageMetaData(pixel_size_x=pixel_size, pixel_time=pixel_time)
+            return ImageMetadata(pixel_size_x=pixel_size, pixel_time=pixel_time)
         else:
-            return ImageMetaData()
+            return ImageMetadata()
 
     @property
     def resolution(self):
-        """
-        :return: X, Y resolution in pixels per cm followed by unit specification accepted by Tifffile
-        """
+        """X, Y resolution in pixels per cm followed by unit specification accepted by Tifffile"""
         # TIFF only supports centimeters and inches as valid units, hence we convert from nm => cm
         return 1e7 / self._pixel_size_x, 1e7 / self._pixel_size_y, "CENTIMETER"
 
     @property
     def metadata(self):
-        """
-        :return: Dictionary with metadata
-        """
+        """Dictionary with metadata"""
         pixel_time = self._pixel_time * 1e-3  # ms => s
         return {'PixelTime': pixel_time, 'PixelTimeUnit': 's'}
 
@@ -144,7 +141,7 @@ def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, red
         return pixels.reshape(-1, lines_per_frame, pixels_per_line).squeeze()
 
 
-def save_tiff(image, filename, dtype, clip=False, metadata=ImageMetaData()):
+def save_tiff(image, filename, dtype, clip=False, metadata=ImageMetadata()):
     """Save an RGB `image` to TIFF
 
     This is a thin wrapper around `tifffile` with additional safety checks
@@ -160,7 +157,7 @@ def save_tiff(image, filename, dtype, clip=False, metadata=ImageMetaData()):
     clip : bool
         If enabled, the photon count data will be clipped to fit into the desired `dtype`.
         This option is disabled by default: an error will be raise if the data does not fit.
-    metadata : ImageMetaData
+    metadata : ImageMetadata
     """
     import tifffile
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -90,29 +90,39 @@ def round_up(size, n):
...
+    if not clip and (np.min(image) < info.min or np.max(image) > info.max):
+        raise RuntimeError(f"Can't safely export image with `dtype={dtype.__name__}` channels."
+                           f" Switch to a larger `dtype` in order to safely store everything"
+                           f" or pass `force=True` to clip the data.")
+
+    # TIFF only supports centimeters and inches as valid units
+    pixel_size = float(pixel_size) * 1e-7       # nm => cm
+    pixel_time = pixel_time * 1e-3              # ms => s
+
+    tifffile.imsave(filename, image.astype(dtype), resolution=(1/pixel_size, 1/pixel_size, "CENTIMETER"),
```
mwilkowski: nitpick: 1/pixel_size -> 1 / pixel_size

lumicks/pylake/detail/image.py
```diff
@@ -90,29 +90,39 @@ def round_up(size, n):
...
+        This option is disabled by default: an error will be raise if the data does not fit.
+    pixel_size : float
+        Pixel size [nm]
+    pixel_time: float
+        How long does pixel acquisition take [ms]
+    """
+    import tifffile
+
+    info = np.finfo(dtype) if np.dtype(dtype).kind == "f" else np.iinfo(dtype)
+    if not clip and (np.min(image) < info.min or np.max(image) > info.max):
```
mwilkowski: nitpick: I think it could improve the readability to wrap np.min(image) < info.min into parenthesis so it's neatly separated then you have:
`( (np.min(image) < info.min) or (np.max(image) > info.max) ) `
and the precedence is now obvious. But that maybe my issue with python operators.

dean0x7d: The precedence is obvious. We should not be adding redundant parenthesis. In the long run, it's much more readable, just like `a * b + c` vs `(a * b) + c`.

lumicks/pylake/detail/image.py
```diff
@@ -90,29 +90,39 @@ def round_up(size, n):
...
-        import tifffile
-
-        info = np.finfo(dtype) if np.dtype(dtype).kind == "f" else np.iinfo(dtype)
-        if not clip and (np.min(image) < info.min or np.max(image) > info.max):
-            raise RuntimeError(f"Can't safely export image with `dtype={dtype.__name__}` channels."
-                               f" Switch to a larger `dtype` in order to safely store everything"
-                               f" or pass `force=True` to clip the data.")
-
-        tifffile.imsave(filename, image.astype(dtype))
+def save_tiff(image, filename, dtype, clip=False, pixel_size=1.0, pixel_time=1.0):
```
dean0x7d: I would suggest making a dedicated class for metadata which can be passed to this function so that we can more easily extend it with more metadata without needing to add a lot of function arguments here. E.g. `pixel_size` does not necessarily need to be the same for x and y so that's already an extra argument. And we'll definitely be adding more in the future.

You can make that metadata class have properties which return the arguments expected by `tifffile.imsave()`.

lumicks/pylake/detail/image.py
dean0x7d: The pixel size can be different for x and y.

lumicks/pylake/detail/image.py
JoepVanlier: Makes sense. I've moved it into its own class, with a from_data that takes the json containing the metadata. Also added a test for it.

lumicks/pylake/detail/image.py
```diff
@@ -3,6 +3,60 @@
 import numpy as np
 
 
+class ImageMetaData:
```
dean0x7d: Metadata is usually one word, so `ImageMetadata`.

lumicks/pylake/detail/image.py
```diff
@@ -3,6 +3,60 @@
...
+    def __init__(self, pixel_size_x=1.0, pixel_size_y=None, pixel_time=1.0):
+        self._pixel_time = pixel_time
+        self._pixel_size_x = pixel_size_x
+        if pixel_size_y:
+            self._pixel_size_y = pixel_size_y
+        else:
+            self._pixel_size_y = pixel_size_x
+
+    @staticmethod
+    def from_dataset(json=None):
```
dean0x7d: I don't think that the default value here makes sense. If someone wants to create a defaulted instance, they should just use the constructor directly: `ImageMetadata()` instead of `ImageMetadata.from_dataset()`.

lumicks/pylake/detail/image.py
```diff
@@ -3,6 +3,60 @@
...
+        """
+        Fetch metadata from json structure
+
+        Parameters
+        ----------
+        json : json structure containing kymograph metadata
```
dean0x7d: Type goes after the colon. The description goes on the next line.

lumicks/pylake/detail/image.py
```diff
@@ -3,6 +3,60 @@
...
+            pixel_size = json["scan volume"]["scan axes"][0]["pixel size (nm)"]
+            pixel_time = json["scan volume"]["pixel time (ms)"]
+            return ImageMetaData(pixel_size_x=pixel_size, pixel_time=pixel_time)
+        else:
+            return ImageMetaData()
+
+    @property
+    def resolution(self):
+        """
+        :return: X, Y resolution in pixels per cm followed by unit specification accepted by Tifffile
```
dean0x7d: `:return:` is not part of `numpydoc`. For functions, you should use the `Returns` section. For properties, you can just say what it returns without any extra qualification. E.g. here it would be simply:
```python
@property
    def resolution(self):
        """X, Y resolution in pixels per cm followed by unit specification accepted by Tifffile"""
```

lumicks/pylake/kymo.py
```diff
@@ -171,4 +171,6 @@ def save_tiff(self, filename, dtype=np.float32, clip=False):
             If enabled, the photon count data will be clipped to fit into the desired `dtype`.
             This option is disabled by default: an error will be raise if the data does not fit.
         """
-        save_tiff(self.rgb_image, filename, dtype, clip)
+        metadata = ImageMetaData.from_dataset(self.json)
+
+        save_tiff(self.rgb_image, filename, dtype, clip, metadata)
```
dean0x7d: Nitpick: you can just inline the `metadata` variable here.

lumicks/pylake/detail/image.py
JoepVanlier: Fair point. Modified.

lumicks/pylake/detail/image.py
JoepVanlier: Amended. 👍

lumicks/pylake/detail/image.py
JoepVanlier: Alright. Thanks. I'll keep that in mind.

lumicks/pylake/detail/image.py
JoepVanlier: This was modified along with the refactor.

lumicks/pylake/detail/image.py
JoepVanlier: Modified.

lumicks/pylake/kymo.py
JoepVanlier: Done.

PR #16, opened 2019-09-03
Ng 1888 - Implement slicing behaviour for kymographs

JoepVanlier:
- Adds slicing behaviour to kymograph objects.

- Adds tests and fixture for kymo loading from file.

- Slightly modifies kymograph constructor interface to bring it more in line with other files (constructing from data using from_dataset rather than directly in the constructor).

- Adds tests for kymograph slicing.

- Adds convenience option to directly specify a plotting range using xlim and ylim for the kymograph plotting function (even without slicing).

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -85,6 +85,24 @@ def reconstruct_num_frames(infowave, pixels_per_line, lines_per_frame):
     return math.ceil(num_pixels / pixels_per_frames)
 
 
+def line_timestamps_image(time_stamps, infowave, pixels_per_line):
+    """Determine the starting timestamps for lines in the kymograph
+
+    Parameters
+    ----------
+    time_stamps : array_like
+        Timestamps corresponding to the infowave.
+    infowave : array_like
+        The infamous infowave.
+    pixels_per_line : int
+        The number of pixels on the fast axis of the scan.
+    """
+    pixel_start_idx = np.flatnonzero(infowave == InfowaveCode.pixel_boundary)
+    pixel_start_idx = np.concatenate(([0], pixel_start_idx + 1))
+
+    return time_stamps[pixel_start_idx[0:-1:pixels_per_line]]
+
+
 def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, reduce=np.sum):
     """Reconstruct a scan or kymograph image from raw data
 
@@ -93,7 +111,7 @@ def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, red
     data : array_like
         Raw data to use for the reconstruction. E.g. photon counts or force samples.
     infowave : array_like
-        The famous infowave.
+        The infamous infowave.
     pixels_per_line : int
         The number of pixels on the fast axis of the scan.
     lines_per_frame : Optional[int]
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -143,9 +143,7 @@ class Kymo(PhotonCounts):
                 Field to look for
             """
             if field in args:
-                tmp = args[field]
-                args.pop(field)
-                return tmp
+                return args.pop(field)
             return None
 
         width_um = self.json["scan volume"]["scan axes"][0]["scan width (um)"]
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -2,7 +2,8 @@ import json
 import numpy as np
 
 from .detail.mixin import PhotonCounts
-from .detail.image import reconstruct_image, save_tiff, ImageMetadata
+from .detail.image import reconstruct_image, save_tiff, ImageMetadata, line_timestamps_image
+from .detail.timeindex import to_timestamp
 
 
 class Kymo(PhotonCounts):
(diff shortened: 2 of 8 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -93,7 +111,7 @@ def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, red
     data : array_like
         Raw data to use for the reconstruction. E.g. photon counts or force samples.
     infowave : array_like
-        The famous infowave.
+        The infamous infowave.
```
mwilkowski: (:

lumicks/pylake/kymo.py
```diff
@@ -2,31 +2,72 @@
...
+        line_timestamps = np.append(line_timestamps, timestamps[-1])
+
+        i_min = np.searchsorted(line_timestamps, start, side='left')
+        i_max = np.searchsorted(line_timestamps, stop, side='left')
+
+        if i_min >= len(line_timestamps):
+            raise IndexError(f"Slice selection out of range ({(start-self.start)/1e9} sec >= {(self.stop-self.start)/1e9} sec)")
+
+        if i_min >= i_max:
+            raise IndexError(f"Slice would result in empty slice. Valid slice range is from {(start-self.start)/1e9} sec to {(self.stop-self.start)/1e9} sec")
```
mwilkowski: Is it customary in python to throw in such cases? It's going to be empty but it's not an error state. Genuinely curious!

lumicks/pylake/kymo.py
```diff
@@ -91,6 +132,22 @@ def timestamps(self) -> np.ndarray:
...
+            Parameters
+            ----------
+            args : dict
+                Dictionary with function arguments
+            field : string
+                Field to look for
+            """
+            if field in args:
+                tmp = args[field]
+                args.pop(field)
```
mwilkowski: can be tmp = args.pop(field) ?

lumicks/pylake/detail/image.py
tiagobonetti: Not the change we needed, but the change we deserved. :bat:

JoepVanlier: 😆

lumicks/pylake/kymo.py
JoepVanlier: Nope, it's not.

I have changed it to returning an empty kymograph now.

lumicks/pylake/kymo.py
JoepVanlier: Good point. Could even return args.pop(field) directly. Fixed 👍

lumicks/pylake/kymo.py
```diff
@@ -99,11 +154,19 @@ def _plot(self, image, **kwargs):
             aspect=(image.shape[0] / image.shape[1]) * (duration / width_um)
         )
 
+        x_lims = fetch_custom_property(kwargs, "xlim")
+        y_lims = fetch_custom_property(kwargs, "ylim")
+
```
tiagobonetti: I would do the extracting of custom arguments before preparing the default arguments.

lumicks/pylake/kymo.py
```diff
@@ -91,6 +132,20 @@ def timestamps(self) -> np.ndarray:
     def _plot(self, image, **kwargs):
         import matplotlib.pyplot as plt
 
+        def fetch_custom_property(args, field):
```
tiagobonetti: I would call it something like `extract_custom_property` to make the intent clearer. I took me a while to understand this was modifying the arguments dictionary.

Another possibility is to do a list comprehension with conditional expression.
```py
x_lims, y_lims = (args.pop(f) if f in args else None) for f in ("xlim", "ylim")
```
Not sure we're not hurting readability on this one. Maybe a mix of both could do the job better.

lumicks/pylake/tests/conftest.py
```diff
@@ -115,6 +124,51 @@ def h5_file(tmpdir_factory, request):
...
+                                        '"cereal_class_version": 1, '\
+                                        '"num of pixels": 5, '\
+                                        '"pixel size (nm)": 10, '\
+                                        '"scan time (ms)": 0, '\
+                                        '"scan width (um)": 36.07468112612217'\
+                                    '}'\
+                                ']'\
+                            '}'\
+                        '}'\
+                    '}'
```
tiagobonetti: Wouldn't it be cleaner to create a JSON from dictionary and dump it as a string?

lumicks/pylake/kymo.py
```diff
@@ -2,31 +2,72 @@
...
+    name : str
+        Kymograph name
     file : lumicks.pylake.File
-        The parent file. Used to loop up channel data
+        Parent file. Contains the channel data.
+    start : int
+        Start point in the relevant info wave.
+    stop : int
+        End point in the relevant info wave.
+    json : dict
```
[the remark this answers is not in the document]
JoepVanlier: 👍 Good idea. Renamed.

lumicks/pylake/kymo.py
```diff
@@ -2,31 +2,72 @@
...
+        if not isinstance(item, slice):
+            raise IndexError("Scalar indexing is not supported, only slicing")
+        if item.step is not None:
+            raise IndexError("Slice steps are not supported")
+
+        start, stop = item.start, item.stop
+        if start is None:
+            start = self.start
+        if stop is None:
+            stop = self.stop
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, fair point. Modified. 👍 

Could also use this same contraction in slice actually.

lumicks/pylake/kymo.py
JoepVanlier: Alright. I swapped it around. Just out of curiosity, why are custom ones typically handled first?

lumicks/pylake/tests/conftest.py
JoepVanlier: Hahaha, yes... I don't know why it didn't occur to me to do it that way actually 🤔

Anyways, it starts its life as a dictionary now.

lumicks/pylake/kymo.py
JoepVanlier: That's somewhat terse. If we go with just the list comprehension, maybe I'd add a comment above it. As an alternative, I just pushed a version with a list comprehension, but with the function defined before. I'm good with either so take your pick :)

lumicks/pylake/kymo.py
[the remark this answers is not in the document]
tiagobonetti: Really like the proactive fixes :heart:

lumicks/pylake/kymo.py
tiagobonetti: I was too vague here.

The motivation was to bring the usage and the declaration of `extract_custom_properties` together not specifically to move it before `default_kwargs`. 

Usually is good to strive for minimizing the mental workload of reading code, so declarations should happen only when needed. But putting the function between the other arguments (`width_um`, `duration` and `default_kwargs`) and their usage would be also confusing, so the best solution would be to move the custom properties to after the declaration.

lumicks/pylake/kymo.py
JoepVanlier: 🥇 Oh man, that is great actually. Readable with +2, -16 is too good to resist :D

lumicks/pylake/kymo.py
JoepVanlier: Ah okay. Makes sense yes. With the new one liner with default argument you suggested I would say this block is nicely cleaned up now :)

lumicks/pylake/kymo.py
```diff
@@ -91,19 +129,26 @@ def timestamps(self) -> np.ndarray:
...
-        width_um = self.json["scan volume"]["scan axes"][0]["scan width (um)"]
+        width_um = self.metadata["scan volume"]["scan axes"][0]["scan width (um)"]
         duration = (self.stop - self.start) / 1e9
 
         default_kwargs = dict(
             extent=[0, duration, 0, width_um],
             aspect=(image.shape[0] / image.shape[1]) * (duration / width_um)
         )
 
+        x_lims, y_lims = [kwargs.pop(f, None) for f in ("xlim", "ylim")]
```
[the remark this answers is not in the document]
JoepVanlier: Fair enough. I can remove the functionality again in the ongoing regression PR. :)

PR #18, opened 2019-09-10
Ng 1929 - As a user, I would like to be able to see the keys in a group when printing either the group, or requesting its keys via keys()

JoepVanlier:
- Adds more informative __repr__ for groups (shows the members and makes navigation of the h5 tree easier).

- Adds keys() for Group.

- Adds iterator for Group.

- Adds tests for above functionality (full coverage).

Note that the tests are failing because of an actual error in the tests that was fixed in #17 . Rebasing after that PR is accepted should fix this.

### DIFF
--- a/lumicks/pylake/group.py
+++ b/lumicks/pylake/group.py
@@ -12,7 +12,6 @@ class Group:
     """
     def __init__(self, h5py_group):
         self.h5 = h5py_group
-        self._idx = -1
 
     def __getitem__(self, item):
         """Return a subgroup or a bluelake timeline channel"""
@@ -25,22 +24,15 @@ class Group:
 
     def keys(self):
         """Return group names at this level"""
-        return list(self.h5.keys())
+        return self.h5.keys()
 
     def __iter__(self):
-        self._idx = -1
-        return self
+        return self.h5.__iter__()
 
     def __next__(self):
-        """Return key names for the iterator analogously to HDF5"""
-        if self._idx >= len(self.keys()) - 1:
-            raise StopIteration
-
-        self._idx += 1
-        return self.keys()[self._idx]
+        return self.h5.__next__()
 
     def __repr__(self):
-        """Return type name and members of the group"""
-        name = self.__class__.__name__
-        members = ''.join(f"{e}, " for e in self.keys())
-        return f"{name} (members: {members[:-2]})"
+        """Return formatted representation of group keys"""
+        group_keys = ", ".join(f"'{k}'" for k in self.keys())
+        return f"{{{group_keys}}}"
--- a/lumicks/pylake/group.py
+++ b/lumicks/pylake/group.py
@@ -43,7 +43,4 @@ class Group:
         """Return type name and members of the group"""
         name = self.__class__.__name__
         members = ''.join(f"{e}, " for e in self.keys())
-        if members:
-            return f"{name} (members: {members[:-2]})"
-        else:
-            return f"{name}"
+        return f"{name} (members: {members[:-2]})"
--- a/lumicks/pylake/group.py
+++ b/lumicks/pylake/group.py
@@ -22,10 +22,6 @@ class Group:
             cls = channel_class(thing)
             return cls.from_dataset(thing)
 
-    def keys(self):
-        """Return group names at this level"""
-        return self.h5.keys()
-
     def __iter__(self):
         return self.h5.__iter__()
 
@@ -34,5 +30,5 @@ class Group:
 
     def __repr__(self):
         """Return formatted representation of group keys"""
-        group_keys = ", ".join(f"'{k}'" for k in self.keys())
+        group_keys = ", ".join(f"'{k}'" for k in self.h5)
         return f"{{{group_keys}}}"
--- a/lumicks/pylake/group.py
+++ b/lumicks/pylake/group.py
@@ -12,6 +12,7 @@ class Group:
     """
     def __init__(self, h5py_group):
         self.h5 = h5py_group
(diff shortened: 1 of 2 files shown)


lumicks/pylake/group.py
```diff
@@ -21,3 +22,25 @@ def __getitem__(self, item):
         else:
             cls = channel_class(thing)
             return cls.from_dataset(thing)
+
+    def keys(self):
+        """Return group names at this level"""
+        return list(self.h5.keys())
```
dean0x7d: I'm a bit torn about this. On one hand, it makes it more convenient for users because they can see the items right away in a notebook cell. On the other hand, it goes against Python conventions and it can cause problems in case there are lots of keys. Usually, the returned keys are just a view which can be iterated without using memory for all the elements. When you call `list()` on it, it actually iterates over all the elements and puts them in a list which can be bad if you have a lot of them.

The downside is that users need to call `list(group.keys())` in order to see everything in a notebook, but that's the same thing they need to do if they were using `h5py` directly or anything that adopts a `dict`-like interface in Python.

If the goal is specifically to make it easier for users to see the thing, the better approach if what you're doing with `__repr__` below. Note that notebooks also support rich representations, e.g `_repr_html_`: https://ipython.readthedocs.io/en/stable/config/integrating.html#rich-display

lumicks/pylake/group.py
```diff
@@ -21,3 +22,25 @@ def __getitem__(self, item):
...
+        if self._idx >= len(self.keys()) - 1:
+            raise StopIteration
+
+        self._idx += 1
+        return self.keys()[self._idx]
+
+    def __repr__(self):
+        """Return type name and members of the group"""
+        name = self.__class__.__name__
+        members = ''.join(f"{e}, " for e in self.keys())
```
dean0x7d: `members = ", ".join(str(k) for k in self.keys())` -- then you don't need to remove the trailing comma with `[:-2]`

lumicks/pylake/group.py
```diff
@@ -21,3 +22,25 @@ def __getitem__(self, item):
...
+        return f"{name} (members: {members[:-2]})"
```
dean0x7d: If we're displaying 'keys' we should not call them 'members'. We should also consider the most idiomatic way to display this data. If we consider that a group is like a `dict`, then `{'key': 'value', ...}` would make sense. `value` is a bit problematic because it might be too much information, so maybe just `{'key1', 'key2', ...}`. That's more like a `set` than a `dict` but at least we're sticking closer to the language idioms.

lumicks/pylake/tests/test_file.py
```diff
@@ -76,6 +76,25 @@ def test_properties(h5_file):
...
+def test_groups(h5_file):
+    f = pylake.File.from_h5py(h5_file)
+    if f.format_version == 1:
+        assert set(f.keys()) == set(f.h5.keys())
+        assert str(f["Force HF"]) == "Group (members: Force 1x, Force 1y)"
+
+        for x in range(0, 2):
+            t = []
+            for name in f:
+                t.append(name)
```
dean0x7d: This can be simplified to `t = [name for name in f]` or even just `list(f)`.

lumicks/pylake/tests/test_file.py
```diff
@@ -76,6 +76,25 @@ def test_properties(h5_file):
     assert f.fdcurves == {}
 
 
+def test_groups(h5_file):
+    f = pylake.File.from_h5py(h5_file)
+    if f.format_version == 1:
```
dean0x7d: Not related to this PR, but just a note for the future. We should start thinking about a better way to do file fixture versioning since the conditions are going to get out of hand.

lumicks/pylake/group.py
tiagobonetti: If I'm not mistaken (and in python, I'm often is)  this should return an object that is view on the group. Couldn't we implement _repr_ (or _repr_html_) on that view object that display all the possible keys?

dean0x7d: It may have been lost in my long comment above, but that is what I was trying to suggest :)

lumicks/pylake/group.py
```diff
@@ -21,3 +22,25 @@ def __getitem__(self, item):
         else:
             cls = channel_class(thing)
             return cls.from_dataset(thing)
+
+    def keys(self):
+        """Return group names at this level"""
+        return list(self.h5.keys())
+
+    def __iter__(self):
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, you're absolutely right.

lumicks/pylake/group.py
JoepVanlier: 👍

lumicks/pylake/group.py
JoepVanlier: Alright. {'key1', 'key2', ...} it is.

lumicks/pylake/tests/test_file.py
JoepVanlier: Ah, yes, I'll use the list comprehension.

lumicks/pylake/group.py
JoepVanlier: Fair enough.

Actually, considering that list just iterates, this thing has an iterator, and you could just construct a list with list(group) now, I don't see a need for keys anymore and removed it.

lumicks/pylake/tests/test_file.py
JoepVanlier: Yep. Noted.

PR #22, opened 2019-12-05
Add documentation CorrelatedStack and check whether stack overlaps with channel

JoepVanlier:
- Adds documentation for the new functionality.

- Add some input checking for downsampled_over. Checks whether stack overlaps with channel being downsampled, since this can give very confusing errors otherwise.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,17 @@
 # Changelog
 
+## v0.4.0 | 2019-12-05
+
+* Add calibration data as attribute of force channels.
+* Take into account slicing of channels when returning force calibration data.
+* Fixed slicing behaviour when slicing continuous time channels within a timestep.
+* Export pixel size and dwell time to TIFFs exported with pylake.
+* Implement slicing for kymographs.
+* Show keys in a group when printing the group.
+* Allow iteration over groups.
+* Fix bug which could cause axes to be inverted for images reconstructed in pylake.
+* Add functionality to correlate images recorded from cameras with timeline data.
+
 ## v0.3.1 | 2019-03-27
 
 * Loading scans and kymographs is now much faster
--- a/changelog.md
+++ b/changelog.md
@@ -2,15 +2,14 @@
 
 ## v0.4.0 | 2019-12-05
 
-* Add calibration data as attribute of force channels.
-* Take into account slicing of channels when returning force calibration data.
-* Fixed slicing behaviour when slicing continuous time channels within a timestep.
-* Export pixel size and dwell time to TIFFs exported with pylake.
-* Implement slicing for kymographs.
+* Add calibration data as attribute of force channels (see docs tutorials section: Files and Channels).
+* Fixed bug which produced shifted timestamps when slicing continuous channel with time values between two data points.
+* Export pixel size and dwell time to exported TIFFs.
+* Implement slicing for kymographs (see docs tutorials section: Kymographs).
 * Show keys in a group when printing the group.
 * Allow iteration over groups.
-* Fix bug which could cause axes to be inverted for images reconstructed in pylake.
-* Add functionality to correlate images recorded from cameras with timeline data.
+* Fix bug which could cause axes to be inverted for reconstructed images.
+* Add functionality to correlate images recorded from cameras with timeline data (see docs tutorial section: Correlated stacks).
 
 ## v0.3.1 | 2019-03-27
 
--- a/docs/tutorial/file.rst
+++ b/docs/tutorial/file.rst
@@ -148,30 +148,17 @@ Once you access the raw data, those are regular arrays which use regular array i
     channel_slice = file.force1x['1.5s':'20s']  # timestamps
     data_slice = file.force1x.data[20:40]  # indices into the array
 
-Calibration for force slices can be found by checking the calibration member::
+Calibrations
+------------
 
-    file.force1x.calibration
+Calibration information for force channels can be found by checking the calibration member. This gives a list of calibrations::
 
+    >>> print(file.force1x.calibration)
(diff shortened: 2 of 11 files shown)


changelog.md
```diff
@@ -1,5 +1,17 @@
 # Changelog
 
+## v0.4.0 | 2019-12-05
+
+* Add calibration data as attribute of force channels.
```
dean0x7d: It would be good to include an example in the tutorial.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Take into account slicing of channels when returning force calibration data.
```
dean0x7d: I don't think that this is new compared to v0.3. This is just part for the calibration feature stated above.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Fixed slicing behaviour when slicing continuous time channels within a timestep.
```
dean0x7d: I don't understand this one.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Export pixel size and dwell time to TIFFs exported with pylake.
```
dean0x7d: "with pylake" is not necessary. It's assumed since this is the pylake changelog.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Implement slicing for kymographs.
```
dean0x7d: A short tutorial entry would be nice. And to point to that from the changelog. When someone comes from a previous version and looks at the changelog, if they see an interesting feature, they would like to use it. And it can be disorienting if there is no hint of how to use the new feature.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Show keys in a group when printing the group.
+* Allow iteration over groups.
+* Fix bug which could cause axes to be inverted for images reconstructed in pylake.
```
dean0x7d: As above, no "in pylake" needed.

changelog.md
```diff
@@ -1,5 +1,17 @@
...
+* Add functionality to correlate images recorded from cameras with timeline data.
```
dean0x7d: Point to the docs entry for this. A simple, "look at the docs section titled <name here>" is enough. No need for a direct link.

docs/tutorial/file.rst
```diff
@@ -147,3 +147,31 @@ Once you access the raw data, those are regular arrays which use regular array i
...
     channel_slice = file.force1x['1.5s':'20s']  # timestamps
     data_slice = file.force1x.data[20:40]  # indices into the array
+
+Calibration for force slices can be found by checking the calibration member::
+
+    file.force1x.calibration
+
+
+CorrelatedStacks
+----------------
```
dean0x7d: I think that this is a big feature which deserves an entire page for itself.

docs/tutorial/file.rst
```diff
@@ -147,3 +147,31 @@ Once you access the raw data, those are regular arrays which use regular array i
 
     channel_slice = file.force1x['1.5s':'20s']  # timestamps
     data_slice = file.force1x.data[20:40]  # indices into the array
+
+Calibration for force slices can be found by checking the calibration member::
+
+    file.force1x.calibration
```
dean0x7d: It would be good to given an example of getting a specific parameter out of the calibration object.

docs/tutorial/file.rst
```diff
@@ -147,3 +147,31 @@ Once you access the raw data, those are regular arrays which use regular array i
...
+
+
+CorrelatedStacks
+----------------
+
+Bluelake has the ability to export videos from the camera's.
+These videos can be opened and sliced using `CorrelatedStack`::
+
+    stack = pylake.CorrelatedStack.from_file("example.tiff")  # Loading a stack.
```
dean0x7d: I think that this is outdated syntax. It doesn't use `.from_file()` anymore. Just the constructor itself.

lumicks/pylake/correlated_stack.py
```diff
@@ -68,16 +69,17 @@ class CorrelatedStack:
...
-        Loading a stack.
-            from lumicks import pylake
-            stack = pylake.CorrelatedStack.from_file("example.tiff")
+        from lumicks import pylake
 
-        Making a plot where force is correlated to images in the stack.
-            file = pylake.File("example.h5")
-            stack.plot_correlated(file.force1x)
+        # Loading a stack.
+        stack = pylake.CorrelatedStack.from_file("example.tiff")
```
dean0x7d: Same as above.

changelog.md
JoepVanlier: Removed.

changelog.md
JoepVanlier: I've rephrased it to make it more explicit ``Fixed bug which produced incorrect timestamps when slicing continuous channel with time values between two data points.``

changelog.md

changelog.md

lumicks/pylake/correlated_stack.py
JoepVanlier: 👍

docs/tutorial/file.rst
JoepVanlier: 👍

docs/tutorial/file.rst
JoepVanlier: Moved to its own page.

changelog.md

docs/tutorial/file.rst
JoepVanlier: Added 👍

changelog.md
JoepVanlier: Added.

changelog.md
JoepVanlier: Done. I made a separate page for kymos. Despite that this duplicates some stuff from scans, I think it is nice to have a head called kymographs, since their implementation is bound to diverge.

PR #28, opened 2020-02-27
Deal with unrecognized fields

JoepVanlier:
In ```group.__getitem__```, we check what type of data we are actually trying to index. Modern bluelake files have a "kind" field for channel data, but there's a fallback for older versions which don't have this field, which returns Continuous.

Not all datasets are channel data however, and when it gets called on one of these, the construction of Continuous fails providing an error that to the user has little or nothing to do with what they were trying to do. Instead, we now throw an exception when the user accesses something pylake isn't built to understand.

In this PR:
- Provides API for markers.
- Allows printing property fields available in the h5 in human readable format.
- When data access is attempted through the fields directly (such as h5file["Calibration"]) the user is redirected to the appropriate API.

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -354,7 +354,14 @@ def channel_class(dset):
         else:
             raise RuntimeError("Unknown channel kind " + str(kind))
     elif dset.dtype.fields is None:
-        # For compatibility with Bluelake HDF5 files v1
-        return Continuous
+        # Version 1 Bluelake HDF5 files do not have a kind field which indicates the type of data they store. These
+        # older files typically contain either Continuous or TimeSeries channel data. Continuous channel data contains
+        # the attribute "Sample rate (Hz)". Newer Bluelake HDF5 files also have fields which do not have the kind
+        # attribute, but which are not Continuous data. Direct access to these fields is not supported as they are
+        # typically accessed through dedicated API classes.
+        if "Sample rate (Hz)" in dset.attrs.keys():
+            return Continuous
+        else:
+            raise IndexError("Direct access to this field is not supported.")
     else:
         return TimeSeries
--- a/lumicks/pylake/file.py
+++ b/lumicks/pylake/file.py
@@ -10,6 +10,7 @@ from .group import Group
 from .kymo import Kymo
 from .point_scan import PointScan
 from .scan import Scan
+from .marker import Marker
 
 __all__ = ["File"]
 
@@ -182,3 +183,7 @@ class File(Group, Force, DownsampledFD, PhotonCounts, PhotonTimeTags):
     @property
     def fdcurves(self) -> Dict[str, FDCurve]:
         return self._get_object_dictionary("FD Curve", FDCurve)
+
+    @property
+    def markers(self) -> Dict[str, Marker]:
+        return self._get_object_dictionary("Marker", Marker)
--- a/lumicks/pylake/file.py
+++ b/lumicks/pylake/file.py
@@ -160,17 +160,18 @@ class File(Group, Force, DownsampledFD, PhotonCounts, PhotonTimeTags):
     def _get_photon_time_tags(self, name):
         return TimeTags.from_dataset(self.h5["Photon Time Tags"][name], "Photon time tags")
 
+    def _get_object_dictionary(self, field, cls):
+        if field not in self.h5:
+            return dict()
+        return {name: cls.from_dataset(dset, self) for name, dset in self.h5[field].items()}
+
     @property
     def kymos(self) -> Dict[str, Kymo]:
-        if "Kymograph" not in self.h5:
-            return dict()
-        return {name: Kymo.from_dataset(dset, self) for name, dset in self.h5["Kymograph"].items()}
+        return self._get_object_dictionary("Kymograph", Kymo)
 
     @property
     def point_scans(self) -> Dict[str, Scan]:
-        if "Point Scan" not in self.h5:
-            return dict()
-        return {name: PointScan(dset, self) for name, dset in self.h5["Point Scan"].items()}
+        return self._get_object_dictionary("Point Scan", PointScan)
 
     @property
     def scans(self) -> Dict[str, Scan]:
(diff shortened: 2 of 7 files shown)


lumicks/pylake/file.py
```diff
@@ -48,6 +49,12 @@ def _check_file_format(self):
         if ff_version not in File.SUPPORTED_FILE_FORMAT_VERSIONS:
             raise Exception(f"Unsupported Bluelake file format version {ff_version}")
 
+        self.blacklist = {"Calibration": "file.force1x.calibration",
+                          "Marker": "file.markers",
+                          "FD Curve": "file.fdcurves",
+                          "Kymograph": "file.kymos",
+                          "Scan": "file.scans"}
```
dean0x7d: `blacklist` doesn't really fit as a name for the use in the printing function below.

lumicks/pylake/file.py
```diff
@@ -104,6 +111,8 @@ def print_dataset(dset, name, indent):
             r += f"{space}- Size: {dset.size}\n"
             return r
 
+        blacklist = self.blacklist
```
dean0x7d: I don't think that this is needed. You can just use `self.blacklist` in the single use of the variable below.

lumicks/pylake/file.py
```diff
@@ -114,10 +123,24 @@ def print_group(group, name="", indent=-2):
...
+
+        def print_force(field_name):
+            field = getattr(self, field_name)
+            calibration = '  .calibration\n' if field.calibration else ''
+            return f'.{field_name}\n{calibration}' if field else ''
+
+        return print_attributes(self.h5) + "\n" + print_group(self.h5) + \
+            ''.join((print_dicts(field) for field in ['kymos', 'fdcurves', 'scans', 'markers'])) + "\n" + \
+            ''.join((print_force(field) for field in ['force1x', 'force1y', 'force2x', 'force2y',
+                                                      'force3x', 'force3y', 'force4x', 'force4y']))
```
dean0x7d: There's some unnecessary duplication here. The `blacklist` already knows about the marker fields. It would be great if you could put all the into a single mapping so that we can just add a field there in order to get the correct printing and `__getitem__` behavior in the future.

lumicks/pylake/file.py
```diff
@@ -160,26 +183,29 @@ def _get_photon_count(self, name):
     def _get_photon_time_tags(self, name):
         return TimeTags.from_dataset(self.h5["Photon Time Tags"][name], "Photon time tags")
 
+    def _get_object_dictionary(self, field, cls):
+        if field not in self.h5:
+            return dict()
+        return {name: cls.from_dataset(dset, self) for name, dset in self.h5[field].items()}
```
dean0x7d: 👍

lumicks/pylake/file.py
```diff
@@ -160,26 +183,29 @@ def _get_photon_count(self, name):
...
-        return {name: Scan.from_dataset(dset, self) for name, dset in self.h5["Scan"].items()}
+        return self._get_object_dictionary("Scan", Scan)
 
     @property
     def fdcurves(self) -> Dict[str, FDCurve]:
-        if "FD Curve" not in self.h5:
+        return self._get_object_dictionary("FD Curve", FDCurve)
+
+    @property
+    def markers(self) -> Dict[str, dict]:
```
dean0x7d:
```suggestion
    def markers(self) -> Dict[str, Marker]:
```

lumicks/pylake/file.py
```diff
@@ -160,26 +183,29 @@ def _get_photon_count(self, name):
...
+        if "Marker" not in self.h5:
             return dict()
-        return {name: FDCurve.from_dset(dset, self) for name, dset in self.h5["FD Curve"].items()}
+        return {name: Marker.from_dataset(dset) for name, dset in self.h5["Marker"].items()}
```
dean0x7d: It would be nice to also make the `Marker.from_dataset()` consistent with the rest and use `_get_object_dictionary()`.

lumicks/pylake/channel.py
```diff
@@ -355,6 +355,9 @@ def channel_class(dset):
             raise RuntimeError("Unknown channel kind " + str(kind))
     elif dset.dtype.fields is None:
         # For compatibility with Bluelake HDF5 files v1
-        return Continuous
+        if "Sample rate (Hz)" in dset.attrs.keys():
+            return Continuous
+        else:
+            raise IndexError("Direct access to this field is not supported.")
```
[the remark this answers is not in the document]

lumicks/pylake/file.py

lumicks/pylake/file.py
JoepVanlier: Yeah. 👍

lumicks/pylake/file.py
JoepVanlier: Agree. Renamed it to ```redirect_list``` since we're basically redirecting to an API class.

lumicks/pylake/file.py
JoepVanlier: Good point. I will take the redirected ones from the list directly. I will also swap the dict out for an ordered one to ensure order stability in the print statement (as this is also explicitly tested). I know that in 3.7 dictionaries are ordered by the language standard, but in 3.6 it still seems to be an implementational detail (even though they are ordered for CPython), hence probably best to be explicit about order preservation.

The force fields are a bit special, since we don't really need to discourage people from using those directly. I think for now, having these two separate cases is ok force fields versus the API classes? If we get more complicated redirects, I can expand the logic.

lumicks/pylake/file.py
JoepVanlier: It did cross my mind, but having a dummy argument felt very icky. I wasn't 100% sure which one was the road of least groan in this case:

```    @staticmethod
    def from_dataset(h5py_dset, file):
        """
        Construct Marker class from dataset.

        Parameters
        ----------
        h5py_dset : h5py.Dataset
            The original HDF5 dataset containing Marker information
        file : lumicks.pylake.File
            The parent file. Unused, but part of the interface.
        """
        del file
        return Marker(h5py_dset.attrs)```

dean0x7d: I would recommend that you add a `.file` attribute to `Marker`. That makes it consistent with the other types. Even if it's not used internally, it's nice to have as an attribute so that you can always trace back where a certain item came from.

PR #29, opened 2020-03-20
Preparation for release v.0.4.1 

JoepVanlier:
This PR aims to get pylake ready for release v0.4.1

- Removes matplotlib < 3 requirement. Despite conda-forge's recommendation to go with matplotlib-base, it seems more sensible to stick with matplotlib > 2.2, since matplotlib-base only starts at 3.0 and > 3.0 may be a harder constraint than base (leading to more difficult environment solving).
- Modifies the documentation to be in accordance with the current output of print.
- Updates the changelog.
- Implements a plotting bugfix on CorrelatedStack.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.4.1 | 2020-03-23
+* Drop `matplotlib` < 3 requirement.
+* Add functionality which redirects users to API when accessing particular fields, e.g. accessing `file["FD curve"]` will throw an error and redirect users to use `file.fdcurves`.
+* Add API for markers, i.e. `file.markers` returns a dictionary of markers (see docs tutorials section: Files and Channels).
+* Bugfix `CorrelatedStack.plot()` which resulted in the function throwing an error rather than showing a single frame.
+
 ## v0.4.0 | 2020-01-21
 
 * Add calibration data as attribute of force channels (see docs tutorials section: Files and Channels).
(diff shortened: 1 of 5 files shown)


setup.py
```diff
@@ -52,6 +52,6 @@ def read(filename):
     packages=PEP420PackageFinder.find(include=["lumicks.*"]),
     python_requires='>=3.6',
     install_requires=['pytest>=3.5, <4.0', 'h5py>=2.9, <3.0', 'numpy>=1.14, <2',
```
dean0x7d: I think that we can drop the `pytest` upper bound as well. It's up to v5 at this point and lots of packages use it.

changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.4.1 | 2020-03-23
+* Drop matplotlib < 3 requirement.
```
dean0x7d: Style 🚓: new line after section, `matplotlib` package name in backticks.

changelog.md
```diff
@@ -1,5 +1,11 @@
...
+* Add functionality which redirects users to API when accessing particular fields.
```
dean0x7d: Not sure this is going to clear to readers. An example would be nice: accessing `file["FD curves"]` will throw an error and redirect users to use `file.fd_curves`.

changelog.md
```diff
@@ -1,5 +1,11 @@
...
+* Add API for markers.
```
dean0x7d: Definitely needs a very short example.

changelog.md
```diff
@@ -1,5 +1,11 @@
...
+* Bugfix CorrelatedStack.plot().
```
dean0x7d: What did the bug prevent people from doing? If someone had this bug and they went to the changelog, they should be able to recognize that it's been fixed.

setup.py
JoepVanlier: 👍

changelog.md

changelog.md
JoepVanlier: Changed the line to: 
* Add functionality which redirects users to API when accessing particular fields, e.g. accessing `file["FD curve"]` will throw an error and redirect users to use `file.fdcurves`.

changelog.md
JoepVanlier: Expanded changelog line and added short tutorial in the docs.
* Add API for markers, i.e. `file.markers` returns a dictionary of markers (see docs tutorials section: Files and Channels).

changelog.md
JoepVanlier: * Bugfix CorrelatedStack.plot() which resulted in the function throwing an error rather than showing a single frame.

This ties in with a bigger issue, which is that it would be nice to at least test whether graphical functions work without throwing. This would prevent stuff like this. I will investigate this, but don't really want to do this as part of this PR.

PR #30, opened 2020-04-01
reconstruct_image: make a copy when in-place resize fails

JoepVanlier:
Three try catches :(.

I could try to put this in a function, but then we wouldn't be able to use try/catch, since passing the reference into the function would bump the reference count by one. There isn't an equivalent of move semantics, is there? :)

We could consider switching to concatenate altogether (since it doesn't seem to have a huge performance impact), but I'm not sure how much memory it takes compared to an in-place resize.

Time measurement (performed in randomized sequential order):
In place: 1.5232401490211487 +- 0.050069897091749985 (N=20)
Concatenate: 1.6158243656158446 +- 0.0414880556443715 (N=20)
np.resize: 2.3809279084205626 +- 0.06063061274418485 (N=20)

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -146,17 +146,18 @@ def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, red
         """Round up `size` to the nearest multiple of `n`"""
         return int(math.ceil(size / n)) * n
 
-    data = data[valid_idx]
-    data.resize(round_up(data.size, pixel_size))
-
-    pixels = reduce(data.reshape(-1, pixel_size), axis=1)
+    resized_data = np.zeros(round_up(infowave.size, pixel_size))
+    resized_data[:infowave.size] = data[valid_idx]
+    pixels = reduce(resized_data.reshape(-1, pixel_size), axis=1)
 
     if lines_per_frame is None:
-        pixels.resize(round_up(pixels.size, pixels_per_line))
-        return pixels.reshape(-1, pixels_per_line)
+        resized_pixels = np.zeros(round_up(pixels.size, pixels_per_line))
+        resized_pixels[:pixels.size] = pixels
+        return resized_pixels.reshape(-1, pixels_per_line)
     else:
-        pixels.resize(round_up(pixels.size, pixels_per_line * lines_per_frame))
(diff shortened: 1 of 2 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -147,15 +147,28 @@ def round_up(size, n):
         return int(math.ceil(size / n)) * n
 
     data = data[valid_idx]
-    data.resize(round_up(data.size, pixel_size))
```
dean0x7d, line 150: Maybe get a win-win situation here and avoid the `try` block and while even increasing performance. Something along the lines of:
```python
data[:infowave.size] = data[valid_idx]
data = data[:round_up(data.size, pixel_size)]
```
So we only shrink the data once instead of shrinking it and growing it. This would also avoid `.resize()` and the debugger reference issues. The above is just a quick idea I had -- there could be some major flaw that I haven't considered.

JoepVanlier: I like it, but there's a potential risk here. This means that we truly modify data in-place, which means that the outside behaviour changes (as we are truly modifying the incoming array). Before the initial `data = data[valid_idx]` made a copy.

It's a bit of a concern, since it could cause some difficult to track down bugs. If we do it, we should document it very clearly. I did a quick scan of all the other functions using reconstruct_image, and for now, it seems that the behaviours wouldn't change as the data for this function typically comes from _get_photon_count (both for timestamps and counts), which just makes a Continuous from the h5, which always creates a new numpy array. However, if we were to cache this in the future, we'd get in trouble though.

Maybe slightly less sexy, but:
```
    resized_data = np.zeros(round_up(infowave.size, pixel_size))
    resized_data[:infowave.size] = data[valid_idx]
    pixels = reduce(resized_data.reshape(-1, pixel_size), axis=1)
```

Would still make a copy of the input.

I don't know, what do you think? Speed or safety?

Another thing, but minor, data.size should be infowave.size in the round_up.

dean0x7d: No, definitely don't modify the input data. I like your suggestion. After that, the other two `.resize()` calls might be omitted the same way.

JoepVanlier: It felt a bit like a C-api, modifying the input ☠️ 

Alright, I changed it to just pre-allocate now.

PR #32, opened 2020-05-20
Return int64 for time from channel.downsampled_over

JoepVanlier:
Timestamps returned from downsampled_over weren't defined as int64. This PR fixes that.

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -121,7 +121,7 @@ class Slice:
         if where != 'center' and where != 'left':
             raise ValueError("Invalid argument for where. Valid options are center and left")
 
-        t = np.zeros(len(range_list))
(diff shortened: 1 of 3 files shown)


lumicks/pylake/channel.py
```diff
@@ -121,7 +121,7 @@ def downsampled_over(self, range_list, reduce=np.mean, where='center'):
         if where != 'center' and where != 'left':
             raise ValueError("Invalid argument for where. Valid options are center and left")
 
-        t = np.zeros(len(range_list))
+        t = np.zeros(len(range_list), dtype=np.int64)
```
dean0x7d, line 124: This might make it more obvious that the aim is to match the timestamp `dtype`:
```suggestion
        t = np.zeros(len(range_list), dtype=self._src.timestamps.dtype)
```

JoepVanlier: Good point, thanks!

PR #33, opened 2020-06-03
FdFit 1.0

JoepVanlier:
This pull requests adds F, d fitting capabilities to pylake. This framework allows global fitting of multiple datasets simultaneously, keeping some parameters the same between datasets and varying others. It also includes several polymer models that are typically used in literature. Where possible, models were provided in analytical form.

To improve performance, model Jacobians are provided and propagated via the chain rule for composite and inverted models. For the inverse models where the inverse is not available as an analytical function, the inversion was performed using least_squares (again using the analytical jacobian if supplied), or, in the case of the tWLC interpolation. 

Tests were added for each of the models as well as the complete fitting infrastructure.

It also adds API documentation, a tutorial and two examples.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,7 @@
 # Changelog
 
-## v0.4.2 | t.b.d.
+## v0.5.0 | 2020-06-08
+* Added F, d Fitting functionality (beta, see docs tutorial section `Fd Fitting` and examples `Twistable Worm-Like-Chain Fitting` and `RecA Fd Fitting`).
 * Fixed an issue which prevented images from being reconstructed when a debugger is attached. Problem resided in `reconstruct_image` which threw an exception when attempting to resize a `numpy` array while the debugger was holding a reference to it.
 * Fixed bug that lead to timestamps becoming floating point values when using `channel.downsampled_over`.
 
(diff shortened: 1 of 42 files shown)


JoepVanlier:
> Wow! You really went above and beyond with the history cleanup. Looks excellent, as does then fitting functionality!
> 
> I only have two comments, but consider them completely optional:
> 
> * The `rename to lk` commit is also secretly introducing the entire fitting module :) (The addition to `__init__.py`, changelog and docs table of contents). Consider making that a separate commit if possible. The tutorial ToC definitely belongs with the tutorial commit. That can be done with interactive rebase + edit, but it's not straightforward git so as I said: completely optional. The PR is fine to merge as-is.
> * All the commit messages mention `fdfit` but in the code the module is called `fitting`. I haven't gone through the docs looking for this but it would be good to keep it consistent. Again, optional. But it would still be good to double-check naming consistency in the docs.

Thanks man! I did the last few fixups. I do agree it was better to split those out. I set the changelog date for v0.5.0 to Monday. I'll merge it now then and then do the release procedure coming Monday.

Also, thanks a lot for all your work on the review! 👍

PR #34, opened 2020-06-09
Add literature page

JoepVanlier:
This PR adds a literature/references page to rtd.

It adds bibtex as a dependency for building the docs (tested on rtd).

Note that the page is called `zref` since it has to be the last page parsed (requirement from sphinxcontrib-bibtex).

I piggybacked a fixup in the docs for `Fit.plot()` since its a small fix and also affects docs.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## v0.5.1 | t.b.d.
 * Add literature page to the documentation.
+* Fix docstring for `Fit.plot()`.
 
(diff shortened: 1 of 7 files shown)


JoepVanlier:
> Looks really great! (Esp. neat that the references are stored readily as BibTeX citations, I didn't know you could do that with Sphinx!)

I didn't know that either until yesterday, but it's pretty convenient that that extension exists. :)

PR #37, opened 2020-08-06
scan: fix reconstruction bug when fast axis in y direction

JoepVanlier:
Fixes issue in [LAKE-95](https://lumicks.atlassian.net/browse/LAKE-95) pertaining to the reconstruction of scans.

The reconstruction of images with a fast axis in y-direction failed due to changes made in https://github.com/lumicks/pylake/pull/19 . The axis flip should have been implemented post-reconstruction as data is acquired in the order specified in the json (fast axis being the first axis in the list).

This PR fixes the issue as well as hardens some of the scan tests (by including multi-frame scans as this almost caused a regression while doing the fix).

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -81,6 +81,14 @@ class Kymo(PhotonCounts, ExcitationLaserPower):
         return first(self.json["scan volume"]["scan axes"], lambda x: x["axis"] == axis)
 
     @property
+    def _fast_axis_metadata(self):
+        return self.json["scan volume"]["scan axes"][0]
+
+    @property
+    def fast_axis(self):
+        return "X" if self._fast_axis_metadata["axis"] == 0 else "Y"
+
+    @property
     def has_fluorescence(self) -> bool:
         return self.json["fluorescence"]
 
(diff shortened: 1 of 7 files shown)


lumicks/pylake/scan.py
```diff
@@ -43,18 +42,32 @@ def num_frames(self):
 
     @property
     def lines_per_frame(self):
-        return self._get_axis_metadata(1)["num of pixels"]
+        return self.json["scan volume"]["scan axes"][1]["num of pixels"]
+
+    def _flip_axis(self, data):
```
lpingas, line 47: Maybe this could be called `_flip_if_needed` or something else that implies that the flip only happens sometimes.

Or you could also go with something like `_to_spatial` as you could then explain that the reconstruction returns a `[slow, fast]` structure and we want to return a `[Y, X]` one.

JoepVanlier: `to_spatial` sounds good to me. I improved the docstring a bit as well.

lumicks/pylake/kymo.py
```diff
@@ -80,6 +80,10 @@ def _get_photon_count(self, name):
     def _get_axis_metadata(self, axis=0):
         return first(self.json["scan volume"]["scan axes"], lambda x: x["axis"] == axis)
 
+    @property
+    def fast_axis(self):
+        return self.json["scan volume"]["scan axes"][0]["axis"]
+
```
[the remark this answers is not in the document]
JoepVanlier: Yes on all accounts.

PR #38, opened 2020-08-07
Fd widget

JoepVanlier:
This PR adds a widget for F,d curves.

It can be used as follows:
```python
   file = lk.File("file.h5")
   fdcurves = file.fdcurves
   selector = lk.FdRangeSelector(fdcurves)
```

After selection, you can get the timestamps or fd curve segments that you selected.

```python
    for t_start, t_stop in selector.ranges["Fd pull #6"]:
        plt.figure()
        plt.plot(fdcurves["Fd pull #6"].f[t_start:t_stop].data)
```

```python
    for fdcurve in selector.fdcurves["Fd pull #6"]:
        plt.figure()
        fdcurve.plot_scatter()
```

Note that it also adds a slice operator to `FDCurve` (gets forwarded to the `f` and `d` caches, making sure to generate a deepcopy). Also added tests for this.

callback_test

You can find a build of the docs here (so you don't have to build it):
https://lumicks-pylake.readthedocs.io/en/fd_widget/tutorial/nbwidgets.html

### DIFF
--- a/lumicks/pylake/nb_widgets/fd_selector.py
+++ b/lumicks/pylake/nb_widgets/fd_selector.py
@@ -4,8 +4,8 @@ import time
 
 
 class FdRangeSelectorWidget:
-    def __init__(self, fd_curve, ax=None, show=True):
-        self._ax = ax if ax else plt.axes(label=f"lk_widget_{time.time()}")
+    def __init__(self, fd_curve, axes=None, show=True):
+        self._axes = axes if axes else plt.axes(label=f"lk_widget_{time.time()}")
         self.fd_curve = fd_curve
         self.time = self.fd_curve.f.timestamps
         self.connection_id = None
@@ -14,7 +14,7 @@ class FdRangeSelectorWidget:
         self.open = show
 
         if show:
-            self._ax.get_figure().canvas.mpl_connect('close_event', self.close)
+            self._axes.get_figure().canvas.mpl_connect('close_event', self.close)
             self.update_plot()
 
     def close(self, event):
@@ -28,6 +28,8 @@ class FdRangeSelectorWidget:
         return (timestamp - self.time[0]) / 1e9
 
     def _add_point(self, fd_timestamp):
+        """Adds a point intended to be a range endpoint. Once two points are selected the new range selection is
+        committed and the range appended."""
         self.current_range.append(fd_timestamp)
 
         # We have a range selected. Append it to the list.
@@ -38,9 +40,10 @@ class FdRangeSelectorWidget:
 
         # Draw a vertical line for some immediate visual feedback
         plt.axvline(self.to_seconds(fd_timestamp))
-        self._ax.get_figure().canvas.draw()
+        self._axes.get_figure().canvas.draw()
 
-    def _remove_point(self, fd_timestamp):
+    def _remove_range(self, fd_timestamp):
+        """Removes a range if the provided timestamp falls inside it."""
         in_range = [start < fd_timestamp < stop for start, stop in self.ranges]
         selected = np.nonzero(in_range)[0]
         if len(selected) > 0:
@@ -50,7 +53,7 @@ class FdRangeSelectorWidget:
         self.update_plot()
 
     def handle_button_event(self, event):
-        if event.inaxes != self._ax:
+        if event.inaxes != self._axes:
             return
 
         # Check if we aren't using a figure widget function like zoom.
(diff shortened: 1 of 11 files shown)


lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -0,0 +1,136 @@
+import matplotlib.pyplot as plt
+import numpy as np
+import time
+
+
+class FdRangeSelectorWidget:
+    def __init__(self, fd_curve, ax=None, show=True):
+        self._ax = ax if ax else plt.axes(label=f"lk_widget_{time.time()}")
```
tiagobonetti, line 8: Is `_ax` supposed to mean axes? If so maybe we should consider just renaming it to `_axes`

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -0,0 +1,136 @@
...
+        self._ax.get_figure().canvas.draw()
+
+    def _remove_point(self, fd_timestamp):
+        in_range = [start < fd_timestamp < stop for start, stop in self.ranges]
+        selected = np.nonzero(in_range)[0]
+        if len(selected) > 0:
+            selected_ranges = [self.ranges[x] for x in selected]
+            smallest_segment = np.argmin([stop - start for start, stop in selected_ranges])
+            self.ranges.pop(selected[smallest_segment])
+        self.update_plot()
```
tiagobonetti, line 50: At first glance, I don't understand this function.  
It seems to remove the smallest range that intersects with the point we pass in, but the name implies removing a point not a range, so I'm not sure what is intended.

I think either a comment or maybe renaming the function would help.

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -0,0 +1,136 @@
...
+            self._remove_point(event_timestamp)
+
+    def disconnect_click_callback(self):
+        """Disconnects the clicking callback if it exists"""
+        if self.connection_id:
+            self._ax.get_figure().canvas.mpl_disconnect(self.connection_id)
+
+    def connect_click_callback(self):
+        """Connects a callback for clicking the axes"""
+        self.connection_id = self._ax.get_figure().canvas.mpl_connect('button_press_event', self.handle_button_event)
```
tiagobonetti, line 77: nitpicking 9000: maybe reorder those two functions so connect show up first, it's minimal but I can say it crossed my mind for a second: "Disconnect? I didn't even know we were connected to anything."

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -0,0 +1,136 @@
...
+    def __init__(self, fd_curves):
+        try:
+            import ipywidgets
+        except ImportError:
+            raise RuntimeError("This widget requires ipywidgets to be installed to work. Please install it.")
+
+        if "ipympl" not in plt.get_backend():
+            raise RuntimeError(("Please enable the widgets backend for this plot to work. You can do this by invoking "
+                                "%matplotlib widget. Please note that you may have to restart the notebook kernel for "
+                                "this to work."))
```
tiagobonetti, line 110: Great error message.

tiagobonetti (commented):
Looks good to me.
Simple enough to understand even with little python experience.  
I added some minor comments where I had a harder time understanding stuff.

I won't approve because I think my seal of approval is not good enough to approve a python PR but it deserves a `Honk!`.

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: Yes. 👍
Done!

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: Good point. Done.

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: Thanks. Yes, I agree the function name was a bit of a red :fish:. Renamed to `remove_range` to avoid the confusion. Also added a little docstring: ```"""Removes a range if the provided timestamp falls inside it."""```

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: Thanks! Hopefully pre-empting some support there :D

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -0,0 +1,139 @@
...
+
+        plt.figure()
+        self.axes = plt.axes()
+        self.active_plot = None
+        self.selectors = {key: FdRangeSelectorWidget(curve, self.axes, show=False) for key, curve in fd_curves.items()}
+        self.keys = [key for key, curve in fd_curves.items()]
+        ipywidgets.interact(self.update_plot,
+                            curve_idx=ipywidgets.IntSlider(min=0, max=len(fd_curves) - 1, step=1, value=0))
```
lpingas, line 121: As we've discussed, indeed, it seems that it should be simple to modify this to a Combobox with: 
```python
interact(self.update_plot, curve=self.keys);
```
and: 
```python
def update_plot(self, curve):
    if self.active_plot:
        self.active_plot.disconnect_click_callback()

    self.active_plot = self.selectors[curve]
    self.active_plot.update_plot()
```
This would make the widget easier to use.

lpingas (approved):
Looks great! Just make the change we've discussed about the curve selection and adapt the docs with it. The Combobox should provide a nicer experience for users.

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: Definitely! Will do.

PR #39, opened 2020-08-18
release: readying for releasing v0.6

JoepVanlier:
Updates changelog with the date and extended description of the fix (with warnings). Bumps version number and copyright notice.


changelog.md
```diff
@@ -1,12 +1,13 @@
 # Changelog
 
-## v0.6.0 | t.b.d.
+## v0.6.0 | 2020-08-18
+* Plot and return images and timestamps for scans using physical coordinate system rather than fast and slow scanning axis. **Note that this is a potentially breaking change for scans with the fast axis in y direction!**
```
dean0x7d, line 4: Leave a blank line after section titles.

changelog.md
```diff
@@ -1,12 +1,13 @@
...
+* `Scan.pixels_per_line` and `Kymo.pixels_per_line` also return the number of pixels along the fast axis now. **Note that this is a potentially breaking change for scans/kymos with the fast axis in y direction!**
+* Verify starting timestamp when reconstructing `Kymo` or `Scan`. In those cases, scans cannot be reliably reconstructed and an error is thrown. For kymos, the first (partial) line is omitted and a warning is issued. **Note that this is a potentially breaking change for scans and kymos where the scan/kymo was initiated before the exported time window!**
```
dean0x7d, line 6: `potentially breaking` sounds scary, but mostly because it's vague. Instead of `potentially breaking` just say `breaking` but specify the exact case when it does break something and how to resolve it. `potentially breaking` leaves a lot of room for uncertainty and will definitely prevent people from updating to the new version to avoid breaking their workflow. Specifying the exact scenario when the break happens and pointing out that this is not a common one at all should put a lot of readers at ease and also be more helpful to people who are actually affected by the change.

dean0x7d:
Feel free to include code examples in the changelog. This document should be as helpful as possible to people who are upgrading to a new version of `pylake`.

changelog.md
JoepVanlier: Done.

changelog.md
JoepVanlier: Good point. Removed potentially.

JoepVanlier:
> Feel free to include code examples in the changelog. This document should be as helpful as possible to people who are upgrading to a new version of `pylake`.

Do you have something specific in mind here? The API didn't change so you'd use the things in the same way, it's just that now they do what it says on the tin. I'll refer to the docs tutorial section on `Confocal Images` and `Kymographs`.

changelog.md
```diff
@@ -1,19 +1,23 @@
 # Changelog
 
-## v0.6.0 | t.b.d.
+## v0.6.0 | 2020-08-18
+
+* Plot and return images and timestamps for scans using physical coordinate system rather than fast and slow scanning axis. See docs tutorial section on `Confocal images` and `Kymographs` on how to use the API. **Note that this is a breaking change for scans with the fast axis in y direction!**
```
dean0x7d, line 5: What's the nature of the breaking change? Are users going to suddenly get wrong results or were their old results wrong? What do they do?

changelog.md
```diff
@@ -1,19 +1,23 @@
...
+* `Scan.pixels_per_line` and `Kymo.pixels_per_line` return the number of pixels along the fast axis now. **Note that this is a breaking change for scans/kymos with the fast axis in y direction!**
```
dean0x7d, line 6: I feel like this should be rolled into the previous point.

changelog.md
```diff
@@ -1,19 +1,23 @@
...
+* Verify starting timestamp when reconstructing `Kymo` or `Scan`. In those cases, scans cannot be reliably reconstructed and an error is thrown. For kymos, the first (partial) line is omitted and a warning is issued. **Note that this is a breaking change for scans and kymos where the scan/kymo was initiated before the exported time window!**
```
dean0x7d, line 7: I'm not sure that this is really a breaking change. The old scans were clearly wrong. Unless I'm missing something, I'd remove the breaking notice here and just mention that before this fix some partial scans may be been wrong.

changelog.md
JoepVanlier: Ah, okay. Yeah, I see the lack of explanation now, thanks. I've tried to make it more explicit.

changelog.md
JoepVanlier: Done.

changelog.md
JoepVanlier: Yes, good point. I feel the warning could still use a bold though in case someone suddenly gets exceptions and wants to know why :)

PR #41, opened 2020-08-26
fdfit: Added simplified Marko Siggia model for protein studies

JoepVanlier:
For protein studies (and the notebook we want to release), the simple (non-extensible) version of the Marko Siggia model is often used to model the protein state. It's basically the Marko Siggia model without the stretch modulus (which is equivalent to a stretch modulus of infinity).

We already have both forms of the full Marko Siggia model in the FdFitter. By both forms, I mean one version of the model with force as independent parameter, and one version of the model with distance as independent parameter. While it's technically possible to just set the stretch modulus to a very large value, this can lead to numerical problems during optimization and is just a generally icky thing to do.

We also already have the simplified Marko Siggia model in the FdFitter.

However, when working with composite models, we would like to do something like `(lk.odijk() + lk.inverted_marko()).invert()` and it would be really bad form (and result in terrible performance) to invert the simplified Marko Siggia model twice when there's an analytic form for its inverse.

Hence this PR introduces the inverted Marko Siggia model.

### DIFF
--- a/lumicks/pylake/fitting/detail/model_implementation.py
+++ b/lumicks/pylake/fitting/detail/model_implementation.py
@@ -2,6 +2,7 @@ from ..parameters import Parameter
 from .derivative_manipulation import invert_function, invert_jacobian, invert_function_interpolation
 from .utilities import latex_sqrt, latex_frac, solve_formatter, solve_formatter_tex
 import numpy as np
+import warnings
 
 
 class Defaults:
@@ -56,6 +57,9 @@ def Marko_Siggia(d, Lp, Lc, kT):
         1. J. Marko, E. D. Siggia. Stretching dna., Macromolecules 28.26,
         8759-8770 (1995).
     """
+    if np.any(d > Lc):
+        warnings.warn("Marko Siggia model is only defined properly up to the contour length (d = Lc)", RuntimeWarning)
+
     d_div_Lc = d / Lc
     return (kT/Lp) * (.25 * (1.0-d_div_Lc)**(-2) + d_div_Lc - .25)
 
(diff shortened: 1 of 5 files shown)


lumicks/pylake/fitting/detail/model_implementation.py
```diff
@@ -732,6 +736,66 @@ def marko_sigga_ewlc_solve_distance_equation_tex(f, Lp, Lc, St, kT=4.11):
                                f"{dLc} - {FSt}", "d", lhs)
 
 
+def inverted_simplified_marko_sigga_coefficients(f, Lp, Lc, kT):
```
lpingas, line 739: The implementation functions have `sigga` as a typo.

lumicks/pylake/fitting/models.py
```diff
@@ -169,6 +170,43 @@ def marko_siggia_simplified(name):
...
+    name : str
+        Name for the model. This name will be prefixed to the model parameter names.
+    """
+    from .model import Model
+    from .detail.model_implementation import (
+        inverted_simplified_marko_sigga,
+        inverted_simplified_marko_sigga_jac,
+        inverted_simplified_marko_sigga_derivative,
+        inverted_simplified_marko_sigga_equation,
+        inverted_simplified_marko_sigga_equation_tex,
```
lpingas, line 192: I feel a little bit confused about the naming of the detail functions for the Marko Siggia models: We have `marko_sigga_ewlc_*`,  `Marko_Siggia_*` and now `inverted_simplified_marko_siggia_*`.

I suggest following the public API names, which means:
- Adding an extra commit to rename the forward simplified functions from `Marko_Siggia_*` to  `marko_siggia_simplified_*`.

- Add a commit to fix the typo in `marko_sigga_ewlc_*` functions

- Change the new ones from `inverted_simplified_marko_siggia_*` to `inverted_marko_siggia_simplified_*`.

lpingas (commented):
Nice addition! Consider the comment about the implementation function names.

lumicks/pylake/fitting/detail/model_implementation.py
JoepVanlier: Whoops! Fixed!

lumicks/pylake/fitting/models.py
JoepVanlier: Yes, this is a legacy remnant. Marko_Siggia was the first model.

I've applied your suggestions :)

lumicks/pylake/fitting/models.py
```diff
@@ -134,7 +134,8 @@ def marko_siggia_ewlc_distance(name):
 
 
 def marko_siggia_simplified(name):
-    """Markov Siggia's Worm-like Chain model based on only entropic contributions (valid for F << 10 pN).
+    """Markov Siggia's Worm-like Chain model based on only entropic contributions (valid for F << 10 pN). This model
```
lpingas, line 137: typo: `Markov`

lumicks/pylake/fitting/models.py
```diff
@@ -147,22 +148,59 @@ def marko_siggia_simplified(name):
...
+        eqn=marko_siggia_simplified_equation,
+        eqn_tex=marko_siggia_simplified_equation_tex,
+        kT=Defaults.kT,
+        Lp=Defaults.Lp,
+        Lc=Defaults.Lc,
+    )
+
+
+def inverted_marko_siggia_simplified(name):
+    """Markov Siggia's Worm-like Chain model based on only entropic contributions (valid for F << 10 pN). This model
```
lpingas, line 174: typo: `Markov`

lpingas (approved):
After our conversation, it looks good to go! 👍

PR #42, opened 2020-08-31
change deprecated field access result from exception to warning

JoepVanlier:
To avoid breaking people's code, this PR changes an `IndexError` exception that we raised upon accessing a field covered by the API by a `FutureWarning`.


dean0x7d (approved):
LGTM!

PR #48, opened 2020-09-07
docs: remove old calibration syntax from cst example

JoepVanlier:
Tiny PR which removes deprecated use of the raw calibration groups from the docs and replaces it with the recommended method.


lpingas (approved):
LGTM!

PR #50, opened 2020-09-18
widgets: make unfocused fdwidget plot to correct axes

JoepVanlier:
This solves an annoying bug that can occur when a different figure got focus in a Jupyter notebook in the meantime.

Before:
bugged

After:
fixed


lpingas (approved):
:+1:

PR #51, opened 2020-09-18
kymo: fix coordinates displayed on axis

JoepVanlier:
This PR flips the displayed coordinates to correctly reflect position in `Kymo.plot_red()`, `Kymo.plot_green()`, `Kymo.plot_blue()` and `Kymo.plot_rgb()`.

While the origin of the image data has always been at the top left, previously the coordinate y=0 was displayed at the bottom left, incorrectly indicating that the origin was on the bottom left.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 * Add slicing (by time) for `FDCurve`.
 * Add widget to slice Fd's with from Jupyter Notebooks.
 * Fixed bug in `FdRangeSelectorWidget` that prevented drawing to the correct axes when other axes has focus.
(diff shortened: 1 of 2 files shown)


dean0x7d:
> the origin of the image data has always been at the top left

How? The data is oriented?

JoepVanlier:
> > the origin of the image data has always been at the top left
> 
> How? The data is oriented?

Poor phrasing. Maybe this is better:

Flip displayed coordinates to correctly reflect position in `Kymo.plot_red()`, `Kymo.plot_green()`, `Kymo.plot_blue()` and `Kymo.plot_rgb()`. The data origin (`kymo.red_image[0, 0]`) is displayed on the top left of the image in these plots, whereas previously this was not reflected correctly in the plot (placing the coordinate origin at the bottom left)..

Before patch:
image

After patch:
image

It seems unnecessarily confusing to plot the origin differently than where it actually is.

dean0x7d (approved):
I guess we're just stuck with top-left being the origin for Bluelake and all Bluelake-related data 😢

PR #53, opened 2020-09-24
Allow exporting h5

JoepVanlier:
One issue with uploading data to Harbor is that data files are often quite big. Very often, one doesn't need the high frequency channels for a particular dataset, or one might want to compress using a higher compression ratio. For this, `export_h5` was added to the `File` interface.

See also: https://lumicks-pylake.readthedocs.io/en/export_h5/tutorial/file.html

### DIFF
--- a/lumicks/pylake/detail/h5_helper.py
+++ b/lumicks/pylake/detail/h5_helper.py
@@ -1,0 +1,40 @@
+import h5py
+from fnmatch import fnmatch
+
+
+def write_h5(h5_file, output_filename, compression_level=5, omit_data={}):
+    """Write a modified h5 file to disk.
+
+    Parameters
+    ----------
+    h5_file : h5py.File
+        loaded h5 file
+    output_filename : str
+        Output file name.
+    compression_level : int
+        Compression level for gzip compression.
+    omit_data : Set[str]
+        Which data sets to omit. Should be a set of h5 paths.
+    """
+
+    with h5py.File(output_filename, "w") as out_file:
+
+        def traversal_function(name, node):
+            if any([fnmatch(name, o) for o in omit_data]):
+                print(f"Omitted {name} from export")
+                return
+
+            if isinstance(node, h5py.Dataset):
+                if node.dtype.kind == "O":
+                    # Non-numerical data doesn't support compression
+                    out_file.create_dataset(name, data=node)
+                else:
+                    # Numerical data can benefit a lot from compression
+                    out_file.create_dataset(name, data=node, compression="gzip", compression_opts=compression_level)
+            else:
+                out_file.create_group(f'{name}')
+
+            out_file[name].attrs.update(node.attrs)
+
+        h5_file.visititems(traversal_function)
+        out_file.attrs.update(h5_file.attrs)
--- a/lumicks/pylake/file.py
+++ b/lumicks/pylake/file.py
@@ -4,8 +4,9 @@ from typing import Dict
 
 from collections import OrderedDict
 from .calibration import ForceCalibration
-from .channel import Slice, Continuous, TimeSeries, TimeTags, channel_class
+from .channel import Slice, Continuous, TimeSeries, TimeTags
 from .detail.mixin import Force, DownsampledFD, PhotonCounts, PhotonTimeTags
+from .detail.h5_helper import write_h5
 from .fdcurve import FDCurve
 from .group import Group
 from .kymo import Kymo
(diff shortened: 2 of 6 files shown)


lumicks/pylake/file.py
```diff
@@ -209,3 +210,34 @@ def fdcurves(self) -> Dict[str, FDCurve]:
     @property
     def markers(self) -> Dict[str, Marker]:
         return self._get_object_dictionary("Marker", Marker)
+
+    def export_h5(self, filename, compression_level=5, omit_data={}):
```
dean0x7d, line 214: Maybe re-export instead of export? Since we aren't really exporting to a different format.
`reexport_h5()`?

lumicks/pylake/file.py
```diff
@@ -209,3 +210,34 @@ def fdcurves(self) -> Dict[str, FDCurve]:
...
+        """Write a modified h5 file to disk.
+
+        When transferring data, it can be beneficial to omit some channels from the h5 file, or use a higher compression
+        ratio. In particular, high frequency channels tend to take up a lot of space, and aren't always necessary for
+        every single analysis.
```
dean0x7d, line 219: It would be nice to mention here that Bluelake exports files at compression level 1 for performance reasons so this function can help reduce the size further if needed.

lumicks/pylake/file.py
```diff
@@ -209,3 +210,34 @@ def fdcurves(self) -> Dict[str, FDCurve]:
...
+        compression_level : int
+            Compression level for gzip compression (default: 5).
+        omit_data : Set[str]
+            Which data sets to omit. Should be a set of h5 paths (e.g. {"Force HF/Force 1y"}).
+
+        Examples
+        --------
+        ::
+
+            from lumicks import pylake
```
dean0x7d, line 234: Aren't we doing `import lumicks.pylake as lk` as the new default?

lumicks/pylake/detail/h5_helper.py
```diff
@@ -0,0 +1,39 @@
...
+    compression_level : int
+        Compression level for gzip compression.
+    omit_data : Set[str]
+        Which data sets to omit. Should be a set of h5 paths.
+    """
+
+    with h5py.File(output_filename, "w") as out_file:
+
+        def traversal_function(name, node):
+            if any([o in name for o in omit_data]):
```
dean0x7d, line 22: I'm afraid that this may be a bit too permissive. That's going to match any substring. That said, I'm not sure I have a better suggestions. Regex is probably overkill, but maybe some simple glob-like matching? I'm sure Python has something builtin for that.

lumicks/pylake/tests/test_file.py
```diff
@@ -257,3 +257,61 @@ def test_invalid_access(h5_file):
 
         with pytest.raises(IndexError):
             m["Kymo1"]
+
+
+def test_h5_export(tmpdir_factory, h5_file):
```
dean0x7d, line 262: The testing of the private and public API is kinda mixed up. If you want to test both, consider parameterizing on that. For example:

```py
def _private_api(file, *args, **kwargs):
    return write_h5(file.h5, *args, **kwargs)

def _public_api(file, *args, **kwargs):
    return file.export_h5(*args, **kwargs)

@pytest.mark.parametrize("func", [_internal_api, _public_api])
def test_h5_export(tmpdir_factory, h5_file, func):
```

dean0x7d (commented):
Cool idea! It should be very useful.

lumicks/pylake/file.py
JoepVanlier: 👍

lumicks/pylake/file.py
JoepVanlier: Yup! 👍

lumicks/pylake/detail/h5_helper.py
JoepVanlier: Yeah, fair point. Went with `fnmatch` for this (e.g. `Force HF/*`, `*/Force 1y`)

lumicks/pylake/tests/test_file.py
JoepVanlier: Fair point. Changed! 👍

lumicks/pylake/file.py
JoepVanlier: How about `save_as`? It's a fairly user-oriented name, and since `h5` is already implied we don't have to put it in the function name. If we ever decide to support more file types, we could always add a keyword argument with a default.

dean0x7d: `save_as()` 👍

lumicks/pylake/detail/h5_helper.py
dean0x7d: Excellent! Please just mention in the parameter docs that it's using `fnmatch` patterns.

dean0x7d (approved):
Very nice! This is going to be super useful. Please just mention in the docs that `fnmatch` patterns are used for omitting fields. You already have good examples in there. This is just for completeness.

PR #54, opened 2020-09-25
Lk51 timeprop

rpauszek:
Added time property to Slice class. Converts absolute timestamps (in nanoseconds) to relative time (in seconds) to help with data visualization

### DIFF
--- a/docs/tutorial/file.rst
+++ b/docs/tutorial/file.rst
@@ -104,6 +104,12 @@ The channels have a few convenient methods, like `.plot()` which make it easy to
     f1x_timestamps = file.force1x.timestamps
     plt.plot(f1x_timestamps, f1x_data)
 
+The `timestamps` attribute returns absolute values in nanoseconds.
+The relative time values in seconds can also be accessed directly::
+
+    f1x_seconds = file.force1x.seconds
+    plt.plot(f1x_time, f1x_data)
+
 The above examples use the `force1x` channel.
 A full list of available channels can be found on the :class:`~lumicks.pylake.File` reference page.
 
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -63,6 +63,11 @@ class Slice:
         return self._src.timestamps
 
     @property
+    def seconds(self):
+        """Relative time (in seconds) that corresponds to the channel data"""
+        return (self._src.timestamps - self._src.start) * 1e-9
+
+    @property
     def calibration(self) -> list:
         """Calibration data slicing is deferred until calibration is requested to avoid
         slicing values that may be needed."""
@@ -155,8 +160,7 @@ class Slice:
         """
         import matplotlib.pyplot as plt
 
-        seconds = (self.timestamps - self.timestamps[0]) / 1e9
-        plt.plot(seconds, self.data, **kwargs)
+        plt.plot(self.seconds, self.data, **kwargs)
         plt.xlabel(self.labels.get("x", "Time") + " (s)")
         plt.ylabel(self.labels.get("y", "y"))
         plt.title(self.labels.get("title", "title"))
--- a/lumicks/pylake/tests/test_channels.py
+++ b/lumicks/pylake/tests/test_channels.py
@@ -333,3 +333,11 @@ def test_downsampling():
         s.downsampled_by(-1)
     with pytest.raises(TypeError):
         s.downsampled_by(1.5)
+
+
+def test_seconds_property():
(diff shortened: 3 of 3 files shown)


lumicks/pylake/channel.py
```diff
@@ -62,6 +62,11 @@ def timestamps(self):
         """Absolute timestamps (since epoch) which correspond to the channel data"""
         return self._src.timestamps
 
+    @property
+    def time(self):
+        """Relative time (in seconds) that corresponds to the channel data"""
+        return (self._src.timestamps-self._src.timestamps[0])/1e9
```
JoepVanlier, line 68: Very tiny nitpick, but a space between the variables and the operator would be nice.

`(self._src.timestamps - self._src.timestamps[0])/1e9`

JoepVanlier (commented):
Looks good to me 👍 

I reckon squashing these commits into one probably makes sense for this, since it's a relatively small feature.

lumicks/pylake/channel.py
```diff
@@ -62,6 +62,11 @@ def timestamps(self):
         """Absolute timestamps (since epoch) which correspond to the channel data"""
         return self._src.timestamps
 
+    @property
+    def time(self):
+        """Relative time (in seconds) that corresponds to the channel data"""
```
dean0x7d, line 67: Time for some name bikeshedding :) I think it isn't very obvious that `timestamps` == absolute timestamps in nanoseconds and `time` == relative timestamps in seconds. I would like to make the naming a bit more obvious and differentiate the two properties. Maybe `.relative_time` or `.relative_seconds` would be nice. What do you think?

docs/tutorial/file.rst
```diff
@@ -104,6 +104,11 @@ The channels have a few convenient methods, like `.plot()` which make it easy to
     f1x_timestamps = file.force1x.timestamps
     plt.plot(f1x_timestamps, f1x_data)
 
+The time values in seconds can also be accessed directly::
+
+    f1x_time = file.force1x.time
+    plt.plot(f1x_time, f1x_data)
```
dean0x7d, line 110: It might be also nice to quickly mention here that `.timestamps` returns absolute values in ns.

lumicks/pylake/channel.py
dean0x7d, line 68: I would suggest using `.start` instead of `.timestamps[0]`. The reason for that can be found in `Continuous.timestamps`: it generates a new `np.arange()` every time that it's called. That means that it would create two identical arrays one after the other. That can be very wasteful and slow for large datasets. On the other hand, `.start` is fast for all 3 channel kinds.

I would also suggest multiplying by `1e-9` rather than dividing by `1e9`. Again, for large datasets, multiplication is pretty much always faster than division. I would not say that you should go out of your way to always multiply instead of divide but this is a nice default when possible.
```suggestion
        return (self._src.timestamps - self._src.start) * 1e-9
```

dean0x7d (commented):
First PR! 🎉 

And it's awesome that you also update the docs! You'll find below that my main comment is in regard to API naming. You'll get used to the endless naming discussion. I promise :)

Other than that, a few tips related to git:

1. Make sure that the email address that's in your commit messages matches the email address in your GitHub account so that GitHub can property link them. You either need to change the address in your git config/client or add that email to you GitHub account. You can add multiple email addresses to your GitHub and it will then properly recognize your commit with all addresses.

2. As Joep mentioned, you can easily squash all this down to a single commit. In general, it's always good to commit a feature and it's associated tests & docs as one commit (when possible). That makes it easier to read through and immediately see how new features are used/tested. Tests and docs are really an integral part of features so they could be committed together. An exception would be when the scope of the changes is just too big and things need to be split up for clarity. But for these small changes, keeping things together is nicer, both for review and later for looking through git history. (You can squash and force-push to this branch to update the PR. Let me know if you need any help with this.)

3. Check out our guidelines for git commit messages: https://github.com/lumicks/coding_guidelines/blob/master/git.md It suggests a format and explains what makes a good commit title and description.

lumicks/pylake/channel.py
rpauszek: Good point. For an attribute like this that is likely to be used frequently by users (ie, for plotting) I like to keep the name as short as possible which is why I picked `time` originally.  How about just `seconds`? Still different, descriptive and short

dean0x7d: `.seconds` sounds great!

dean0x7d (approved):
Excellent! Just squash to a single commit and merge away!

JoepVanlier (approved):
LGTM!

PR #55, opened 2020-09-25
scan: allow exporting plots to video formats

JoepVanlier:
This PR adds a feature to `scan` that's been requested a few times, namely the ability to export multi-frame scans to movie formats such as `.avi`, .`mp4` and `.gif`.

I decided to use the `matplotlib` animation module for this, since it allows us to easily export the axes with and doesn't require any fiddling with ffmpeg or other converters ourselves.

One thing to note is that I update the `imshow` data rather than replot it entirely every time. This is a lot faster than the alternative.

By default it uses `Pillow` (which already ships with `mpl`), and will allow you to generate GIFs. If `ffmpeg` is installed it will use that instead since it supports far more formats. I added some conda instructions to the installation instructions for installing `ffmpeg`. While that binary is GPL-3.0, it's really just a separate binary that we would not be shipping or linking.

Relevant readthedocs here: https://lumicks-pylake.readthedocs.io/en/movies/tutorial/images.html

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -98,6 +98,19 @@ If you already have Pylake installed and you want to update to the latest versio
     conda update lumicks.pylake
 
 
+.. _ffmpeg_installation:
+
+Optional dependencies
+---------------------
+
+.. rubric:: ffmpeg
+
+Exporting to compressed video formats requires an additional dependency named ffmpeg which must be installed separately.
+When using conda, ffmpeg can be installed as follows::
+
+    conda install -c conda-forge ffmpeg
+
+
 Troubleshooting
 ---------------
 
--- a/docs/tutorial/images.rst
+++ b/docs/tutorial/images.rst
@@ -26,6 +26,11 @@ Or just pick a single one::
     scan = file.scans["name"]
     scan.plot_red()
 
+If a few pixels dominate the image, one might want to set the scale by hand. We can pass an extra argument to `plot_red`
+named `vmax` to accomplish this. This parameter gets forwarded to :func:`matplotlib.pyplot.imshow`::
+
+    scan.plot_red(vmax=5)
+
 Access the raw image data::
 
     rgb = scan.rgb_image  # matrix with `shape == (h, w, 3)`
@@ -51,3 +56,16 @@ Multi-frame scans are also supported::
     print(scan.rgb_image.shape)  # (self.num_frames, h, w, 3) -> three color channels
 
     scan.plot(frame=3)  # plot the third frame -- defaults to the first frame if no argument is given
+
+Scans can also be exported to video formats.
+Exporting the red channel of a multi-scan GIF can be done as follows for example::
+
+    scan.export_video_red("test_red.gif")
+
+Or if we want to export a subset of frames (the first frame being 10, and the last frame being 40) of all three channels
+at a frame rate of 40 frames per second, we can do this::
+
+    scan.export_video_rgb("test_rgb.gif", start_frame=10, end_frame=40, fps=40)
+
+For other video formats such as `.mp4` or `.avi`, ffmpeg must be installed. See
+:ref:`installation instructions <ffmpeg_installation>` for more information on this.
--- a/lumicks/pylake/scan.py
+++ b/lumicks/pylake/scan.py
@@ -75,7 +75,21 @@ class Scan(Kymo):
 
         return self._to_spatial(timestamps)
 
-    def _plot(self, image, frame=1, **kwargs):
+    def _plot(self, image, frame=1, image_handle=None, **kwargs):
+        """Plot a scan frame.
+
+        Parameters
+        ----------
+        image : array_like
+            Image data
+        frame : int
+            Frame index (starting at 1).
+        image_handle : `matplotlib.image.AxesImage` or None
+            Optional image handle which is used to update plots with new data rather than reconstruct them (better for
+            performance).
+        **kwargs
+            Forwarded to :func:`matplotlib.pyplot.imshow`
+        """
         import matplotlib.pyplot as plt
 
         frame = np.clip(frame, 1, self.num_frames)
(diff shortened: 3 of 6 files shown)


docs/tutorial/images.rst
```diff
@@ -51,3 +56,16 @@ Multi-frame scans are also supported::
...
+
+Scans can also be exported to movie formats.
+Exporting the red channel of a multi-scan GIF can be done as follows for example::
+
+    scan.export_movie_red("test_red.gif")
+
+Or if we want to export a subset of frames (the first frame being 10, and the last frame being 40) of all three channels
+at a frame rate of 40 frames per second, we can do this::
+
+    scan.export_movie_rgb("test_rgb.gif", start_frame=10, end_frame=40, fps=40)
```
dean0x7d, line 68: I think that this should be called `video` instead of `movie`. "Movie" makes me think of a story and cinema. "Video" is a better fit.

I know that "movie" is the popular term for this in academia but it's not really the right term. A similar pet peeve of mine is calling everything a "cartoon" when something like "illustration" would be a better fit.

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
...
+        **kwargs
+            Forwarded to :func:`matplotlib.pyplot.imshow`."""
+        from matplotlib import animation, rcParams
+        import matplotlib.pyplot as plt
+
+        metadata = dict(title=self.name)
+        if "ffmpeg" in animation.writers:
+            writer = animation.writers['ffmpeg'](fps=fps, metadata=metadata)
+        elif "pillow" in animation.writers:
+            writer = animation.writers['pillow'](fps=fps, metadata=metadata)
```
dean0x7d, line 133: Mixing `""` and `''`.

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
...
+        else:
+            raise RuntimeError("You need either ffmpeg or pillow installed to export movies.")
+
+        start_frame = start_frame if start_frame else 1
+        end_frame = end_frame if end_frame else self.num_frames
+        end_frame += 1
```
dean0x7d, line 139: Why unconditionally add 1? Shouldn't it only be for the default value? `self.num_frames + 1`

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
...
+
+        # This prevents font anti-aliasing from being disabled on the notebook backends (at the cost of not having
+        # transparent regions outside the axes part of figures).
+        rcParams['figure.facecolor'] = "White"
```
dean0x7d, line 143: Indicate why it needs to be force-enabled. Workaround for some bug or some other reason? "Why" is one of the most important things to answer in comments and commit messages.

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
             aspect=(image.shape[0] / image.shape[1]) * (x_um / y_um)
         )
 
-        plt.imshow(image, **{**default_kwargs, **kwargs})
+        if not image_handle:
+            image_handle = plt.imshow(image, **{**default_kwargs, **kwargs})
+        else:
+            # Updating the image data in an existing plot is a lot faster than re-plotting with `imshow`.
+            image_handle.set_data(image)
```
dean0x7d, line 98: It would be good to make this an argument docstring instead of a comment. It would explain why `image_handle` is even needed as an argument.

dean0x7d (commented):
A lot of people are going to be super happy to have this feature! The code looks great overall. The main two comments are about a possible off-by-one error and maybe a small optimization.

lumicks/pylake/scan.py
JoepVanlier: good catch 👍

docs/tutorial/images.rst
JoepVanlier: Yeah, makes sense!

docs/install.rst
```diff
@@ -98,6 +98,19 @@ If you already have Pylake installed and you want to update to the latest versio
     conda update lumicks.pylake
 
 
+.. _ffmpeg_installation:
+
+Additional dependencies
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/scan.py
JoepVanlier: Good catch! Fixing!

lumicks/pylake/scan.py
JoepVanlier: Yeah. This ticket seems related and started my search for a solution in the realm of transparency: https://github.com/matplotlib/matplotlib/issues/5335
The actual tag `nbagg.transparent` has since then been deprecated.

I could make it a bit more specific (only override when there's transparency) and restore the color after export.

```python
        # On some notebook backends, figures render with a transparent background by default. This leads to very
        # poor image quality, since it prevents font anti-aliasing (at the cost of not having transparent regions
        # outside the axes part of figures).
        face_color = to_rgba(rcParams["figure.facecolor"])
        if face_color[3] < 1.0:
            rcParams["figure.facecolor"] = (face_color[0], face_color[1], face_color[2], 1.0)
        else:
            face_color = None
```

lumicks/pylake/scan.py
JoepVanlier: Fair enough. I've added a docstring to the function.

```python
        """Plot a scan frame.

        Parameters
        ----------
        image : array_like
            Image data
        frame : int
            Frame index (starting at 1).
        image_handle : `matplotlib.image.AxesImage` or None
            Optional image handle which is used to update plots with new data rather than reconstruct them (better for
            performance).
        **kwargs
            Forwarded to :func:`matplotlib.pyplot.imshow`
        """
```

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
...
+        end_frame : int
+            Last frame.
+        fps : int
+            Frames per second.
+        **kwargs
+            Forwarded to :func:`matplotlib.pyplot.imshow`.
+        """
+        self._export_movie("rgb", file_name, start_frame, end_frame, fps, **kwargs)
+
+    def export_movie_red(self, file_name, start_frame=None, end_frame=None, fps=15, **kwargs):
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, I agree. This API setup is going to get out of hand at some point. Will think about it.

lumicks/pylake/scan.py
```diff
@@ -91,10 +91,137 @@ def _plot(self, image, frame=1, **kwargs):
...
+        plot_func = getattr(self, f"plot_{plot_type}")
+        image_handle = plot_func(frame=start_frame, **kwargs)
+
+        def plot(num):
+            plot_func(frame=start_frame + num, image_handle=image_handle, **kwargs)
+            return plt.gca().get_children()
+
+        line_ani = animation.FuncAnimation(fig, plot, end_frame - start_frame, interval=1, blit=True)
+        line_ani.save(file_name, writer=writer)
+        plt.close(fig)
```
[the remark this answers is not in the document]
JoepVanlier: Yep works as intended 👍

lumicks/pylake/scan.py
```diff
@@ -91,10 +106,146 @@ def _plot(self, image, frame=1, **kwargs):
...
+            raise RuntimeError("You need either ffmpeg or pillow installed to export videos.")
+
+        start_frame = start_frame if start_frame else 1
+        end_frame = end_frame if end_frame else self.num_frames + 1
+
+        # On some notebook backends, figures render with a transparent background by default. This leads to very
+        # poor image quality, since it prevents font anti-aliasing (at the cost of not having transparent regions
+        # outside the axes part of figures).
+        face_color = rcParams["figure.facecolor"]
+        face_color_rgba = to_rgba(face_color)
```
dean0x7d, line 159: You imported all `matplotlib` bits at the function level, except this one. Any reason for the difference?

lumicks/pylake/scan.py
```diff
@@ -91,10 +106,146 @@ def _plot(self, image, frame=1, **kwargs):
...
+        if face_color_rgba[3] < 1.0:
+            rcParams["figure.facecolor"] = (face_color_rgba[0], face_color_rgba[1], face_color_rgba[2], 1.0)
```
dean0x7d, line 161: There's a couple of ways to do this without repetition:
```suggestion
            rcParams["figure.facecolor"] = (*face_color_rgba[:3], 1.0)
```
```suggestion
            rcParams["figure.facecolor"] = face_color_rgba[:3] + (1.0,)
```

lumicks/pylake/scan.py
dean0x7d: It would be nice to include the link in the comment as well.

lumicks/pylake/scan.py
dean0x7d: I think that you can completely remove the comment now. It's clear from the docstring.

lumicks/pylake/scan.py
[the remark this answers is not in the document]
dean0x7d:
```py
def export_video(self, file_name, channels="rgb", ...):
    ...
```
```py
scan.export_video("scan.gif")
scan.export_video("scan.gif", "r")
scan.export_video("scan.gif", "b")
scan.export_video("scan.gif", "br")
scan.export_video("scan.gif", "bgr")
```
This could be an option but the question is what to do with arbitrary color ordering?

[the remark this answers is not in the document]
JoepVanlier: I mean, initially, we could just constrain it to the types we already have in plot (i.e. "red", "green", "blue" and "rgb").
I'm not sure whether there are any applications for arbitrary orders or combinations of colors at this point. Otherwise this PR will turn into a complete refactor of `scan` where we also break backward compatibility.

I would say "br" is undefined since you don't know in which channel you want them.

If we are to go with that layout, I'd always make it a tuple of three.
e.g.
"r00"  // Currently "Red"
"0g0" // Currently "Green"
"rg0" // Currently not possible
"rrr"  // Grayscale "Red"
"rgb" // Just "rgb"

That would give users complete control over what is mapped to which channel exactly. But this would require reworking of `Scan` and `Kymo` as well.

lumicks/pylake/scan.py
JoepVanlier: Yeah, that's better. Fixed! 👍

lumicks/pylake/scan.py
JoepVanlier: Yeah, fair point. I moved it to the function scope, since we really only use that functionality there and this function is likely used rarely.

Do we have any standard guidelines for imports?

dean0x7d: For me, there are two possible reasons to put imports at function-level instead of module-level:
1. If it's an optional dependency -- the user will only get the import error if they try to use the feature that requires it.
2. If the module load time is long we may want to defer its initialization until someone actually uses the feature. I think that I noticed this with `scipy` or `matplotlib` on Windows. Although, with our use case, I think we almost always need `matplotlib` anyway. Not a super important point unless there's some clear data to indicate slowdowns.

lumicks/pylake/scan.py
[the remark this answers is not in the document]
dean0x7d: Oh, no, whatever happens, make it a separate PR. I just wanted to discuss it briefly here.

dean0x7d (approved):
`pr.export_video_rgb("approved.mrg", fps="yes")`

PR #57, opened 2020-10-06
CorrelatedStack: align image

rpauszek:
This PR adds feature to `CorrelatedStack` object to reconstruct an RGB image from the alignment matrices calculated in Bluelake. Functionality also added for plotting 16-bit RBG images in addition to the grayscale 8-bit images already implemented.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## v0.6.3 | t.b.d.
 
+* Add `align` kwarg to `CorrelatedStack`. If enabled, this option will automatically align the color channels for multi-channel images, provided that the alignment metadata from Bluelake is available.
 * Add `save_as` to `File` for exporting compressed HDF5 files, or omitting specific channels from an HDF5 file See tutorial on [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#exporting-h5-files) for more information.
 * Add `Scan.export_video_red`, `Scan.export_video_green`, `Scan.export_video_blue` and `Scan.export_video_rgb` to export multi-frame videos to video files or GIFs. See tutorial on [confocal images](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/images.html) for more information.
 * Expose low-level kymograph API (still in alpha status!). See tutorial on [kymotracking](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html)
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,8 @@
 
 ## v0.6.3 | t.b.d.
 
-* Add `align` kwarg to `CorrelatedStack`. If enabled, this option will automatically align the color channels for multi-channel images, provided that the alignment metadata from Bluelake is available.
+* Add support for RGB images to `CorrelatedStack`.
+* Add image alignment to `CorrelatedStack`. Image alignment is enabled by default and will automatically align the color channels for multi-channel images, provided that the alignment metadata from Bluelake is available. Image alignment can be disabled by specifying `align=False` when loading the `CorrelatedStack`.
 * Add `save_as` to `File` for exporting compressed HDF5 files, or omitting specific channels from an HDF5 file See tutorial on [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#exporting-h5-files) for more information.
 * Add `Scan.export_video_red`, `Scan.export_video_green`, `Scan.export_video_blue` and `Scan.export_video_rgb` to export multi-frame videos to video files or GIFs. See tutorial on [confocal images](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/images.html) for more information.
 * Expose low-level kymograph API (still in alpha status!). See tutorial on [kymotracking](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html)
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -6,8 +6,6 @@ import cv2
 import tifffile
 import warnings
 
-from .channel import Slice, TimeSeries
-
 
 class TiffFrame:
     """Thin wrapper around a TIFF frame stack. For camera videos timestamps are stored in the DateTime tag in
@@ -294,7 +292,7 @@ class CorrelatedStack:
         """
         import matplotlib.pyplot as plt
 
-        downsampled = channel_slice.downsampled_over(self.timestamps, where='left')
+        downsampled = channel_slice.downsampled_over(self.timestamps, where='left', reduce=reduce)
 
         if len(downsampled.timestamps) < len(self.timestamps):
             warnings.warn("Only subset of time range available for selected channel")
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -1,7 +1,12 @@
 import numpy as np
 import os
 import re
+import json
+import cv2
+import tifffile
 import warnings
+
+from .channel import Slice, TimeSeries
 
 
 class TiffFrame:
@@ -13,12 +18,70 @@ class TiffFrame:
     page : tifffile.tifffile.TiffPage
         Tiff page recorded from a camera in Bluelake.
     """
-    def __init__(self, page):
+    def __init__(self, page, align):
         self._src = page
+        self._align = align
+
+    def _align_image(self):
+        """ reconstruct image using alignment matrices from Bluelake 
+            return aligned image as a NumPy array 
+        """
+
+        def correct_alignment_offset(alignment, x_offset, y_offset):
+            # translate the origin of the image so that it matches that of the original transform
+            translation = np.eye(3)
+            translation[0, -1] = -x_offset
+            translation[1, -1] = -y_offset
+            # apply the original transform to the translated image. 
+            # it only needs to be resized from a 2x3 to a 3x3 matrix
+            original = np.vstack((alignment, [0, 0, 1]))
+            # translate the image back to the original origin. 
+            # takes into account both the offset and the scaling performed by the first step
+            back_translation = np.eye(3)
+            back_translation[0, -1] = original[0, 0] * x_offset
+            back_translation[1, -1] = original[1, 1] * y_offset
+            # concatenate the transforms by multiplying their matrices and ignore the unused 3rd row
+            return np.dot(back_translation, np.dot(original, translation))[:2, :]
+
+        try:
+            description = json.loads(self._src.description)
+        except json.decoder.JSONDecodeError:
+            warnings.warn("File does not contain metadata. Only raw data is available")
+            return self.raw_data
+
+        try:
+            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2,3))
+                        for color in ('red', 'blue')]
+            align_roi = np.array(description["Alignment region of interest (x, y, width, height)"])[:2]
+            roi = np.array(description["Region of interest (x, y, width, height)"])[:2]
+        except KeyError:
+            warnings.warn("File does not contain alignment matrices. Only raw data is available")
+            return self.raw_data
+        
+        x_offset, y_offset = align_roi - roi
+        if not (x_offset == 0 and y_offset == 0):
+            align_mats = [correct_alignment_offset(mat, x_offset, y_offset) for mat in align_mats]
+
+        img = self.raw_data
+        rows, cols, _ = img.shape
+        for mat, channel in zip(align_mats, (0, 2)):
+            img[:,:,channel] = cv2.warpAffine(img[:,:,channel], mat, (cols,rows),
+                                              flags=(cv2.INTER_LINEAR | cv2.WARP_INVERSE_MAP),
+                                              borderMode=cv2.BORDER_CONSTANT, borderValue=0)
+        return img
+        
 
     @property
     def data(self):
+        return self._align_image() if (self.is_rgb and self._align) else self._src.asarray()
+
+    @property
+    def raw_data(self):
         return self._src.asarray()
+
+    @property
+    def is_rgb(self):
+        return self._src.tags["SamplesPerPixel"].value == 3
 
     @property
     def start(self):
@@ -39,16 +102,16 @@ class TiffStack:
     tiff_file : tifffile.TiffFile
         TIFF file recorded from a camera in Bluelake.
     """
-    def __init__(self, tiff_file):
+    def __init__(self, tiff_file, align):
         self._src = tiff_file
+        self._align = align
 
     def get_frame(self, frame):
-        return TiffFrame(self._src.pages[frame])
+        return TiffFrame(self._src.pages[frame], align=self._align)
 
     @staticmethod
-    def from_file(image_file):
-        import tifffile
-        return TiffStack(tifffile.TiffFile(image_file))
+    def from_file(image_file, align):
+        return TiffStack(tifffile.TiffFile(image_file), align=align)
 
     @property
     def num_frames(self):
@@ -63,6 +126,10 @@ class CorrelatedStack:
     ----------
     image_name : str
         Filename for the image stack. Typically a TIFF file recorded from a camera in Bluelake.
+
+    align : bool
+        If enabled, multi-channel images will be reconstructed from the image alignment metadata
+        from Bluelake. The default value is `True`.
 
     Examples
     --------
@@ -80,8 +147,8 @@ class CorrelatedStack:
         # Determine the force trace averaged over frame 2...9.
         file.force1x.downsampled_over(stack[2:10].timestamps)
     """
-    def __init__(self, image_name):
-        self.src = TiffStack.from_file(image_name)
+    def __init__(self, image_name, align=True):
+        self.src = TiffStack.from_file(image_name, align=align)
         self.name = os.path.splitext(os.path.basename(image_name))[0]
         self.start_idx = 0
         self.stop_idx = self.src.num_frames
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -23,9 +23,7 @@ class TiffFrame:
         self._align = align
 
     def _align_image(self):
-        """ reconstruct image using alignment matrices from Bluelake 
-            return aligned image as a NumPy array 
-        """
+        """ reconstruct image using alignment matrices from Bluelake; return aligned image as a NumPy array"""
 
         def correct_alignment_offset(alignment, x_offset, y_offset):
             # translate the origin of the image so that it matches that of the original transform
@@ -50,8 +48,8 @@ class TiffFrame:
             return self.raw_data
 
         try:
-            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2,3))
-                        for color in ('red', 'blue')]
+            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2, 3))
+                          for color in ('red', 'blue')]
             align_roi = np.array(description["Alignment region of interest (x, y, width, height)"])[:2]
             roi = np.array(description["Region of interest (x, y, width, height)"])[:2]
         except KeyError:
@@ -65,11 +63,10 @@ class TiffFrame:
         img = self.raw_data
         rows, cols, _ = img.shape
         for mat, channel in zip(align_mats, (0, 2)):
-            img[:,:,channel] = cv2.warpAffine(img[:,:,channel], mat, (cols,rows),
-                                              flags=(cv2.INTER_LINEAR | cv2.WARP_INVERSE_MAP),
-                                              borderMode=cv2.BORDER_CONSTANT, borderValue=0)
+            img[:, :, channel] = cv2.warpAffine(img[:, :, channel], mat, (cols, rows),
+                                                flags=(cv2.INTER_LINEAR | cv2.WARP_INVERSE_MAP),
+                                                borderMode=cv2.BORDER_CONSTANT, borderValue=0)
         return img
-        
 
     @property
     def data(self):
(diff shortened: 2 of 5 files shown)


lumicks/pylake/correlated_stack.py
```diff
@@ -46,9 +99,21 @@ def __init__(self, tiff_file):
...
+        """ reconstruct all pages using alignment matrices from Bluelake 
+            return as new TiffStack
+        """
+        buffer = io.BytesIO()
+        with tifffile.TiffWriter(buffer) as writer:
+            for j in range(self.num_frames):
+                frame = self.get_frame(j)
+                aligned_img = frame.align()
+                writer.save(aligned_img)
+        buffer.seek(0)
```
JoepVanlier, line 112: While I do like keeping the internals very much the same, and providing a TIFF image, there's a fairly critical issue with doing it this way; namely that other TIFF properties (such as tags) get lost.

In the current implementation the metadata gets lost (e.g. `.start` and `.stop` no longer work post alignment). Note that there are a few tests regarding correlated stacks (checking `start`, `stop` etc), and it would be good to see whether these pass post alignment as well. If you do it like this, it would be good to add a test for this (or rather, somehow parameterize the existing `CorrelatedStack` test to optionally include an alignment procedure so that future additions to `CorrelatedStack` also immediately test the aligned stacks to see whether all the attributes still work). 

Alternatively, I wonder whether it isn't preferable to lazily perform this alignment when the data is requested, rather than having a function which returns a copy of the whole TIFF stack with modified frames. Perhaps the frame `.data` can return an aligned image when alignment information is available and we add a field named `raw_data` or `unaligned_data` which doesn't apply the processing. As for plotting, there could be a named keyword `aligned` or something. I would assume that in most cases, users want the aligned images.

Thoughts @facutuesca @dean0x7d ?

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -119,3 +138,87 @@ def test_correlation():
     cc = channel.Slice(channel.Continuous(np.arange(10, 80, 2), 1588267266006287100, 1000))
     ch = cc.downsampled_over([(1588267266006287100, 2588267266006287100)], where='left')
     assert int(ch.timestamps[0]) == 1588267266006287100
+
+
+def make_alignment_image_data(spots, Tx_red, Ty_red, theta_red, Tx_blue, Ty_blue, theta_blue):
+    def make_transform_matrix(Tx, Ty, theta):
+        M = np.eye(3)
+        M[0,-1] = Tx
```
JoepVanlier, line 146: Nitpick: Space after comma.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -14,6 +15,18 @@ def value(self):
 
         self.tags = {"DateTime": Tag()}
```
JoepVanlier, line 17: nitpick: two linebreaks before a new top level class def (https://www.python.org/dev/peps/pep-0008/#blank-lines)

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -119,3 +138,87 @@ def test_correlation():
...
+    # test that bad alignment matrix gives high error compared to correct matrix
+    description["Alignment red channel"][2] = 25
+    bad_fake_tiff = TiffStack(MockTiffRgb(img, [["10", "18"]], description))
+    stack = CorrelatedStack.from_data(bad_fake_tiff)        
+    aligned = stack.align()
+    fr = aligned._get_frame(0)
+    maxabserr_r = np.max(np.abs(fr.data[:,:,1]-fr.data[:,:,0]))
+    maxabserr_b = np.max(np.abs(fr.data[:,:,1]-fr.data[:,:,2]))
+
+    assert (maxabserr_r > 0.05)
```
JoepVanlier, line 224: nitpick: finish file with line break

JoepVanlier (changes requested):
Overall it looks good, there's some issues with the current implementation however. The biggest issue is that in the current implementation tags are lost; which means it would no longer be possible to time-correlate frames post-alignment.

I'm also not entirely sure whether having to explicitly call `align` and then return a copy is preferable; or whether we should just handle it transparently and have a property for when raw unaligned data is specifically desired.

Last thing, is that cvode needs to be added to `setup.py` 👍

lumicks/pylake/correlated_stack.py
JoepVanlier: Addendum: I discussed it briefly with Roeland, and the aligned image should be the default; which kinda makes me think the lazy evaluation with aligned as default makes more sense.

lumicks/pylake/correlated_stack.py
```diff
@@ -17,10 +21,59 @@ class TiffFrame:
...
     def data(self):
         return self._src.asarray()
 
+    @property
+    def is_rgb(self):
+        return self.data.ndim > 2
+
+    @property
+    def plot_data(self):
+        return (self.data/256).astype(np.uint8) if self.is_rgb else self.data
```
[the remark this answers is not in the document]
rpauszek: I agree, the vmax doesn't do anything for rgb images since there's no colormap. But, as I noticed during testing, sometimes the images are quite dark if the full dynamic range isn't used. Perhaps including a special implementation for `vmax` in a `plot_rgb()` method that could renormalize the data to a specified range? I like the idea of including the single-channel plotting functions too

lumicks/pylake/correlated_stack.py
```diff
@@ -17,10 +21,59 @@ class TiffFrame:
...
+            # translate the image back to the original origin. 
+            # takes into account both the offset and the scaling performed by the first step
+            transform3 = np.eye(3)
+            transform3[0,-1] = transform2[0,0]*x_offset
+            transform3[1,-1] = transform2[1,1]*y_offset
+            # concatenate the transforms by multiplying their matrices and ignore the unused 3rd row
+            return np.dot(transform3, np.dot(transform2, transform1))[:2,:]
+
+
+        description = json.loads(self._src.description)
```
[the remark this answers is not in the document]
rpauszek: Perhaps just a warning that only the unaligned data is available? It might not be ideal for visualization, but as of right now there are no other methods I can see that depend on the presence of the aligned data.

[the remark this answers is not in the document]
JoepVanlier: Yeah, considering that after discussing with Roeland we concluded that aligned should be the default, it should probably not warn when alignment is not possible, since you'd be getting that warning a lot on older datasets. This function should probably only be called if aligned data is available. Could even consider making this function private (at least initially), since if we do it lazily, we don't really need users to explicitly call it.

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +19,103 @@ class TiffFrame:
...
+    def _get_plot_data(self, channel="rgb", vmax=None): 
+        """ return data an numpy array, convert dtype=float in range [0,1] 
+            if data is grayscale or channel in ('red', 'green', 'blue')
+                return data as is
+            if channel is 'rgb', correct for optional vmax argument:
+                None   : return data as is
+                'auto' : normalize data to max signal of all channels
+                float  : normalize data to vmax value
+        """
+        data = (self.data/(2**self.bit_depth-1)).astype(np.float)
```
JoepVanlier, line 98: I think it might be nicer to do the scaling in one step. The issue with doing it like this is that now `vmax` will be in the float scale from `0.0f - 1.0f`, whereas if you do it all at once, you'd be able to just have a look at the histogram of the raw data. What do you think?

lumicks/pylake/correlated_stack.py
```diff
@@ -129,36 +224,48 @@ def from_data(cls, data, name=None, start_idx=0, stop_idx=None):
...
-        if self.num_frames == 1:
-            plt.title(self.name)
-        else:
-            plt.title(f"{self.name} [frame {frame}/{self.num_frames}]")
+        if show_title:
+            if self.num_frames == 1:
+                plt.title(self.name)
+            else:
+                # display with 1-based index for frames and total frames
+                plt.title(f"{self.name} [frame {frame+1}/{self.num_frames}]")
```
JoepVanlier, line 261: why optional?

rpauszek: While I was testing, it was easiest to plot a grid of images using different conditions for the vmax options and the titles kept getting in the way. Rather than setting the title to an empty string each time, it seemed useful to be able to turn the auto-generated title off. It's a small change and the default is still the same as before; but we can get rid of the option if you think that's best

tiagobonetti (commented):
Looks pretty good. I added some suggestions on readability.

JoepVanlier (changes requested):
I think it's really shaping up! Would be good to up the test coverage a bit and I have a few other suggestions.

Other than that, the only thing I was still wondering about one thing. How would you feel about making the default behaviour for `vmax` `auto`. So when `vmax` is None, it means that for RGB images, it will just scale them as `auto` does now.

The reason I'm asking, is because I think it would simplify the logic quite a bit. Then you could deal with `vmax` transparently in `plot_correlated` (just passing the `**kwargs` along, with defaults specified, making it completely analogous to `plot`).

i.e.
```python
        default_kwargs = dict(
            cmap='gray',
            vmax=None
        )
        kwargs = {**default_kwargs, **kwargs}
```

It would also get rid of this niggle:
```python
        if isinstance(kwargs["vmax"], str):
            kwargs.pop("vmax")
```
in plot.

What do you think?

lumicks/pylake/correlated_stack.py
```diff
@@ -173,6 +280,14 @@ def plot_correlated(self, channel_slice, frame=0, reduce=np.mean):
             The function which is going to reduce multiple samples into one. The default is
             :func:`numpy.mean`, but :func:`numpy.sum` could also be appropriate for some cases
             e.g. photon counts.
+        channel : 'rgb', 'red', 'green', 'blue', None; optional
+            Channel to plot for RGB images (None defaults to 'rgb')
+            Not used for grayscale images
+        vmax : None, 'auto', float; optional
```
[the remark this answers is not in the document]
JoepVanlier: I agree with what you say regarding the name being kinda weird, but `vmax` is what matplotlib uses for `imshow`, so I'd keep it for consistency.

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +19,103 @@ class TiffFrame:
...
+        self._align = align
 
+    def _align_image(self):
+        """ reconstruct image using alignment matrices from Bluelake 
+            return aligned image as a NumPy array 
+        """
+
+        def correct_alignment_offset(alignment, x_offset, y_offset):
+            # translate the origin of the image so that it matches that of the original transform
+            transform1 = np.eye(3)
```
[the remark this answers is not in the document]
rpauszek: I think it's pretty common knowledge, especially since it's also a Matlab function. It has more flexibility than identity because you can choose which diagonal you want filled, so it's my go-to

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +19,103 @@ class TiffFrame:
...
+        data = (self.data/(2**self.bit_depth-1)).astype(np.float)
+        
+        if not self.is_rgb:
+            return data
+
+        if (channel is None) or (channel.lower() == "rgb"):
+            if vmax is None:
+                return data
+            elif isinstance(vmax, (int, float)):
+                return data/vmax
```
JoepVanlier, line 107: This approach works, but ends up showing a ton of warnings from imshow which mention that the data is clamped. Would be nice to handle that more gracefully.

lumicks/pylake/correlated_stack.py
JoepVanlier: One thing I'm worried about is the behavioural change that happens here for single channel data. I think it would be wise to not change the single channel data. The normalization is not needed there as you can just use the `vmax` argument of `imshow`. When hovering over the data, you can see the numeric values of the pixels and with single channel data, it would be nice if those still corresponded to raw signal values. I'd sooner treat `rgb` as a special case, like we do in `Scan`. It would be nice to have a uniform API with `Scan` as well.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -1,33 +1,96 @@
...
-    def __init__(self, times):
-        self.pages = []
-        for r in times:
-            self.pages.append(TiffPage(r[0], r[1]))
+class MockTiffPage:
+    def __init__(self, data, start_time, end_time, description="", bit_depth=8):
+        self._data = data
+        bit_depth = bit_depth if data.ndim == 2 else (bit_depth, bit_depth, bit_depth)
+        self.tags = {"DateTime": MockTag(f"{start_time}:{end_time}"),                   
+                     "ImageDescription": MockTag(description),#json.dumps(description)),
```
tiagobonetti, line 23: I guess this is probably a forgotten comment from development?

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -119,3 +186,239 @@ def test_correlation():
...
+                "bit_depth": 8}
+
+    # grayscale
+    img0, img, description = make_alignment_image_data(**img_args)
+    fake_tiff = TiffStack(MockTiffFile(data=[img[:, :, 0]], times=[["10", "18"]], 
+                                       description=json.dumps(description), bit_depth=8), 
+                          align=True)
+    stack = CorrelatedStack.from_data(fake_tiff)        
+    fr = stack._get_frame(0)
+    # normed_data = fr.raw_data(fr.raw_data/(2**fr.bit_depth-1)).astype(np.float)
```
tiagobonetti, line 332: Forgotten development comment?

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -119,3 +186,239 @@ def test_correlation():
...
+            tmp_spots[0] += offsets[0]
+            tmp_spots[1] += offsets[1]
+    
+    img = make_image(red_spots, spots, blue_spots, bit_depth)
+    description = make_description(m_red, m_blue, offsets)
+
+    return img0, img, description            
+
+
+def test_image_reconstruction():
```
tiagobonetti, line 310: This test feels like a bunch of small tests in a single big test.   This is not ideal since one test can observe side effects from another.

Maybe we could split them?
The comments that already indicate different things being tested are a great reference point.
Apparently, the suggested thing to group them is to put them in a class [docs](https://pytestguide.readthedocs.io/en/latest/pytestGuide/#test-code).

Same thing might apply to `test_rgb_correlated_stack`.

tiagobonetti (commented):
Looks pretty good to me.
I added some small questions and suggestions.

lumicks/pylake/correlated_stack.py
```diff
@@ -129,36 +223,45 @@ def from_data(cls, data, name=None, start_idx=0, stop_idx=None):
...
+                # display with 1-based index for frames and total frames
+                plt.title(f"{self.name} [frame {frame+1}/{self.num_frames}]")
 
     def _get_frame(self, frame=0):
         if frame >= self.num_frames or frame < 0:
             raise IndexError("Frame index out of range")
         return self.src.get_frame(self.start_idx + frame)
 
-    def plot_correlated(self, channel_slice, frame=0, reduce=np.mean):
+    def plot_correlated(self, channel_slice, frame=0, channel="rgb", vmax="auto", reduce=np.mean):
```
JoepVanlier, line 264: Would it make sense to make this more uniform with `plot` (e.g. forwarding `vmax` to `imshow`). Would mean removing `auto` as option as well. Was the `auto` option tested? Because it doesn't seem to be supported by `_get_plot_data` anymore?

JoepVanlier (changes requested):
Kind of looks like there's still a remnant of the `auto` behaviour for `plot_correlated`?

lumicks/pylake/tests/test_correlated_stack.py
rpauszek: Agreed. I left `test_rgb_correlated_stack` as it because it's the same format as `test_correlated_stack`. Now that the testing is growing, it might be worthwhile to reorganize the whole file in another ticket

lumicks/pylake/correlated_stack.py
rpauszek: Missed this one. I completely took out the `vmax` kwarg for this method since other optional plotting parameters aren't supported at this point (`cmap`, etc..) So for `channel="rgb"` the data will be normalized and all other options the data will be returned as is.

JoepVanlier: Yeah, it happens. I guess I forgot to add a test for the actual plots at the time. Could add them in a separate test which just spins up the plots to prevent a regression in the future, That or maybe make a separate ticket for it so we don't forget.

We could consider adding the other `kwargs` that `plot` has to this one and handle it similarly, forwarding `kwargs` to `imshow`. Then `vmax` could also be used with `plot_correlated`. It'd be pretty nice to have, since you said by default they can be very dim. But if you don't want the scope creep, that's also fine, we can make a separate ticket out of that. Maybe that's better, since this one has been in flight for a while now :)

Just one general comment, since we don't know whether people are using named arguments, adding an argument in the middle could cause issues for some people. We should either put new arguments at the end or mention very clearly in the changelog that an argument was added in the middle.

rpauszek: Yes, it'd be nice to have the optional plotting arguments forwarded to `imshow` but I figured save that for a separate ticket since it's not really in the scope of this one (and there's already some things out of scope potentially that have made their way into this PR)

And I see what you mean about the argument order, I'll move `channel` to the end

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +18,102 @@ class TiffFrame:
...
+        return self._align_image() if (self.is_rgb and self._align) else self._src.asarray()
+
+    @property
+    def raw_data(self):
         return self._src.asarray()
 
+    @property
+    def bit_depth(self):        
+        bit_depth = self._src.tags["BitsPerSample"].value
+        if self.is_rgb: # (int r, int g, int b)
```
JoepVanlier, line 84: Extra space before `#`.

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +18,102 @@ class TiffFrame:
...
     """
-    def __init__(self, page):
+    def __init__(self, page, align):
         self._src = page
+        self._align = align
+
+    def _align_image(self):
+        """ reconstruct image using alignment matrices from Bluelake 
+            return aligned image as a NumPy array 
+        """
```
JoepVanlier, line 28:
```suggestion
        """reconstruct image using alignment matrices from Bluelake return aligned image as a NumPy array"""
```

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +18,102 @@ class TiffFrame:
...
+
+        try:
+            description = json.loads(self._src.description)
+        except json.decoder.JSONDecodeError:
+            warnings.warn("File does not contain metadata. Only raw data is available")
+            return self.raw_data
+
+        try:
+            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2,3))
+                        for color in ('red', 'blue')]
```
JoepVanlier, line 54:
```suggestion
            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2,3))
                          for color in ('red', 'blue')]
```

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +18,102 @@ class TiffFrame:
...
+            return self.raw_data
+        
+        x_offset, y_offset = align_roi - roi
+        if not (x_offset == 0 and y_offset == 0):
+            align_mats = [correct_alignment_offset(mat, x_offset, y_offset) for mat in align_mats]
+
+        img = self.raw_data
+        rows, cols, _ = img.shape
+        for mat, channel in zip(align_mats, (0, 2)):
+            img[:,:,channel] = cv2.warpAffine(img[:,:,channel], mat, (cols,rows),
```
JoepVanlier, line 68:
```suggestion
            img[:, :, channel] = cv2.warpAffine(img[:, :, channel], mat, (cols, rows),
```

lumicks/pylake/correlated_stack.py
```diff
@@ -14,13 +18,102 @@ class TiffFrame:
...
+            return np.dot(back_translation, np.dot(original, translation))[:2, :]
+
+        try:
+            description = json.loads(self._src.description)
+        except json.decoder.JSONDecodeError:
+            warnings.warn("File does not contain metadata. Only raw data is available")
+            return self.raw_data
+
+        try:
+            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2,3))
```
JoepVanlier, line 53:
```suggestion
            align_mats = [np.array(description[f"Alignment {color} channel"]).reshape((2, 3))
```

lumicks/pylake/correlated_stack.py
```diff
@@ -129,36 +226,45 @@ def from_data(cls, data, name=None, start_idx=0, stop_idx=None):
...
+                # display with 1-based index for frames and total frames
+                plt.title(f"{self.name} [frame {frame+1}/{self.num_frames}]")
 
     def _get_frame(self, frame=0):
         if frame >= self.num_frames or frame < 0:
             raise IndexError("Frame index out of range")
         return self.src.get_frame(self.start_idx + frame)
 
-    def plot_correlated(self, channel_slice, frame=0, reduce=np.mean):
+    def plot_correlated(self, channel_slice, frame=0, reduce=np.mean, channel="rgb"):
```
JoepVanlier, line 267: Not the fault of this PR, but I had a look at this function again, and reduce is never forwarded to `downsampled_over`.

JoepVanlier (commented):
Looks pretty solid to me. I only have a few minor nitpicks that showed up in my linter this time around.

changelog.md
```diff
@@ -2,6 +2,7 @@
 
 ## v0.6.3 | t.b.d.
 
+* Add `align` kwarg to `CorrelatedStack`. If enabled, this option will automatically align the color channels for multi-channel images, provided that the alignment metadata from Bluelake is available.
```
JoepVanlier, line 5: Very minor, but it seems that two things were actually added:
* Add support for RGB images to `CorrelatedStack`.
* Add image alignment to `CorrelatedStack`. Image alignment is enabled by default and will automatically align the color channels for multi-channel images, provided that the alignment metadata from Bluelake is available. Image alignment can be disabled by specifying `align=False` when loading the `CorrelatedStack`.

JoepVanlier (approved):
Awesome! It looks good to me!

PR #60, opened 2020-10-13
downsample Slice to frequency

rpauszek:
this PR adds downsampling to a specific frequency for both Continuous and (possibly variable frequency) TimeSeries channels

### DIFF
--- a/docs/tutorial/file.rst
+++ b/docs/tutorial/file.rst
@@ -113,6 +113,9 @@ The relative time values in seconds can also be accessed directly::
 The above examples use the `force1x` channel.
 A full list of available channels can be found on the :class:`~lumicks.pylake.File` reference page.
 
+Slicing
+^^^^^^^
+
 By default, entire channels are returned from a file::
 
     everything = file.force1x
@@ -147,6 +150,47 @@ Once you access the raw data, those are regular arrays which use regular array i
 
     channel_slice = file.force1x['1.5s':'20s']  # timestamps
     data_slice = file.force1x.data[20:40]  # indices into the array
+
+Downsampling
+^^^^^^^^^^^^
+
+A slice can be downsampled using various methods.
+
+To downsample to a specific frequency use `downsampled_to` with the desired frequency in Hz::
+
+    channel = file.force1x # original frequency 78125 Hz
+    timestep = np.diff(channel.timestamps[:2]) * 1e-9        # timestep 12.8 us
+
+    ds_channel = channel.downsampled_to(3125)
+    ds_timestep = np.diff(ds_channel.timestamps[:2]) * 1e-9  # timestep 320 us
+
+By default, this method will take the mean of every N samples where N is defined as the ratio between the two sampling times. 
+This can cause issues when N isn't an integer, leading to an unequal number of points contributing to each point in the 
+downsampled channel. To automatically find the nearest higher frequency that will fulfill this requirement, use the `method="ceil"`::
+
+    ds_channel2 = channel.downsampled_to(3126, method="ceil")
+    ds_timestep2 = np.diff(ds_channel2.timestamps[:2]) * 1e-9  # timestep 307.2 us
+
+For data that is recorded with variable sampling frequencies, it is usually not possible to downsample to a 
+single sample rate, while maintaining an equal number of samples per downsampled sample. To force downsampling 
+to a single frequency in the case of variable sample rates, use `method="force"`::
+
+    variable_channel = file.downsampled_force1x 
+    variable_ds_channel = variable_channel.downsampled_to(3125, method="force")
+
+Note that this same flag can also be used to force a specific downsampling rate for non-integer downsampling rates.
+
+A slice can also be downsampled over arbitrary time segments by using `downsampled_over` and supplying a 
+list of `(start, stop)` tuples indicating over which ranges to apply the function.
+
+Finally, a slice that contains equally spaced timestamps can be downsampled by a specific factor using `downsampled_by`
+*(note that the ratio of the original/final sampling frequencies must be an integer.)*::
+
+    channel = file.force1x # original frequency 78125 Hz
+    timestep = np.diff(channel.timestamps[:2]) * 1e-9        # timestep 12.8 us
+
+    ds_channel = channel.downsampled_by(5)
+    ds_timestep = np.diff(ds_channel.timestamps[:2]) * 1e-9  # timestep 64 us
 
 Calibrations
 ------------
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -88,6 +88,10 @@ class Slice:
         except AttributeError:
             return None
 
+    @property
+    def _timesteps(self):
+        return self._src._timesteps
+
     def downsampled_over(self, range_list, reduce=np.mean, where='center'):
         """Downsample channel data based on timestamp ranges. The downsampling function (e.g. np.mean) is evaluated for
         the time between a start and end time of each block. A list is returned that contains the data corresponding to
@@ -104,7 +108,7 @@ class Slice:
             cases, e.g. photon counts.
         where : str
             Where to put the final time point.
-            'center' time point is put at start + stop / 2
+            'center' time point is put at (start + stop) / 2
             'left' time point is put at start
 
         Examples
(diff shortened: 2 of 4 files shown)


JoepVanlier (changes requested):
Testing this actually made me realize an issue in `downsampled_by` :cry: . I was testing by comparing the timestamps of `downsampled_by` on `Continuous` to `downsampled_to` for an integer fraction of the samplerate (which leads to an integer downsampling factor).

The data produced is identical (good), but the timestamps are not (sad).

Rather than `(start + stop)//2` we should've used `(start + subset.timestamps[-1])//2` there, since `stop` is beyond the sample in question.

To prevent feature creep on this one, I suggest separating that issue from this one though. I'll make a ticket for it, but it would be one that I think we should pick up before release and make a warning about in the changelog.

Also, please add a changelog item to this ticket.

lumicks/pylake/channel.py
```diff
@@ -131,6 +131,32 @@ def downsampled_over(self, range_list, reduce=np.mean, where='center'):
 
         return Slice(TimeSeries(d, t), self.labels)
 
+    def downsampled_to(self, frequency, reduce=np.mean, where='center'):
+        """Return a copy of this slice which is downsampled by `factor`
+
+        Parameters
+        ----------
+        frequency : int
+            The desired downsampled frequency downsampled (per second)
```
JoepVanlier, line 140: `The desired downsampled frequency (Hz)`?

lumicks/pylake/channel.py
```diff
@@ -131,6 +131,32 @@ def downsampled_over(self, range_list, reduce=np.mean, where='center'):
...
+        reduce : callable
+            The `numpy` function which is going to reduce multiple samples into one.
+            The default is `np.mean`, but `np.sum` could also be appropriate for some
+            cases, e.g. photon counts.
+        where : str
+            Where to put the final time point.
+            'center' time point is put at start + stop / 2
+            'left' time point is put at start
```
JoepVanlier, line 148: Maybe a small warning about non-integer ratios not giving an equal number of samples per resulting sample may be in order?

JoepVanlier (changes requested):
Mostly looks good to me. Would be nice to add some docs for this. Maybe under `Files and Channels`.

lumicks/pylake/channel.py
rpauszek: oops, copy/paste from `downsampled_by` for doc consistency but forgot to change that line

JoepVanlier (changes requested):
I discussed the issues with this downsampling ticket a bit more with the others and we do think it should even be an exception and not just a warning. The reason for this is that we must also cater to more novice users who may not be thinking about issues that could arise from non-integer downsampling ratios. Based on that I would propose the following:

Essentially, there are three types of channels to deal with.
1. Continuous.
2. Continuous-like (timeseries with just one time step). Most time series will actually just have one timestep, and therefore, making a special case for this would be beneficial, since that would allow more data to be resampled safely.
3. Time series.

### 1, 2. Continuous(-like).
Desired behavior: Exception should be raised for non-integer ratios of time steps, which should probably print the nearest integer ratio below and above so users can easily find one that works. A field should be added to override the exception. Perhaps `force=True` or something of the like. Upsampling should never be allowed, even with `force=True`.

### 3. Time series.
Desired behavior: Time series with more than one time step can never be `downsampled_to` completely safely, so the user should be made aware of what's happening here. So in this case, it should always throw unless the force flag is active.

### Time step
Regarding the check for whether it has a unique timestep, a private member timestep which returns `np.unique(np.diff(self.timestamps))` should do the trick to detect this. This adds some pretty serious overhead, but for timeseries data (which is usually low frequency) this should be OK. For the true continuous channels however, this might be a limiting problem, but here you can just specialize it and return the time step based on the samplerate (which is a lot cheaper than computing timestamps first).

What do you think?

lumicks/pylake/channel.py
```diff
@@ -131,6 +135,53 @@ def downsampled_over(self, range_list, reduce=np.mean, where='center'):
...
+        force : bool
+            if `True`, will allow downsampling for variable frequency slices and
+            for non-integer ratios of original and new frequencies.
+        """
+
+        def get_good_frequencies(frequency, target):
+            possible = np.arange(frequency)+1
+            choices = np.extract(np.mod(frequency, possible)==0, possible).astype(np.int)
+            ix = int(np.argwhere(choices < target)[-1])
```
JoepVanlier, line 161: This can fail when no lower frequency exists e.g. `get_good_frequencies(1e9/850, 400)`

But actually, maybe we could have a second parameter named method rather than force.

With `force` pushing for whatever frequency the user specified. And `nearest` just getting the next achievable timestep. It would also sidestep the issue of the floating point math and get you a much closer achievable sample rate than an integer one that's suggested:
```
def get_next_timestep_size(source_timestep, target_timestep):
    remainder = target_timestep % source_timestep
    return target_timestep - remainder
```

What do you think?

JoepVanlier (changes requested):
There are some issues with FP math that can cause problems with this implementation. Let's discuss it and wrap it up together this evening.

lumicks/pylake/channel.py
```diff
@@ -131,6 +135,58 @@ def downsampled_over(self, range_list, reduce=np.mean, where='center'):
...
+        reduce : callable
+            The `numpy` function which is going to reduce multiple samples into one.
+            The default is `np.mean`, but `np.sum` could also be appropriate for some
+            cases, e.g. photon counts.
+        where : str
+            Where to put the final time point.
+            'center' time point is put at (start + stop) / 2
+            'left' time point is put at start
+        method : str
+            How to handle non-integer frequency ratio.
```
JoepVanlier, line 154: Maybe express it in sample time (since that's what we are looking at for the criteria).
```
How to handle target sample times that are not exact multiples of the current sample time.

'safe' new sample time must be an exact multiple of the current sample time, else an exception is raised.
'ceil' rounds the sample rate up to the nearest frequency which fulfills this condition
'force' downsample data with the target input frequency; this will result in variable sample times and a variable number of sampling contributing to each target sample, but must be used for variable-frequency data
```

lumicks/pylake/tests/test_channels.py
```diff
@@ -333,3 +333,99 @@ def test_downsampling():
...
+    assert np.allclose(s2b.data, [2.5, 6.5, 10.0, 12.5, 14.5, 16.5, 18.5])
+
+    # to 3333 ns step
+    s3a = s.downsampled_to(3e5, where="left", method="force")
+    assert np.all(np.equal(s3a.timestamps, [0, 3333, 6666, 9999]))
+    assert np.allclose(s3a.data, [4.0, 10.0, 14.0, 17.5])
+    s3b = s.downsampled_to(3e5, where="center", method="force")
+    assert np.all(np.equal(s3b.timestamps, [1666, 4999, 8332, 11665]))
+    assert np.allclose(s3b.data, [4.0, 10.0, 14.0, 17.5])
+
```
JoepVanlier, line 431: last line should be empty (also no whitespace) :)

docs/tutorial/file.rst
```diff
@@ -142,6 +145,31 @@ Once you access the raw data, those are regular arrays which use regular array i
...
+A slice can be downsampled using various methods.
+
+To downsample to a specific frequency use `downsampled_to` with the desired frequency in Hz::
+
+    channel = file.force1x # original frequency 78125 Hz
+    timestep = np.diff(channel.timestamps[:2]) * 1e-9        # timestep 12.8 us
+
+    ds_channel = channel.downsampled_to(3125)
+    ds_timestep = np.diff(ds_channel.timestamps[:2]) * 1e-9  # timestep 320 us
+
```
JoepVanlier, line 160: Could add an example of `force` and `ceil`.

JoepVanlier (commented):
From my side, code-wise I think it's pretty much there :)

Just some minor comments on the exceptions presented to the user and docs.

docs/tutorial/file.rst
```diff
@@ -148,6 +151,40 @@ Once you access the raw data, those are regular arrays which use regular array i
...
+This method can only be applied successfully if the ratio of the current and target frequencies
+is an integer. To automatically find the nearest higher frequency that will fulfill this requirement,
```
JoepVanlier, line 168: Might be nice to specify what successfully means here:

`By default, this method will take the mean of every N samples where N is defined as the ratio between the two sampling times. This can cause issues when N isn't an integer, leading to an unequal number of points contributing to each point in the downsampled channel.` and then continue with the To automatically you already have

docs/tutorial/file.rst
```diff
@@ -148,6 +151,40 @@ Once you access the raw data, those are regular arrays which use regular array i
...
+use the `method="ceil"`::
+
+    ds_channel2 = channel.downsampled_to(3126, method="ceil")
+    ds_timestep2 = np.diff(ds_channel2.timestamps[:2]) * 1e-9  # timestep 307.2 us
+
+For data that is recorded with variable sampling frequencies, use `method="force"`.
```
JoepVanlier, line 174: `For data that is recorded with variable sampling frequencies, it is usually not possible to downsample to a single sample rate, while maintaining an equal number of samples per downsampled sample. To force downsampling to a single frequency in the case of variable sample rates, use method="force"::`

then the code example. And the maybe something to mention that this same flag can also be used to force the downsampling rate in the other case:

`Note that this same flag can also be used to force a specific downsampling rate for non-integer downsampling rates.`

JoepVanlier (approved):
Looking good man! I have some minor comments on the docs, mostly around making them a little more explicit so people understand the why better (for the integer downsampling ratios).

PR #61, opened 2020-10-16
Downsampled over fix

JoepVanlier:
**Why this PR?**

This PR fixes two issues:

- `downsampled_over` used to be inconsistent with `downsampled_by`. The two did not produce the same timestamps when downsampling over integer frequency ratios. When downsampling the new timestamp should not have been `(subset.start + subset.stop)/2` but `(subset.timestamps[0] + subset.timestamps[-1])/2`

- The second one came  up during WIT testing, which is that when the h5 is smaller than the tiff exported, you can't use `correlatedstack` at all anymore. This PR fixes that by omitting those points in the downsampled series and raises a warning when this is the case.

- Tests were added to prevent the same inconsistency from popping up again.

### DIFF
--- a/lumicks/pylake/tests/test_channels.py
+++ b/lumicks/pylake/tests/test_channels.py
@@ -8,14 +8,14 @@ def test_calibration_timeseries_channels():
     time_field = 'Stop time (ns)'
     mock_calibration = ForceCalibration(time_field=time_field,
                                         items=[
-                                           {'Calibration Data': 50, time_field: 50},
-                                           {'Calibration Data': 20, time_field: 20},
-                                           {'Calibration Data': 30, time_field: 30},
-                                           {'Calibration Data': 40, time_field: 40},
-                                           {'Calibration Data': 80, time_field: 80},
-                                           {'Calibration Data': 90, time_field: 90},
-                                           {'Calibration Data': 120, time_field: 120},
-                                       ])
+                                            {'Calibration Data': 50, time_field: 50},
+                                            {'Calibration Data': 20, time_field: 20},
+                                            {'Calibration Data': 30, time_field: 30},
+                                            {'Calibration Data': 40, time_field: 40},
+                                            {'Calibration Data': 80, time_field: 80},
+                                            {'Calibration Data': 90, time_field: 90},
+                                            {'Calibration Data': 120, time_field: 120},
+                                        ])
 
     # Channel should have calibration points 40, 50 since this is the only area that has force data.
     cc = channel.Slice(channel.TimeSeries([14, 15, 16, 17], [40, 50, 60, 70]),
@@ -247,6 +247,7 @@ def test_timetags_indexing():
 def test_time_indexing():
     """String time-based indexing"""
     s = channel.Slice(channel.TimeSeries([1, 2, 3, 4, 5], [1400, 2500, 16e6, 34e9, 122 * 1e9]))
+
     # --> in time indices: ['0ns', '1100ns', '15.9986ms', '33.99s', '2m 2s']
 
     def assert_equal(actual, expected):
@@ -346,7 +347,7 @@ def test_seconds_property():
 def test_continuous_downsampling_to():
     # Continuous
     d = np.arange(1, 24)
-    s = channel.Slice(channel.Continuous(d, 0, 500)) # 2 MHz
+    s = channel.Slice(channel.Continuous(d, 0, 500))  # 2 MHz
 
     # to 1000 ns step
     s2a = s.downsampled_to(1e6, where='left')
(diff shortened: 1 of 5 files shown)


lumicks/pylake/tests/test_channels.py
```diff
@@ -349,15 +349,15 @@ def test_downsampling_to():
     assert np.all(np.equal(s2a.timestamps, [0, 20, 40, 60, 80, 100, 120]))
     assert np.allclose(s2a.data, [2.5, 6.5, 10.0, 12.5, 14.5, 16.5, 18.5])
     s2b = s.downsampled_to(0.5e8, where="center")
-    assert np.all(np.equal(s2b.timestamps, [10, 30, 50, 70, 90, 110, 130]))
+    assert np.all(np.equal(s2b.timestamps, np.array([15/2, 55/2, 90/2, 130/2, 170/2, 210/2, 250/2], dtype=np.int)))
```
rpauszek, line 352: The behavior here to me seems odd. As a user, I would expect that the `downsampled_to` method would return a continuous slice with timesteps corresponding to the requested frequency. However, in this implementation (and this test specifically) the timesteps are not consistent (20, 17.5, 20, 20, 20, 20). I think for long slices, repeated offsets like this could result in the end of the slice being "out of phase" in a sense with the beginning of the slice

lumicks/pylake/tests/test_channels.py
```diff
@@ -372,14 +372,55 @@ def test_downsampling_to():
...
 
+
+def test_downsampling_consistency():
+    d = np.arange(1, 24)
+    s = channel.Slice(channel.Continuous(d, 0, 10))
+
+    # Multiple of 5 should downsample to the same irrespective of the method
+    # Source frequency was 1e9 / 10 Hz. So we go to .2e8 Hz.
+    s1 = s.downsampled_to(.2e8)
+    s2 = s.downsampled_by(5)
```
rpauszek, line 394: Maybe include a test with a not so well-behaved timestep (like 7ns)? I have a feeling this might also bring out the odd behavior mentioned in `test_downsampling_to()`

rpauszek (commented):
Overall I think it looks good, and I understand why we would want to have consistent behavior between `downsampled_by()` and `downsampled_over()` but, as noted below, I think it can cause some strange results for the `downsampled_to()` method. Maybe it'd be easier to just completely decouple those 2 methods and leave everything here as is?

lumicks/pylake/tests/test_channels.py
JoepVanlier: I don't really think so. Please correct me if I'm wrong, but while the variable delta timestamp is an annoying inconvenience, if you just use the timestamps as they are computed now, the new way is the more correct one (see figures below). I agree with you that if you assume a constant timestep, where there is none you'd go out of phase, but this doesn't happen if you make sure you use the actual timestamps. This is also why `downsampled_over` always returns a `TimeSeries` and not a `Continuous`.

If we approach the limit of low samples per new sample we get the following:

Old_ds

New_ds

Code to reproduce the plots:
```python
version = 'New'
t = np.arange(0, 20) * 1e9
y = np.sin(t * 1e-9)

s = Slice(TimeSeries(y, t))
ds = s.downsampled_to(.75)
ds2 = s.downsampled_to(.66)

plt.figure(figsize=(10, 7))
plt.subplot(2, 1, 1)
plt.title(f'{version} implementation')
plt.plot(s.timestamps, s.data)
plt.plot(ds.timestamps, ds.data, '.', markersize=12)
plt.plot(ds2.timestamps, ds2.data, '.', markersize=12)

plt.subplot(2, 1, 2)
plt.title('Residual between interpolated high frequency and downsampled')
plt.plot(ds.timestamps, ds.data - np.interp(ds.timestamps, s.timestamps, s.data))
plt.plot(ds2.timestamps, ds2.data - np.interp(ds2.timestamps, s.timestamps, s.data))

plt.savefig(f'{version}_ds.png')
```

For more samples per new sample, we can't use sines as test signals anymore, because the effect of the boxcar filtering starts suppressing the gain. But if we just use a single slope, we get:

Old_ds

New_ds

I do have worries about this whole non-integer resampling business though, which is:
1. What non-integer resampling rates do to the noise for low sample per point counts (and thereby to statistical models that may assume a single standard deviation for all points), since a different number of samples go into each sample (noise scaling with `1/sqrt(n)`). 
2. People using `.data` without realizing the timestamps are not equidistant. There should probably be a big warning about this in the docs. I would also consider emitting a warning for non-integer ratios in `downsampled_to`.
3. `np.sum` makes little sense when you don't guarantee the same sample count. It raises the question whether this setup with the `reduce` function makes much sense in this specific function.
4. Whether we should be using better downsampling methods that have better signal preservation below nyquist and better rejection above nyquist (brickwall rather than a simple boxcar filter). A higher quality resampling filter would be nice, but while this is pretty easy for `Continuous` channels, it can get very hairy for non-uniformly sampled `TimeSeries`.

But this worry applies to both methods for setting the timestamps.

lumicks/pylake/tests/test_channels.py
JoepVanlier: I can add one with 7 ns source timestep. As long as the `downsampled_by` ratio is integer, everything should be fine in the new implementation. For non-integer fractions of the original samplerate you cannot get consistency by definition, only tradeoffs, as you're going into resampling territory.

rpauszek (approved):
Looks good to me

PR #62, opened 2020-10-20
Missing meta

rpauszek:
This PR fixes a bug where if a file contains a scan that is missing metadata, none of the scans could be loaded. Now a warning is generated and the scan that is missing metadata is ignored while all other scans are accessible.

_**Update Nov 2**_
This PR is now based on #68

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -281,7 +281,11 @@ class Kymo(PhotonCounts, ExcitationLaserPower):
         start = h5py_dset.attrs["Start time (ns)"]
         stop = h5py_dset.attrs["Stop time (ns)"]
         name = h5py_dset.name.split("/")[-1]
-        json_data = json.loads(h5py_dset[()])["value0"]
+        try:
+            json_data = json.loads(h5py_dset[()])["value0"]
+        except KeyError:
+            raise KeyError(f"Scan '{name}' is missing metadata and cannot be loaded")
+
         return cls(name, file, start, stop, json_data)
 
 
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@
 * Add widget to graphically slice a `Slice` in Jupyter Notebooks. It can be opened by calling `channel.range_selector`. For more information, see [notebook widgets](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html).
 * Fixed bug in `CorrelatedStack.plot_correlated` which lead to the start index of the frame being added twice when the movie had been sliced.
 * Fixed bug in `File.point_scans` to return `PointScan` instances. Previously, attempts to access this property would cause an error due to missing `PointScan.from_dataset` method. **Note that the `__init__` method arguments of `PointScan` have been changed to be in line with `Kymo` and `Scan`!**
+* Fixed bug in `File.scans` so that a warning is generated when a scan is missing metadata. Other scans that can be loaded properly are still accessible.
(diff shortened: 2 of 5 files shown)


mwilkowski (commented):
No complaints about the code. What we do in bluelake with commit messages and what I think is also happening in pylake is that we usually prefix them with the module name. So in this case I would expect something in the vein of `docs: Update changelog, and `kymo: Add missing metadata unit test`

JoepVanlier (approved):
Looks good to me. 👍

lumicks/pylake/kymo.py
```diff
@@ -281,7 +281,12 @@ def from_dataset(cls, h5py_dset, file):
         start = h5py_dset.attrs["Start time (ns)"]
         stop = h5py_dset.attrs["Stop time (ns)"]
         name = h5py_dset.name.split("/")[-1]
-        json_data = json.loads(h5py_dset[()])["value0"]
+        try:
+            json_data = json.loads(h5py_dset[()])["value0"]
+        except KeyError:
+            warnings.warn(f"Scan '{name}' is missing metadata and cannot be loaded")
+            return None
```
dean0x7d, line 288: An argument can be made here that his should actually re-raise the exception with more information, rather than translate that exception into a `None` result:
```py
        try:
            json_data = json.loads(h5py_dset[()])["value0"]
        except KeyError:
            raise KeyError(f"Scan '{name}' is missing metadata and cannot be loaded")
```

```py
        def try_from_dataset(*args):
            try:
                return Scan.from_dataset(*args)
            except Exception as e:
                warnings.warn(str(e))
                return None

        scan_objects = [(name, try_from_dataset(dset, self)) for name, dset in self.h5["Scan"].items()]
        return {name: scan for name, scan in scan_objects if scan is not None}
```

There are two reasons:
1. Returning `None` for a specific set of errors means that we get one set of errors the usual way (exceptions) and the other as `None` results. However, this is pretty surprising. We usually expect to get all errors the same way unless there's a really good reason -- in which case it should be mentioned in the docstring (it's part of the API and error handling changes are also API breaking changes).
2. Previously, `Scan.from_dataset()` would always return a proper object or raise an exception. Now it can return `None` which means that it has `Optional` semantics. Since `.from_dataset()` is a constructor this can imply that type itself has optional semantics. If that's true, it shifts the burden to users to check `if obj is None`.

Essentially:
```py
try:
    scan = Scan.from_dataset(...)
    if scan is not None:
        # good path
    else:
        # failure path 1
except:
    # failure path 2
```
Having two failure paths is sometime justified, but it's more of a thing where you need to explicitly prove "yes, this is the best API for this type". It should not be done accidentally because it does increase the API area and it adds more stuff to document and more stuff for users to learn.

JoepVanlier: Good point!

rpauszek: I see your point here. One of the issues is that if `Scan.from_dataset()` throws an exception, then any code requesting the list of available objects will break (`File.__str__()` and `File.scans`) Maybe the best thing to do is throw the exception from `from_dataset()` and handle that in the other two methods? Should be a pretty simple fix. What do you think?

rpauszek: which now re-reading your suggestion is exactly what I just said...sorry. I'll go ahead and change it up

dean0x7d: Indeed, that's what I was suggesting with the `try_from_dataset()` helper function, but I left it a bit obscure as the example code.

dean0x7d (approved):
LGTM!

PR #63, opened 2020-10-23
channel: add range slicer

JoepVanlier:
This PR generalizes the interactive range slicer to be used on channel data in addition to on F,d curves.

Why this PR:
- Makes it easier to extract chunks of channel data that one is interested in for further analysis.
- Extending this feature from just F,d to channels in general was requested.
- Makes sure that the widget gets a copy of the timestamp ranges involved rather than a reference (so that if the user messes with the ranges, the widget doesn't actually break)

image

Docs here: https://lumicks-pylake.readthedocs.io/en/slice_selector/tutorial/nbwidgets.html

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -58,3 +58,13 @@ Kymotracking
     track_greedy
     track_lines
     filter_lines
+
+Notebook widgets
+----------------
+
+.. autosummary::
+    :toctree: _api
+
+    FdRangeSelector
+    nb_widgets.fd_selector.SliceRangeSelectorWidget
+    nb_widgets.fd_selector.FdRangeSelectorWidget
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -66,5 +66,5 @@ Notebook widgets
     :toctree: _api
 
     FdRangeSelector
-    nb_widgets.fd_selector.SliceRangeSelectorWidget
+    nb_widgets.fd_selector.SliceRangeSelector
     nb_widgets.fd_selector.FdRangeSelectorWidget
--- a/lumicks/pylake/nb_widgets/fd_selector.py
+++ b/lumicks/pylake/nb_widgets/fd_selector.py
@@ -3,7 +3,7 @@ import numpy as np
 import time
 
 
-class SliceRangeSelectorWidget:
+class SliceRangeSelector:
     def __init__(self, channel_slice, axes=None, show=True):
         self._axes = axes if axes else plt.axes(label=f"lk_slice_widget_{time.time()}")
         self.slice = channel_slice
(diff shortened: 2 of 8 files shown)


docs/api.rst
```diff
@@ -58,3 +58,13 @@ Kymotracking
...
+
+Notebook widgets
+----------------
+
+.. autosummary::
+    :toctree: _api
+
+    FdRangeSelector
+    nb_widgets.fd_selector.SliceRangeSelectorWidget
+    nb_widgets.fd_selector.FdRangeSelectorWidget
```
mcsmurf, line 70: nitpick: We can lose the last `Widget` in the name

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -3,14 +3,14 @@
...
-        self.time = self.fd_curve.f.timestamps
+class SliceRangeSelectorWidget:
+    def __init__(self, channel_slice, axes=None, show=True):
+        self._axes = axes if axes else plt.axes(label=f"lk_slice_widget_{time.time()}")
+        self.slice = channel_slice
+        self.time = self.slice.timestamps
         self.connection_id = None
         self.current_range = []
-        self.ranges = []
+        self._ranges = []
```
mcsmurf, line 13: Has anything changed here? If the meaning has changed, a more descriptive name would be preferred.

lumicks/pylake/nb_widgets/fd_selector.py
```diff
@@ -145,9 +162,10 @@ def update_plot(self, curve):
 
     @property
     def ranges(self):
+        """Return list of range timestamps"""
```
mcsmurf, line 165: Perhaps this can now be `range_timestamps` since it's not really `ranges` being returned

mcsmurf (approved):
Good stuff! A few comments about naming but looks great otherwise :rocket:

docs/api.rst
JoepVanlier: I can indeed kill `Widget` for `SliceRangeSelector`. Will do! 👍

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: There's an attribute that grants access to a copy of ranges now (named `ranges`). The issue with allowing direct access to it is that the user can edit them and this might cause issues for the widget if the users changes them to some invalid format. This way, we communicate that `_ranges` is private and that once they edit the actual source data, all bets are off.

API-wise, nothing changes other than that they get a copy rather than a reference of these values.

lumicks/pylake/nb_widgets/fd_selector.py
JoepVanlier: This is part of the public API already and there are notebooks out that depend on this name. I could change it, but then we'd still need to keep `ranges` around with a deprecation notice.

PR #65, opened 2020-10-27
GitHub actions

JoepVanlier:
**What's this PR?**
- Switching from Travis to GitHub actions for Continuous Integration.

**Why?**
- Travis queue is getting prohibitively long.
- The virtual environment hack we have to use on Travis is unfortunate.
- Why not?


dean0x7d (approved):
Faster and better Python version coverage. Yes, please!

PR #66, opened 2020-10-28
CorrelatedStack: bugfix in plot_correlated to prevent misalignment

JoepVanlier:
**Why this PR?**
- There is a critical bug in `CorrelatedStack.plot_correlated` that happens when a `TiffStack` is sliced and then plotted correlated to `Channel` data. In this case, the start index is added twice since `get_frame` already takes into account the start index.

**What this PR adds**
- A fix for the bug and a regression test so that this does not happen again.


mwilkowski (approved):
Cool tests

PR #68, opened 2020-10-31
Pointscan fromdata

rpauszek:
This PR extends PR #62, fixing a bug in creating `PointScan` instances from the `File.point_scans` property. Previously the `from_dataset` class method was missing from `PointScan` and attempts to list or create point scans in `File` would throw an exception. 

* `from_dataset` was added to `PointScan` and the `__init__` argument list was changed to match those of `Kymo` and `Scan`
* unit tests were added for `point_scan` module
* `File.scans` was changed to utilize the `_get_object_dictionary` method

**_Update Nov 2._**
This PR no longer relies on PR #62 . Instead, this PR should be merged first and #62 adds exception handling for missing metadata


lumicks/pylake/tests/test_point_scan.py
```diff
@@ -0,0 +1,40 @@
...
+    if f.format_version == 2:
+        ps = f.point_scans["PointScan1"]
+        for plot_func in (ps.plot_red, ps.plot_green, ps.plot_blue):
+            plot_func()
+            assert np.allclose(np.sort(plt.xlim()), [0, 0.0625 * 64])
+            assert np.allclose(np.sort(plt.ylim()), [0, 8])
+
+        ps.plot_rgb(lw=5)
+        assert np.allclose(np.sort(plt.xlim()), [0, 0.0625 * 64])
+        assert np.allclose(np.sort(plt.ylim()), [0, 8])
```
JoepVanlier, line 40: nitpick: End the file with a newline.

lumicks/pylake/point_scan.py
```diff
@@ -1,4 +1,5 @@
 import json
+import warnings
```
JoepVanlier, line 2: I don't see where this is used. If not, please remove it.

changelog.md
```diff
@@ -13,6 +13,7 @@
 * Add widget to graphically slice a `Slice` in Jupyter Notebooks. It can be opened by calling `channel.range_selector`. For more information, see [notebook widgets](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html).
 * Fixed bug in `CorrelatedStack.plot_correlated` which lead to the start index of the frame being added twice when the movie had been sliced.
 * Fixed bug in `File.scans` so that a warning is generated when a scan is missing metadata. Other scans that can be loaded properly are still accessible.
+* Fixed bug in `File.point_scans` to return `PointScan` instances. **Note that the `__init__` method arguments of `PointScan` have been changed to be in line with `Kymo` and `Scan`!**
```
JoepVanlier, line 16: 👍 for the warning

JoepVanlier (approved):
Looks good to me. I only have two very minor nitpicks. Thanks for picking this up when you noticed it!

lumicks/pylake/file.py
```diff
@@ -187,7 +187,8 @@ def _get_photon_time_tags(self, name):
     def _get_object_dictionary(self, field, cls):
         if field not in self.h5:
             return dict()
-        return {name: cls.from_dataset(dset, self) for name, dset in self.h5[field].items()}
+        object_list = [(name, cls.from_dataset(dset, self)) for name, dset in self.h5[field].items()]
+        return {name: obj for name, obj in object_list if obj is not None}
```
dean0x7d, line 191: AFAIK `cls.from_dataset()` can't ever return `None`. Or at least it shouldn't. It should only signal errors by raising an exception, not returning `None`. I think that this was a mistake in the original `.scans`. If so, you could just omit this part.

dean0x7d (changes requested):
Very nice catch! I'm just requesting changes here because that `None` part is not clear to me.

lumicks/pylake/file.py
rpauszek: This is the result of #62 that this PR is based on. Because a `Scan` might have missing metadata it can't be loaded and `from_dataset()` returns None. I was debating on whether that check should also be done for `PointScan` for consistency with `Scan` and `Kymo`. Either way, having all three call `_get_object_dictionary()` is best to avoid duplicated code

dean0x7d: Ah, I missed that part. Then things are fine here. Just maybe take a look at the changelog line.

dean0x7d (approved):

dean0x7d (approved):
Great catch and it's great to have it covered with tests now.

JoepVanlier (approved):
👍

PR #69, opened 2020-11-02
ci: switch to headless backend for mpl tests

JoepVanlier:
**Why this PR?**
Builds are occassionally failing on GitHub actions because of Tk dependencies (see error below).
`_tkinter.TclError: Can't find a usable tk.tcl in the following directories`

This is not something we want to explicitly test. In this PR we switch to a headless `Agg` backend.


rpauszek (approved):
LGTM

PR #70, opened 2020-11-03
ci import time tests

JoepVanlier:
This PR adds a test which measures how much time it takes to import Pylake.

**Why this PR?**
As we are pulling in more and more dependencies, we might want to keep an eye on import times.

image

**Why this implementation?**
- Option A. Downside: Measures time taken to spin up python itself.
```python
def import_pylake():
  subprocess.run(['python', '-c', 'import lumicks.pylake'])

  times = timeit.repeat(import_pylake, number=1, repeat=3)
  print(f"Pylake import time: {np.mean(times)} +- {np.std(times)}")
```

- Option B. Downside: Little control over output string. Reparsing the output of timeit would have been messy since it specifies time with prefixes to the units.
```
subprocess.run(['python', '-m', 'timeit', '--number=1', '--repeat=1', 'import lumicks.pylake'])
```

- Option C. This is more or less the implementation used now.
```
code = dedent("""\
  import time
  tic = time.time()
  import lumicks.pylake
  print(time.time() - tic)
""")

times = [float(subprocess.check_output(['python', '-c', code])) for i in range(repeats)]
print(f"  pylake import time: {np.mean(times)} +- {np.std(times)}")
```

### DIFF
--- a/lumicks/pylake/tests/test_import_time.py
+++ b/lumicks/pylake/tests/test_import_time.py
@@ -1,0 +1,21 @@
+from textwrap import dedent
+import numpy as np
+import subprocess
+import sys
+import pytest
+
+
+@pytest.mark.slow
+def test_disabling_capturing(report_line):
+    repeats = 3
+
+    code = dedent("""\
+        import time
+        tic = time.time()
+        import lumicks.pylake
+        print(time.time() - tic)
+        """)
+
+    times = [float(subprocess.check_output([f'{sys.executable}', '-c', code])) for i in range(repeats)]
+
+    report_line(f"Module import time: {np.mean(times):.2f} +- {np.std(times):.2f} seconds (N={repeats})")
--- a/.github/workflows/pylake_test.yml
+++ b/.github/workflows/pylake_test.yml
@@ -21,5 +21,4 @@ jobs:
         python -m pip install --upgrade pip
         pip install .
     - name: Pytest
-      run: |
-        pytest --color=yes
+      run: pytest lumicks/pylake --color=yes --runslow
--- a/lumicks/pylake/tests/conftest.py
+++ b/lumicks/pylake/tests/conftest.py
@@ -5,9 +5,25 @@ import json
 import matplotlib.pyplot as plt
 
 
-def pytest_configure():
+def pytest_addoption(parser):
+    parser.addoption(
+        "--runslow", action="store_true", default=False, help="run slow tests"
+    )
+
+
+def pytest_collection_modifyitems(config, items):
+    if config.getoption("--runslow"):
+        return
+    skip_slow = pytest.mark.skip(reason="need --runslow option to run")
+    for item in items:
+        if "slow" in item.keywords:
+            item.add_marker(skip_slow)
+
+
+def pytest_configure(config):
     # Use a headless backend for testing
     plt.switch_backend('agg')
+    config.addinivalue_line("markers", "slow: mark test as slow to run")
 
 
 # We generate mock data files for different versions of the Bluelake HDF5 file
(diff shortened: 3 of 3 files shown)


lumicks/pylake/tests/test_import_time.py
```diff
@@ -0,0 +1,21 @@
...
+        import time
+        tic = time.time()
+        import lumicks.pylake
+        print(time.time() - tic)
+        """)
+
+    times = [float(subprocess.check_output(['python', '-c', code])) for i in range(repeats)]
+
+    with capsys.disabled():
+        print(os.environ['TEST_IMPORT_TIME'])
```
dean0x7d, line 20: You can use `pytest.mark` for this. E.g. see `pytest.mark.slow`: https://pytest.org/en/stable/example/simple.html#control-skipping-of-tests-according-to-command-line-option

lumicks/pylake/tests/test_import_time.py
```diff
@@ -0,0 +1,21 @@
...
+    repeats = 3
+
+    code = dedent("""\
+        import time
+        tic = time.time()
+        import lumicks.pylake
+        print(time.time() - tic)
+        """)
+
+    times = [float(subprocess.check_output(['python', '-c', code])) for i in range(repeats)]
```
dean0x7d, line 17: Use `sys.executable` to get the same Python as you're running in now. Otherwise, you play the OS lottery of what Python (if any) you'll get.
```suggestion
    times = [float(subprocess.check_output([f'{sys.executable}', '-c', code])) for i in range(repeats)]
```

lumicks/pylake/tests/test_import_time.py
JoepVanlier: 👍

lumicks/pylake/tests/test_import_time.py
JoepVanlier: 👍

lumicks/pylake/tests/test_import_time.py
```diff
@@ -0,0 +1,23 @@
...
+        tic = time.time()
+        import lumicks.pylake
+        print(time.time() - tic)
+        """)
+
+    times = [float(subprocess.check_output([f'{sys.executable}', '-c', code])) for i in range(repeats)]
+
+    with capsys.disabled():
+        sys.stdout.write(f"\n \033[92mImport takes: "
+                         f"{np.mean(times):.2f} +- {np.std(times):.2f} seconds (N={repeats})\033[0m")
```
dean0x7d, line 23: Consider dropping the newline and `Import takes` to better line up the results (the file name `test_import_time` already says what the time is). For example, `f": {mean} +-{std} seconds (N={repeats})"` would produce this:
```
lumicks/pylake/tests/test_image.py ......                                [ 83%]
lumicks/pylake/tests/test_import_time.py: 0.64 +-0.01 seconds (N=3).     [ 84%]
lumicks/pylake/tests/test_kymotracker.py ...................             [ 94%]
```

JoepVanlier: You get this instead:
image

JoepVanlier: How about abusing atexit? 👼 

image

dean0x7d: Do it! 🚀

JoepVanlier: I set it up as a fixture now. That way we can have different tests use the same reporting mechanism if this is needed again in the future. Or do you think this is premature and should I just keep it inside that one test for now?

dean0x7d: Sounds perfect!

dean0x7d (approved):
LGTM! Just squash the git history.

PR #71, opened 2020-11-03
[wip] code coverage

JoepVanlier:
Observations so far:
- Codecov.io seems ok, but slow.
- GitHub artefacts can't be browsed. You get a zip file with the results, which is not so useful for a coverage report.


mwilkowski:
is this still wip ?

PR #72, opened 2020-11-03
release: readying for release v0.7.0

JoepVanlier:
**Why this PR?**
Would like to release v0.7.0


dean0x7d (commented):
Wow, big release! 🎉 

A few suggestions for the changelog:
* When it gets this long it's good to split it into sections (I think that the `###` level in markdown should work). It find that it's good to have "new features", "bug fixes" and "breaking changes" as the 3 sections.
* Order the points in order of relevance to most users. E.g. right now, the top 2 points are about RGB images and alignment which is only relevant for a couple of users (only the latest systems with specific features). On the other hand `Slice.seconds`, `Slice.downsampled_to()`, `save_as()`, the video export is going to be useful for almost everyone. Those should be at the top.
* Be consistent with tenses: I see a mix of `Add` and `Fixed`. It should all be past tense: `Added`, `Fixed`.

JoepVanlier:
> Wow, big release! 🎉
> 
> A few suggestions for the changelog:
> 
> * When it gets this long it's good to split it into sections (I think that the `###` level in markdown should work). It find that it's good to have "new features", "bug fixes" and "breaking changes" as the 3 sections.
> * Order the points in order of relevance to most users. E.g. right now, the top 2 points are about RGB images and alignment which is only relevant for a couple of users (only the latest systems with specific features). On the other hand `Slice.seconds`, `Slice.downsampled_to()`, `save_as()`, the video export is going to be useful for almost everyone. Those should be at the top.
> * Be consistent with tenses: I see a mix of `Add` and `Fixed`. It should all be past tense: `Added`, `Fixed`.

All good points! Amended!

dean0x7d (approved):
Very nice!

PR #73, opened 2020-11-04
Kymo plot time limits

rpauszek:
This PR fixes a bug in `kymo._plot()` to ensure plotted time limits exclude fly-in/out times. Previously time limits were calculated from `kymo.start` and `kymo.stop` which can lead to subtle time discrepancies when comparing with force channels. Here the time limits are adjusted to the center timestamp of the first and last scan lines.

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -187,7 +187,8 @@ class Kymo(PhotonCounts, ExcitationLaserPower):
         import matplotlib.pyplot as plt
 
         width_um = self._ordered_axes()[0]["scan width (um)"]
-        duration = (self.stop - self.start) / 1e9
+        ts = self.timestamps
+        duration = (ts[0, -1] - ts[0, 0]) / 1e9
 
         default_kwargs = dict(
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymo.py
```diff
@@ -181,7 +181,9 @@ def _plot(self, image, **kwargs):
         import matplotlib.pyplot as plt
 
         width_um = self._ordered_axes()[0]["scan width (um)"]
-        duration = (self.stop - self.start) / 1e9
+        # Compute average pixel time along kymo line (time at the center of the scanline)
+        mean_times = np.mean(self.timestamps, axis=0).astype(np.int64)
+        duration = (mean_times[-1] - mean_times[0]) / 1e9
```
dean0x7d, line 186: I think that the `mean()` is redundant work here since we only care about the difference. This could be simplified to:
```py
        ts = self.timestamps
        duration = (ts[0, -1] - ts[0, 0]) / 1e9
```

dean0x7d (approved):
LGTM!

PR #74, opened 2020-11-06
scan/kymo: workaround for non-quantized start timestamp

JoepVanlier:
**What is this PR?**
Due to an issue with the implementation of the sted delay, the starting timestamp of a scan is not necessarily quantized to a multiple of the samplerate. This leads to an exception in the function which checks whether the start of the scan was truncated. In this PR, we allow for a small discrepancy here (within a time step) to accomodate for the quantization mishap (since the channel data can start a fraction of a sample after the scan start time).

**Why this PR?**
Since systems shipped that have this delay mechanism, we need to handle it gracefully in Pylake for both scans and kymos.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## v0.7.1 | t. b. d.
+* Add workaround for `Scan` and `Kymo` which could prevent valid scans and kymos from being opened when the `start` timestamp of a scan or kymo had a value before the actual start of the timeline channels. The cause of this subsample time difference was the lack of quantization of a delay when acquiring STED images.
+
 ## v0.7.0 | 2020-11-04
 
 #### New features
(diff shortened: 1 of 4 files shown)


rpauszek (approved):
The code itself looks fine to me; however I just want to make sure I'm understanding why this is still ok (since I haven't encountered this error or worked with exporting from Bluelake yet). Is it assumed that the discrepancy only occurs during the flyin time and therefore won't affect the actual image data? If so, then I think it's good to go

JoepVanlier:
> The code itself looks fine to me; however I just want to make sure I'm understanding why this is still ok (since I haven't encountered this error or worked with exporting from Bluelake yet). Is it assumed that the discrepancy only occurs during the flyin time and therefore won't affect the actual image data? If so, then I think it's good to go

The start timestamp of the `scan` is not quantized to the channel sample rate, which means it won't be the same. The workaround here is to snap it to the next sample. Yes, this will typically be during the fly in time.

PR #75, opened 2020-11-07
Kymo plot with downsampled force

rpauszek:
This PR adds functionality for automatically downsampling a force channel over the time ranges of the kymograph scan lines and plotting the image and downsampled force.

* added `Kymo.plot_with_force()`
* added `Kymo._downsample_channel()` to automatically calculate scanline time ranges and return appropriately downsampled channel

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -255,6 +255,58 @@ class Kymo(PhotonCounts, ExcitationLaserPower):
         image = image / np.max(image)
         return self._plot(image, **kwargs)
 
+    def _downsample_channel(self, n, xy, reduce=np.mean):
+        force = self.file._get_force(n, xy)
+        # downsample exactly over the scanline time range
+        min_times = self.timestamps[0, :].astype(np.int64)
+        max_times = self.timestamps[-1, :].astype(np.int64)
+        time_ranges = [(mini, maxi) for mini, maxi in zip(min_times, max_times)]
+        return force.downsampled_over(time_ranges, reduce=reduce, where="center")
+
+    def plot_with_force(self, force_channel, color_channel, aspect_ratio=0.25,
+                        reduce=np.mean, **kwargs):
+        """Plot kymo with force channel downsampled over scan lines
+
+        Parameters
+        ----------
+        force_channel: str
+            name of force channel to downsample and plot
+        color_channel: str
+            color channel of kymo to plot ('red', 'green', 'blue', 'rgb')
+        aspect_ratio: float
+            aspect ratio of the axes (i.e. ratio of y-unit to x-unit)
+        reduce : callable
+            The `numpy` function which is going to reduce multiple samples into one.
+            Forwarded to :func:`Slice.downsampled_over`
+        **kwargs
+            Forwarded to :func:`Slice.plot`.
+        """
+        def set_aspect_ratio(axis, ar):
+            """This function forces a specific aspect ratio, can be useful when aligning figures"""
+            axis.set_aspect(ar * np.abs(np.diff(axis.get_xlim())[0] / np.diff(axis.get_ylim()))[0])
+
+        import matplotlib.pyplot as plt
+        _, (ax1, ax2) = plt.subplots(2, 1)
+        
+        # plot kymo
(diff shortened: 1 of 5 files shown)


lumicks/pylake/kymo.py
```diff
@@ -249,6 +249,56 @@ def plot_rgb(self, **kwargs):
...
+        """Plot kymo with force channel downsampled over scan lines
+
+        Parameters
+        ----------
+        force_channel: str
+            name of force channel to downsample and plot
+        color_channel: str
+            color channel of kymo to plot ('red', 'green', 'blue', 'rgb')
+        aspect_ratio: float
+            axes aspect ratio
```
lpingas, line 271: I'm not an avid user of `matplotlib`, so I could not make sense of what a single value `aspect_ratio` meant.

I suggest you add something in the lines of "Set the aspect of the axis scaling, i.e. the ratio of y-unit to x-unit", which I shamelessly copied from https://matplotlib.org/3.1.1/api/_as_gen/matplotlib.axes.Axes.set_aspect.html  : D

lumicks/pylake/kymo.py
```diff
@@ -248,6 +249,56 @@ def plot_rgb(self, **kwargs):
...
+        **kwargs
+            Forwarded to :func:`Slice.plot`.
+        """
+        def set_aspect_ratio(axis, ar):
+            """This function forces a specific aspect ratio, can be useful when aligning figures"""
+            axis.set_aspect(ar*np.diff(axis.get_xlim())[0] / np.abs(np.diff(axis.get_ylim()))[0])
+
+        import matplotlib.pyplot as plt
+
+        force = self._downsample_channel(force_channel[-2], force_channel[-1])
```
lpingas, line 287: `where` and `reduce` are not being propagated down.

lumicks/pylake/tests/test_kymo.py
```diff
@@ -116,3 +118,29 @@ def test_damaged_kymo(h5_file):
...
+
+@cleanup        
+def test_plotting_with_force(h5_file):
+    f = pylake.File.from_h5py(h5_file)
+    if f.format_version == 2:
+        kymo = f.kymos["Kymo1"]
+
+        ds = kymo._downsample_channel(2, "x", reduce=np.mean, where="center")
+        assert np.allclose(ds.data, [30, 30, 10, 10])
+        assert np.allclose(ds.timestamps, kymo.timestamps[2])
```
lpingas, line 142: For timestamps `np.equal` should do the job.

lumicks/pylake/kymo.py
```diff
@@ -248,6 +249,56 @@ def plot_rgb(self, **kwargs):
...
+            Forwarded to :func:`Slice.downsampled_over`
+        where : str
+            Where to put the final time point.
+            Forwarded to :func:`Slice.downsampled_over`
+        **kwargs
+            Forwarded to :func:`Slice.plot`.
+        """
+        def set_aspect_ratio(axis, ar):
+            """This function forces a specific aspect ratio, can be useful when aligning figures"""
+            axis.set_aspect(ar*np.diff(axis.get_xlim())[0] / np.abs(np.diff(axis.get_ylim()))[0])
```
lpingas, line 283: Nitpick: This assumes that `np.diff(axis.get_xlim())[0]` will never be negative, which makes sense for timestamps, but you could change it to `ar * np.abs(np.diff(axis.get_xlim())[0] / np.diff(axis.get_ylim())[0])` for some extra robustness. 
Personally, I find it a bit more readable as one can quickly observe that `np.diff(axis.get_<>lim)[0]` is calculating the range of the `<>` axis. 

Nitpick cause I'm talking about generalizing a nested function.

lpingas (commented):
I'm sorry about the long delay! I've left some comments for you to consider.

lumicks/pylake/kymo.py
rpauszek: good catch ... fixed it for `reduce` and after some discussion with @JoepVanlier it doesn't make sense to give the `where` option to users since we are defining the line time at the central timestamp

lpingas (approved):
Looks great! 👍

PR #76, opened 2020-11-10
channel: add function to downsample like bluelake

JoepVanlier:
**Why this PR?**
It was requested to have a function which downsamples exactly like Bluelake does.

**How does it work?**
Bluelake averages backward in time from the end (i.e. `[t-dt : t]`) using the frame time. We can reconstruct this frame time from the consecutive time stamps.

When the framerate changes, it takes some time for new images to arrive. The code shouldn't average that entire time and instead just use the new framerate based on the timesteps that follow.

**Why not via downsampled_over?**
The difference in how the change in framerate is handled warranted a separate function.

Test with actual bluelake data with varying framerate (first samples cannot be reconstructed since there is no HF data there):
image

The residual when they overlap:
image

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -173,7 +173,6 @@ class Slice:
                     sample times and a variable number of sampling contributing to each target sample, 
                     but must be used for variable-frequency data
         """
-
         if method not in ("safe", "ceil", "force"):
             raise ValueError(f"method '{method}' is not recognized")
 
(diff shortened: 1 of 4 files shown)


lumicks/pylake/channel.py
```diff
@@ -231,6 +231,9 @@ def downsampled_like(self, other_slice, reduce=np.mean):
             The default is `np.mean`, but `np.sum` could also be appropriate for some
             cases, e.g. photon counts.
         """
+        assert isinstance(other_slice._src, TimeSeries), "This function is intended to downsample like a LF channel"
```
mcsmurf, line 234:
```suggestion
        assert isinstance(other_slice._src, TimeSeries), "This function is intended to downsample like an LF channel"
```

lumicks/pylake/tests/test_channels.py
```diff
@@ -505,3 +505,17 @@ def test_downsampling_over_subset():
     sd = s.downsampled_over([(20, 40), (40, 60), (60, 80)])
     np.allclose(sd.data, [(3+4)/2, (4+5)/2, (5+6)/2])
     np.allclose(sd.timestamps, [(20+30)/2, (40+50)/2, (60+70)/2])
+
+
+def test_downsampling_like():
+    #    0  2  4  6  8  10 12 14 16 18 20 22 24 26 28 30 32 34 36 38 40 42 44 46 48 50
```
mcsmurf, line 511: I suppose this will be removed?

mcsmurf (approved):
:rocket:

lumicks/pylake/tests/test_channels.py
JoepVanlier: yes

lumicks/pylake/channel.py

Actually decided to clarify the error a bit more.
`You did not pass a low frequency channel to serve as reference channel`
and
`Downsampled_like is only available for high frequency channels`

mcsmurf (approved):
LGTM!

lumicks/pylake/channel.py
mcsmurf: Generally good to suggest a solution when reporting the error, so yeah. Much better :+1:

PR #77, opened 2020-11-12
Allow plotting with respect to a different timestamp

JoepVanlier:
**Why this PR?**
Quite often, you want to plot two slices of something together. It can be inconvenient to use a different starting point for the time axes. For example, when comparing downsampled data to the normal data, it can be pretty annoying that there is an apparent time shift when the initial sample is different. Or when you want to plot two slices together, you might want to be able to align them quickly and conveniently.

This PR exposes `start` and `stop` on `Slice` and allows you to `.plot()` with a custom starting timestamp.

```python
first_slice = file.force1x['5s':'10s']
second_slice = file.force1x['15s':'20s']
first_slice.plot()
second_slice.plot(start=first_slice.start)  # we want to use the start of first_slice as time point "zero"
```


dean0x7d (approved):
Great feature!

JoepVanlier:
> Great feature!

thanks 😄

PR #78, opened 2020-11-12
Refactor confocal

rpauszek:
**Why this PR?**
Currently `Scan` inherits from `Kymo`, while this relationship is not necessarily correct or straightforward (same if the inheritance were reversed). Additionally, `PointScan` has its own implementation for many methods/attributes that are duplicated directly from `Kymo`. 

This PR refactors the class hierarchy such that all three classes now inherit from a single `BaseScan` superclass, which takes care of object initialization and shared methods (such as public plotting methods). A second mixin class `ConfocalImage` is also introduced to gather methods/properties needed for both `Scan` and `Kymo`, mostly involved in reconstructing images from infowaves. Overall, this results in significantly less duplicated code and sets up a clearer hierarchy that should simplify future additions of new experiments (ie, 3D scans, diagonal scans, ...)

**Potential Issues**
As currently implemented by mixin, the order of superclasses is important for `Scan` and `Kymo` to ensure that the correct version of both `_plot_color()` and `plot_rgb()` are available. An argument could be made that `ConfocalImage` should inherit directly from `BaseScan` -- in which case both `Scan` and `Kymo` would inherit directly from `ConfocalImage` -- which would take care of this.


lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,241 @@
...
+
+    def plot_red(self, **kwargs):
+        """Plot an image of the red photon channel
+
+        Parameters
+        ----------
+        **kwargs
+            Forwarded to :func:`matplotlib.pyplot.imshow`.
+        """
+        return self._plot_color("red", **kwargs) # ?? need to return here?
```
JoepVanlier, line 82: Regarding the comment, yes, see: https://github.com/lumicks/pylake/pull/55/files

You can add that it's used to maintain a handle to images so they can be updated in existing plots.

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,241 @@
...
+        filename : str
+            The name of the TIFF file where the image will be saved.
+        dtype : np.dtype
+            The data type of a single color channel in the resulting image.
+        clip : bool
+            If enabled, the photon count data will be clipped to fit into the desired `dtype`.
+            This option is disabled by default: an error will be raise if the data does not fit.
+        """
+        if self.rgb_image.size > 0:
+            save_tiff(self.rgb_image, filename, dtype, clip, ImageMetadata.from_dataset(self.json))
```
JoepVanlier, line 187: `save_tiff` and `ImageMetadata` aren't imported, so this won't work

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,241 @@
...
+    def plot_rgb(self, **kwargs):
+        """Plot all color channels
+
+        Parameters
+        ----------
+        **kwargs
+            Forwarded to `~matplotlib.pyplot.plot`.
+        """
+        for color in ["red", "green", "blue"]:
+            self._plot_color(color, **kwargs)
```
JoepVanlier, line 113: This confused me at first, but it's overridden by `ConfocalImage` I guess. It just struck me as odd since in most cases, I wouldn't be using this one. I'd almost think that leaving it non-implemented here and overridding it explicitly in `PointScan` might make sense. I'm not sure really.

lumicks/pylake/scan.py
```diff
@@ -1,28 +1,27 @@
...
+        Scan name
     file : lumicks.pylake.File
         Parent file. Contains the channel data.
     start : int
         Start point in the relevant info wave.
     stop : int
         End point in the relevant info wave.
     json : dict
-        Dictionary containing kymograph-specific metadata.
+        Dictionary containing scan-specific metadata.
```
JoepVanlier, line 21: Good catch on the docstrings 🥇

JoepVanlier (changes requested):
Overall, I do like it. I think it's great to deduplicate a lot of this stuff. I left a few comments here and there.

I was initially a bit confused by some of the things already being implemented in the `BaseScan`. I wonder if moving those to `Confocal` and `Point` would indeed make sense as you mentioned.

Also, be careful with the tiff export / metadata, since it's not covered.

lumicks/pylake/confocal.py
rpauszek: good catch

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,241 @@
...
+    def _get_photon_count(self, name):
+        photon_count = getattr(self.file, f"{name}_photon_count".lower())[self.start:self.stop]
+
+        try:
+            if self._has_incorrect_start(photon_count._src.start, photon_count._src.dt):
+                photon_count = getattr(self.file, f"{name}_photon_count".lower())[self.start:self.stop]
+        except AttributeError:
+            pass
+
+        return photon_count
```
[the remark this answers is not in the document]
rpauszek: yes, actually I was pretty confused by this when I first came across it. I ended up just including the common error checking in the actual method, and leaving the specific implementation of the fix as the `_fix_incorrect_start()` method for each subclass. I think it's a lot cleaner this way

lumicks/pylake/confocal.py
```diff
@@ -55,15 +55,25 @@ def from_dataset(cls, h5py_dset, file):
         return cls(name, file, start, stop, json_data)
 
     def _get_photon_count(self, name):
-        photon_count = getattr(self.file, f"{name}_photon_count".lower())[self.start:self.stop]
+        timeline_start = self.file.h5["Photon count"][name].attrs["Start time (ns)"]
+        timeline_sample_rate = self.file.h5["Photon count"][name].attrs["Sample rate (Hz)"]
+        timeline_dt = int(1e9 /timeline_sample_rate)
```
JoepVanlier, line 60: Is there a reason you go with the raw `h5` here? I'd prefer using the dt from the higher level API.

rpauszek: My thought was to avoid grabbing the `photon_count` slice twice, because as you mentioned before it's a bit confusing and looks like duplicated code. I also thought, performance wise, it was somewhat wasteful to construct the whole slice just to grab the `start` and `dt`, but thinking about it more I guess it's not a particularly complicated operation.

I'm fine with going back to the original version, but also perhaps would it be nice to have a small helper method to grab the time info without constructing the whole slice? Would this be useful elsewhere maybe?

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,231 @@
...
+
+    def _get_photon_count(self, name):
+        timeline_start = self.file.h5["Photon count"][name].attrs["Start time (ns)"]
+        timeline_sample_rate = self.file.h5["Photon count"][name].attrs["Sample rate (Hz)"]
+        timeline_dt = int(1e9 /timeline_sample_rate)
+
+        # Workaround for a bug in the STED delay mechanism which could result in scan start times ending up within
+        # the sample time.
+        if timeline_start - timeline_dt < self.start < timeline_start:
+            self.start = timeline_start
```
JoepVanlier, line 65: This is not correct. The timeline can start well before the `Scan` and in most use cases does. Pre-refactor the timeline was sliced, after which you'd get the appropriate start from `Slice.start` (it being the first timestamp after `Scan.start`).

However, for the STED thing, there was an error that could sometimes lead to a small temporal displacement of the `Scan` timestamp that wasn't quantized to whole timestamp. When you slice the timeline then, you will end up with the next sample that's valid, but it won't line up with the `Scan.start`. The reason the test didn't pick this up is because the test didn't incorporate data prior to the `Scan`. My bad on the test really.

Please have a peek at branch `stedbug_redux`. I've put a hardened test there which should pick up this regression. Would be nice to include it.

lumicks/pylake/confocal.py
JoepVanlier: Yeah, I see what you mean. Unfortunately, for the sted workaround, you need the first timestamp after or equal to `Scan.start`, so I don't think there's a nice way around the slicing it already. `Continuous.slice` is actually fairly lightweight, it doesn't construct all the timestamps to just give you the start.

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,241 @@
...
+    @property
+    def has_fluorescence(self) -> bool:
+        return self.json["fluorescence"]
+
+    @property
+    def has_force(self) -> bool:
+        return self.json["force"]
+
+
+class ConfocalImage():
```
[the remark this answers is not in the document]
dean0x7d: Not a nitpick -- redundant parentheses are the root of all evil.

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,232 @@
...
+        if timeline_start - timeline_dt < self.start < timeline_start:
+            self.start = timeline_start
+
+        # Checks whether the scan or kymograph starts before the timeline information. 
+        # If this is the case, it will lead to an incorrect reconstruction. 
+        # If implemented, resolve the problem.
+        if timeline_start > self.start:
+            self._fix_incorrect_start()
+
+        return getattr(self.file, f"{name}_photon_count".lower())[self.start:self.stop]
```
dean0x7d, line 74: This ends up reading the photon counts twice, even if there's no need to fix the incorrect start.

lumicks/pylake/confocal.py
```diff
@@ -0,0 +1,232 @@
...
+
+    def _fix_incorrect_start(self):
+        raise NotImplementedError
+
+    def _plot_color(self, color, **kwargs):
+        raise NotImplementedError
```
dean0x7d, line 80: I would put the docstring here instead of in the individual implementations. This is because this point defines what is expected of this function (and where we expect it to be implemented). The implementations can contain additional comments but this defines the interface and it's the point that should be documented.

dean0x7d (commented):
Very nice refactor! It's really good to see the duplicate code going away as well as the confusion surrounding it.

lumicks/pylake/kymo.py
```diff
@@ -8,7 +8,7 @@
 from .detail.timeindex import to_timestamp
 
 
-class Kymo(BaseScan, ConfocalImage):
+class Kymo(ConfocalImage, BaseScan):
```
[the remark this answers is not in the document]
rpauszek: I decided to go with making `ConfocalImage` a subclass of `BaseScan` for a couple reasons:
- many of the methods in `ConfocalImage` depend on the attributes initialized in `BaseScan` so it makes sense to make the relationship concrete
- I think it's best to keep `_plot_color()` as a not-implemented method in `BaseScan` because the other plotting functions defined there depend on it explicitly. So subclassing is really the only way to get around the MRO issue in this case

JoepVanlier (approved):
:shipit:

dean0x7d (approved):
Looks good! 🚢 

It would be great if you could also clean up the git history when you rebase this for the final merge (drop all the 'cleanup' commits). Let me know if you need any help with juggling the git history/interactive rebase.

PR #80, opened 2020-11-16
kymotracker: add refinement pt. 1

JoepVanlier:
**Why this PR?**
When performing Kymotracking using the greedy algorithm, one strings together a sequence of points that fall above a certain threshold into a trace. This means gaps can occur in the trace. This PR adds a method which interpolates the traces, and then tweaks those points based on the local brightness (using the brightness weighted centroid).

Before:
image

After:
image

**Why part 1?**
Ideally this method is followed up by a more extensive fitting procedure that will take into account the Poissonian nature of the measurement.

Docs: https://lumicks-pylake.readthedocs.io/en/kymo_refinement_pt1/tutorial/kymotracking.html

I added a paragraph on the refinement (look for refine_lines_centroid).

### DIFF
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -1,4 +1,5 @@
-from lumicks.pylake.kymotracker.detail.peakfinding import peak_estimate, refine_peak_based_on_moment, merge_close_peaks
+from lumicks.pylake.kymotracker.detail.peakfinding import peak_estimate, refine_peak_based_on_moment, merge_close_peaks, \
+    KymoPeaks
 from lumicks.pylake.kymotracker.detail.trace_line_2d import detect_lines, points_to_line_segments
 from lumicks.pylake.kymotracker.detail.scoring_functions import kymo_score
 import numpy as np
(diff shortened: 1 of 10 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -175,3 +175,38 @@ def filter_lines(lines, minimum_length):
...
+      dx = (1 / m0) * sum_{i <= w} i(I(x + i))
+
+    Parameters
+    ----------
+    lines : List[pylake.KymoLine]
+        Detected traces on a kymograph.
+    line_width : int
+        Line width
+    """
+    lines = [line.interpolate() for line in lines]
```
lpingas, line 196: Given that `lines` is already a parameter, it would be better to have this as `interpolated_lines` or `it_lines` (if you need a shorter name) for readability. 

But now I'm curious: Doesn't this modify the parameter as it is taken by reference?

lpingas (commented):
This is looking great! Left a question before we merge it.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Yeah, that's fair. I'll rename it to `interpolated_lines`.

And no, it shouldn't. If you look at `interpolate`, you can see it explicitly makes a new `KymoLine`. Just to verify:
image

Do you want me to add an explicit test for this?

lpingas: >And no, it shouldn't. If you look at interpolate, you can see it explicitly makes a new KymoLine. Just to verify:

Oh, but I meant assigning the list comprehension to `lines`.

>Do you want me to add an explicit test for this?

I don't believe this is needed.

lpingas (approved):
👍

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: > Oh, but I meant assigning the list comprehension to `lines`.

Ah, no, the list comprehension just made a new list and assigned it to that variable name.

image

PR #81, opened 2020-11-18
Refactor image reconstruction

rpauszek:
**Why this PR?**
Currently, there is a lot of duplicated code in `reconstruct_image()` and `reconstruct_image_sum()`. Additionally these functions have an optional `lines_per_frame` kwarg to handle reconstructing 2D images, which leads to different call signatures from `Scan` and `Kymo`

- duplicated code in `reconstruct_image()` and `reconstruct_image_sum()` is pulled out into separate helper functions
- a private `_shape` property is added to `Scan` and `Kymo` which can be used directly in the image reconstruction functions rather than the current `pixels_per_line` and optional `lines_per_frame` arguments
- this allows for a homogenous call to the reconstruction functions; therefore `_image()` and `_timestamps()` are moved to `ConfocalImage` mixin class rather than being implemented separately in `Kymo` and `Scan`

**NOTE**
Because the call signature for the image reconstruction functions changed, the arguments in `test_image.py` for these functions needed to be flipped.

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -60,6 +60,21 @@ class InfowaveCode(enum.IntEnum):
     pixel_boundary = 2  # useful data and marks the last data sample in a pixel
 
 
+def discard_zeros(infowave):
+    # Example infowave:
+    #  1 0 0 1 0 1 2 0 1 0 0 1 0 1 0 1 0 0 1 0 2 0 1 0 1 0 0 1 0 1 0 1 2 1 0 0 1
+    #              ^ <-----------------------> ^                       ^
+    #                       one pixel
+    valid_idx = infowave != InfowaveCode.discard
+    subset = infowave[valid_idx]
+
+    # After discard:
+    #  1 1 1 2 1 1 1 1 1 2 1 1 1 1 1 2 1 1 1
+    #        ^ <-------> ^           ^
+    #         pixel_size (i.e. data samples per pixel)
+    return subset, valid_idx
+
+
 def reconstruct_num_frames(infowave, pixels_per_line, lines_per_frame):
     """Reconstruct the number of frames in a continuous scan
 
@@ -109,9 +124,8 @@ def seek_timestamp_next_line(infowave):
     infowave = infowave.data
 
     # Discard unused acquisition
-    mask = infowave != InfowaveCode.discard
-    infowave = infowave[mask]
-    time = time[mask]
+    infowave, valid_idx = discard_zeros(infowave)
+    time = time[valid_idx]
 
     # Fetch start of pixels
     pixel_ends, = np.where(infowave == InfowaveCode.pixel_boundary)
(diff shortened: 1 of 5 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -129,7 +129,40 @@ def seek_timestamp_next_line(infowave):
...
+    return subset, mask
+
+
+def round_up(size, n):
+        """Round up `size` to the nearest multiple of `n`"""
+        return int(math.ceil(size / n)) * n
+
+
+def reshape_reconstructed_image(pixels, shape):
+    # shape = (<<lines_per_frame>>, pixels_per_line)
```
dean0x7d, line 156: I'd make this part of the function's docstring since it's the expectation for one of the arguments. Also, a general docstring would nice here to indicate the purpose of this function -- something short.

lumicks/pylake/detail/image.py
```diff
@@ -129,7 +129,40 @@ def seek_timestamp_next_line(infowave):
...
+    resized_pixels = np.zeros(round_up(pixels.size, np.prod(shape)))
+    resized_pixels[:pixels.size] = pixels
+    if len(shape) > 1:
+        return resized_pixels.reshape(-1, *shape).squeeze()
+    else:
+        return resized_pixels.reshape(-1, *shape)
```
dean0x7d, line 162: Could an unconditional `.squeeze()` work here? I don't exactly remember why it would not. It would be good to either always do `.squeeze()` or add a comment indicating why there's a difference.

lumicks/pylake/detail/image.py
```diff
@@ -192,40 +207,12 @@ def reconstruct_image(data, infowave, pixels_per_line, lines_per_frame=None, red
...
-    # But for now we assume that every pixel consists of the same number of samples
-    pixel_size = np.argmax(infowave) + 1
-
-    def round_up(size, n):
-        """Round up `size` to the nearest multiple of `n`"""
-        return int(math.ceil(size / n)) * n
-
-    resized_data = np.zeros(round_up(infowave.size, pixel_size))
-    resized_data[:infowave.size] = data[valid_idx]
+    subset, mask = process_infowave(infowave)
```
dean0x7d, line 210: Inevitable naming comment:
* `process_infowave()` is super generic. "Process" could be doing anything. I'd suggest `discard_zeros(infowave)`.
* I tend to get confused about which half is `mask` masking. IMO `valid_idx` was a bit clearer.

lumicks/pylake/confocal.py
```diff
@@ -133,6 +133,20 @@ def _ordered_axes(self):
         """Returns axis indices in spatial order"""
         return sorted(self.json["scan volume"]["scan axes"], key=lambda x: x["axis"])
 
+    def _to_spatial(self, data):
+        raise NotImplementedError
```
dean0x7d, line 137: Add a docstring here and for `_shape` so we know what this needs to do. The implementations can differ but the intent is the same.

dean0x7d (commented):
Negative line PR! Those are the best. Very nice deduplication of code. I left a few suggestions.

P.S. with these dependent PRs, it's good to indicate in the description that this can't be merged until PR #XYZ is merged.

lumicks/pylake/detail/image.py
```diff
@@ -129,7 +129,40 @@ def seek_timestamp_next_line(infowave):
...
+    resized_pixels[:pixels.size] = pixels
+    if len(shape) > 1:
+        return resized_pixels.reshape(-1, *shape).squeeze()
+    else:
+        return resized_pixels.reshape(-1, *shape)
+
+
+def reconstruct_image_sum(data, infowave, shape):
     """Rapidly reconstruct a scan or kymograph image from raw data summing pixels where the infowave equates to 1.
     See reconstruct_image for more information.
```
JoepVanlier, line 167: I should have added this at the time, but given that you're working on this code now, I think it wouldn't hurt to add a little warning to this function `reconstruct_image_sum` mentioning that this one should not be used for timestamps since they will overflow.

JoepVanlier (commented):
I only have a minor request. Otherwise lgtm. 👍

lumicks/pylake/detail/image.py
rpauszek: From the tests it seems like the only time `squeeze` should not be called is for a single-line Kymo (data expected to always be >= 2D) so I moved that check to the `_to_spatial()` method in the Kymo class. I think it's a lot cleaner this way

JoepVanlier: I wonder if we can get something more controlled than squeeze because now it's kind of offloading dimensionality checks to the end user.

If you have a 2D scan with one line, now, you suddenly get 1D data instead of 2D. If you have a 3D scan with only one plane, you get a 2D array. I don't like that the dimensionality turns dynamic depending on how much data you acquired (rather than the expected dimensionality from shape). It would make code using the result from reconstructions having to check the dimensionality.

dean0x7d: That's a good point, Joep. Maybe it would be best to flip it: don't squeeze by default in `reshape_reconstructed_image()`. Only do it in the end class if it makes sense.

JoepVanlier (approved):
image

image

LGTM though! :shipit:

dean0x7d (approved):

PR #82, opened 2020-11-19
Better caching

rpauszek:
**Why this PR?**
Currently, caching for expensive calculations is handled manually (and not implemented for timestamps). Would be better to let a dedicated caching package handle this.

- caching is handled through `cachetools` method decorator 
- caching is implemented for `timestamps` along with `{color}_image`
- all calculations are handled through a single method `_from_infowave` (and `_timestamps` and `_image` have been removed)

### DIFF
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -1,5 +1,6 @@
 import json
 import numpy as np
+import cachetools
 
 from .mixin import PhotonCounts
 from .mixin import ExcitationLaserPower
@@ -140,16 +141,24 @@ class ConfocalImage(BaseScan):
         """Implements any necessary post-processing actions after image reconstruction from infowave """
         raise NotImplementedError
 
-    def _image(self, color):
-        if color not in self._cache:
-            photon_counts = getattr(self, f"{color}_photon_count").data
-            self._cache[color] = reconstruct_image_sum(photon_counts, self.infowave.data, self._shape)
-            self._cache[color] = self._to_spatial(self._cache[color])
-        return self._cache[color]
+    @cachetools.cachedmethod(lambda self: self._cache)
+    def _image(self, channel):
+        channel_data = getattr(self, f"{channel}_photon_count").data
+        raw_image = reconstruct_image_sum(channel_data, self.infowave.data, self._shape)
+        return self._to_spatial(raw_image)
 
-    def _timestamps(self, sample_timestamps):
-        timestamps = reconstruct_image(sample_timestamps, self.infowave.data, self._shape, reduce=np.mean)
-        return self._to_spatial(timestamps)
+    @cachetools.cachedmethod(lambda self: self._cache)
+    def _timestamps(self, channel):
+        assert channel == "timestamps"
+        # Uses the timestamps from the first non-zero-sized photon channel
+        for color in ("red", "green", "blue"):
+            channel_data = getattr(self, f"{color}_photon_count").timestamps
+            if len(channel_data) != 0: 
+                break
+        else:
+            raise RuntimeError("Can't get pixel timestamps if there are no pixels")
+        raw_image = reconstruct_image(channel_data, self.infowave.data, self._shape, reduce=np.mean)
+        return self._to_spatial(raw_image)
 
     def _plot_color(self, color, **kwargs):
         from matplotlib.colors import LinearSegmentedColormap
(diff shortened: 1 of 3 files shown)


lumicks/pylake/confocal.py
```diff
@@ -136,16 +137,29 @@ def _ordered_axes(self):
...
-        if color not in self._cache:
-            photon_counts = getattr(self, f"{color}_photon_count").data
-            self._cache[color] = reconstruct_image_sum(photon_counts, self.infowave.data, self._shape)
-            self._cache[color] = self._to_spatial(self._cache[color])
-        return self._cache[color]
-
-    def _timestamps(self, sample_timestamps):
-        timestamps = reconstruct_image(sample_timestamps, self.infowave.data, self._shape, reduce=np.mean)
-        return self._to_spatial(timestamps)
+    @cachetools.cachedmethod(operator.attrgetter('_cache'))
```
dean0x7d, line 140: Consider this syntax instead. I find it a little bit more relatable. However, neither one is exactly intuitive. Either way is fine in the end.
```suggestion
    @cachetools.cachedmethod(lambda self: self._cache)
```

lumicks/pylake/confocal.py
```diff
@@ -136,16 +137,29 @@ def _ordered_axes(self):
...
+    def _from_infowave(self, channel):
+        if not channel in ("timestamps", "red", "green", "blue"):
+            raise ValueError(f"{channel} channel is not recognized.")
```
dean0x7d, line 143: I would avoid the upfront check here, especially for implementation details. This is unlikely to ever come up as a regular error. Only when a developer wants to add an extra channel in which case it's just more bookkeeping.

lumicks/pylake/confocal.py
```diff
@@ -136,16 +137,29 @@ def _ordered_axes(self):
...
+                                 for color in ("red", "green", "blue")):
+                if len(channel_data) != 0: 
+                    break
+            else:
+                raise RuntimeError("Can't get pixel timestamps if there are no pixels")
+            fcn = reconstruct_image
+            kwargs["reduce"] = np.mean
+        else:
+            channel_data = getattr(self, f"{channel}_photon_count").data
+            fcn = reconstruct_image_sum
```
dean0x7d, line 159: Given how many differences there are between these two paths, I feel like it might actually be simpler to split it into 2 separate functions (colors and timestamps) rather than force them together to reuse just 1 or 2 lines.

I guess that you joined the 2 functions in order to share the single cache but I don't think that this is strictly necessary.

dean0x7d (commented):
Great to have a more convenient and robust caching mechanism!

lumicks/pylake/confocal.py
```diff
@@ -29,7 +30,7 @@ def __init__(self, name, file, start, stop, json):
         self.name = name
         self.json = json
         self.file = file
-        self._cache = {}
+        self._cache = cachetools.LRUCache(maxsize=4)
```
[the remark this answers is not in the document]
rpauszek: It seems that you can't have an unlimited cache for this class. I increased the size to 16 (could go higher); I suppose we could just use a plain `dict` as before if you think that would be better

lumicks/pylake/confocal.py
rpauszek: True. I was trying to simplify the internals (everything that uses the cache calls the same method) but the method itself did indeed get a bit convoluted. Turns out you can use the same cache for different methods though if they have the same signature

lumicks/pylake/confocal.py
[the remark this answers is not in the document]
dean0x7d: I think that a plain `dict` would be fine here. It communicates the intent much more clearly than an arbitrary maximum.

lumicks/pylake/confocal.py
```diff
@@ -136,16 +137,24 @@ def _ordered_axes(self):
...
 
-    def _timestamps(self, sample_timestamps):
-        timestamps = reconstruct_image(sample_timestamps, self.infowave.data, self._shape, reduce=np.mean)
-        return self._to_spatial(timestamps)
+    @cachetools.cachedmethod(lambda self: self._cache)
+    def _timestamps(self, channel):
+        assert channel == "timestamps"
+        # Uses the timestamps from the first non-zero-sized photon channel
+        for channel_data in (getattr(self, f"{color}_photon_count").timestamps 
+                                for color in ("red", "green", "blue")):
```
dean0x7d, line 151: This looks pretty messy to me. Consider one of these two alternatives:
```suggestion
        for color in "red", "green", "blue":
            channel_data = getattr(self, f"{color}_photon_count").timestamps
```
```suggestion
        channels = (getattr(self, f"{color}_photon_count").timestamps 
                    for color in "red", "green", "blue")
        for channel_data in channels:
```

dean0x7d (approved):
I just have one optional esthetic gripe (see below). Other than that, this is a great improvement!

JoepVanlier (approved):
Looks solid! Looking forward to also using this in other places.

PR #85, opened 2020-11-19
ci: support python 3.9

JoepVanlier:
**Why this PR?**
To support Python 3.9 we need wheels for `h5py`. We need to start using the new `h5py` to still get wheels.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.7.2 | t.b.d.
+
+* Support Python 3.9 on Windows (this required bumping the `h5py` requirement to >= 3.0).
+
 ## v0.7.1 | 2020-11-19
 
 * Add `start` and `stop` property to `Slice`. These return the timestamp in nanoseconds.
--- a/setup.py
+++ b/setup.py
@@ -51,8 +51,15 @@ setup(
 
     packages=PEP420PackageFinder.find(include=["lumicks.*"]),
     python_requires='>=3.6',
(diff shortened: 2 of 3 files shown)


changelog.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## v0.7.2 | t.b.d.
+* Change `h5py` requirement that the version should be smaller than 3 to bigger or equal to three.
```
dean0x7d, line 4: * Add a blank like after Markdown (section) titles
* The changelog should really mention something about adding Python 3.9 support on Windows (I assume that things worked on Linux before?)
* The sentence is a bit confusing. Maybe simply: `Bumped h5py requirement to >= 3.0`

setup.py
```diff
@@ -51,7 +51,7 @@ def read(filename):
 
     packages=PEP420PackageFinder.find(include=["lumicks.*"]),
     python_requires='>=3.6',
-    install_requires=['pytest>=3.5', 'h5py>=2.9, <3.0', 'numpy>=1.14, <2',
+    install_requires=['pytest>=3.5', 'h5py>=3.0, <4', 'numpy>=1.14, <2',
```
dean0x7d, line 54: This might be a good time to re-format this list to one entry per line in order to get nicer diffs.

changelog.md
JoepVanlier: How about `* Support Python 3.9 on Windows (this required bumping the `h5py` requirement to >= 3.0).`?

setup.py
JoepVanlier: Yes. Done.

dean0x7d (approved):
Awesome!

PR #86, opened 2020-11-24
Don't force users to install ipympl + faq

JoepVanlier:
**Why this PR?**
This PR hopes to make it simpler for our users to get the interactive widgets to work.

**How?**
`Pylake` checked explicitly whether the backend is `ipympl` and threw an exception if it wasn't. In Jupyter notebook, the backend `nbagg` is installed by default and would also be suitable for displaying our interactive plots.

The second change introduced in this PR is that `ipywidgets` is added as an explicit dependency. Some of our users may not know how to install it, and it'd be nice to just have it dealt with.

I also added a few FAQ issues to the docs that come up frequently.

**Docs:**
https://lumicks-pylake.readthedocs.io/en/simplify_widgets/install.html#faq

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -117,3 +117,73 @@ Troubleshooting
 If you run into any errors after installation, try updating all conda packages to the latest versions using the following command::
 
     conda update --all
+
+
+Frequently asked questions
+--------------------------
+
+**Why are the plots in my notebook not interactive?**
+
+To enable interactive plots, you have to invoke the correct `magic commands <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_
+in the notebook. When using Jupyter notebook, the following command will switch the `matplotlib` backend from the inline
+one (which renders images) to the interactive backend::
+
+    %matplotlib notebook
+
+You can also choose to install `ipympl`, which can perform better in some cases. You can install it with `pip`::
+
+    pip install ipympl
+
+or `conda`::
+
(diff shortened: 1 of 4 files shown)


docs/install.rst
```diff
@@ -117,3 +117,73 @@ Troubleshooting
...
 
     conda update --all
+
+
+Frequently asked questions
+--------------------------
+
+**Why are the plots in my notebook not interactive?**
+
+To enable interactive plots, you have to invoke the correct magic commands in the notebook. When using Jupyter notebook,
```
dean0x7d, line 127: Maybe link `magic commands` to this https://ipython.readthedocs.io/en/stable/interactive/magics.html So that readers don't think this is just us being funny :)

docs/install.rst
```diff
@@ -117,3 +117,73 @@ Troubleshooting
...
+
+or `conda`::
+
+    conda install -c conda-forge ipympl
+
+The `ipympl` backend can be activated by invoking the following magic command in a notebook::
+
+    %matplotlib widget
+
+*Note that switching backend, typically requires you to restart the Jupyter kernel*. When using JupyterLab, `ipympl` is
```
dean0x7d, line 145:
```suggestion
*Note that switching backends typically requires you to restart the Jupyter kernel*. When using JupyterLab, `ipympl` is
```

dean0x7d (approved):
This is great!

PR #88, opened 2020-12-03
kymo: fix plot time extent

rpauszek:
**Why this PR?**
Currently, the plotting functions for `Kymo` incorrectly set the time extent of the image, leading to a subtle bug (see image below)

By adjusting the time extent (essentially adding an extra line time) the line time points now correspond with the center of the pixel. 

Shown below is a simple dataset of 12 pixels with a timestep of 1

image

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -189,10 +189,12 @@ class Kymo(PhotonCounts, ExcitationLaserPower):
         width_um = self._ordered_axes()[0]["scan width (um)"]
         ts = self.timestamps
         duration = (ts[0, -1] - ts[0, 0]) / 1e9
+        linetime = (ts[0, 1] - ts[0, 0]) / 1e9
 
         default_kwargs = dict(
             # With origin set to upper (default) bounds should be given as (0, n, n, 0)
-            extent=[0, duration, width_um, 0],
+            # pixel center aligned with mean time per line
+            extent=[-0.5 * linetime, duration + 0.5 * linetime, width_um, 0],
             aspect=(image.shape[0] / image.shape[1]) * (duration / width_um)
         )
 
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 * Added `refine_lines_centroid` for refining lines detected by the kymotracking algorithm. See [kymotracking](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html) for more information.
 * Added `Kymo.plot_with_force` for plotting a kymograph and corresponding force channel downsampled to the same time ranges of the scan lines.
 * Fixed exception message in `downsampled_to` which erroneously suggested to use `force=True` when downsampling a variable frequency channel, while the correct argument is `method="force"`.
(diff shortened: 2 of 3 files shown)


lumicks/pylake/kymo.py
```diff
@@ -189,10 +189,13 @@ def _plot(self, image, **kwargs):
         width_um = self._ordered_axes()[0]["scan width (um)"]
         ts = self.timestamps
         duration = (ts[0, -1] - ts[0, 0]) / 1e9
+        dt = (ts[1, 0] - ts[0, 0]) / 1e9
+        linetime = ts.shape[0] * dt
```
JoepVanlier, line 193: Naively, I would have expected `ts[0, 1] - ts[0, 0]`. Is this different?

JoepVanlier (commented):
As discussed before, I agree with fixing this behaviour. I only have a small question.

lumicks/pylake/kymo.py
rpauszek: Yes, that worked fine for the 1D case, but when I expanded the test out to 2D data the times were still slightly offset from the center of the pixels. 

I believe the issue comes down to the fact that the `extent` argument is setting the outer edges whereas `ts[0, 1] - ts[0, 0` gives you the number of bins. The middle panel below is the result if you calculate the line time how you suggest

image

JoepVanlier: That's weird. I observe the opposite?

```python
plt.close('all')
plt.figure(figsize=(7, 7))

def plot_test(method):
    row_time = 5
    t = np.array([[0, 1, 2], 
                  np.array([0, 1, 2]) + row_time, 
                  np.array([0, 1, 2]) + 2 * row_time, 
                  np.array([0, 1, 2]) + 3 * row_time]).T

    duration = t[0, -1] - t[0, 0]
    width_um = 2

    line_time = t[0, 1] - t[0, 0] if method == 0 else (t[1, 0] - t[0, 0]) * t.shape[0]
    title = 't[0, 1] - t[0, 0]' if method == 0 else '(t[1, 0] - t[0, 0]) * shape[0]'
    plt.title(title)

    plt.imshow(t)
    kwargs = dict(
        extent=[-0.5 * line_time, duration + 0.5 * line_time, width_um, 0],
        aspect=(t.shape[0] / t.shape[1]) * (duration / width_um)
    )
    plt.imshow(t, **kwargs)
    
plt.subplot(2, 1, 1)
plot_test(0)
plt.subplot(2, 1, 2)
plot_test(1)
plt.tight_layout()```

![image](https://user-images.githubusercontent.com/19836026/101162906-3e37ac00-3633-11eb-8cf3-8ef6b284e168.png)

rpauszek: You're right. I wasn't taking into account the flyin time between lines in my simple test data, and the discrepancy for the real data (large dataset) was minimal so it looked ok. Fixed now

JoepVanlier: 👍

JoepVanlier (approved):
Looks good to me 👍

PR #89, opened 2020-12-03
kymotracker: save tracked kymos to file

JoepVanlier:
**Why this PR?**
Being able to write kymograph traces to csv was requested by multiple customers. We already have this functionality on Harbor, but it would be nice to add a nicer API for it. This PR is also preparation for the kymotracker notebook widget PR which relies on this functionality being present.

Note that only the `save` functionality is exposed. The `load` functionality is also implemented (and tested), and needed for the widget.

Originally, this was going to be a PR with the widget as well, but I decided to split it up seeing how this refactor of how KymoLines are handled made it a bit bigger than intended already.

**Why KymoLines?**
I debated between just having save to csv a free function, but ended up choosing to write a small API to collect multiple `KymoLine` instances. The reason for this is is that I suspect that this will be less error prone to deal with, and will eventually help us when we start having more complicated calibration procedures. It also constrains the number of potential mistakes a user can make when it comes to passing things to save.

**Why did you piggyback a test on this?**
Because I needed to ensure that `filter_lines` stays functional and it had somehow escaped testing the first time.

Documentation can be found here: https://lumicks-pylake.readthedocs.io/en/kymotracker_saveload/tutorial/kymotracking.html#exporting-kymograph-traces

### DIFF
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -99,3 +99,33 @@ class KymoLine:
 
     def __len__(self):
         return len(self.coordinate_idx)
+
+
+class KymoLineGroup:
+    """Kymograph lines"""
+    def __init__(self, kymo_lines):
+        self._src = kymo_lines
+
+    def __iter__(self):
+        return self._src.__iter__()
+
+    def __getitem__(self, item):
+        if isinstance(item, slice):
+            return KymoLineGroup(self._src[item])
+        else:
+            return self._src[item]
+
+    def __setitem__(self, item, value):
+        raise NotImplementedError("Cannot overwrite KymoLines.")
+
+    def __len__(self):
+        return len(self._src)
+
+    def extend(self, other):
+        if isinstance(other, self.__class__):
+            self._src.extend(other._src)
+        elif isinstance(other, self._src[0].__class__):
+            self._src.extend([other])
+        else:
+            raise TypeError(f"You can only extend a {self.__class__} with a {self.__class__} or "
+                            f"{self._src[0].__class__}")
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -1,6 +1,80 @@
 import numpy as np
 
 
+def export_kymolinegroup_to_csv(filename, kymoline_group, dt, dx, delimiter, sampling_width):
+    """Export KymoLineGroup to a csv file.
+
+    Parameters
+    ----------
+    filename : str
+        Filename to output KymoLineGroup to.
+    kymoline_group : KymoLineGroup
+        Kymograph traces to export.
+    dt : float
+        Calibration for the time axis.
+    dx : float
+        Calibration for the coordinate axis.
+    delimiter : str
+        Which delimiter to use in the csv file.
+    sampling_width : int or None
+        When supplied, this will sample the source image around the kymograph line and export the summed intensity with
+        the image. The value indicates the number of pixels in either direction to sum over.
+    """
+    if not kymoline_group:
+        raise RuntimeError("No kymograph traces to export")
+
+    idx = np.hstack([np.full(len(line), idx) for idx, line in enumerate(kymoline_group)])
+    coords_idx = np.hstack([line.coordinate_idx for line in kymoline_group])
+    times_idx = np.hstack([line.time_idx for line in kymoline_group])
+
+    data, header, fmt = [], [], []
+
+    def store_column(column_title, format_string, new_data):
+        data.append(new_data)
+        header.append(column_title)
+        fmt.append(format_string)
+
+    store_column("line index", "%d", idx)
+    store_column("time (pixels)", "%.18e", times_idx)
+    store_column("coordinate (pixels)", "%.18e", coords_idx)
+
+    if dt:
+        store_column("time", "%.18e", [times_idx * dt])
+
+    if dx:
+        store_column("coordinate", "%.18e", [coords_idx * dx])
+
+    if sampling_width is not None:
+        store_column(f"counts (summed over {2 * sampling_width + 1} pixels)", "%d",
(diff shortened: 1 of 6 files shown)


lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -1,6 +1,67 @@
...
+    delimiter : str
+        Which delimiter to use in the csv file.
+    sampling_width : int or None
+        When supplied, this will sample the source image around the kymograph line and export the summed intensity with
+        the image. The value indicates the number of pixels in either direction to sum over.
+    """
+    if not kymo_lines:
+        raise RuntimeError("No kymolines to export")
+
+    idx = np.hstack([idx * np.ones(len(line.time_idx)) for idx, line in enumerate(kymo_lines)])
```
rpauszek, line 26: Couple minor suggestions:
- I think `len(line)` is clearer than `len(line.time_idx)`
- I always forget about this function, but using `np.full()` might be a bit more clear
```suggestion

    idx = np.hstack([np.full(len(line), idx) for idx, line in enumerate(kymo_lines)])
```

lumicks/pylake/tests/test_kymotracker.py
```diff
@@ -404,3 +404,13 @@ def test_kymolines():
...
+
+
+def test_filter_lines():
+    k1 = KymoLine([1, 2, 3], [1, 2, 3])
+    k2 = KymoLine([2, 3], [1, 2])
+    k3 = KymoLine([2, 3, 4, 5], [1, 2, 4, 5])
+    lines = KymoLines([k1, k2, k3])
+    assert len(filter_lines(lines, 5)) == 0
+    assert len(filter_lines(lines, 3)) == 2
+    assert len(filter_lines(lines, 2)) == 3
```
mwilkowski, line 416: I don't think it's enough to test that `filter_lines` is doing it's job right. I'd rather check the actual values instead of `len`.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -1,6 +1,67 @@
...
+
+
+def import_kymolines_from_csv(filename, image, delimiter=';'):
+    data = np.loadtxt(filename, delimiter=delimiter)
+    assert len(data.shape) == 2, "Invalid file format"
+    assert data.shape[0] > 2, "Invalid file format"
+
+    indices = data[:, 0]
+    lines = np.unique(indices)
+    return KymoLines(KymoLine(data[indices == k, 1], data[indices == k, 2], image) for k in lines)
```
rpauszek, line 62: Maybe add a comment or docstring indicating the labels for each column of the data. I find it helpful rather than having to refer to back to the saving function.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -99,3 +160,55 @@ def extrapolate(self, forward, n_estimate, extrapolation_length):
 
     def __len__(self):
         return len(self.coordinate_idx)
+
+
+class KymoLines:
+    """Kymograph lines"""
```
rpauszek, line 166: :thumbsup: From a user standpoint, I'm a big fan of simple custom classes to handle managing lists of specific objects

mwilkowski (commented):
Added a few comments but I think it's a bit too arcane for me to confidently click approve

lumicks/pylake/kymotracker/tests/test_io.py
```diff
@@ -0,0 +1,78 @@
...
+
+    # Test round trip through the API
+    testfile = f"{tmpdir_factory.mktemp('pylake')}/test.csv"
+    lines.save(testfile, dt, dx, delimiter, sampling_width)
+    read_file = import_kymolines_from_csv(testfile, test_img, delimiter=delimiter)
+
+    # Test raw fields
+    data = read_txt(testfile, delimiter)
+
+    for l1, l2 in zip(lines, read_file):
```
rpauszek, line 57: nitpick: I'm all for using short variable names, but I really try to avoid using a single `l`. Personally I would prefer `line1` and `line2` instead?

rpauszek (approved):
Couple of very minor comments on clarity, but otherwise looks great

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -99,3 +99,33 @@ def extrapolate(self, forward, n_estimate, extrapolation_length):
 
     def __len__(self):
         return len(self.coordinate_idx)
+
+
+class KymoLines:
```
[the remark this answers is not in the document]
JoepVanlier: Good point. `KymolineGroup`.

lumicks/pylake/tests/test_kymotracker.py
JoepVanlier: Yes. Amending! :)

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Thanks. That one wasn't on my radar, but it is indeed nicer.

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Yeah, I'll add it. Only risk is going out of date with the implementation, but we'll just have to be vigilant of this. Added the following docstring:

```python
    """Import kymolines from csv

    Parameters
    ----------
    filename : str
        filename to import from
    image : array_like
        2D image that these lines were tracked from
    delimiter : str
        A delimiter that delimits the column data.

    The file format contains a series of columns as follows:
    line index, time (pixels), coordinate (pixels), time (optional), coordinate (optional), sampled_counts (optional)"""
```

lumicks/pylake/kymotracker/tests/test_io.py
JoepVanlier: I only do it when the scope is one or two lines, but I'm alright with avoiding it. I'll change it to `line1` and `line2` :)

PR #90, opened 2020-12-03
nbwidgets: mouse drag

JoepVanlier:
**Why this PR?**
It is required for the kymotracker widget to work.

**What is it?**
It's a small handler to deal with mouse drag events on a plot where the axis limits respond to dragging.
I have also factored out the mocked mouse event to be able to reuse it in different tests.


lpingas (approved):
LGTM!

PR #93, opened 2020-12-11
Kymo plot with histogram

rpauszek:
**Why this PR?**
User request for histograms plotted alongside a kymo (for example, to visualize binding events)
- can histogram along either time or position axes
- user inputs the number of pixels to bin together
Example:
image

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -3,7 +3,7 @@ import numpy as np
 import warnings
 
 from .detail.confocal import ConfocalImage
-from .detail.image import reconstruct_image_sum, reconstruct_image, line_timestamps_image, seek_timestamp_next_line
+from .detail.image import reconstruct_image_sum, reconstruct_image, line_timestamps_image, seek_timestamp_next_line, histogram_rows
 from .detail.timeindex import to_timestamp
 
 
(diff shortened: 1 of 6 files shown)


lumicks/pylake/kymo.py
```diff
@@ -150,6 +150,87 @@ def set_aspect_ratio(axis, ar):
...
+
+        n_bins = n_rows // pixels_per_bin
+        remainder = n_rows % pixels_per_bin
+        if remainder != 0:
+            warnings.warn(f"{n_rows} pixels is not divisible by {n_bins} bins, skipping last {remainder} pixels")
+        image = image[:(n_bins * pixels_per_bin)].reshape((n_bins, -1))
+
+        counts = image.sum(axis=1)
+        edges = np.arange(n_bins) * bin_width
+        return edges, counts, bin_width
```
JoepVanlier, line 167: This function doesn't use self, so it could just be a free function, which would also make it more testable.

Would be nice to have a test for both divisible summation counts and non-divisible with a tiny image where you know what needs to come out.

Nicely vectorized btw!

lumicks/pylake/tests/test_kymo.py
```diff
@@ -145,6 +145,33 @@ def test_plotting_with_force(h5_file):
         kymo.plot_with_force(force_channel="2x", color_channel="red")
         assert np.allclose(np.sort(plt.xlim()), [-0.5, 3.5], atol=0.05)
         assert np.allclose(np.sort(plt.ylim()), [10, 30])
+
+
+@cleanup        
+def test_plotting_with_histograms(h5_file):
```
JoepVanlier, line 151: It's good that it tests whether the plot spins up and detects bad input, but I'm missing a test which actually tests the histogram data.

lumicks/pylake/kymo.py
```diff
@@ -150,6 +150,87 @@ def set_aspect_ratio(axis, ar):
...
+        bin_width = pixels_per_bin * pixel_width # in physical units
+        n_rows = image.shape[0]
+        if pixels_per_bin > n_rows:
+            raise ValueError("bin size is larger than the available pixels")
+
+        n_bins = n_rows // pixels_per_bin
+        remainder = n_rows % pixels_per_bin
+        if remainder != 0:
+            warnings.warn(f"{n_rows} pixels is not divisible by {n_bins} bins, skipping last {remainder} pixels")
+        image = image[:(n_bins * pixels_per_bin)].reshape((n_bins, -1))
```
JoepVanlier, line 163: I find the skipping for the overhanging pixel a strange choice from a user perspective. The warning is good, but seeing the last pixel be empty is strange to me when I have the data right there. I don't feel that strongly about it, but thought I'd just mention it.

image

JoepVanlier (changes requested):
Cool! This is a great feature to have!

I also appreciate how you re-used the histogram part. Very nice!

I have some requests, mostly pertaining to additional tests.

JoepVanlier (approved):
Looks good to me!

PR #94, opened 2020-12-23
Include downsampled_force## when printing File

rpauszek:
**Why this PR?**
to make clear that the `downsampled_force##` attributes exist in the file


lpingas (approved):
👍

PR #99, opened 2021-01-15
channel: fix exception downsampled_over over gap

JoepVanlier:
*Why this PR?*
Sometimes, `TimeSeries` data can contain a large gap. When you `downsample_over` a gap you get an exception (since some of the slices are empty). This can happen for example when you have data where the distance lock is turned off and then on.

This PR fixes that by ignoring samples where there is no data. Note that since we don't guarantee that a `Slice` of `TimeSeries` is contiguous, it should be evident that the returned `Slice` doesn't have to be either.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.7.3 | t.b.d.
+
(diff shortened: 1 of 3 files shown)


rpauszek (approved):
looks good

PR #100, opened 2021-01-19
FDCurve - distance range selector widget

rpauszek:
**Why this PR?**
User request for a notebook widget to select a range based on distance rather than time

**Key Points & Notes**
- `fd_selector.py` was renamed to `range_selector.py` (internal module)
- selector widgets code was refactored to make more general, easier to add new widgets
- some internal widget classes renamed to be more consistent
- noise in distance measurement can lead to premature truncation of the time channel if the longest contiguous block of data is selected from a pure logical mask (see figure, left and middle columns). An optional argument allows for small gaps caused by noise in the data (figure, right column)
try_slicing_result
the desired distance threshold is displayed as the dashed black line in relevant plots

See the docs [here](https://lumicks-pylake.readthedocs.io/en/range_selector/tutorial/nbwidgets.html)

### DIFF
--- a/docs/tutorial/nbwidgets.rst
+++ b/docs/tutorial/nbwidgets.rst
@@ -44,6 +44,9 @@ plot all of our selections in separate plots for instance, we can do the followi
 F,d selection
 -------------
 
+Range selection by time
+^^^^^^^^^^^^^^^^^^^^^^^
+
 Assume we have an F,d curve we want to analyze. We know that this file contains one F,d curve which should be split up
 into two segments that we should be analyzing separately::
 
@@ -78,6 +81,9 @@ This produces a separate plot for each selection. There's also a more direct way
 
 .. image:: fd_widget3.png
 
+Processing multiple files
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
 Now let's say our experiment is split up over multiple files, each containing a few F,d curves. We would like to load
 these curves all at once and make our selections. We can do this using automatically using `glob`. With `glob.glob`
 we grab a list of all `.h5` files in the directory `my_directory`. We then iterate over this list and open each file.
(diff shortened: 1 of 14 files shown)


lumicks/pylake/fdcurve.py
```diff
@@ -79,6 +80,31 @@ def __getitem__(self, item):
...
+        # logical mask of data lying within selected distance range
+        mask_in_range = np.logical_and(min_dist <= self.d.data, self.d.data <= max_dist).astype(np.bool)
+
+        # find runs of data that are not in selected distance range
+        # runs shorter than the allowed gap should be allowed
+        ranges, lengths = find_contiguous(~mask_in_range)
+        for rng, ln in zip(ranges, lengths):
+            if ln < max_gap:
+                # ignore first and last runs of data, as total length is not defined
+                if rng[0] == 0 or rng[1] == self.d.data.size:
```
JoepVanlier, line 94: I'm not sure I'd understand from the comment what this condition is trying to catch.

Maybe something like 
`# We want to include short sections of data that fall outside the selected distance range, when these excursions are caused by noise. However, we do not want to extend to the start or end of the curve, since we cannot know whether we are simply going out of the range because of noise, or whether we have just actually left the range.`

lumicks/pylake/nb_widgets/range_selector.py
```diff
@@ -3,34 +3,53 @@
...
+        self.range_conversion_fcn = range_conversion_fcn
         self.open = show
 
         if show:
             self._axes.get_figure().canvas.mpl_connect('close_event', self.close)
             self.update_plot()
 
+    @property
+    def xdata(self):
+        raise NotImplementedError
```
JoepVanlier, line 35: Public?

lumicks/pylake/fdcurve.py
```diff
@@ -79,6 +80,31 @@ def __getitem__(self, item):
...
+
+        # find longest contiguous run of data and return corresponding FD curve
+        ranges, lengths = find_contiguous(mask_in_range)
+        longest_range = ranges[np.argmax(lengths)]
+        start = self.d.timestamps[longest_range[0]]
+        try:
+            stop = self.d.timestamps[longest_range[-1]]
+            return self[start:stop]
+        except IndexError: # slicing to the end of data
+            return self[start:]
```
JoepVanlier, line 106: Can we make this a bit more specific? Why not just use an if-statement that checks whether the end of the range is beyond the end of the timestamps? The `try` potentially catches more `IndexErrors` that may go silently unnoticed now.

JoepVanlier (changes requested):
This'll be a really nice feature to have. Worth considering putting it in the notebooks on Harbor too. Also, thanks for the plots in the PR description!

I gave it a once over, but I need to have another look at the refactor part of this PR. Looks cool though. Other than the comments so far, I'm missing documentation and a changelog entry too.

JoepVanlier (commented):
Addendum: Refactor part looks good to me.

lumicks/pylake/detail/utilities.py
```diff
@@ -35,3 +35,11 @@ def lighten_color(c, amount):
     hsv = mpl.colors.rgb_to_hsv(mpl.colors.to_rgb(c))
     hsv[2] = np.clip(hsv[2]+amount, 0.0, 1.0)
     return mpl.colors.hsv_to_rgb(hsv)
+
+def find_contiguous(mask):
+    """Find [start,stop] indices and lengths of contiguous blocks where mask is True."""
+    padded = np.hstack((0, np.asarray(mask, dtype=bool) == True, 0))
```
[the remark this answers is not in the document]
rpauszek: yes, I think it was extraneous code from when I was trying to get it working. I still cast the array to `np.bool`, this way if a numeric array is passed as the argument, it effectively finds contiguous ranges of non-zero values

lumicks/pylake/nb_widgets/range_selector.py
rpauszek: 👍 wasn't sure if it was necessary to make it internal since these classes are internal anyway, but better safe than sorry

docs/tutorial/nbwidgets.rst
```diff
@@ -102,3 +108,42 @@ dictionary of curve sets, and not the keys (which in our case are the curve name
...
+    plt.legend()
+
+    plt.subplot(2, 1, 2)
+    original.f.plot()
+    sliced.f.plot(start=original.start)
+
+.. image::  fd_dist_widget2.png
+
+The returned F,d curves correspond to the longest contiguous (in time) stretch of data that falls 
+within the distance thresholds. However, noise in the distance measurement can lead to short gaps of the time
```
JoepVanlier, line 140: `within the distance thresholds` -> `within the selected distance range` ?

JoepVanlier (approved):
Looks good to me.

JoepVanlier (approved):
LGTM :shipit:

PR #102, opened 2021-01-28
Moving force calibration into Pylake

JoepVanlier:
**Why this PR?**
We want to move force calibration into Pylake. We may want to refactor some of the code, but for this, we first need to make sure the code is tested. This first PR just imports the code as-is and adds testing.

**What's in this PR?**
The current version of the force calibration script.
- Tests for the power spectrum computation
- Tests for the power spectrum blocking
- Integration tests for power spectrum fitting based on a theoretical Lorentzian
- Integration test with real BL data.
- A reference spectrum in an npz file which is stored on git-lfs.
- An extra step in ci that also checks out the lfs materials.

**What's not in this PR?**
- A nice API for Pylake.
- Refactoring of `lumicks/pylake/force_calibration/powerspectrum_calibration.py`. Please defer comments on this file until we actually start refactoring it.

If you're interested in the theory, see: https://www.photonics.ethz.ch/fileadmin/user_upload/Courses/NanoOptics/Opticaltweezrs_Berg_2004.pdf

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -21,3 +21,4 @@
 *.bat text eol=crlf merge=union
 *.png filter=lfs diff=lfs merge=lfs -text
 *.ipynb filter=lfs diff=lfs merge=lfs -text
+*.npz filter=lfs diff=lfs merge=lfs -text
--- a/.github/workflows/pylake_release.yml
+++ b/.github/workflows/pylake_release.yml
@@ -11,6 +11,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+      with:
+        lfs: true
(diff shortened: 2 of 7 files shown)


lpingas (approved):
It is looking good! It is always nice to see unexpected behavior captured by tests.

I would just suggest renaming files from `*powerspectrum*` to `*power_spectrum*` as it would look better.

JoepVanlier:
> It is looking good! It is always nice to see unexpected behavior captured by tests.
> 
> I would just suggest renaming files from `*powerspectrum*` to `*power_spectrum*` as it would look better.

👍

PR #103, opened 2021-01-28
Document access to scan metadata

rpauszek:
**Why this PR?**
Let users know what type of metadata is available for scans and kymos

**What's in this PR?**

* added `center_point_um` and `scan_width_um` properties
* added documentation for all public properties in `Scan` and `Kymo` classes. This should hopefully avoid many users from directly accessing the `.json` metadata
* cleaned up documenation for Confocal Images and Kymographs

Updated docs for [Scan](https://lumicks-pylake.readthedocs.io/en/scan_props/tutorial/images.html) and [Kymo](https://lumicks-pylake.readthedocs.io/en/scan_props/tutorial/kymographs.html)

### DIFF
--- a/docs/tutorial/images.rst
+++ b/docs/tutorial/images.rst
@@ -26,12 +26,10 @@ Or just pick a single one::
     scan = file.scans["name"]
     scan.plot_red()
 
-If a few pixels dominate the image, one might want to set the scale by hand. We can pass an extra argument to `plot_red`
-named `vmax` to accomplish this. This parameter gets forwarded to :func:`matplotlib.pyplot.imshow`::
+Scan data and details
+---------------------
 
-    scan.plot_red(vmax=5)
-
-Access the raw image data::
+You can access the raw image data::
 
     rgb = scan.rgb_image  # matrix with `shape == (h, w, 3)`
     blue = scan.blue_image  # single color so `shape == (h, w)`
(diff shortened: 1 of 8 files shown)


lumicks/pylake/scan.py
```diff
@@ -97,8 +97,7 @@ def _plot(self, image, frame=1, image_handle=None, **kwargs):
             image = image[frame - 1]
 
         axes = self._ordered_axes()
```
JoepVanlier, line 99: Consider moving this in closer proximity to the plot command, since that is now the first use of these axes. Helps to keep variables more local to their use.

docs/tutorial/images.rst
```diff
@@ -45,9 +43,25 @@ For an even lower-level look at data, the raw photon count samples can be access
     photons = scan.red_photons
     plt.plot(photons.timestamps, photons.data)
 
-The images can also be exported in the TIFF format::
+There are also several properties available for convenient access to the scan metadata
```
JoepVanlier, line 46: End with a `:`?

I know that this is not part of this PR, but line 40 has always irked me a bit. I'd suggest changing it from:
```The images contain pixel data where each pixel is made up a multiple photon count samples collected by the scanner.```
to:
```The images contain pixel data where each pixel represents summed photon count samples collected by the scanner.```

or even just `summed photon counts`

File: docs/tutorial/kymographs.rst [hunk 3]
```diff
@@ -32,6 +35,23 @@ Access the raw image data::
     # Plot manually
     plt.imshow(rgb)
 
+Kymographs can also be sliced in order to obtain a specific time range.
+For example, one can plot the region of the kymograph between 175 and 180 seconds using::
+
+    kymo["175s":"180s"].plot_red()
+
+There are also several properties available for convenient access to the kymograph metadata
```
JoepVanlier, line 43: End with a `:`?

JoepVanlier (commented):
Cool! Better docs, always good stuff. 👍 
There are a few more fields in the `json` afaik. `has_fluorescence`, `has_force`? I don't know if anyone uses those directly though. I'm still missing documentation on the pixel time, which is in the json?

docs/tutorial/kymographs.rst
```diff
@@ -32,6 +35,23 @@ Access the raw image data::
...
+    kymo["175s":"180s"].plot_red()
+
+There are also several properties available for convenient access to the kymograph metadata:
+
+* `kymo.center_point_um` provides a dictionary of the central x, y, and z coordinates of the scan in micrometers relative to the brightfield field of view
+* `kymo.scan_width_um` provides a list of scan widths in micrometers along the axes of the scan
+* `kymo.pixelsize_um` provides the pixel size in micrometers
+* `kymo.pixels_per_line` provides the number of pixels in each line of the kymograph
+* `kymo.fast_axis` provides the axis that was scanned (x or y)
+* `kymo.line_time_seconds` provides the scan time per line in seconds
```
JoepVanlier, line 50: Nitpick, but maybe good for correctness: "provides the time between successive lines"; since the scan time is shorter (flyback).

JoepVanlier (approved):
Thanks for this! I left one last tiny comment on the docs, but I think it's pretty much good to go.

PR #104, opened 2021-02-02
Fixes for deprecated functionalities in numpy

JoepVanlier:
**Why this PR?**
Some functionalities were deprecated in numpy. We'd like to stay up to date.

**Why the addition to setup.cfg?**
h5py is still using some deprecated behaviour. They already have a fix for this (see: https://github.com/h5py/h5py/pull/1780/files ) but until that's out there, it would be nice to silence this one error.

### DIFF
--- a/lumicks/pylake/tests/test_file.py
+++ b/lumicks/pylake/tests/test_file.py
@@ -283,11 +283,9 @@ def test_invalid_access(h5_file):
 def test_missing_metadata(h5_file_missing_meta):
     f = pylake.File.from_h5py(h5_file_missing_meta)
     if f.format_version == 2:
-        with pytest.warns(UserWarning) as record:
+        with pytest.warns(UserWarning, match="Scan 'fast Y slow X no meta' is missing metadata and cannot be loaded"):
             scans = f.scans
             assert len(scans) == 1
-            assert len(record) == 1
-            assert record[0].message.args[0] == "Scan 'fast Y slow X no meta' is missing metadata and cannot be loaded"
 
 
 def _internal_h5_export_api(file, *args, **kwargs):
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -359,7 +359,7 @@ class Continuous:
 
     @property
     def _timesteps(self):
-        return np.asarray([self.dt], dtype=np.int)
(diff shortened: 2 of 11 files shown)


lumicks/pylake/tests/test_file.py
```diff
@@ -283,11 +283,16 @@ def test_invalid_access(h5_file):
...
             assert len(scans) == 1
-            assert len(record) == 1
-            assert record[0].message.args[0] == "Scan 'fast Y slow X no meta' is missing metadata and cannot be loaded"
+            assert np.any(
+                [
+                    record.message.args[0]
+                    == "Scan 'fast Y slow X no meta' is missing metadata and cannot be loaded"
+                    for record in records
+                ]
+            )
```
rpauszek, line 295: what caused this change to be necessary? it wasn't clear to me from the numpy changlelog

rpauszek (approved):
I'm just confused as to why the one change was necessary, but all else looks fine to me

lumicks/pylake/tests/test_file.py
JoepVanlier: Yeah, I also didn't expect it initially, but the reason for this change is that you get multiple warnings when accessing that scan (including non-`UserWarnings`):
``` DeprecationWarning('Passing None into shape arguments as an alias for () is deprecated.'), category : 'DeprecationWarning', filename : 'C:\\Conda\\envs\\test_pylake_latest\\lib\\site-packages\\h5py\\_hl\\dataset.py', lineno : 767, line : None}
{message : DeprecationWarning('Passing None into shape arguments as an alias for () is deprecated.'), category : 'DeprecationWarning', filename : 'C:\\Conda\\envs\\test_pylake_latest\\lib\\site-packages\\h5py\\_hl\\dataset.py', lineno : 767, line : None}
{message : UserWarning("Scan 'fast Y slow X no meta' is missing metadata and cannot be loaded"), category : 'UserWarning', filename : 'D:\\widget_hacks\\pylake\\lumicks\\pylake\\file.py', lineno : 201, line : None}```

All we want to really test is whether our warning is there, so that's why I changed it. Anyways, I just had another look and seems like pytest has a nice API for actually checking whether the string matches too, so I updated it to use that.

PR #105, opened 2021-02-05
Export (aligned) tiff files

rpauszek:
**Why this PR?**
Users would like to export the aligned images for use in other software or presentation

**What's in this PR?**
* added `CorrelatedStack.export_tiff()` method. `roi` kwarg allows for exporting a specific region
* some refactoring of the tests, adding fixtures to deal with test image and file generation
* works with either RGB or grayscale stacks, single- or multi-frame stacks

### DIFF
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -32,3 +32,14 @@ using timestamps obtained from the `CorrelatedStack`::
 
     # Determine the force trace averaged over frame 2...9.
     file.force1x.downsampled_over(stack[2:10].timestamps)
+
+The aligned image stack can also be exported to tiff format::
+
+    stack.export_tiff("aligned_stack.tiff")
+    stack[5:20].export_tiff("aligned_short_stack.tiff") # export a slice of the CorrelatedStack
+
+Generally, the edges of an aligned image can become corrupted due interopolation artefacts. 
+In this case, we can export a cropped region of interest by supplying the `roi` kwarg in the form
+`[min_x_pixel, max_x_pixel, min_y_pixel, max_y_pixel]`::
+
+    stack.export_tiff("aligned_cropped_stack.tiff", roi=[20, 280, 20, 180])
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -28,10 +28,10 @@ class TiffFrame:
             translation = np.eye(3)
             translation[0, -1] = -x_offset
             translation[1, -1] = -y_offset
-            # apply the original transform to the translated image. 
+            # apply the original transform to the translated image.
             # it only needs to be resized from a 2x3 to a 3x3 matrix
             original = np.vstack((alignment, [0, 0, 1]))
-            # translate the image back to the original origin. 
+            # translate the image back to the original origin.
             # takes into account both the offset and the scaling performed by the first step
             back_translation = np.eye(3)
             back_translation[0, -1] = original[0, 0] * x_offset
@@ -53,7 +53,7 @@ class TiffFrame:
         except KeyError:
             warnings.warn("File does not contain alignment matrices. Only raw data is available")
             return self.raw_data
-        
+
         x_offset, y_offset = align_roi - roi
         if not (x_offset == 0 and y_offset == 0):
             align_mats = [correct_alignment_offset(mat, x_offset, y_offset) for mat in align_mats]
@@ -75,7 +75,7 @@ class TiffFrame:
         return self._src.asarray()
 
     @property
-    def bit_depth(self):        
+    def bit_depth(self):
         bit_depth = self._src.tags["BitsPerSample"].value
         if self.is_rgb: # (int r, int g, int b)
             return bit_depth[0]
@@ -86,7 +86,7 @@ class TiffFrame:
     def is_rgb(self):
         return self._src.tags["SamplesPerPixel"].value == 3
 
-    def _get_plot_data(self, channel="rgb", vmax=None): 
+    def _get_plot_data(self, channel="rgb", vmax=None):
         """ return data an numpy array, appropriate for use by `imshow`
             if data is grayscale or channel in ('red', 'green', 'blue')
                 return data as is
@@ -94,7 +94,7 @@ class TiffFrame:
                 None  : normalize data to max signal of all channels
                 float : normalize data to vmax value
         """
-        
+
         if not self.is_rgb:
             return self.data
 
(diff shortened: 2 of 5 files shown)


lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -419,6 +419,56 @@ def rgb_image_data_offset(spot_coordinates, warp_parameters):
...
+            tag_datetime = (306, 's', len(str_datetime), str_datetime, False)
+            tif.save(frame, 
+                    description=json.dumps(description, indent=4),
+                    software="Bluelake Unknown",
+                    metadata=None, contiguous=False,
+                    extratags=(tag_orientation, tag_sample_format, tag_datetime))
+
+
+@pytest.fixture(scope="session")
+def rgb_file(tiff_dir, rgb_image_data):
```
tiagobonetti, line 346: I would suggest referring to these fixtures as tiff files, or maybe just tiffs, so is very clear what kind of files they are.

docs/tutorial/correlatedstacks.rst
```diff
@@ -32,3 +32,14 @@ using timestamps obtained from the `CorrelatedStack`::
 
     # Determine the force trace averaged over frame 2...9.
     file.force1x.downsampled_over(stack[2:10].timestamps)
+
+The aligned image stack can also be exported to tiff format::
+
+    stack.export_tiff("aligned_stack.tiff")
+    stack[5:20].export_tiff("aligned_short_stack.tiff") # export a slice of the CorrelatedStack
```
JoepVanlier, line 39: Nice! 👍

lumicks/pylake/correlated_stack.py
```diff
@@ -340,6 +340,59 @@ def select_frame(event):
...
+            xmax = self._get_frame(0).data.shape[1]
+
+        # re-name alignment matrices fields in image description
+        # to reflect the fact that the image has already been processed
+        description = self._get_frame(0)._src.description
+        if description != "":
+            description = json.loads(description)
+            for color in ("blue", "green", "red"):
+                key = f"Alignment {color} channel"
+                description["Applied "+key] = description.pop(key)
```
JoepVanlier, line 380: What happens if a user opens a file that was exported in this manner with `CorrelatedStack`?

I guess it wouldn't be aligned again, which is what we want (though they'd be getting warnings all the time). But what if they export that file again, I guess then the alignment matrices could be lost?

tiagobonetti (commented):
Great stuff I added some questions and comments.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -361,24 +362,65 @@ def make_description(m_red, m_blue, offsets):
...
+    img_args = {"spots": spot_coordinates,
+                **warp_parameters,
+                "bit_depth": 16}
+    img0, img, description = make_alignment_image_data(**img_args)
+    description["Alignment red channel"][2] = 25
+    return img0, img, description, img_args["bit_depth"]
 
+
+@pytest.fixture(scope="session")
+def rgb_image_data_offset(spot_coordinates, warp_parameters):
```
JoepVanlier, line 314: Maybe `rgb_alignment_image_data`? For the others too. To drive home the point that these are intended for alignment purposes?

lumicks/pylake/correlated_stack.py
```diff
@@ -340,6 +340,59 @@ def select_frame(event):
...
+            xmin, xmax, ymin, ymax = roi
+        else:
+            ymin, xmin = 0, 0
+            ymax = self._get_frame(0).data.shape[0]
+            xmax = self._get_frame(0).data.shape[1]
+
+        # re-name alignment matrices fields in image description
+        # to reflect the fact that the image has already been processed
+        description = self._get_frame(0)._src.description
+        if description != "":
```
JoepVanlier, line 376: Is the `!= ""` needed here?

JoepVanlier (commented):
Still need to have a closer look at the tests tomorrow. Have a few questions already though.

lumicks/pylake/correlated_stack.py
```diff
@@ -340,6 +340,59 @@ def select_frame(event):
...
+            description = json.loads(description)
+            for color in ("blue", "green", "red"):
+                key = f"Alignment {color} channel"
+                description["Applied "+key] = description.pop(key)
+
+        # add pylake to Software tag
+        software = self._get_frame(0)._src.tags["Software"].value
+        software += f", pylake v{version}"
```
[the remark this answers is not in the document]
rpauszek: I don't think so, it should be set to the Bluelake version when exported from there

lumicks/pylake/correlated_stack.py
```diff
@@ -340,6 +340,59 @@ def select_frame(event):
...
+
+        # write frames sequentially
+        with tifffile.TiffWriter(file_name) as tif:
+            for frame in self:
+                tif.save(frame.data[ymin:ymax, xmin:xmax], 
+                         description=json.dumps(description, indent=4),
```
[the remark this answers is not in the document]
rpauszek: `tifffile` only saves the description on the first page. I thought it was cleaner to just keep it in the loop instead of having a separate conditional statement to check for the first page

lumicks/pylake/correlated_stack.py
rpauszek: I had some test data where the description tag was missing, and therefore returned as an empty string when opened with `tifffile`. Without the check, you get an exception with `json.loads()`. I don't know that this would ever actually happen from a stable version of Bluelake though.

JoepVanlier: No, don't get rid of the check. I meant that an empty string usually behaves as `False` when used as a conditional statement.

lumicks/pylake/correlated_stack.py
```diff
@@ -340,7 +340,7 @@ def select_frame(event):
 
         fig.canvas.mpl_connect('button_press_event', select_frame)
 
-    def export_tiff(self, file_name, roi):
+    def export_tiff(self, file_name, roi=None):
```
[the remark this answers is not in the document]
rpauszek: I realized I had forgotten to put in the default value halfway through the 2nd commit. I think this can all be squashed though because it's really just one feature with a bunch of testing

lumicks/pylake/correlated_stack.py
rpauszek: correct, it's not aligned again because the metdata field names have changed on export. If you try to re-load the exported file, it does indeed issue a warning, but I don't think that's necessarily a bad thing. It reminds the user that they are not working with the original raw data. I fixed it so the metadata is not changed on re-export (the fields remain "Applied alignment color channel")

lumicks/pylake/correlated_stack.py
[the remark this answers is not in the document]
tiagobonetti: I'm not very knowledgeable about python but my impression would be that this code would call `json.dumps(...)` every iteration of the loop, and send the result to the function.

My suggestion would be to avoid this doing this (no conditional needed):
```python
            description = json.dumps(description, indent=4)
            for frame in self:
                tif.save(frame.data[ymin:ymax, xmin:xmax],
                         description=description,
```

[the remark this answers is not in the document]
tiagobonetti: I'm not very knowledgeable about python but my impression would be that this code would call `json.dumps(...)` every iteration of the loop, and send the result to the function.

My suggestion would be to avoid this doing this (no conditional needed):
```python
            description = json.dumps(description, indent=4)
            for frame in self:
                tif.save(frame.data[ymin:ymax, xmin:xmax],
                         description=description,
```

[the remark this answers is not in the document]
rpauszek: oh, I thought you were talking about something else ... good catch

tiagobonetti (approved):
Everything seems pretty `honk` on my side. :+1:

PR #107, opened 2021-02-10
Deprecate confocal json public attribute

rpauszek:
**Why this PR?**
Currently the public `json` attribute in `confocal.BaseScan` is a regular Python dict read from the json metadata exported by Bluelake. This attribute should be deprecated so users are warned not to use it.

**What changed?**
* `BaseScan.json` is now a property decorated with `@deprecated`
* internal access to the metadata is still available as before through `BaseScan._json` 
* deprecated `BaseScan.has_force` and `BaseScan.has_fluorescence`. These fields were never really useful, now return hard-coded boolean values

### DIFF
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -1,6 +1,7 @@
 import json
 import numpy as np
 import cachetools
+from deprecated.sphinx import deprecated
 
 from .mixin import PhotonCounts
 from .mixin import ExcitationLaserPower
@@ -28,7 +29,7 @@ class BaseScan(PhotonCounts, ExcitationLaserPower):
         self.start = start
         self.stop = stop
         self.name = name
-        self.json = json
+        self._json = json
         self.file = file
         self._cache = {}
 
@@ -54,6 +55,13 @@ class BaseScan(PhotonCounts, ExcitationLaserPower):
 
         return cls(name, file, start, stop, json_data)
 
+    @property
+    @deprecated(reason=("Access to raw metadata will be removed in a future release. "
+                        "Use accessor properties instead. (see docs)"),
+                        action="always", version="0.8.0")
+    def json(self):
+        return self._json
+
     def _get_photon_count(self, name):
         """Grab the portion of the photon count that overlaps with the scan."""
         photon_count = getattr(self.file, f"{name}_photon_count".lower())[self.start:self.stop]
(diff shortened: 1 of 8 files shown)


lumicks/pylake/tests/test_kymo.py
```diff
@@ -18,8 +18,9 @@ def test_kymo_properties(h5_file):
                                         [2.084375e+10, 2.187500e+10, 2.284375e+10, 2.387500e+10]], np.int64)
 
         assert repr(kymo) == "Kymo(pixels=5)"
-        assert kymo.has_fluorescence
-        assert not kymo.has_force
+        with pytest.warns(DeprecationWarning):
+            assert kymo.has_fluorescence
+            assert not kymo.has_force
```
lpingas, line 23: I would suggest using [`pytest.deprecated_call()`](https://docs.pytest.org/en/stable/warnings.html#ensuring-code-triggers-a-deprecation-warning) for this.

Also, if the intention is to test the deprecation, I believe that each statement requires its own `with` clause since `has_force` could omit the warning if `has_flourescence` already raised one.

Lastly, you could also test if `.json()` also raises the warning.

lpingas (commented):
It is looking great! Way simpler than I expected : D

I left a comment on the tests.

lumicks/pylake/detail/confocal.py
```diff
@@ -1,6 +1,7 @@
 import json
 import numpy as np
 import cachetools
+from deprecated.sphinx import deprecated
```
dean0x7d, line 4: Oh, that's cool! I didn't know they added Sphinx integration. That's great!

lumicks/pylake/tests/test_kymo.py
dean0x7d: Indeed, each statement needs its own check.

rpauszek: 👍 I wasn't aware of `pytest.deprecated_call()`, thanks!

lumicks/pylake/tests/test_kymo.py
```diff
@@ -103,8 +107,9 @@ def test_kymo_slicing(h5_file):
             empty_kymograph.plot_rgb()
 
         assert empty_kymograph.red_image.shape == (5, 0)
-        assert empty_kymograph.has_fluorescence
-        assert not empty_kymograph.has_force
+        with pytest.warns(DeprecationWarning):
+            assert empty_kymograph.has_fluorescence
+            assert not empty_kymograph.has_force
```
dean0x7d, line 112: You missed one here.

rpauszek: d'oh!

dean0x7d (approved):
```py
from deprecated.sphinx import versionadded
from pr import approved
```
Just update the PR title to better reflect the latest contents!

PR #108, opened 2021-02-10
Ldat refactor (pt. 1)

JoepVanlier:
**Why?**
Would be nice to allow different `ForceCalibration` models to be used.

ldat2

**What is all in this PR?**
`PowerSpectrum` is in its own file, with its own tests. I have moved keeping track of how much averaging has taken place into `PowerSpectrum`, since this guy owns that information.

Instead of a stateful class, the calibration is now a function that raises an exception when things go awry. You do however want access to the `PowerSpectrum` used for fitting if it fails. Instead, you now compute the `PowerSpectrum` with its own API, and pass it and an appropriate model to the calibration function. I've added some plotting functionality to it as well, which is probably not so relevant for automation/Bluelake, but nice to have in Pylake when adding new `ForceCalibration` models.

In the future, we want to add additional models which do additional things. I have defined a `CalibrationModel` interface of functions which should be implemented for that to work. This meant modifying the class with fixed attributes named `CalibrationResults`. Instead, I have replaced it with a class which uses `__getitem__` to provide dictionary-like access (akin to `ForceCalibration` items in Pylake). If these need to be attributes for some reason, we could change this.

I have wrapped parameters into a little tuple with their description. Initially, I had a std error inside this tuple, because I think it would be nice to pack these two elements of data together. While I think that would be nicer, unfortunately, this is not how the force calibration structure we have in Pylake is set up at the moment. Instead, I went with a parameter layout that for static calibration produces exactly the same fields as what we have in Pylake now (except for one exception listed below). Hopefully, this will lead to users being instantly familiar with the new API.

Some things I shuffled back and forth a bit.

One thing I was unsure about is the duplicate field:
```
     'Response (pN/V)': 1265.2669140456192,
     'Rf (pN/V)': 1265.2669140456192,
```

Do we need both of them?

I have moved the parameter scaling that was happening into its own class so that it is separated from the actual model. This is a (mostly) feature parity PR, so I have not yet removed the non-linear rescaling used to introduce parameter bounds yet (as this PR is already pretty chunky). Since `curve_fit` supports bounds natively nowadays, I'd prefer to use this in pt 2 of this refactor.

There's a few items of information that do not make it to the CalibrationResults, which Bluelake probably fills. These are:
```
    {'Fit range (max.) (Hz)': 23000.0,
     'Fit range (min.) (Hz)': 100.0,
     'Kind': 'Full calibration',
     'Number of samples': 781250.0,
     'Offset (pN)': 0.0,
     'Sign': 1.0,
     'Start time (ns)': 1611320223633490800,
     'Stop time (ns)': 1611320233633490800,
     }
```

Also worth nothing is that the field `Max iterations` in Calibration fields seems to be incorrect and is actually maximum function evaluations. I kept this naming mishap for now, but maybe we should fix that name.

One thing I doubted about was in `CalibrationResults` and the way it interacts with the model. One thing I notice is that all of the calibration models have a different post-processing step. Here they calculate values (which may be interesting to keep in the calibration) from the parameters obtained in the power spectrum fit. I pass these along to `CalibrationResults` to be presented to the user.

When `CalibrationResults` is created, I check whether the mandatory parameters exist in the params field (the ones required to use the calibration (kappa for instance)). This is to make sure that new models actually expose these parameters in the correct form. Alternatively, one thing we could do is have a regular (mandatory) named parameter `kappa` (and whichever other mandatory parameters exist) for `CalibrationResults`, along with a `get_kappa()` on the models. It would mean storing mutable state in `CalibrationModels` though. I'm not entirely sure how much we want to enforce the structure of these parameters at this point already. We'll probably be the only ones writing these `CalibrationModels`. We'd know the consequences of exposing different things.

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -1,17 +1,23 @@
 import math
 import scipy
-from collections import namedtuple
 from .detail.power_models import passive_power_spectrum_model, sphere_friction_coefficient
+from .power_spectrum_calibration import CalibrationParameter
 
 
 class CalibrationModel:
-    def __init__(self):
-        pass
+    def __init__(self, bead_diameter):
+        self.bead_diameter = bead_diameter
+
+    def __name__(self):
+        raise NotImplementedError
 
     def __call__(self):
         raise NotImplementedError
 
-    def calibration_parameters(self, params):
+    def __repr__(self):
+        return f"{self.__name__()}({''.join([f'{k}={v}, ' for k, v in vars(self).items()])[:-2]})"
+
+    def calibration_parameters(self, fc, diffusion_constant):
         raise NotImplementedError
 
 
@@ -43,8 +49,11 @@ class PassiveCalibrationModel(CalibrationModel):
     temperature : float, optional
         Liquid temperature [Celsius].
     """
+    def __name__(self):
+        return "PassiveCalibrationModel"
+
     def __init__(self, bead_diameter, viscosity=1.002e-3, temperature=20):
-        self.bead_diameter = bead_diameter
+        super().__init__(bead_diameter)
         self.viscosity = viscosity
         self.temperature = temperature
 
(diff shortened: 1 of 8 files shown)


mwilkowski (commented):
Looks good to my layman eye but I sure do not have the courage to click that approve button. Sorry :(

agom (approved):
I'm not an expert in either python or force calibration but looks great and it is nicely tested. Just left a couple of minor comments.

PR #115, opened 2021-02-17
ci: test installation

JoepVanlier:
**Why this PR?**
Because I don't want to release a broken pylake again. During the v0.8.0 we had the problem that one of the files needed in the test wasn't being shipped. This wasn't picked up by CI, because CI doesn't actually test the installed package, but just whether pytest passes in the current working directory.

**Why this solution?**
This solution installs pylake using pip and runs the tests there.

Originally, I opted for the cleaner
`pytest --runslow --color=yes --pyargs lumicks.pylake`

But it wouldn't recognise our custom option `runslow`. Therefore this solution has the preference.

Note that I have moved the warning configuration from `setup.cfg` to `conftest.py`. The reason for this is that we want our testing configured when calling pytest in the imported module as well. This seems like a cleaner solution than adding a `pytest.ini` to the list of things installed with the package (which could be shadowed by an ini the user has).

There were two cases in which certain files were not packaged. Either leaving `include_package_data` out of `setup.py` or leaving the file out of the manifest). I have tested both cases that would lead to incorrect packaging and verified that both of them would now not pass the testing ci.

### DIFF
--- a/.github/workflows/pylake_test.py
+++ b/.github/workflows/pylake_test.py
@@ -1,0 +1,7 @@
+import lumicks.pylake as lk
+import sys
+import os
+import pathlib
+
+os.chdir(str(pathlib.Path(lk.__file__).parent))
+sys.exit(lk.pytest(args=["--runslow", "--color=yes"]))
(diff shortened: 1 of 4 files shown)


tiagobonetti (approved):
Everything seems to make sense, but I don't have much experience with Github Actions so you might want to wait for a second opinion.

dean0x7d (approved):

PR #116, opened 2021-02-18
Configure`black` and reformat the codebase

dean0x7d:
As we talked about. `tests` and `force_calibration` are excluded for now. The reformat commit was done by a machine user to avoid confusing the git history. Let me know if you find anything weird.

The `black` docs have a good page about editor integration for automatic formatting on save: https://github.com/psf/black/blob/master/docs/editor_integration.md


JoepVanlier (approved):
Looks good to me. 👍 

There are the minor weirdnesses that we discussed before, but nothing else pops out at me as being annoying.

PR #117, opened 2021-02-23
force_calibration: remove extra downsample factor conversions

JoepVanlier:
**Why this PR?**
While writing the docs I noticed two things that needed fixups. One, we  use `sample_rate` instead of `sampling_rate`.

The second is that how you specify how much the spectra need to be downsampled is a bit strange. The API takes a number of points to use for averaging. That is then converted to a number of blocks, which is then converted back to a number of points. By changing the API, we get more predictable behaviour.

It also deduplicates some code between channel and force calibration (and allows both to be covered by the same test of the basic functionality).

One thing that's worth noting is that this is part of the reason the original LDAT was using the wrong `n_points_per_block` in the actual fit (since it usually wasn't what you actually got and it was used straight from `self.settings.n_points_per_block` rather than from `PowerSpectrum`, which at the time didn't keep track of how often things had been downsampled).

**Why did the test change?**
When specifying a number of blocks, you can end up with a nearest downsampling factor that would allow more blocks than you asked for. They are then just dropped.

The test in question has 50 data points at a samplerate of 4000, where we request 7 blocks from the API. 50 datapoints leads to a frequency spectrum that is 26 long. We want 7 blocks, so that's a downsampling factor of 3 (floor of 3.71). At that downsampling factor we can actually get 8 blocks, so specifying a number of blocks would end up throwing away data we could have had. This way, it's also more consistent with the rest of the Pylake API.

### DIFF
--- a/lumicks/pylake/detail/utilities.py
+++ b/lumicks/pylake/detail/utilities.py
@@ -1,6 +1,7 @@
 import numpy as np
 import matplotlib.pyplot as plt
 import matplotlib as mpl
+import math
 
 
 def first(iterable, condition=lambda x: True):
@@ -44,3 +45,27 @@ def find_contiguous(mask):
     ranges = np.argwhere(change_points == 1).reshape(-1, 2)
     run_lengths = np.diff(ranges, axis=1).squeeze()
     return ranges, np.atleast_1d(run_lengths)
+
+
+def downsample(data, factor, reduce):
+    """This function is used to downsample data in blocks of size `factor`.
+
+    This function downsamples blocks of size `factor` using the function specified in reduce. If
+    there are insufficient points to fill a final block, this last block is discarded.
+
+    Parameters
+    ----------
+    data : array_like
+        Input data
+    factor : int
+        Factor to downsample by
+    reduce : callable
+        Function to use for reducing the data
+    """
+
+    def round_down(size, n):
+        """Round down `size` to the nearest multiple of `n`"""
+        return int(math.floor(size / n)) * n
+
+    data = data[: round_down(data.size, factor)]
+    return reduce(data.reshape(-1, factor), axis=1)
--- a/lumicks/pylake/force_calibration/power_spectrum_calibration.py
+++ b/lumicks/pylake/force_calibration/power_spectrum_calibration.py
@@ -131,7 +131,7 @@ def guess_f_diode_initial_value(ps, guess_fc, guess_D):
         A good initial value for the parameter `f_diode`, for fitting the full
         power spectrum.
     """
-    f_nyquist = ps.sampling_rate / 2
+    f_nyquist = ps.sample_rate / 2
     P_aliased_nyq = (guess_D / (2 * math.pi ** 2)) / (f_nyquist ** 2 + guess_fc ** 2)
     if ps.P[-1] < P_aliased_nyq:
         dif = ps.P[-1] / P_aliased_nyq
@@ -140,14 +140,14 @@ def guess_f_diode_initial_value(ps, guess_fc, guess_D):
         return 2 * f_nyquist
 
 
-def calculate_power_spectrum(data, sampling_rate, fit_range=(1e2, 23e3), num_points_per_block=350):
+def calculate_power_spectrum(data, sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350):
     """Compute power spectrum.
 
     Parameters
     ----------
     data : np.array
         Data used for calibration.
-    sampling_rate : float
+    sample_rate : float
         Sampling rate [Hz]
     fit_range : tuple (f_min, f_max), optional
         Tuple of two floats, indicating the frequency range to use for the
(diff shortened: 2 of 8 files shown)


lumicks/pylake/detail/utilities.py
```diff
@@ -44,3 +45,12 @@ def find_contiguous(mask):
     ranges = np.argwhere(change_points == 1).reshape(-1, 2)
     run_lengths = np.diff(ranges, axis=1).squeeze()
     return ranges, np.atleast_1d(run_lengths)
+
+
+def downsample(data, factor, reduce):
```
dean0x7d, line 50: It would be good to give it a docstring and tests at this point. Previously, it was used as a private function in the same file so the usage was the main thing defining it. Now that it's become (internally) reusable, it needs a docstring and a short unit test. (Internal API users are very similar to public API users.)

lumicks/pylake/force_calibration/tests/test_power_model.py
```diff
@@ -59,7 +58,7 @@ def test_fit_analytic(reference_models, corner_frequency, diffusion_constant, nu
     fit_range = (0, 15000)
     ps = PowerSpectrum(data, sample_rate=len(data))
     ps = ps.in_range(*fit_range)
-    ps = ps.block_averaged(num_blocks=ps.P.size // num_points_per_block)
+    ps = ps.downsampled_by(num_points_per_block)
```
dean0x7d, line 61: I think that you got in a weird git state because this change is in the second commit but the actual API is added in the third commit.

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -153,16 +153,14 @@ def calculate_power_spectrum(data, sample_rate, fit_range=(1e2, 23e3), num_point
...
         Default: 350.
     """
     if not isinstance(data, np.ndarray) or (data.ndim != 1):
         raise TypeError('Argument "data" must be a numpy vector')
     power_spectrum = PowerSpectrum(data, sample_rate)
     power_spectrum = power_spectrum.in_range(*fit_range)
-    power_spectrum = power_spectrum.block_averaged(
-        num_blocks=power_spectrum.P.size // num_points_per_block
-    )
+    power_spectrum = power_spectrum.downsampled_by(num_points_per_block)
```
dean0x7d, line 163: That's a lot nicer!

lumicks/pylake/tests/test_utilities.py
```diff
@@ -69,3 +69,17 @@ def check_blocks_are_true(mask, ranges):
...
+    [
+        [np.arange(10), 2, [0.5, 2.5, 4.5, 6.5, 8.5], [0.5, 0.5, 0.5, 0.5, 0.5]],
+        [np.arange(0, 10, 2), 1, [0.0, 2.0, 4.0, 6.0, 8.0], [0.0, 0.0, 0.0, 0.0, 0.0]],
+        [np.arange(0, 10, 2), 2, [1.0, 5.0], [1.0, 1.0]],
+        [np.arange(0, 11, 2), 2, [1.0, 5.0, 9.0], [1.0, 1.0, 1.0]],
+    ],
+)
+def test_downsample(data, factor, avg, std):
+    assert np.allclose(avg, downsample(data, factor, reduce=np.mean))
+    assert np.allclose(std, downsample(data, factor, reduce=np.std))
```
dean0x7d, line 85: Well, never mind my previous comment about adding tests for `downsample()`. Just add the docstring then. But your commit contents definitely got shuffled around between the seconds and third commits.

lumicks/pylake/force_calibration/tests/test_power_spectrum.py
```diff
@@ -19,21 +19,21 @@ def test_blocking_separate(data, num_blocks, avg, std):
...
     """Testing the attributes of power spectra"""
-    data = np.sin(2.0 * np.pi * frequency / sampling_rate * np.arange(num_data))
-    power_spectrum = PowerSpectrum(data, sampling_rate)
+    data = np.sin(2.0 * np.pi * frequency / sample_rate * np.arange(num_data))
+    power_spectrum = PowerSpectrum(data, sample_rate)
 
-    df = sampling_rate / num_data
-    frequency_axis = np.arange(0, .5 * sampling_rate + 1, df)
+    df = sample_rate / num_data
+    frequency_axis = np.arange(0, .5 * sample_rate + 1, df)
```
rpauszek, line 22: super nit-picky but I would use `0.5` instead of `.5` as the decimal is easy to miss as far as readability

lumicks/pylake/detail/utilities.py
rpauszek: agreed, but I do like that it's separated out now

lumicks/pylake/force_calibration/tests/test_power_spectrum.py
JoepVanlier: Fixed! :)

lumicks/pylake/detail/utilities.py
JoepVanlier: Yes, will add a docstring :)

lumicks/pylake/force_calibration/tests/test_power_model.py
JoepVanlier: Yep. Must've messed something up rebasing. I was actually planning to merge those two commits into one deduplication commit anyhow, so let me do that now. :)

lumicks/pylake/force_calibration/power_spectrum_calibration.py
JoepVanlier: Thanks! :)

dean0x7d (approved):
Good catch with the sample rate API naming! But I especially like the downsampling cleanup -- much nicer now!

PR #118, opened 2021-02-23
force_calibration: docs

JoepVanlier:
**Why this PR?**
To illustrate how to use the Force Calibration functions.

**Open issues**
- How should we deal with the fact that the calibration is slightly different now? Should we document it? Should I redo the reference calibration in the docs such that the two agree (assuming BL will be updated)?

The reason for the difference is that the original Force Calibration used `num_points_per_block` from the `CalibrationSettings` class even though the actual downsampling provided him with a different actual downsampling. The refactored version fixes this bug leading to differences in the calibration. For the important parameters the differences are very small (fourth digit), but for the `chi_squared_per_deg` the difference can be more than 10%.

See also: https://github.com/lumicks/pylake/pull/120 to see how the erroneous behaviour can be reproduced.

- The default for `num_points` was 350 in the original Force Calibration settings struct, however I see in most bluelake calibrations a `num_points` of 2000 (the function default in the original Force Calibration). Which one should be used as default?

Docs build here: https://lumicks-pylake.readthedocs.io/en/ldat_docs/

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -1,0 +1,95 @@
+Force calibration
+=================
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+Force calibration is used to convert the raw voltages that Bluelake records to actual forces.
+In this tutorial, we will be redoing a calibration that we already performed in Bluelake, to provide an example of using the force calibration module::
+
+    f = lk.File("calibration.h5")
+
+In this tutorial, we want to reproduce calibration item 1.
+We grab the chunk of data that was used for the calibration last time::
+
+    start = f.force1x.calibration[1]["Start time (ns)"]
+    stop = f.force1x.calibration[1]["Stop time (ns)"]
+    force_slice = f.force1x[start:stop]
+
+To be able to calibrate our forces, we first convert back our data to raw voltages.
+We can do this using the previous calibration performed in Bluelake (in this case, the first one in the file)::
+
+    volts = force_slice.data / f.force1x.calibration[0]["Rf (pN/V)"]
+
+Force calibration models are fit to power spectra. To compute a power spectrum from our data::
+
+    power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate)
+
+This function returns a power_spectrum which we can plot::
+
+    power_spectrum.plot()
+
+.. image:: force_calibration_blocked_spectrum.png
+
+Note that the computation of the power spectrum involves some downsampling.
+If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
+
+    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
+
+To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+This model can be found in `PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
+
+.. math::
+
+    P(f) = \frac{D}{2 \pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha)
+
+where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the fitted cutoff. The second term :math:`g` takes into account the slower response of the position detection system and is given by:
+
+.. math::
+
+    g(f, f_{diode}, \alpha) = \alpha^2 + \frac{1 - \alpha ^ 2}{1 + (f / f_{diode})^2}
+
+Here :math:`\alpha` corresponds to the fraction of the signal response that is instantaneous, while :math:`f_{diode}` characterizes the frequency response of the diode.
+
+To convert the parameters obtained from this spectral fit to a trap stiffness, the following is computed:
+
+.. math::
+
+    \kappa = 2 \pi \gamma f_c
+
+Here :math:`\kappa` then represents the final trap stiffness. The parameter :math:`\gamma` corresponds to the friction coefficient of a sphere, which is given by:
+
+.. math::
+
+    \gamma = 3 \pi \eta d
+
+where :math:`\eta` corresponds to the dynamic viscosity and :math:`d` represents the bead diameter.
+
+We use the bead diameter found in the calibration performed in Bluelake.
+You can optionally also provide a viscosity (in Pa/s) and temperature (in degrees Celsius)::
+
+    bead_diameter = f.force1x.calibration[1]["Bead diameter (um)"]
+    force_model = lk.PassiveCalibrationModel(bead_diameter, viscosity=0.001002, temperature=20)
+
+To fit this model to the data, you can now invoke::
+
+    calibration = lk.fit_power_spectrum(power_spectrum, force_model)
+    calibration
+
+This will produce a table with your fitted calibration parameters.
+
+.. image:: force_calibration_table.png
+
+These parameters can be accessed analogously to the calibrations in Bluelake files::
+
+    >>> print(calibration["kappa (pN/nm)"])
+    >>> print(f.force1x.calibration[1]["kappa (pN/nm)"])
+    0.17432391259341345
+    0.17431947810792106
+
+We can plot the calibration by calling::
+
+    calibration.plot()
+
+.. image:: force_calibration_fit.png
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -37,7 +37,7 @@ This function returns a power_spectrum which we can plot::
 Note that the computation of the power spectrum involves some downsampling.
 If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
 
-    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
+    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000)
 
 To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
 This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -20,7 +20,9 @@ We grab the chunk of data that was used for the calibration last time::
 To be able to calibrate our forces, we first convert back our data to raw voltages.
 We can do this using the previous calibration performed in Bluelake (in this case, the first one in the file)::
 
-    volts = force_slice.data / f.force1x.calibration[0]["Rf (pN/V)"]
+    offset = f.force1x.calibration[0]["Offset (pN)"]
+    response = f.force1x.calibration[0]["Rf (pN/V)"]
+    volts = (force_slice.data - offset) / response
 
 Force calibration models are fit to power spectra. To compute a power spectrum from our data::
 
@@ -38,7 +40,7 @@ If you wish to change the amount of downsampling applied or which range to compu
     lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
 
 To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
-This model can be found in `PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
+This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
 
 .. math::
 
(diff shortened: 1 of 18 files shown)


docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,69 @@
...
+    start = f.force1x.calibration[1]["Start time (ns)"]
+    stop = f.force1x.calibration[1]["Stop time (ns)"]
+    force_slice = f.force1x[start:stop]
+
+To be able to calibrate our forces, we first convert back our data to raw voltages.
+We can do this using the previous calibration performed in Bluelake (in this case, the first one in the file)::
+
+    volts = force_slice.data / f.force1x.calibration[0]["Rf (pN/V)"]
+
+Force calibration models are fit to power spectra. Let's compute a power spectrum from our data::
```
mcsmurf, line 25: Nothing wrong with these, just alternatives to avoid the contractions.
```suggestion
Force calibration models are fit to power spectra. To compute a power spectrum from our data::
```

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,69 @@
...
+
+Note that the computation of the power spectrum involves some downsampling.
+If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
+
+    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
+
+To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+This model needs to be provided with a bead diameter in microns.
+
+Let's use the bead diameter found in the calibration performed in Bluelake.
```
mcsmurf, line 43:
```suggestion
We use the bead diameter found in the calibration performed in Bluelake.
```

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,69 @@
...
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+Force calibration is used to convert the raw voltages that Bluelake records to actual forces.
+In this tutorial, we will be redoing a calibration that we already performed in Bluelake, to provide an example of using the force calibration module::
+
+    f = lk.File("calibration.h5")
+
+In this tutorial, we want to reproduce calibration item 1.
+Let's grab the chunk of data that was used for the calibration last time::
```
mcsmurf, line 14:
```suggestion
We grab the chunk of data that was used for the calibration last time::
```

mcsmurf (approved):
LGTM, left a few optional suggestions. Go for launch 🚀

docs/tutorial/force_calibration.rst
JoepVanlier: Done! :)

docs/tutorial/force_calibration.rst
JoepVanlier: Amended!

docs/tutorial/force_calibration.rst
JoepVanlier: Amended!

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+
+    \kappa = 2 \pi \gamma f_c
+
+Here :math:`\kappa` then represents the final trap stiffness. The parameter :math:`\gamma` corresponds to the friction coefficient of a sphere, which is given by:
+
+.. math::
+
+    \gamma = 3 \pi \eta d
+
+Where :math:`\eta` corresponds to the dynamic viscosity and :math:`d` represents the bead diameter.
```
mcsmurf, line 67:
```suggestion
where :math:`\eta` corresponds to the dynamic viscosity and :math:`d` represents the bead diameter.
```

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
+
+To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+This model can be found in `PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
+
+.. math::
+
+    P(f) = \frac{D}{2 \pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha)
+
+Where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the fitted cutoff. The second term :math:`g` takes into account the slower response of the position detection system and is given by:
```
mcsmurf, line 47:
```suggestion
where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the fitted cutoff. The second term :math:`g` takes into account the slower response of the position detection system and is given by:
```

mcsmurf (approved):
No contractions this time 💯 😄

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+We grab the chunk of data that was used for the calibration last time::
+
+    start = f.force1x.calibration[1]["Start time (ns)"]
+    stop = f.force1x.calibration[1]["Stop time (ns)"]
+    force_slice = f.force1x[start:stop]
+
+To be able to calibrate our forces, we first convert back our data to raw voltages.
+We can do this using the previous calibration performed in Bluelake (in this case, the first one in the file)::
+
+    volts = force_slice.data / f.force1x.calibration[0]["Rf (pN/V)"]
```
dean0x7d, line 23: You should also subtract the offset. And probably also mention that the offset is not calculated by the force calibration procedure. Instead, it's a user-defined zero level.

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+
+Force calibration models are fit to power spectra. To compute a power spectrum from our data::
+
+    power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate)
```
dean0x7d, line 27: This function returns an instance of the `PowerSpectrum` class, but that class is in the `detail` namespace and not found in the API reference. So users don't know what they can do with this result (apart from what's described in this tutorial). I would suggest moving that class out of `detail` and adding it to the API reference.

Notes on the `PowerSpectrum` API itself: 
* The uppercase `.P` as an attribute goes against Python conventions. I'm not sure how consistent we are with this. There is an argument to be made about keeping it close to scientific notation. What I don't like is that this messes with most Python tools. E.g. even simple syntax highlighting ends up marking that as a type instead of a variable. I would sooner go with `.power` and `.frequency` for those attributes. But that's my opinion. It might be good if a few of us had a chat about these conventions and maybe write down a quick style guide to define what we want in situations like these.
* I see that `PowerSpectrum` still has `as_dict()`. You can drop that.
* The docstring for `calculate_power_spectrum()` doesn't mention that it returns a `PowerSpectrum`.

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+
+.. image:: force_calibration_blocked_spectrum.png
+
+Note that the computation of the power spectrum involves some downsampling.
+If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
+
+    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350)
+
+To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+This model can be found in `PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
```
dean0x7d, line 41: Use the Sphinx notation to link this class to its API reference page.

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,95 @@
...
+
+We use the bead diameter found in the calibration performed in Bluelake.
+You can optionally also provide a viscosity (in Pa/s) and temperature (in degrees Celsius)::
+
+    bead_diameter = f.force1x.calibration[1]["Bead diameter (um)"]
+    force_model = lk.PassiveCalibrationModel(bead_diameter, viscosity=0.001002, temperature=20)
+
+To fit this model to the data, you can now invoke::
+
+    calibration = lk.fit_power_spectrum(power_spectrum, force_model)
```
dean0x7d, line 77: * `fit_power_spectrum` does not document its return type. `CalibrationResults` is not listed in the API reference.
* What's the difference between the parameters that go into `CalibrationModel` vs the ones that go into `CalibrationSettings`? Do we expect that the settings will apply to any and all models?
* As far as I can tell, there's no way users can actually set the `settings` argument because `CalibrationSettings` is not a public type.
* You can drop the `CalibrationError` type. It's not public so no one can catch it. Just forwarding the original `ValueError` will do and probably be more informative.

dean0x7d (changes requested):
I found a few undocumented bits of the API.

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
JoepVanlier: > The uppercase .P as an attribute goes against Python conventions. I'm not sure how consistent we are with this. There is an argument to be made about keeping it close to scientific notation. What I don't like is that this messes with most Python tools. E.g. even simple syntax highlighting ends up marking that as a type instead of a variable. I would sooner go with .power and .frequency for those attributes. But that's my opinion. It might be good if a few of us had a chat about these conventions and maybe write down a quick style guide to define what we want in situations like these.
I also feel `.power` and `.frequency` is better.
> I see that PowerSpectrum still has as_dict(). You can drop that.
Dropped.
> The docstring for calculate_power_spectrum() doesn't mention that it returns a PowerSpectrum
Amended!

docs/tutorial/force_calibration.rst
JoepVanlier: > fit_power_spectrum does not document its return type. CalibrationResults is not listed in the API reference.

Good point! I will list it.

> What's the difference between the parameters that go into CalibrationModel vs the ones that go into CalibrationSettings? Do we expect that the settings will apply to any and all models?

`CalibrationSettings` are optimization parameters. The other refers to model parameters (as in physical quantities that relate to one particular model). I'd expect the `CalibrationSettings` to universally apply yes.

> As far as I can tell, there's no way users can actually set the settings argument because CalibrationSettings is not a public type.

That's a good point. I'd expect them to be changed rarely, but you're totally right that they should be public and documented.

> You can drop the CalibrationError type. It's not public so no one can catch it. Just forwarding the original ValueError will do and probably be more informative.

I wasn't sure whether Bluelake was interacting with this specific exception, so I left it. I'd rather replace it with a `RuntimeError`? Seems more appropriate to me.

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
JoepVanlier: Welp! That's a good one! I would've totally forgotten about that.

docs/tutorial/force_calibration.rst
```diff
@@ -0,0 +1,96 @@
...
+
+    start = f.force1x.calibration[1]["Start time (ns)"]
+    stop = f.force1x.calibration[1]["Stop time (ns)"]
+    force_slice = f.force1x[start:stop]
+
+To be able to calibrate our forces, we first convert back our data to raw voltages.
+We can do this using the previous calibration performed in Bluelake (in this case, the first one in the file)::
+
+    offset = f.force1x.calibration[0]["Offset (pN)"]
+    volts = (force_slice.data - offset) / f.force1x.calibration[0]["Rf (pN/V)"]
```
dean0x7d, line 24: Let's do this for conceptual symmetry:
```suggestion
    offset = f.force1x.calibration[0]["Offset (pN)"]
    respose = f.force1x.calibration[0]["Rf (pN/V)"]
    volts = (force_slice.data - offset) / response
```

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -133,16 +131,18 @@ def guess_f_diode_initial_value(ps, guess_fc, guess_D):
...
 
 
-def calculate_power_spectrum(data, sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350,
-                             compatibility_mode=False):
-    """Compute power spectrum.
+def calculate_power_spectrum(
+    data, sample_rate, fit_range=(1e2, 23e3), num_points_per_block=350, compatibility_mode=False
+):
+    """Compute power spectrum and returns it as a
+    `lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum`.
```
dean0x7d, line 145: Use Sphinx notation to make sure that this becomes a link in the docs. I'm pretty sure that you also don't need to make it this long: 
```
:class:`~.PowerSpectrum`
```
That should work. The `.` tell it to look for the name in the current module and `~` tells it to also use this short name for the rendered docs (instead of the usual fully qualified name). See the sphinx docs and experiment with this a bit. It would be good to have all this markup applied consistently in the docs for a good user experience: https://www.sphinx-doc.org/en/master/usage/restructuredtext/roles.html#cross-referencing-syntax

dean0x7d (approved):
LGTM! Just a couple of optional docs tips.

docs/tutorial/force_calibration.rst
dean0x7d: Just propagating the original exception is fine.

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

lumicks/pylake/force_calibration/power_spectrum_calibration.py
JoepVanlier: Thanks. I updated the docs for force calibration in this fashion and will keep a lookout in the future.

PR #121, opened 2021-03-02
kymotracker: fix bug in peakfinding routine

JoepVanlier:
**Why this PR?**
It fixes a bug in the Kymotracker that could occur when a bright structure suddenly ends.

The reason for this bug was that a filter that was being applied to reduce pixelation noise was being applied in two directions rather than just along the scan axis. This meant that the pixel could bleed through to the next column of pixels.

image

Note that while this bug affects peak _detection_, it should not affect peak localization, since this step is followed by a centroid estimation step.

image

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## v0.8.2 | t.b.d.
+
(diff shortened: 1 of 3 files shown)


lpingas (approved):
LGTM! I also really liked the regression test!

PR #122, opened 2021-03-05
Population Dynamics: Gaussian Mixture Models

rpauszek:
**Why this PR?**
As a start to population dynamics analysis workflow, we need some model to label state data. GMM is simple and implemented in scikit-learn, so we can just wrap that

**What's included**
- added `population` sub-package
- model can be intialized from a `Slice` object
- model states are sorted by means
- convenience methods for plotting histogram and state-labeled channel data

Docs found [here](https://lumicks-pylake.readthedocs.io/en/gmm/tutorial/population_dynamics.html)


lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,115 @@
...
+            "lower_bound": self._model.lower_bound_,
+        }
+
+    @property
+    def _map(self):
+        """Indices of sorted means."""
+        return np.argsort(self._model.means_.squeeze())
+
+    @property
+    @as_sorted
```
JoepVanlier, line 50: I like this.

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,115 @@
...
+        output_states = np.argsort(self._map)  # output model state labels in wrapped model order
+        return output_states[labels]  # output model labels
+
+    def bic(self, trace):
+        """Bayesian Information Criterion."""
+        return self._model.bic(trace.data.reshape((-1, 1)))
+
+    def aic(self, trace):
+        """Akaike Information Criterion."""
+        return self._model.aic(trace.data.reshape((-1, 1)))
```
JoepVanlier, line 81: Is it useful to compute the AIC/BIC on a _new_ trace? Shouldn't it always use the data you trained with since it's an in-sample method? If not, wouldn't it make sense to just store this thing after training?

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,115 @@
...
+        plt.ylabel("density")
+        plt.xlabel(trace.labels.get("y", "signal"))
+
+    def plot(self, trace, **kwargs):
+        import matplotlib.pyplot as plt
+
+        kwargs = {"marker": "o", "ls": "none", "ms": 3, **kwargs}
+
+        labels = self.label(trace)
+        trace.plot(c="#c5c5c5")
```
JoepVanlier, line 112: It would be good if the user can still modify this color.

docs/tutorial/population_dynamics.rst
```diff
@@ -0,0 +1,75 @@
...
+The Gaussian Mixture Model (GMM) is a simple probabilistic model that assumes all data points belong
+to one of a fixed number of states, each of which is described by a (weighted) normal distribution.
+
+We can initialize a model from a force channel as follows::
+
+    gmm = force.train_gmm(2)
+    print(gmm.weights)
+    print(gmm.means)
+
+Here the force data is used to train a 2 component GMM. The component weights and means can be accessed with their respective attributes.
```
JoepVanlier, line 49: Maybe briefly mention what "weights" mean in this context (fraction spent in each state).

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,115 @@
...
+        Parameters
+        ----------
+        K : int
+            number of Gaussian components in the model
+        """
+        self.K = K
+        self._model = GaussianMixture(
+            n_components=K, init_params=init_method, n_init=n_init, tol=1e-3, max_iter=100
+        )
+        self._model.fit(data)
```
JoepVanlier, line 29: Would it make sense to keep a reference to the data it was fitted on? In that case, the trace argument of plot could be optional.

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,115 @@
...
+        data = slc.data.reshape((-1, 1))
+        return cls(data, K, init_method="kmeans", n_init=1, tol=1e-3, max_iter=100)
+
+    @property
+    def exit_flag(self):
+        return {
+            "converged": self._model.converged_,
+            "n_iter": self._model.n_iter_,
+            "lower_bound": self._model.lower_bound_,
+        }
```
JoepVanlier, line 42: I find finding out about convergence after the fact a bit weird. When the model fails to converge, shouldn't fit throw?

What value do you get for `converged_` if you did multiple starts?

This property also seems undocumented and untested.

lumicks/pylake/population/tests/test_mixture.py
```diff
@@ -0,0 +1,38 @@
...
+import matplotlib.pyplot as plt
+
+
+def test_gmm(trace_lownoise):
+    trace, statepath, params = trace_lownoise
+
+    m = GaussianMixtureModel.from_slice(trace, params["K"])
+    assert np.all(np.equal(m.label(trace), statepath))
+    assert np.allclose(m.means, params["mu"], atol=0.05)
+    assert np.allclose(m.std, params["sigma"], atol=0.02)
```
JoepVanlier, line 14: `weights`, `pdf` and `aic` should also be tested.

JoepVanlier (changes requested):
I love it.

I quite like the `as_sorted` decorator. I also like the docs.

I have left some comments.

aafkevandenberg (approved):

lumicks/pylake/population/mixture.py
rpauszek: This functionality is offered by sklearn, but I agree it may not be useful. Default is now to return values only for training data

lumicks/pylake/population/mixture.py
rpauszek: I think this is pretty standard behavior for fitting (the values are actually pulled directly from the sklearn GMM instance after it is fit). I think throwing an exception is a bit too strict. This way the results are available to the user and they can decide what to do with it (throw if `converged == False` for example)

docs/tutorial/population_dynamics.rst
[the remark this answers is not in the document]
rpauszek: good catch, should be consistent with 'state' now

docs/tutorial/population_dynamics.rst
```diff
@@ -0,0 +1,75 @@
...
+===================
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+
+The following tools enable analysis of experiments which can be described as a system of discrete states. Consider as an example
+a DNA hairpin which can exist in a folded and unfolded state. At equilibrium, the relative populations of the two states are governed by the interconversion kinetics described by the rate constants \\(k_{fold}\\) and \\(k_{unfold}\\)
+
```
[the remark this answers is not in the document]
rpauszek: I think for here, I'd like to keep the physics background discussion to a minimum and focus on the code features. This may be more applicable when a specific kinetics section is added or a specific example case for instance.

docs/tutorial/population_dynamics.rst
```diff
@@ -0,0 +1,75 @@
...
+.. image:: hairpin_kinetics.png
+
+First let's take a look at some example force channel data. We'll downsample the data in order to speed up the caclulations
+and smooth some of the noise::
```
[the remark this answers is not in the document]
rpauszek: I added a bit of clarification to the discussion at the bottom of the page

lumicks/pylake/population/tests/test_mixture.py
```diff
@@ -0,0 +1,67 @@
...
+    ef = m.exit_flag
+    assert ef["converged"] == True
+    assert ef["n_iter"] == 2
+    assert np.allclose(ef["lower_bound"], 0.215335)
+
+
+def test_pdf(trace_simple):
+    trace, statepath, params = trace_simple
+    m = trace.train_gmm(params["K"])
+    print(m.pdf(np.array([10, 11])))
```
JoepVanlier, line 50: leftover print statement?

JoepVanlier (changes requested):
looks good to me

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,178 @@
...
+
+        lims = (np.min(trace.data), np.max(trace.data))
+        bins = np.linspace(*lims, num=n_bins)
+        x = np.linspace(*lims, num=(n_bins * 5))
+
+        g_components = self.pdf(x)
+
+        plt.hist(trace.data, bins=bins, density=True, **hist_kwargs)
+        # reset color cycle
+        plt.gca().set_prop_cycle(None)
```
dean0x7d, line 149: Why is this needed?

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,178 @@
...
+            method used to initialize parameters
+        n_init : int
+            number of initializations to perform
+        tol : float
+            convergence tolerance
+        max_iter : int
+            maximum number of iterations to perform
+        """
+        self.K = K
+        self._training_data = trace
```
dean0x7d, line 36: Do think about if you really need to store a full `Slice` as the training data. It's a tradeoff:
* The downside is that users will need to pass the slice again for the plotting function.
* The upside is that you don't need to store and think about all the extra stuff in `Slice` that you don't need for the training. It removes the coupling between the GMM and `Slice`. And it makes your testing easier because you no longer need a full `Slice` for unit testing.

lumicks/pylake/population/tests/conftest.py
```diff
@@ -0,0 +1,76 @@
+import pytest
+import numpy as np
+import contextlib
+from lumicks.pylake.channel import Slice, Continuous
+from lumicks.pylake.population.detail.random import temp_seed
+
+
+def generate_trace(K, pi, A, mu, sigma, T, seed):
```
dean0x7d, line 8: Same comment about naming conventions as above. You can also see that GitHub highlights these variables as types which is not great for code consistency and readability.

lumicks/pylake/population/tests/conftest.py
```diff
@@ -0,0 +1,76 @@
...
+
+    with temp_seed(seed):
+        # draw state path
+        sp = np.zeros(T, dtype=int)
+        sp[0] = np.random.choice(K, 1, replace=False, p=pi).squeeze()
+        for t in range(1, T):
+            sp[t] = np.random.choice(K, 1, replace=False, p=A[sp[t - 1]]).squeeze()
+
+        # draw observation vector
+        y = np.random.normal(loc=mu[sp], scale=sigma[sp])
```
dean0x7d, line 40: I'm honestly not sure how portable this seeded data generation is. E.g. if `numpy` change the way they generate random numbers in the future or if they use different generators on different platforms. Would the expected test values change in that case?

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,178 @@
...
+        self._model = GaussianMixture(
+            n_components=K, init_params=init_method, n_init=n_init, tol=1e-3, max_iter=100
```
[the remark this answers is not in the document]
JoepVanlier: damn... good catch!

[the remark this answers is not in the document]
rpauszek: good catch, definitely a copy/paste error

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,178 @@
...
+    """Decorator to return results sorted according to mapping array."""
+
+    def wrapper(self, *args, **kwargs):
+        result = fcn(self, *args, **kwargs)
+        return result[self._map]
+
+    return wrapper
+
+
+class GaussianMixtureModel:
```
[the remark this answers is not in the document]
rpauszek: done

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,178 @@
...
+    def __init__(self, trace, K, init_method, n_init, tol, max_iter):
+        """A wrapper around scikit-learn's GMM
```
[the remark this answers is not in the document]
rpauszek: Added a bit of explanation here. I think it sounds a bit awkward but let's see what you guys think. In my experience it's easier to compare different models (either different number of states or trained on different data) if the parameters are sorted somehow rather than being in random order (by means seems to make the most sense here). Also, this wrapper is specifically implemented for 1D data (thus the `squeeze()`, this way you don't have to worry about output shape)

lumicks/pylake/population/mixture.py
rpauszek: I noticed this when testing and I was plotting both the histogram and labeled data in the same figure. The first call to `plt.hist()` causes the color cycle to increment, however this doesn't happen inside the `GMM.plot()` method (I think because `trace.plot()` is used instead of plotting the data manually). 

At any rate, this ensures that the same state has the same color if using the default color cycle in plots generated either from `GMM.hist()` or `GMM.plot()`. I don't think it's absolutely necessary, but I thought it was a nice feature and only requires 1 line of code.

lumicks/pylake/population/mixture.py
rpauszek: I kept going back and forth on this (see Joep's earlier comments). Looking at it again, I think I agree that it's not worth it to store the `Slice` or training data inside the object.

One thing that I still find awkward here though is now the constructor accepts a `np.array` as training data but other methods (`.plot()`, `.hist()`, `.label()`, etc...) rely on the input argument being `Slice`-like (ie, having a `.data` attribute and sometimes a `.plot()` method).

In addition to force channels, I could see this class being useful also for `PointScan.color_photon_count` which returns `Continuous` or output from `KymoLine.sample_from_image()` which is currently returns a `list`. For handling these situations, I don't know what the best option would be:

- enforce the `Slice`-like API, but `Continuous.plot()` does not exist and `KymoLine.sample_from_image()` would need to return a `Slice` instead of `list`
- have all of the `GMM` methods just accept `np.array`, although I think `GMM.plot(trace.data)` is still a bit awkward
- have some sort of internal check in these methods to handle the different types that could be passed as input arguments

What do you guys think?

lumicks/pylake/population/tests/conftest.py
rpauszek: True, but in this case I don't think it will matter really. I thought it was better to simulate more realistic data rather than hard-coding in a trace. This will especially be better when we get to dealing with HMMs. I think that as long as the simulated trace is long enough, the model fitting should converge to the input parameters, and the simulated statepath is also returned which is what is actually used for checking state assignments in the tests. I also thought setting the RNG seed was better than just leaving it completely uncontrolled.

I'm definitely open to suggestions if you think there is a better way to do this though.

lumicks/pylake/population/mixture.py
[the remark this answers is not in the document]
dean0x7d: I think that you could add some of the info from this comment to the docstring. Sorting is an important aspect of this class and quite different from the original so it's good to point that out to users. Similarly for `squeeze()`.

lumicks/pylake/population/tests/conftest.py
dean0x7d: I think that the extremes are actually better here. If the data can be random, let it be truly random (no seeding). If the test can't tolerate true random data, generate the input once and store it. The middle ground is prone to issues that may show up in obscure circumstances.

lumicks/pylake/population/mixture.py
```diff
@@ -0,0 +1,187 @@
...
+        self._model = GaussianMixture(
+            n_components=n_states,
+            init_params=init_method,
+            n_init=n_init,
+            tol=tol,
+            max_iter=max_iter,
+        )
+        self._model.fit(np.reshape(data, (-1, 1)))
+        self._bic = self._model.bic(np.reshape(data, (-1, 1)))
+        self._aic = self._model.aic(np.reshape(data, (-1, 1)))
```
dean0x7d, line 51: * To avoid needless repetition, you should pull this out into a variable. That way it's easy to see that all 3 of these are getting the same data. Otherwise, readers need to spend more time to confirm that all 3 are indeed identical.
* Are there any performance issues evaluating `bic` and `aic` eagerly here?

lumicks/pylake/population/mixture.py
dean0x7d: The photon counts are returned as `Slice`s as far as I know. Nothing should return `Continuous` since that's an implementation detail and not a public API. If something does return that, it should be fixed.

I think that it's fine to require `Slice` as the input. The plotting functions don't just use the data but also timestamps, seconds, and labels as far as I can see. I do think that that the core processing should be done with raw data, but I don't worry so much about the plotting functions. They are more of a convenience feature and it's hard to make something both convenient and also very generic. If someone needs to plot the raw data, it's easy to see the source code and replicate it.

That said, I also don't like the discrepancy between the core class and the plotting functions. If this was a different language, I would say you should make the plotting functions into free functions that take the model as one of the arguments. But that's difficult to do in Python -- or just as awkward as the current situation. So I would say that this is fine for now and we'll discover more as we get info about how this API will actually be used. Too much guessing for now.

rpauszek: you're right, photon counts is a `Slice`, it's just a very convoluted chain of calls and inheritance so I got confused

lumicks/pylake/population/tests/conftest.py
rpauszek: static file is the way to go I guess, data generation script is included in the test directory

docs/tutorial/population_dynamics.rst
```diff
@@ -0,0 +1,84 @@
...
+
+Gaussian Mixture Models
+-----------------------
+
+The Gaussian Mixture Model (GMM) is a simple probabilistic model that assumes all data points belong
+to one of a fixed number of states, each of which is described by a (weighted) normal distribution.
+
+We can initialize a model from a force channel as follows::
+
+    gmm = force.train_gmm(2)
```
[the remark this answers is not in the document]
rpauszek: oops, good catch

lumicks/pylake/population/mixture.py
rpauszek: Perhaps for extremely large datasets. The alternative though is to keep a reference to the training data, which in the current form I'm not sure is particularly useful either.

We could go back to how I had originally had the API set up where `aic` and `bic` are methods and you would supply the data as an argument. This is how it is implemented in `sklearn` and is more flexible, but as Joep pointed out it doesn't necessarily make sense to calculate the information criteria for data that wasn't used to train the model.

JoepVanlier: I'm skeptical whether `aic` or `bic` for out of sample testing makes sense. This is certainly not how it was used in the papers that introduced these methods. I'd like to see a reference where this is motivated because I see potential for user error if they start supplying it with different data than what they trained on.

As for perf impact, it should only be a matter of evaluating the likelihood once for each (if it is not actually cached by sklearn internally), so I wouldn't expect a big performance overhead.

dean0x7d: OK, if it ends up being a performance issue, it easy enough to make it a property.

lumicks/pylake/population/tests/generate_trace_data.py
```diff
@@ -0,0 +1,67 @@
+import numpy as np
```
dean0x7d, line 1: Last optional remark: consider moving this file and the data into a subfolder together to clearly separate them from the regular tests files.

Before:
```
tests/generate_trace_data.py
tests/trace_data.npy
```

After:
```
tests/data/generate_trace.py
tests/data/trace.npy
```

dean0x7d (approved):
Looks great! Thanks for going along with my many comments. And congrats on cracking open a brand new submodule in pylake!

JoepVanlier (approved):
:shipit:

PR #124, opened 2021-03-11
kymo: fix slicing bug

JoepVanlier:
**Why this PR?**
While fixing something else, I noticed that this behaviour is unexpected.

When slicing a kymo, the intended behaviour was to use the starting timestamps of each line as a determinant of whether that line should be included. Instead, it was using the one beyond the end.

**Why not use `self.timestamps`**?
`self.timestamps` provides means of the timestamps of samples corresponding to each pixel. This can lead to strange issues with a partial pixel being cut off when slicing that way. You really want the first sample of each line.

**How could this one slip through the tests?**
The overarching test of the slicing is not precise enough. This is part of bigger problem which we hope to address soon, but which requires a bit of refactoring.

The test on the actual function grabbing the timestamps was bad. The reference data always had one "ignored" sample which meant that for that particular test it passed. It was not representative of the type of data you'd encounter in a real case. I've made the test a bit more robust by including different `scan settings`. The old code would not pass this test.

### DIFF
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -114,10 +114,8 @@ def line_timestamps_image(time_stamps, infowave, pixels_per_line):
     pixels_per_line : int
         The number of pixels on the fast axis of the scan.
     """
-    pixel_start_idx = np.flatnonzero(infowave == InfowaveCode.pixel_boundary)
-    pixel_start_idx = np.concatenate(([0], pixel_start_idx + 1))
-
-    return time_stamps[pixel_start_idx[0:-1:pixels_per_line]]
+    min_stamps = reconstruct_image(time_stamps, infowave, shape=(pixels_per_line,), reduce=np.min)
+    return min_stamps.astype(np.int64)[:, 0]
 
 
 def seek_timestamp_next_line(infowave):
--- a/lumicks/pylake/tests/test_image.py
+++ b/lumicks/pylake/tests/test_image.py
@@ -32,17 +32,19 @@ def test_metadata_from_json():
     assert image_metadata.metadata['PixelTimeUnit'] == 's'
 
 
-def test_timestamps_image():
-    infowave = np.array([0, 1, 0, 1, 1, 0, 2, 1, 0, 1,  0,  0,  1,  2,  1,  1,  1, 2])
-    time = np.array([1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18])
+@pytest.mark.parametrize("num_lines, pixels_per_line, pad_size", [(5, 3, 3), (5, 4, 2), (4, 7, 5)])
+def test_timestamps_image(num_lines, pixels_per_line, pad_size):
+    line_info_wave = np.tile(np.array([1, 1, 2], dtype=np.int32), (pixels_per_line, ))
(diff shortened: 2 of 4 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -114,10 +114,8 @@ def line_timestamps_image(time_stamps, infowave, pixels_per_line):
     pixels_per_line : int
         The number of pixels on the fast axis of the scan.
     """
-    pixel_start_idx = np.flatnonzero(infowave == InfowaveCode.pixel_boundary)
-    pixel_start_idx = np.concatenate(([0], pixel_start_idx + 1))
-
-    return time_stamps[pixel_start_idx[0:-1:pixels_per_line]]
+    min_stamps = reconstruct_image(time_stamps, infowave, shape=(pixels_per_line,), reduce=np.min)
+    return min_stamps.astype(np.int64)[:, 0]
```
rpauszek, line 118: I'm not sure how often this would come up, but `reconstruct_image()` is a relatively expensive calculation; so if a user creates many slices from a single parent kymograph this calculation would be repeated many times.

With this change, the code is very similar to the `ConfocalImage._timestamps()` method which is cached. Would it make sense to instead make this function a cached method of `Kymo`? As far as I can tell, `Kymo.__getitem__()` is the only place where `line_timestamps_image()` is used.

lumicks/pylake/tests/test_image.py
```diff
@@ -32,17 +32,19 @@ def test_metadata_from_json():
...
-
-    line_stamps = line_timestamps_image(time, infowave, 2)
-    assert line_stamps.shape == (2,)
-    assert np.all(line_stamps == [1, 15])
+@pytest.mark.parametrize("num_lines, pixels_per_line, pad_size", [(5, 3, 3), (5, 4, 2), (4, 7, 5)])
+def test_timestamps_image(num_lines, pixels_per_line, pad_size):
+    line_info_wave = np.tile(np.array([1, 1, 2], dtype=np.int32), (pixels_per_line, ))
+    line_selector = np.zeros(line_info_wave.shape)
+    line_selector[0] = True
+    pad = np.array([0] * pad_size, dtype=np.int32)
```
rpauszek, line 40: couldn't this be `np.zeros(pad_size, dtype=np.int32)`?

rpauszek (commented):
The logic looks good to me; my only concern is with efficiency if a user were to slice the same `Kymo` many times (see my comment below)

lumicks/pylake/tests/test_image.py
JoepVanlier: Yes, good catch. Fixing.

lumicks/pylake/detail/image.py
JoepVanlier: Yeah, I considered that.

When the kymo is sliced a new instance is created, so it wouldn't benefit from the caching when doing repeated slicing. I guess you would benefit if you'd take multiple slices from the same kymo though.

I can make it a cached private member function.

rpauszek: Yes, my concern is if you make like 5 slices from the same Kymo you repeat that calculation 5 times

JoepVanlier: Yeah, it's a good point, should be fixed now!

rpauszek (approved):
👍

PR #126, opened 2021-03-16
confocal: differentiate ui scan_width from actual scan_width

JoepVanlier:
**Why this PR?**
In pylake, the scan width the user entered was available through `scan_width_um`. While a user may be lured into believing that this is the actual scan width of the scan they're dealing with, it is actually not because Bluelake will round the scan to a fixed number of pixels multiplied by the pixel width. It would be better to reserve this property for the actually obtained scan width and have a separate property for the scan width as entered in the UI.

### DIFF
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -265,12 +265,33 @@ class ConfocalImage(BaseScan):
 
     @property
     def pixelsize_um(self):
-        """Returns a `List` of axes dimensions in um. The length of the list corresponds to the number of scan axes."""
+        """Returns a `List` of axes dimensions in um. The length of the list corresponds to the
+        number of scan axes."""
         return [axes["pixel size (nm)"] / 1000 for axes in self._ordered_axes()]
 
     @property
+    def size_um(self):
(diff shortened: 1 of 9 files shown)


lumicks/pylake/detail/confocal.py
```diff
@@ -265,12 +265,23 @@ def timestamps(self) -> np.ndarray:
...
+        return [
+            axes["pixel size (nm)"] * axes["num of pixels"] / 1000 for axes in self._ordered_axes()
+        ]
+
+    @property
+    def scan_width_um_ui(self):
+        """Returns a `List` of scan widths as configured in the Bluelake UI. The length of the list
+        corresponds to the number of scan axes. Note that these widths can deviate from the actual
+        scan widths performed in practice"""
         return [axes["scan width (um)"] for axes in self._ordered_axes()]
```
dean0x7d, line 285: To avoid the breaking change, you could deprecate the old one without changing the name and call the new one `.size_um` since its a multidimensional size. The `scan_` part was redundant (`Scan.scan_size_um`) or confusing (`Kymo.scan_size_um`).

dean0x7d (approved):

PR #129, opened 2021-03-22
Handle new TIF metadata

rpauszek:
**Why this PR?**
The color channel alignment in `CorrelatedStack` relies on the alignment matrices from Bluelake stored in the `ImageDescription` tag of a tif file. These field names changed in Bluelake 1.7 in preparation for 2-color WIT systems. `CorrelatedStack` needs to be able to handle both metadata formats.

**What's included?**
- No changes to the public API.
- Moved backend classes that handle the TIF data to a separate `detail.widefield` module. These were previously included in `correlated_stack.py` which was getting quite long. It seemed sensible to separate the file handling and the actual `CorrelatedStack` functionality.
- Added an `ImageDescription` class to handle the metadata. Previously, parsing the metadata was mixed in with the channel alignment method, which would get even messier with this bugfix. This also provides a clear structure for handling future format changes and/or deprecating fields.


lumicks/pylake/detail/widefield.py
```diff
@@ -18,50 +18,22 @@ class TiffFrame:
 
     def __init__(self, page, align):
         self._src = page
+        self._description = ImageDescription(page)
```
JoepVanlier, line 22: `Description` or `Metadata`? Based on the property I sense doubt :)

lumicks/pylake/detail/widefield.py
```diff
@@ -47,10 +47,6 @@ def _align_image(self):
             )
         return img
 
-    @property
-    def _has_metadata(self):
-        return bool(self._description)
-
```
JoepVanlier: Ignore previous comment. Seems like you already did.

lumicks/pylake/detail/widefield.py
```diff
@@ -75,6 +47,10 @@ def correct_alignment_offset(alignment, x_offset, y_offset):
             )
         return img
 
+    @property
+    def _has_metadata(self):
+        return bool(self._description)
```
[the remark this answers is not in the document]
JoepVanlier: Ignore this comment. Seems like you already did.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -461,7 +470,10 @@ def test_image_reconstruction_rgb_missing_metadata(rgb_alignment_image_data):
     assert record[0].message.args[0] == "File does not contain metadata. Only raw data is available"
```
JoepVanlier, line 470: One minor thing:

Could you perhaps catch this like this?
```python
with pytest.warns(UserWarning, match="File does not contain metadata. Only raw data is available"):
```

JoepVanlier: Also, don't forget a changelog entry! :)

PR #130, opened 2021-03-23
API for MSD-based methods [MSD 3 of N]

JoepVanlier:
**Why this PR?**
With the backend existing, it would be nice to have a user facing API so that users can actually use it.

image
_Figure 1: And this is why you only use a few points! Horrible!_

You can find the docs here:
https://lumicks-pylake.readthedocs.io/en/msd_api/tutorial/kymotracking.html#studying-diffusion-processes

### DIFF
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -248,3 +248,87 @@ The candidate with the lowest score is then selected.
 Since this algorithm is specifically looking for curvilinear structures, it can have issues with structures that are
 more blob-like (such as short-lived fluorescent events) or diffusive traces, where the particle moves randomly rather
 than in a uniform direction.
+
+
+Studying diffusion processes
+----------------------------
+
+To study diffusive processes we can make use of the Mean Squared Displacement (MSD).
+There are multiple ways to estimate this quantity.
+We use the following estimator:
+
+.. math::
+
+    \hat{\rho}[n] = \frac{1}{N - n} \sum_{i=1}^{N-n}\left(x_{i+n} - x_{i}\right)^2
+
+where :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` is the number of time points in the tracked line, and :math:`x_i` is the trace position at time frame :math:`i`.
+
+What we can see in this definition is that it uses the same data points several times, thereby resulting in a well averaged estimate.
+However, the downside of this estimator is that the calculated values are highly correlated :cite:`qian1991single,michalet2010mean,michalet2012optimal` which needs to be accounted for in subsequent analyses.
+
+In the following, we'll use three simulated :class:`~lumicks.pylake.kymotracker.kymoline.KymoLine` instances of length 62 based on a diffusion constant of `10.0`.
+
+With Pylake, we can calculate the MSD from a :class:`~lumicks.pylake.kymotracker.kymoline.KymoLine` with a single command::
+
+    kymolines[0].msd()
+
+This returns a tuple of lags and MSD estimates. If we only wish MSDs up to a certain lag, we can provide a `max_lag` argument::
+
+    >>> kymolines[0].msd(mag_lag = 5)
+    (array([0.16, 0.32, 0.48, 0.64, 0.8 ]), array([12.48593965, 16.34844311, 17.21359513, 27.25210869, 32.34473104]))
+
+MSDs are typically used to calculate diffusion constants.
+With pure diffusive motion (a complete absence of drift) in an isotropic medium, 1-dimensional MSDs can be fitted by the following relation:
+
+.. math::
+
+    \rho[n] = 2 D t + offset
+
+where :math:`D` is the diffusion constant in :math:`um^2/s`, :math:`t` is time, and the offset is determined by the localization accuracy:
+
+.. math::
+
+    offset = 2 \sigma^2 - 4 R D \Delta t
+
+where :math:`\sigma` is the static localization accuracy, :math:`R` is a motion blur constant and :math:`\Delta t` represents the time step.
+
+While it may be tempting to use a large number of MSDs in the diffusion estimation procedure, this actually produces poor estimates of the diffusion constant :cite:`qian1991single,michalet2010mean,michalet2012optimal`.
+There exists an optimal number of lags to fit such that the estimation error is minimal.
+This optimal number of lags depends on the ratio between the diffusion constant and the dynamic localization accuracy:
+
+.. math::
+
+    \epsilon_{localization} = \frac{offset}{slope} = \frac{2 \sigma^2 - 4 R D \Delta t}{2 D \Delta t} = \frac{\sigma^2}{D \Delta t} - 2 R
+
+When localization is infinitely accurate, the optimal number of points is two :cite:`michalet2010mean`.
+At the optimal number of lags, it doesn't matter whether we use a weighted or unweighted least squares algorithm to fit the curve :cite:`michalet2010mean`, and therefore we opt for the latter, analogously to :cite:`michalet2012optimal`.
+With Pylake, you can obtain an estimated diffusion constant by invoking::
+
+    >>> kymolines[0].estimate_diffusion_ols()
+    7.386961688211464
+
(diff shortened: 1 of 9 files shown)


docs/tutorial/kymotracking.rst
```diff
@@ -252,3 +252,85 @@ The candidate with the lowest score is then selected.
...
+where :math:`D` represents the diffusion constant, :math:`t` represents time and the offset is determined by the localization accuracy and is given by:
+
+.. math::
+
+    offset = 2 \sigma^2 - 4 R D \Delta t
+
+Where :math:`\sigma` is the static localization accuracy, :math:`R` is a motion blur constant and :math:`\Delta t` represents the time step.
+
+While it may be tempting to use a large number of MSDs in the diffusion estimation procedure, this actually produces poor estimates of the diffusion constant :cite:`qian1991single,michalet2010mean,michalet2012optimal`.
+There exists an optimal number of lags to incorporate such that the estimation error is minimal.
```
rpauszek, line 298: maybe change "incorporate" to "fit"

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -1,4 +1,6 @@
 import numpy as np
+import matplotlib.pyplot as plt
```
rpauszek, line 2: AFAIK everywhere else we keep the `pyplot` import local to the plotting functions, might want to move this to keep consistent

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -222,6 +224,106 @@ def extrapolate(self, forward, n_estimate, extrapolation_length):
...
+        -------
+        lag_time : array_like
+            array of lag times
+        msd : array_like
+            array of mean square distance estimates for the corresponding lag times
+        """
+        frame_lag, msd = calculate_msd(
+            np.array(self.time_idx, dtype=int),
+            np.array(self.coordinate),
+            max_lag if max_lag else len(self.time_idx),
```
rpauszek, line 255: elsewhere we've used something like `max_lag or len(self.time_idx)`, is there any reason to use that style over the tertiary form or vice versa?

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -222,6 +224,106 @@ def extrapolate(self, forward, n_estimate, extrapolation_length):
...
+        Parameters
+        ----------
+        max_lag : int (optional)
+            Maximum lag to include. When omitted, an optimal number of lags is chosen [1].
+        **kwargs
+            Forwarded to :func:`matplotlib.pyplot.plot`.
+        """
+
+        frame_idx, coordinates = np.array(self.time_idx, dtype=int), np.array(self.coordinate)
+        max_lag = max_lag if max_lag else determine_optimal_points(frame_idx, coordinates)[0]
```
rpauszek, line 282: same as previous comment

rpauszek (commented):
I really like the user facing API, nice and clean considering how complex the backend is. Here I've left mostly some comments on the documentation. I have a pretty specific style of scientific writing, so please feel free to discuss or disagree with my suggested changes.

One thing also to think about is how we might want to handle error estimation for all of this. I'm not saying it needs to be implemented in this PR, but I'm sure users will want to have error estimates on the fitted values and possibly even the MSD estimates. I think this could likely easily be handled by separate methods but I might be missing something, so just wanted to point it out.

docs/tutorial/kymotracking.rst
```diff
@@ -252,3 +252,85 @@ The candidate with the lowest score is then selected.
...
+We use the following estimator:
+
+.. math::
+
+    \hat{\rho}[n] = \frac{1}{N - n} \sum_{i=1}^{N-n}\left(x_{i+n} - x_{i}\right)^2
+
+where :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` reflects the number of samples, :math:`n` represents the lag we are computing and the trace positions at time frame i is given by :math:`x_i`.
+
+The reason we use this definition is because it results in the largest number of displacements being used thereby resulting in a well averaged estimate. The downside of this estimator however, is that it suffers from providing very correlated estimates :cite:`qian1991single,michalet2010mean,michalet2012optimal` which needs to be taken into account when doing subsequent analyses.
+
```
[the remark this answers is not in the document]
JoepVanlier: How about:

> What we can see in this definition is that it uses the same data points several times, thereby resulting in a well averaged estimate.

> However, the downside of this estimator is that the calculated values are highly correlated :cite:`qian1991single,michalet2010mean,michalet2012optimal` which needs to be accounted for in subsequent analyses.

I'm hesitant about calling it an error because there is no error in the MSD estimate itself. The variance just goes up like crazy.

docs/tutorial/kymotracking.rst
```diff
@@ -252,3 +252,85 @@ The candidate with the lowest score is then selected.
...
+With Pylake, we can calculate the MSD from a :class:`~lumicks.pylake.kymotracker.kymoline.KymoLine` with a single command::
+
+    kymolines[0].msd()
+
+This returns a tuple of lags and MSD estimates. If we only wish MSDs up to a certain lag, we can provide a `max_lag` argument::
+
+    >>> kymolines[0].msd(mag_lag = 5)
+    (array([0.16, 0.32, 0.48, 0.64, 0.8 ]), array([12.48593965, 16.34844311, 17.21359513, 27.25210869, 32.34473104]))
+
+MSDs are typically used to calculate diffusion constants. With pure diffusive motion (a complete absence of drift), 1-dimensional MSDs can be fitted by the following relation:
```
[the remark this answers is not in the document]
JoepVanlier: Yeah good point!

docs/tutorial/kymotracking.rst
```diff
@@ -252,3 +252,85 @@ The candidate with the lowest score is then selected.
...
+
+.. math::
+
+    \rho[n] = 2 D t + offset
+
+where :math:`D` represents the diffusion constant, :math:`t` represents time and the offset is determined by the localization accuracy and is given by:
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/kymotracking.rst
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Yeah, let's just use tertiary. I think the `or` is a bad habit I picked up in Lua.

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: same as previous comment

😁

mwilkowski (approved):
Looks good to me 👍

docs/tutorial/kymotracking.rst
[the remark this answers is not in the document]
rpauszek: sounds good

docs/tutorial/kymotracking.rst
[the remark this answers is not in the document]
rpauszek: oy, yes I have no idea why I typed 'cm'

I think it would be good to have the units in the docs because otherwise the user has to go through the code to see what's actually being done in order to figure out the units themselves, which kind of defeats the point of offering the feature in the first place a bit. 

It's definitely something we'll have to keep an eye on in the future, but since the units of the input values (um and sec) are assumed, there should be a definite unit on the return value

rpauszek (approved):
looks good. there's just the one small comment about importing pyplot within the function rather than the full module as we do in other places

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: > looks good. there's just the one small comment about importing pyplot within the function rather than the full module as we do in other places

was a rebase mishap (the import ended up in both places)

PR #131, opened 2021-03-23
force_calibration: drop compatibility mode

JoepVanlier:
**Why this PR?**
Since the differences between the old and new force calibration were so small, it was decided to drop the compatibility mode with the old version (since it only adds more code to maintain).

image

If someone _really really_ needs it, it'll still be in the git history.


agom (approved):
⛩️

PR #132, opened 2021-03-31
Refactor

JoepVanlier:
Rough draft PR to investigate whether we can abstract some implementational details of `Scan` and `Kymo`. The idea is to make them more agnostic to where the data is coming from and contain everything related to reconstruction in its own class. This will hopefully allow us to make WIT kymos and treat them through the same API.

Note it's very rough. It's more an exploration to see what would happen if we put the abstraction at this level.

Aims:
- Make it clear what data we expect from an image, see `ScannedImage`. Essentially new "images" should implement this interface (we could formalize this by a base class).
- Being able to provide another class providing image data would make testing new functionalities easier.
- Being able to generate Kymos from other data sources.

Questions:
- Is the abstraction desirable to be at this level? The advantage is that `Scan` and `Kymo` are treated more uniformly and that `Scan` could also benefit from easier testing.
- Is `BaseScan` still worth it given that it actually doesn't deduplicate that much, but makes the hierarchy more complex.
- Do we really want Kymo deriving from a method that collects `Scan` API calls? This introduces coupling that may be hard to get rid of later (An alternative could be to have Kymo construct a `Scan`, and "hold" a `Scan` internally)?

### DIFF
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -287,6 +287,9 @@ class ScannedImage(BaseScan):
 
     @property
     def lines_per_frame(self):
+        if len(self._shape) == 1:
+            raise NotImplementedError("This method is not defined for 1D scans")
+
         return self._json["scan volume"]["scan axes"][1]["num of pixels"]
 
     @property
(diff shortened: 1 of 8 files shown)


JoepVanlier:
Closing this for now. Gonna change the design a little bit and then reopen.

PR #133, opened 2021-04-01
Kymo.downsampled_by(#) [MSD 2a of N]

JoepVanlier:
## Why this PR?
One of the issues with MSD 2 of N was that we spun out a separate class because we needed to handle downsampling. This is undesirable because it makes things more complex for the user (yet another object to think about).

Instead, this PR attempts to add downsampling to `Kymo` itself. To not make the same mess as before, I have also developed a small mocking setup, which allows us to mock Kymos.

One advantage of this setup is that the downsampling can be done lazily.

One alternative to this is to do something similar to https://github.com/lumicks/pylake/pull/132 and abstract the image in the `Kymo`. Then this `DownsampledKymo` could be a regular `Kymo` but with a different image as source. You'd lose the connection to the original though.

## Notes
To enable this feature, and still be able to plot, I had to "break" one of the existing tests. Essentially, what that test was testing were plot limits for the Kymograph. However, the reason these plot limits were as they were was because of the unequal line times in the test data. I have compared that for typical BL kymographs, the new plot setup actually produces the same result.

Updating the test data would've required fixes in more than 50 tests, therefore I have left that exercise for a future PR.

### DIFF
--- a/docs/tutorial/kymographs.rst
+++ b/docs/tutorial/kymographs.rst
@@ -49,6 +49,20 @@ There are also several properties available for convenient access to the kymogra
 * `kymo.fast_axis` provides the axis that was scanned (x or y)
 * `kymo.line_time_seconds` provides the time between successive lines
 
+Downsampling kymograph
+----------------------
+
+We can downsample a kymograph by invoking::
+
+    kymo_ds = kymo.downsampled_by(time_factor=2)
+
+Note however, that not all functionalities are present anymore when downsampling a kymograph. For
+example, we can no longer access the per pixel timestamps::
+
+    >>> kymo_ds.timestamps
+    AttributeError: Per pixel timestamps are no longer available after downsampling a kymograph since they are not well defined (the downsampling occurs over a non contiguous time window).
+    Line timestamps are still available however. See: `Kymo.line_time_seconds`.
+
 Plotting and exporting
 ----------------------
 
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -8,6 +8,24 @@ from .mixin import ExcitationLaserPower
 from .image import reconstruct_image_sum, reconstruct_image, save_tiff, ImageMetadata
 
 
+def _default_image_factory(self: "ConfocalImage", color):
+    channel_data = getattr(self, f"{color}_photon_count").data
+    raw_image = reconstruct_image_sum(channel_data, self.infowave.data, self._shape)
+    return self._to_spatial(raw_image)
+
+
+def _default_timestamp_factory(self: "ConfocalImage"):
+    # Uses the timestamps from the first non-zero-sized photon channel
+    for color in ("red", "green", "blue"):
+        channel_data = getattr(self, f"{color}_photon_count").timestamps
+        if len(channel_data) != 0:
+            break
+    else:
+        raise RuntimeError("Can't get pixel timestamps if there are no pixels")
+    raw_image = reconstruct_image(channel_data, self.infowave.data, self._shape, reduce=np.mean)
+    return self._to_spatial(raw_image)
+
+
 class BaseScan(PhotonCounts, ExcitationLaserPower):
     """Base class for confocal scans
 
@@ -31,6 +49,8 @@ class BaseScan(PhotonCounts, ExcitationLaserPower):
         self.name = name
         self._json = json
         self.file = file
+        self._image_factory = _default_image_factory
+        self._timestamp_factory = _default_timestamp_factory
         self._cache = {}
 
     @classmethod
@@ -172,22 +192,13 @@ class ConfocalImage(BaseScan):
 
     @cachetools.cachedmethod(lambda self: self._cache)
     def _image(self, channel):
-        channel_data = getattr(self, f"{channel}_photon_count").data
-        raw_image = reconstruct_image_sum(channel_data, self.infowave.data, self._shape)
-        return self._to_spatial(raw_image)
+        assert channel in ("red", "green", "blue")
+        return self._image_factory(self, channel)
 
     @cachetools.cachedmethod(lambda self: self._cache)
     def _timestamps(self, channel):
         assert channel == "timestamps"
-        # Uses the timestamps from the first non-zero-sized photon channel
-        for color in ("red", "green", "blue"):
-            channel_data = getattr(self, f"{color}_photon_count").timestamps
-            if len(channel_data) != 0:
-                break
-        else:
-            raise RuntimeError("Can't get pixel timestamps if there are no pixels")
-        raw_image = reconstruct_image(channel_data, self.infowave.data, self._shape, reduce=np.mean)
-        return self._to_spatial(raw_image)
+        return self._timestamp_factory(self)
 
     def _plot_color(self, color, **kwargs):
         from matplotlib.colors import LinearSegmentedColormap
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -2,6 +2,7 @@ import numpy as np
 import warnings
 import cachetools
 
+from skimage.measure import block_reduce
 from .detail.confocal import ConfocalImage
 from .detail.image import (
     line_timestamps_image,
@@ -9,6 +10,18 @@ from .detail.image import (
     histogram_rows,
 )
 from .detail.timeindex import to_timestamp
+
+
+def _default_line_time_factory(self: "Kymo"):
+    """Line time in seconds"""
+    if self.timestamps.shape[1] > 1:
+        ns_to_sec = 1e-9
+        return (self.timestamps[0, 1] - self.timestamps[0, 0]) * ns_to_sec
+    else:
+        raise RuntimeError(
+            "This kymograph consists of only a single line. It is not possible to determine the "
+            "kymograph line time for a kymograph consisting only of a single line."
+        )
 
 
 class Kymo(ConfocalImage):
@@ -27,6 +40,10 @@ class Kymo(ConfocalImage):
     json : dict
         Dictionary containing kymograph-specific metadata.
     """
+
+    def __init__(self, name, file, start, stop, json):
+        super().__init__(name, file, start, stop, json)
+        self._line_time_factory = _default_line_time_factory
 
     def __repr__(self):
         name = self.__class__.__name__
@@ -95,10 +112,7 @@ class Kymo(ConfocalImage):
     @property
     def line_time_seconds(self):
         """Line time in seconds"""
-        if self.timestamps.shape[1] > 1:
-            return (self.timestamps[0, 1] - self.timestamps[0, 0]) / 1e9
-        else:
-            raise RuntimeError("Line time is not defined for kymograph with only a single line")
+        return self._line_time_factory(self)
 
     def _plot(self, image, **kwargs):
         import matplotlib.pyplot as plt
@@ -235,6 +249,41 @@ class Kymo(ConfocalImage):
         ax_hist.set_xlim(ax_kymo.get_xlim())
         ax_hist.set_ylabel("counts")
         ax_hist.set_title(self.name)
+
+    def downsampled_by(self, time_factor, reduce=np.sum):
+        """Return a copy of this Kymograph which is downsampled by `time_factor`
+
+        Parameters
+        ----------
+        time_factor : int
+            The number of pixels that will be averaged in time.
+        reduce : callable
+            The `numpy` function which is going to reduce multiple pixels into one.
+            The default is `np.sum`.
+        """
+        result = Kymo(self.name, self.file, self.start, self.stop, self._json)
+
+        def image_factory(_, channel):
+            data = self._image(channel)
+            return block_reduce(data, (1, time_factor), func=reduce)[
+                :, : data.shape[1] // time_factor
+            ]
+
+        def timestamp_factory(_):
+            raise AttributeError(
+                "Per-pixel timestamps are no longer available after downsampling a kymograph since "
+                "they are not well defined (the downsampling occurs over a non contiguous time "
+                "window). Line timestamps are still available however. See: "
+                "`Kymo.line_time_seconds`."
+            )
+
+        def line_time_factory(_):
+            return self.line_time_seconds * time_factor
+
+        result._image_factory = image_factory
+        result._timestamp_factory = timestamp_factory
+        result._line_time_factory = line_time_factory
+        return result
 
 
 class EmptyKymo(Kymo):
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -104,14 +104,13 @@ class Kymo(ConfocalImage):
         import matplotlib.pyplot as plt
 
         size_um = self.size_um[0]
-        ts = self.timestamps
-        duration = (ts[0, -1] - ts[0, 0]) / 1e9
-        linetime = (ts[0, 1] - ts[0, 0]) / 1e9
+        duration = self.line_time_seconds * image.shape[1]
+        linetime = self.line_time_seconds
 
         default_kwargs = dict(
             # With origin set to upper (default) bounds should be given as (0, n, n, 0)
             # pixel center aligned with mean time per line
-            extent=[-0.5 * linetime, duration + 0.5 * linetime, size_um, 0],
+            extent=[-0.5 * linetime, duration - 0.5 * linetime, size_um, 0],
             aspect=(image.shape[0] / image.shape[1]) * (duration / size_um),
         )
 
--- a/lumicks/pylake/tests/test_kymo.py
+++ b/lumicks/pylake/tests/test_kymo.py
@@ -140,7 +140,11 @@ def test_plotting(h5_file):
         kymo = f.kymos["Kymo1"]
 
         kymo.plot_red()
-        assert np.allclose(np.sort(plt.xlim()), [-0.5, 3.5], atol=0.05)
+        # The following assertion fails because of unequal line times in the test data. These
+        # unequal line times are not typical for BL data. Kymo nowadays assumes equal line times
+        # which is why the old version of this test fails.
+        # assert np.allclose(np.sort(plt.xlim()), [-0.5, 3.5], atol=0.05)
+        assert np.allclose(plt.xlim(), [-0.515625, 3.609375])
         assert np.allclose(np.sort(plt.ylim()), [0, 0.05])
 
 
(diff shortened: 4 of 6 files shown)


rpauszek (commented):
The data generation functions and de-duplication in the tests is fantastic.

My one concern is having a separate specialized `DownsampledKymo` class. Generally this makes sense, and I don't see an easy way to wrap it into the existing `Kymo`.  I'm worried though that if we end up with a bunch of different types of transformations each needing their own subclass are we going to end up with something unmanageable in the future? Or maybe this won't be a huge problem, definitely would like to hear others' thoughts on this.

lumicks/pylake/kymo.py
```diff
@@ -236,6 +237,62 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+        return DownsampledKymo(
+            self.name,
+            self.file,
+            self.start,
+            self.stop,
+            self._json,
+            self,
+            downsampling_factor,
+            reduce,
+        )
```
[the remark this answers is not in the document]
JoepVanlier: Truncated yes.

lumicks/pylake/kymo.py
```diff
@@ -104,14 +104,13 @@ def _plot(self, image, **kwargs):
...
-        duration = (ts[0, -1] - ts[0, 0]) / 1e9
-        linetime = (ts[0, 1] - ts[0, 0]) / 1e9
+        duration = self.line_time_seconds * image.shape[1]
+        linetime = self.line_time_seconds
 
         default_kwargs = dict(
             # With origin set to upper (default) bounds should be given as (0, n, n, 0)
             # pixel center aligned with mean time per line
-            extent=[-0.5 * linetime, duration + 0.5 * linetime, size_um, 0],
+            extent=[-0.5 * linetime, duration - 0.5 * linetime, size_um, 0],
```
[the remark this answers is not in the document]
JoepVanlier: In the old one, the timestamps corresponded to the center of the first pixel (`0`) and last (`size - 1`). I would argue that before, the duration was actually a misnomer, since it was the duration from the first center to the last center rather than the actual duration of the kymo (which is extent to extent).

In the new setup, for a pixel array with size 4:
`0 | 1 | 2 | 3`

You can see how the extent is `-0.5 dx` to `3 + 0.5 * dx` (or `4 - 0.5 * dx`, or `size - 0.5 * dx`).

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -48,3 +50,103 @@ def generate_scan_json(axes):
             }
         }
     )
+
+
+def generate_image_data(image, samples_per_pixel, line_padding):
+    """Generates the appropriate info_wave and photon_count data for a particular image.
```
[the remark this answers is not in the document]
JoepVanlier: I'll add some short docstrings in the local functions.

lumicks/pylake/tests/conftest.py
```diff
@@ -224,87 +197,64 @@ def h5_file(tmpdir_factory, request):
...
+        ds = mock_file.make_json_data(
+            "Scan",
+            "fast Y slow X",
+            generate_scan_json(
+                [
+                    {"axis": 1, "num of pixels": 4, "pixel size (nm)": 191.0},
+                    {"axis": 0, "num of pixels": 5, "pixel size (nm)": 197.0},
+                ]
+            ),
+        )
```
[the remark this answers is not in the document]
JoepVanlier: Done.

JoepVanlier:
> The data generation functions and de-duplication in the tests is fantastic.
> 
> My one concern is having a separate specialized `DownsampledKymo` class. Generally this makes sense, and I don't see an easy way to wrap it into the existing `Kymo`. I'm worried though that if we end up with a bunch of different types of transformations each needing their own subclass are we going to end up with something unmanageable in the future? Or maybe this won't be a huge problem, definitely would like to hear others' thoughts on this.

One alternative to this is to do something similar to #132 and abstract the image in the `Kymo`. Then you could spin up a new `Kymo` with just an image as source. I closed that PR since it got a lot of pushback on that setup.

dean0x7d:
@JoepVanlier @rpauszek I pushed a suggestion commit. This is what I was suggesting before with an image factory, if you remember. This way, you don't need a different type for the downsampled kymo. You just give it a different factory that produces the image. I think this is the way to go.

(The tests are failing because I pushed before doing the same for `line_time_seconds` but you get the idea.)

JoepVanlier:
> @JoepVanlier @rpauszek I pushed a suggestion commit. This is what I was suggesting before with an image factory, if you remember. This way, you don't need a different type for the downsampled kymo. You just give it a different factory that produces the image. I think this is the way to go.
> 
> (The tests are failing because I pushed before doing the same for `line_time_seconds` but you get the idea.)

Thanks @dean0x7d !

Yeah, your suggestion is a lot nicer. I like how everything related to downsampling is stored compactly in `downsampled_by`.

lumicks/pylake/detail/confocal.py
```diff
@@ -8,6 +8,25 @@
 from .image import reconstruct_image_sum, reconstruct_image, save_tiff, ImageMetadata
 
 
+def _default_image_factory(self, channel):
+    channel_data = getattr(self, f"{channel}_photon_count").data
+    raw_image = reconstruct_image_sum(channel_data, self.infowave.data, self._shape)
+    return self._to_spatial(raw_image)
```
dean0x7d, line 14: I know that I originally factored this out, but looking at it now, it would be good to add a type annotation so that we know what this `self` is intended to be:
```py
def _default_image_factory(self: "ConfocalImage", channel):
```
Same thing for the other factories.

lumicks/pylake/kymo.py
```diff
@@ -237,6 +246,26 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
         ax_hist.set_ylabel("counts")
         ax_hist.set_title(self.name)
 
+    def downsampled_by(self, downsampling_factor, reduce=np.sum):
```
dean0x7d, line 249: * Needs a docstring.
* Rename `downsampling_factor` to just `factor`. It's redundant otherwise. Consider: `kymo.downsampled_by(downsampling_factor=2)` -- it's repeating itself. This is why it's good to use named args in the docs: both as a good first example for users and to see if there are any naming redundancies. (Also: `factor` makes it consistent with `Slice.downsampled_by()`.
* Think about how this API would evolve if the space dimension also needed to be downsampled (only think about it for now).

lumicks/pylake/kymo.py
```diff
@@ -237,6 +246,26 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+        result = Kymo(self.name, self.file, self.start, self.stop, self._json)
```
dean0x7d, line 250: I was originally thinking that this should be `result = copy(self)` which is a lot easier to read and understand. However, I wasn't sure what happens with the `._cache` attribute in that case. We may need to implement a custom `__copy__` to ensure correct behavior. I'd say it's not a priority for this PR but definitely something that should be done to ensure good, predictable behavior of the confocal classes. Both for us and any users that may be using these as part of other structures that might be copied.

lumicks/pylake/kymo.py
```diff
@@ -237,6 +246,26 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+
+        def image_factory(_, channel):
+            data = self._image(channel)
+            return block_reduce(data, (1, downsampling_factor), func=reduce)[
+                :, : data.shape[1] // downsampling_factor
+            ]
+
+        def timestamp_factory(*_):
+            raise NotImplementedError("This method is not implemented for downsampled kymographs.")
```
dean0x7d, line 259: Could use a more descriptive message as I mentioned in my docs comment.

lumicks/pylake/tests/test_kymo.py
```diff
@@ -213,3 +223,43 @@ def test_save_tiff(tmpdir_factory, h5_file):
...
+
+    assert kymo_ds.name == "Mock"
+    assert np.allclose(kymo_ds.red_image, ds)
+    assert np.allclose(kymo_ds.start, 100)
+    assert np.allclose(kymo_ds.pixelsize_um, 1 / 1000)
+    assert np.allclose(kymo_ds.line_time_seconds, 2 * 7 * (5 * 4 + 2 + 2) / 1e9)
+
+    # Per pixel timestamps cannot reasonably be computed for a downsampled image
+    with pytest.raises(NotImplementedError):
+        kymo_ds.timestamps
```
dean0x7d, line 262: The explanation should be part of the error message. And then you can test for a matching message here. Then the test is self-documented and you don't need the comment.

dean0x7d (commented):
Looking good! Thanks for spinning off the test refactor. Makes this much easier to review.

lumicks/pylake/kymo.py
```diff
@@ -11,6 +12,14 @@
 from .detail.timeindex import to_timestamp
 
 
+def _default_line_time_factory(self):
+    """Line time in seconds"""
+    if self.timestamps.shape[1] > 1:
+        return (self.timestamps[0, 1] - self.timestamps[0, 0]) / 1e9
+    else:
+        raise RuntimeError("Line time is not defined for kymograph with only a single line")
```
mikhas, line 20: I would have no idea what to do with that error message. Perhaps link to a definition of `line time`, or explain briefly, so that the uninitiated hacker at least has a fighting chance to understand why this logic is correct?

From reading the Kymo tutorial, I get the idea that it measures the timedelta between two lines, but then I would expect the delta for one line to just be 0. I must be missing something fundamental here why that's that not an acceptable return value.

lumicks/pylake/kymo.py
mikhas: To your last point, that reads as if you'd _want_ to prefix the scaling factors, in case you end up with one for each dimension.

mikhas (commented):
I like the use of factories here and I think this PR is starting to look very solid.

docs/tutorial/kymographs.rst
```diff
@@ -49,6 +49,19 @@ There are also several properties available for convenient access to the kymogra
...
 * `kymo.line_time_seconds` provides the time between successive lines
 
+Downsampling kymograph
+----------------------
+
+We can downsample a kymograph by invoking::
+
+    kymo_ds = kymo.downsampled_by(2)
+
+Note however, that not all functionalities are present anymore when downsampling a kymograph. For
```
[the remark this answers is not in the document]
JoepVanlier: Yes, it's something that may happen in the future; but I'd rather not add it to this PR.

We could call the parameter `factor`, `time_factor` just to future proof it a bit more.

docs/tutorial/kymographs.rst
```diff
@@ -49,6 +49,19 @@ There are also several properties available for convenient access to the kymogra
...
+example, we can no longer access the per pixel timestamps::
+
+    >>> kymo_ds.timestamps
+    NotImplementedError: This method is not implemented for downsampled kymographs.
```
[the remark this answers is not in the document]
JoepVanlier: How about
```python
            raise AttributeError("Per pixel timestamps are no longer available after downsampling "
                                 "a kymograph since they are not well defined (the downsampling"
                                 "occurs over a non contiguous time window). Line timestamps are"
                                 "still available however. See: `Kymo.line_time_seconds`.")
```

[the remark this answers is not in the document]
dean0x7d: Perfect!

lumicks/pylake/kymo.py
JoepVanlier: Agreed. I've made a ticket for it and bumped it to the top of the backlog.

lumicks/pylake/tests/test_kymo.py
JoepVanlier: Good point. :)

lumicks/pylake/kymo.py
JoepVanlier: Yes. Good point.

Regarding your third point, maybe it'd be good to just be explicit. How about `time_factor`? It does mean breaking with `Channel` (which uses `factor`), but then we future proof it in case we end up supporting spatial downsampling.

lumicks/pylake/detail/confocal.py
```diff
@@ -8,6 +8,25 @@
...
+
+
+def _default_timestamp_factory(self, channel):
+    assert channel == "timestamps"
```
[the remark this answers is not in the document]
JoepVanlier: On second thought, I think that assertion should be in the attribute, since it is actually there to protect the caching mechanism. The factory doesn't even need this argument (I'll kill it).

The caching itself and the assertion weren't added in this PR.

lumicks/pylake/detail/confocal.py
```diff
@@ -8,6 +8,25 @@
 from .image import reconstruct_image_sum, reconstruct_image, save_tiff, ImageMetadata
 
 
+def _default_image_factory(self, channel):
```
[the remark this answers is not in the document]
JoepVanlier: It's gone. And I'll add the guard clause in the other attribute.

lumicks/pylake/kymo.py
```diff
@@ -11,6 +12,14 @@
 from .detail.timeindex import to_timestamp
 
 
+def _default_line_time_factory(self):
+    """Line time in seconds"""
+    if self.timestamps.shape[1] > 1:
+        return (self.timestamps[0, 1] - self.timestamps[0, 0]) / 1e9
```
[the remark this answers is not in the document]
JoepVanlier: I'll change it.

lumicks/pylake/kymo.py
JoepVanlier: I think alerting the user that it's one line is better than quietly returning zero. When recording the `Kymo` they still set an intended line time. I would say that this would still be the intended return value.

However, we have no way of knowing whether what we got is one completed line or one incomplete line (the latter is more likely even). Hence there is no way of knowing the actual linetime that was intended.

I do agree that the message could use a touch up though.
```python
        raise RuntimeError("This kymograph consists of only a single line. It is not possible to "
                           "determine the kymograph line time for a kymograph consisting only of a "
                           "single line.")
```

lumicks/pylake/detail/confocal.py
JoepVanlier: Good point.

mikhas (approved):

lumicks/pylake/kymo.py
rpauszek: I'd be ok with `time_factor`. Kymos are inherently higher dimensional than a `Channel` so makes sense that arguments might not be able to be named exactly the same.

lumicks/pylake/kymo.py
```diff
@@ -11,6 +12,18 @@
 from .detail.timeindex import to_timestamp
 
 
+def _default_line_time_factory(self: "Kymo"):
+    """Line time in seconds"""
+    if self.timestamps.shape[1] > 1:
+        ns_factor = 1e9
+        return (self.timestamps[0, 1] - self.timestamps[0, 0]) / ns_factor
```
rpauszek, line 19: ok this might be too nit picky, but as long as we're going to pull the conversion factor out into a named variable I think it might be clearer to specify which way the conversion is going (personally I get tired of having to figure out if it's sec->ns or vice versa). Also, I recall Dean has mentioned to me, as a general rule of thumb, to use multiplication instead of division when it makes sense. So I suggest the following:
```suggestion
        ns_to_sec = 1e-9
        return (self.timestamps[0, 1] - self.timestamps[0, 0]) * ns_to_sec
```

rpauszek (approved):
I definitely like the factories better than a separate downsampled class. Everything looks good to me, just left a tiny naming comment

lumicks/pylake/kymo.py
JoepVanlier: Good point.

docs/tutorial/kymographs.rst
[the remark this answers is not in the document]
dean0x7d: Make sure to also update the message here in the docs.

lumicks/pylake/kymo.py
```diff
@@ -237,6 +250,41 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+            return block_reduce(data, (1, time_factor), func=reduce)[
+                :, : data.shape[1] // time_factor
+            ]
+
+        def timestamp_factory(_):
+            raise AttributeError(
+                "Per pixel timestamps are no longer available after downsampling a kymograph since "
+                "they are not well defined (the downsampling occurs over a non contiguous time "
+                "window). Line timestamps are still available however. See: `Kymo.line_time_"
+                "seconds`."
```
dean0x7d, line 277: Try to avoid splitting single words across lines.

lumicks/pylake/kymo.py
dean0x7d: 👍 👍 👍

lumicks/pylake/tests/test_kymo.py
```diff
@@ -213,3 +224,50 @@ def test_save_tiff(tmpdir_factory, h5_file):
...
+    assert np.allclose(kymo_ds.line_time_seconds, 2 * 7 * (5 * 4 + 2 + 2) / 1e9)
+
+    with pytest.raises(
+            AttributeError,
+            match=re.escape(
+                "Per pixel timestamps are no longer available after downsampling a kymograph since "
+                "they are not well defined (the downsampling occurs over a non contiguous time "
+                "window). Line timestamps are still available however. See: `Kymo.line_time_"
+                "seconds`"
+            ),
```
dean0x7d, line 268: You can keep the test to a substring as long as it makes it clear what's expected:
```suggestion
            match=re.escape(
                "Per pixel timestamps are no longer available after downsampling"
            ),
```

lumicks/pylake/kymo.py
```diff
@@ -237,6 +250,41 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+
+        def image_factory(_, channel):
+            data = self._image(channel)
+            return block_reduce(data, (1, time_factor), func=reduce)[
+                :, : data.shape[1] // time_factor
+            ]
+
+        def timestamp_factory(_):
+            raise AttributeError(
+                "Per pixel timestamps are no longer available after downsampling a kymograph since "
```
dean0x7d, line 274: I think this needs a hyphen since it's an adjective: "Per-pixel timestamps"

dean0x7d (approved):
Looks great! I just noted a couple of error message/wording things, but it should be good to go otherwise.

lumicks/pylake/kymo.py
JoepVanlier: I think you are correct :)

PR #134, opened 2021-04-12
Handle baseline corrected FD curves

rpauszek:
## Why this PR?
Allow users to access FD curves exported with baseline correction from Bluelake.

## Notes
Currently in Bluelake, the baseline is only calculated for the x-component `Fx`. However, as currently implemented in pylake, `FdCurves` by default are constructed using the channel magnitude `sqrt(Fx**2 + Fy**2)`. There are also future considerations (like piezotracking as the distance channel) which are not yet clearly defined. So it seemed most reasonable to add the `FdCurve.with_baseline_correction()` method to return a new instance using the corrected data rather than messing around with object construction from `File`.

~~I chose to add a `FdCurve._baseline_corrected` attribute to keep track of which data is used for each instance. However, I also considered making a separate `BaselineCorrectedFdCurve` subclass. Both options have pros and cons as far as future maintenance and feature extendibility, so any discussion on this is welcome!~~

## Update
After the original PR was opened, found out that BL will not directly export corrected LF data, only corrected HF data. Therefore, the corrected HF data needs to be downsampled like the LF distance channel. In order to do this efficiently, we need functionality from PR #139 .

## Not included  
The following features were originally part of this ticket, however it seemed like the PR would be too large for manageable review. So these will be separate PRs after this one is merged:
- plotting baseline and raw data 
- applying correction to non-corrected data if coefficients are available 
- accessing polynomial coefficients

### DIFF
--- a/docs/tutorial/fdcurves.rst
+++ b/docs/tutorial/fdcurves.rst
@@ -71,3 +71,21 @@ and distance from such an ensemble using::
 
     f = fd_ensemble.f
     d = fd_ensemble.d
+
+Baseline Correction
+-------------------
+
+.. note::
+    By default, FD curves are constructed using the force magnitude :math:`F = \sqrt{F_x^2 + F_y^2}`. However, baseline
+    correction in Bluelake is only calculated for the x-component :math:`F_x`. Therefore, FD curves with baseline
+    correction applied are constructed with only the x-component rather than the full magnitude and may not be directly
+    comparable to the corresponding uncorrected FD curve.
+
+    Additionally, baseline-corrected FD curves are read directly from the source HDF5 file. Therefore, any data processing previously
+    applied to the FD curve used to obtain the baseline corrected curve is lost.
+
+FD curves can also be constructed from baseline corrected force data if the channel was exported from Bluelake with a baseline correction applied::
+
+    file = lk.File("example.h5")
+    fd = file.fdcurves["baseline"]          # low frequency, uncorrected force magnitude
+    fd_bl = fd.with_baseline_corrected_x() # low frequency, baseline corrected force x-component
--- a/lumicks/pylake/fdcurve.py
+++ b/lumicks/pylake/fdcurve.py
@@ -159,6 +159,24 @@ class FdCurve(DownsampledFD):
 
         return new_curve
 
+    def with_baseline_corrected_x(self):
+        """Return a copy of this F,d curve with baseline correction applied to force channel.
+        Note: only the x component is available with baseline correction.
+        Note: All previous data manipulations (eg. subtraction of another curve) will be lost."""
+        new_curve = self.__copy__()
+
+        # get high-frequency, baseline-corrected data
+        corrected_force_hf = getattr(self.file, f"corrected_force{self._primary_force_channel}x")
+        if len(corrected_force_hf) == 0:
+            raise ValueError("baseline correction not found.")
+
+        # downsample to match distance data
+        force, distance = corrected_force_hf.downsampled_like(self.d)
+        new_curve._force_cache = force
+        new_curve._distance_cache = distance
+
+        return new_curve
+
     @property
     def f(self):
         """The primary force channel associated with this FD curve"""
--- a/lumicks/pylake/tests/test_fdcurve.py
+++ b/lumicks/pylake/tests/test_fdcurve.py
@@ -1,6 +1,7 @@
 import numpy as np
 import pytest
 
+from lumicks.pylake import File
 from lumicks.pylake.fdcurve import FdCurve
 from lumicks.pylake.channel import Slice, TimeSeries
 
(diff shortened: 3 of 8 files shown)


JoepVanlier (commented):
Looks mostly alright.

Once we have a bit more clarity on how piezo tracking and such develop, it would be nice to decide on a cleaner API. I'm a little concerned that people may forget to actually take the baseline corrected data.

lumicks/pylake/fdcurve.py
```diff
@@ -159,6 +159,23 @@ def with_offset(self, force_offset=0, distance_offset=0):
 
         return new_curve
 
+    def with_baseline_corrected_x(self):
+        """Return a copy of this F,d curve with baseline correction applied to force channel.
+        Note, only the x component is available."""
```
agom, line 164:
```suggestion
        Note: only the x component is available."""
```

agom (approved):
Looks good! Just a minor docstring correction.

lumicks/pylake/tests/test_fdcurve.py
```diff
@@ -175,3 +178,23 @@ def test_copy_behaviour_with_offset(h5_file):
...
+    assert fd2._primary_force_channel == "1"
+    assert fd2._primary_distance_channel == "2"
+
+
+def test_with_baseline_correction(fd_h5_file):
+    f = File.from_h5py(fd_h5_file)
+    fd = f.fdcurves["fd1"]
+
+    fd2 = fd.with_baseline_correction_x()
+    assert not np.all(np.equal(fd.f.data, fd2.f.data))
```
mcsmurf: This is great.  Since we have the coefficients I'd also add at least one force entry's manual calculation that reverses the corrected force back to the original, then compare that with the `fd.fd.data`, something like:
```
# get the polynomial
p = np.poly1d(f.fdcurves.metadata.coefficients) # Not sure about the actual coefficients variable name

# calculate the original force using the polynomial
x0 = fd.d.data[0]
reverted_f_data_entry = x0 - p(x0)

# compare the manually calculated force against the original
assert np.equal(reverted_f_data_entry, fd.f.data[0])
```

mcsmurf (approved):
Great work! I've added one suggestion for additional tests, but overall it looks good to go.
Go for launch 🚀

lumicks/pylake/tests/test_fdcurve.py
rpauszek: the coefficients are being pushed off to another ticket to try to keep these PRs from becoming too massive. Right now we just read the corrected data channel directly, there's no import for the metadata. I agree this test should be included when that feature is implemented though

lumicks/pylake/fdcurve.py
```diff
@@ -194,6 +216,16 @@ def with_channels(self, force, distance):
...
 
+    def with_baseline_correction(self):
+        """Return a copy of this FD curve with baseline correction applied to force channel.
+        Note, only the x component is available."""
+        new_fd = copy(self)
+        new_fd._baseline_corrected = True
+        new_fd._force_cache = None
+        if len(new_fd.f) == 0:
+            raise ValueError("baseline correction not found.")
+        return new_fd
```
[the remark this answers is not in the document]
rpauszek: now, rather than resetting the caches, the corrected force and distance is read directly from the h5 file and evaluated immediately (similar to what is done in `FdCurve.with_offset()`). I've added an extra note to the docstring and a more explicit description in the tutorial docs.

lumicks/pylake/fdcurve.py
```diff
@@ -194,6 +216,16 @@ def with_channels(self, force, distance):
         new_fd._primary_distance_channel = distance
         return new_fd
 
+    def with_baseline_correction(self):
```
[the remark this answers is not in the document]
rpauszek: name changed to `with_baseline_corrected_x()`

docs/tutorial/fdcurves.rst
```diff
@@ -71,3 +71,21 @@ and distance from such an ensemble using::
...
+Baseline Correction
+-------------------
+
+.. note::
+    By default, FD curves are constructed using the force magnitude :math:`F = \sqrt{F_x^2 + F_y^2}`. However, baseline
+    correction in Bluelake is only calculated for the x-component :math:`F_x`. Therefore, FD curves with baseline
+    correction applied are constructed with only the x-component rather than the full magnitude and may not be directly
+    comparable to the corresponding uncorrected FD curve.
+
+    Additionally, baseline-correded FD curves are read directly from the source HDF5 file. Therefore, any data manipulations
```
JoepVanlier, line 84: `baseline-correded` => `baseline-corrected`

docs/tutorial/fdcurves.rst
```diff
@@ -71,3 +71,21 @@ and distance from such an ensemble using::
...
+    (such as offset correction) that were previously applied to the calling instance will no longer be present.
```
JoepVanlier, line 85: This sentence may be a bit tricky to understand for some. I'm also not a huge fan of the word "manipulation". It sounds evil :D. Maybe something along the lines of:

```Therefore, any data processing previously applied to the `Fd Curve` used to obtain the baseline corrected curve is lost.```

That said, I do like how prominent the note is though.

JoepVanlier (commented):
Mostly looks good. Only have some minor comments.

PR #135, opened 2021-04-13
Putting the Kymo back in Kymotracker [MSD 2b]

JoepVanlier:
## Why this PR?
In the old version of the kymotracker, users had to keep track of units themselves. They would essentially get raw data from `Kymo` and then be left to their own devices. Then, after tracking, they would have to pass their own calibration factors in with whatever functions they'd be using. This is error prone and can lead to subtle mistakes.

## About this PR
- The new design sets up an intermediate class which is essentially a calibrated image. It's just a thin wrapper around an image with a calibration. This is all private, so that if we need to modify some things in the future, we can, without worrying we break people's code.
- Note that the PR is smaller than it looks. The `CalibratedImage` versions of`track_greedy` and `track_lines` were made private (since they need to be accessed directly in `KymoWIdget`) and a wrapper function was added for the user convenient API. There's also quite a bit of documentation that changed because of the API change.

image
_Figure 1: Look at those actual units on the axes of the kymotracker_

## Context
For now, the units are hardcoded to seconds and microns (axis labels and such), but we're going to be adding different axes calibrations in the future. I did not want to feature creep this PR too much, since it is already pretty big.

In the future, we may want to have time be represented in timestamps internally in CalibratedImage, but this requires some additional thought when it comes to how to define it so that it stays consistent.

Docs here: https://lumicks-pylake.readthedocs.io/en/msd_2b/examples/cas9_kymotracking/cas9_kymotracking.html
And here: https://lumicks-pylake.readthedocs.io/en/msd_2b/tutorial/kymotracking.html


lumicks/pylake/kymotracker/detail/tracking.py
```diff
@@ -0,0 +1,180 @@
...
+    peaks = merge_close_peaks(peaks, np.ceil(0.5 * line_width))
+    lines = points_to_line_segments(
+        peaks,
+        kymo_score(vel=vel, sigma=sigma, diffusion=diffusion),
+        window=window,
+        sigma_cutoff=sigma_cutoff,
+    )
+
+    lines = [KymoLine(line.time_idx, line.coordinate_idx, kymograph_channel) for line in lines]
+
```
rpauszek, line 105: would it be worthwhile to have a constructor classmethod that specifically takes `(KymoLineData, CalibratedKymographChannel)` as inputs? On one hand, this clearly couples `KymoLine` and `KymoLineData`, but that might not be a bad thing. Clearly the two classes are related just based on the names. This then defines a very specific interface for converting from one to the other. 

On the other hand it could just end up being unneeded syntactic sugar

rpauszek (commented):
Overall I really like it. While this structure adds a bit of complexity to the internals, you can clearly see from the updated docs that it greatly simplifies things for the user. 

Most of my comments are dealing with docstrings as it can get quite confusing to keep having to figure out if time/spatial attributes are in physical units (sec/um) or pixel units. I think adding a bit more explicit clarification for this will save us some headaches in the future

lumicks/pylake/kymotracker/detail/calibrated_images.py
```diff
@@ -0,0 +1,114 @@
...
+
+class CalibratedKymographChannel:
+    """A kymograph image along with its calibration"""
+
+    def __init__(self, name, data, time_step, pixel_size):
+        self.name = name
+        self.data = data
+        self.time_step = time_step
+        self._pixel_size = pixel_size
+
```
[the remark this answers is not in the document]
JoepVanlier: I've renamed `time_step` to `time_step_ns`?

For the coordinates, it will depend. Soon we will have to support multiple coordinate systems (sequence position etc), so I'd rather not encode it in the name there.

I've also added a little docstring.

lumicks/pylake/kymotracker/tests/test_calibrated_images.py
```diff
@@ -0,0 +1,61 @@
...
+@pytest.fixture(scope="module")
+def test_img():
+    return np.array(
+        [
+            np.uint8([0, 12, 0, 12, 0]),
+            np.uint8([0, 0, 0, 0, 0]),
+            np.uint8([12, 0, 0, 0, 12]),
+            np.uint8([0, 12, 12, 12, 0]),
+        ]
+    )
```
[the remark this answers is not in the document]
JoepVanlier: Certainty could!

lumicks/pylake/kymotracker/tests/test_calibrated_images.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def test_fields_calibrated_kymograph_channel(test_img):
+    calibrated_channel = CalibratedKymographChannel(
+        "test", test_img, time_step=3e9, pixel_size=7
+    )
+    assert calibrated_channel.name == "test"
+    assert np.allclose(calibrated_channel.data, test_img)
```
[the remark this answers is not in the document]
JoepVanlier: Good catch.

lumicks/pylake/tests/test_kymo.py
```diff
@@ -262,3 +262,31 @@ def test_downsampled_kymo():
...
+        samples_per_pixel=5,
+        line_padding=2
+    )
+
+    calibrated_channel = kymo._get_channel("red")
+    assert np.allclose(calibrated_channel.data, image)
+    assert np.allclose(calibrated_channel.time_step, kymo.line_time_seconds * int(1e9))
+    assert np.allclose(calibrated_channel._pixel_size, kymo.pixelsize_um[0])
+    assert np.allclose(calibrated_channel.to_coord(np.array([1, 2, 3])), [7, 14, 21])
+    assert np.allclose(calibrated_channel.to_seconds(np.array([1, 2, 3])), [72.0, 144.0, 216.0])
```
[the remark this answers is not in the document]
JoepVanlier: Fair point. Removed!

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -92,6 +92,10 @@ def __init__(self, time_idx, coordinate_idx, image):
 
     def with_offset(self, time_offset, coordinate_offset):
         """Returns an offset version of the KymoLine"""
+        # Convert from image units to pixels
+        time_offset = self._image.from_seconds(time_offset)
+        coordinate_offset = self._image.from_coord(coordinate_offset)
+
```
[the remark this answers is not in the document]
JoepVanlier: Yeah. Definitely a good idea. I've renamed the variables to the ones you suggested.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -111,17 +115,26 @@ def __getitem__(self, item):
             np.array(np.vstack((self.time_idx[item], self.coordinate_idx[item]))).transpose()
         )
 
+    @property
+    def time(self):
+        return np.array([self._image.to_seconds(t) for t in self.time_idx])
```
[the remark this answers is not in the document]
JoepVanlier: Yes, good one. That would've been a headache in the future.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -111,17 +115,26 @@ def __getitem__(self, item):
...
         rect : Tuple[Tuple[float, float], Tuple[float, float]]
-            Only perform tracking over a subset of the image. Pixel coordinates should be given as:
+            Only perform tracking over a subset of the image. Coordinates should be given as:
             ((min_time, min_coord), (max_time, max_coord)).
         """
-        time_idx = np.array(self.time_idx)
-        coordinate_idx = np.array(self.coordinate_idx)
+        time_idx = np.array([self._image.from_seconds(t) for t in self.time_idx])
+        coordinate_idx = np.array([self._image.from_coord(c) for c in self.coordinate_idx])
+
```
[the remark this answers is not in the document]
JoepVanlier: Oi! You just found a bug. Thanks!

No transformation was necessary here. I suspect this was pulled in when I cherry picked from the other branch I still had open. I will add a regression test for this.

lumicks/pylake/nb_widgets/tests/test_kymotracker_widget.py
```diff
@@ -29,17 +29,21 @@ def test_parameters_kymo(kymograph, mockevent):
...
 @cleanup
 def test_track_kymo(kymograph, region_select):
-    kymo_widget = KymoWidgetGreedy(kymograph, 1, use_widgets=False)
+    kymo_widget = KymoWidgetGreedy(kymograph, "red", 1, use_widgets=False)
     assert len(kymo_widget.lines) == 0
 
+    print("HEYHEYHEY")
+    print("HEYHEYHEY")
+    print(kymograph.line_time_seconds)
+
```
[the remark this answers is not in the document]
JoepVanlier: ...oh that's embarrassing. Removed.

docs/tutorial/kymotracking.rst
```diff
@@ -175,10 +170,10 @@ Exporting kymograph traces to `csv` files is easy. Just invoke `save` on the ret
 
     traces.save("traces.csv")
 
-By default, this saves the kymograph in pixel coordinates. We can also save calibrated coordinates and photon counts by
-passing a calibration factor for the x-axis and y-axis and a width in pixels to sum counts over::
+By default, this saves the kymograph in pixel coordinates. We can also save photon counts by
+passing a width in pixels to sum counts over::
 
-    traces.save("traces_calibrated.csv", dt=kymo.line_time_seconds, dx=kymo.pixelsize_um, sampling_width=3)
+    traces.save("traces_calibrated.csv", sampling_width=3)
```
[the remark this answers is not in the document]
JoepVanlier: It does now :D

lumicks/pylake/kymotracker/detail/calibrated_images.py
```diff
@@ -0,0 +1,114 @@
...
+        if t0_pixels > self.data.shape[1]:
+            raise IndexError(
+                f"Specified minimum time {t0} beyond the time range {self.to_seconds(self.data.shape[1])}"
+            )
+
+        if t0 >= t1:
+            raise IndexError("Please specify rectangle from minimum time to maximum time")
+
+        if p0 >= p1:
+            raise IndexError("Please specify rectangle from minimum position to maximum position")
```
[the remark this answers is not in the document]
JoepVanlier: True. Added (+tests)!

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -180,14 +182,11 @@ def _traverse_line_direction(
     return nodes
 
 
+@dataclass(frozen=True)
 class KymoLineData:
     __slots__ = ["time_idx", "coordinate_idx"]
-
-    def __init__(self, time_idx, coordinate_idx):
-        """Some of the kymotracker classes return numpy arrays. Appending to many small numpy
-        arrays was a bottleneck and therefore we explicitly use lists."""
```
rpauszek: We should probably still have a docstring for `KymoLineData` and this comment could be included as I think it's useful. Definitions for `time_idx` and `coordinate_idx` would be good too for consistency in documentation and also to clearly state expected units (pixels)

lumicks/pylake/kymotracker/detail/calibrated_images.py
```diff
@@ -67,12 +67,18 @@ def get_rect(self, rect):
 
     @property
     def line_time_seconds(self):
-        return self.time_step / 1e9
+        return self.time_step_ns / 1e9
 
     def from_coord(self, coord):
+        """Convert from coordinates to pixels. Note that it rounds the position.
+
+        Note that it rounds the pixel position."""
```
rpauszek, line 91: the `"Note that it rounds the position"` on the first line should be removed

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -86,6 +86,10 @@ def __init__(self, time_idx, coordinate_idx, image):
         self.coordinate_idx = list(coordinate_idx)
         self._image = image
 
+    @classmethod
+    def from_kymolinedata(cls, kymolinedata, image):
+        return cls(kymolinedata.time_idx, kymolinedata.coordinate_idx, image)
+
```
rpauszek: I think this is good, it makes a clear connection between the two classes. Should it maybe be a private method? I don't think we want users using this directly since `KymoLineData` is meant only for internal use, but `KymoLine` is exposed as results from the kymotracker.

lumicks/pylake/kymotracker/detail/calibrated_images.py
```diff
@@ -59,6 +60,9 @@ def get_rect(self, rect):
         t0_pixels, t1_pixels = self.from_seconds(t0), self.from_seconds(t1)
         p0_pixels, p1_pixels = self.from_coord(p0), self.from_coord(p1)
 
+        if np.any(np.array([t0_pixels, t1_pixels, p0_pixels, p1_pixels]) < 0):
+            raise IndexError(f"Selection needs to be in bounds of the image")
+
```
rpauszek, line 65: I think the error message should be more specific something like `"negative indices are not supported."`. This would make it clear that we purposefully do not support negative indices, unlike lists or numpy arrays

lumicks/pylake/kymotracker/tests/test_calibrated_images.py
```diff
@@ -42,19 +42,23 @@ def test_in_rect(test_img, rect, valid_result):
...
+        ((5, 20), (10, 10)),
+        ((5, 16), (10, 18)),
+        ((26, 5), (28, 7)),
+        ((-1, 8), (22, 20)),
+        ((5, -1), (22, 20)),
+        ((5, 8), (-1, 20)),
+        ((5, 8), (22, -1))
+    ],
+)
+def test_rect_errors(rect):
```
rpauszek, line 58: nice

lumicks/pylake/kymotracker/detail/calibrated_images.py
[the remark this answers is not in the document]
rpauszek: fair enough. I think we will need to add some internal attribute to store the calibration type (um, bp, etc...) but probably best to hold off on that until we actually have to implement another spatial coordinate

rpauszek (commented):
a few relatively minor suggestions, but overall looking good

lumicks/pylake/kymotracker/detail/trace_line_2d.py
JoepVanlier: added a proper docstring for it

lumicks/pylake/kymotracker/detail/calibrated_images.py
JoepVanlier: Good catch. Not sure why I didn't see it.

lumicks/pylake/kymotracker/detail/calibrated_images.py
JoepVanlier: I wouldn't talk about indices, since we're selecting by coordinate here, not index.

I'll add the phrase `indexing by negative coordinates not supported`.

lumicks/pylake/kymotracker/detail/calibrated_images.py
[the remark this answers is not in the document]
JoepVanlier: Yeah, I'm in favor of that.  It's also not unlikely that `from_coord` may have two coordinates in the future (if we have dynamic calibrations); but I didn't want to scope creep this PR too much by already introducing the other units.

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Private now.

lumicks/pylake/kymotracker/detail/calibrated_images.py
```diff
@@ -0,0 +1,139 @@
...
+        Line time [nanoseconds].
+    pixel_size : float
+        Pixel calibration.
+    """
+
+    def __init__(self, name, data, time_step_ns, pixel_size):
+        self.name = name
+        self.data = data
+        self.time_step_ns = time_step_ns
+        self._pixel_size = pixel_size
```
rpauszek, line 24: Since we're using them anyway, we could make this a `dataclass` also. Not necessary, but might be good to have consistency for all the classes that just store inputs from `__init__()`

rpauszek (approved):
looks good to me. just a minor comment about dataclasses, but should be a simple fix if you think it's worthwhile

lumicks/pylake/kymotracker/detail/calibrated_images.py
JoepVanlier: I'd like to hold off on this for now. Especially since `_pixel_size` is private for now.

mwilkowski (approved):
🧜‍♀️

lumicks/pylake/kymo.py
```diff
@@ -250,6 +251,10 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
         ax_hist.set_ylabel("counts")
         ax_hist.set_title(self.name)
 
+    def _get_channel(self, channel):
+        """Fetch a calibrated Kymograph channel"""
+        return CalibratedKymographChannel.from_kymo(self, channel)
```
dean0x7d, line 256: This private getter just creates circular coupling between the two modules. Why is it even needed? You could very easily call `.from_kymo()` where needed.

lumicks/pylake/kymotracker/detail/tracking.py
```diff
@@ -0,0 +1,180 @@
...
+    computation to achieve subpixel accuracy. After peak detection the feature points are linked
+    together using a greedy forward search analogous to [2]. This in contrast with the linking
+    algorithm in [1] which uses a graph based optimization approach.
+
+    The linking step traverses the kymograph, tracing lines starting from each frame. It starts
+    with the highest line and proceeds to lines with lower signal intensity. For every point along
+    the line, the algorithm makes a prediction for where the particle will be in the next frame.
+    Points are considered candidates for line membership when they fall within a cone parameterized
+    by a sigma and diffusion constant. The candidate point closest to the prediction is chosen and
+    connected to the line. When no more candidates are available the line is terminated.
```
dean0x7d, line 40: Is this a copy of the entire public docstring? It would be good to avoid a full copy/paste. Instead, just point to the other function and explain what the differences are: why is this one private, how does the API differ.

lumicks/pylake/kymotracker/detail/tracking.py
```diff
@@ -0,0 +1,180 @@
...
+    line width and then attempts to find curvilinear sections. Based on eigenvalue decomposition of
+    the local Hessian it finds the principal direction of the line. It then computes subpixel
+    accurate positions by computing the maximum perpendicular to the line using local Taylor
+    expansions. Pixels where this subpixel position estimate falls within the pixel are considered
+    lines. The initial step leads to lines with a width of one. These lines are then traced. When
+    ambiguity arises, on which point to connect next, a score comprised of the distance to the next
+    subpixel minimum and angle between the successive normal vectors is computed. The candidate with
+    the lowest score is then selected.
+
+    For more information, please refer to the paper.
```
dean0x7d, line 139: Same as above.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -45,88 +20,80 @@ def track_greedy(
...
+        kymograph._get_channel(channel),
+        line_width,
+        pixel_threshold,
+        window,
+        sigma,
+        vel,
+        diffusion,
+        sigma_cutoff,
+        rect,
     )
```
dean0x7d, line 91: I'm not sure that the private API has a purpose here. I know that we talked about having a low-level private API and a high-level public API, but this isn't really what I had in mind. I figured that the private one would be uncalibrated and it would allow you to do unit tests more easily (more data without mocking). The high-level public API would add the calibration.

But this isn't really what ended up here. The private one does pretty much the same thing. Only one argument is different. I don't think that there is any advantage to doing this. It just seems like a lot of code duplication now.

Sorry if I wasn't clear when we talked about this way back. I didn't mean to imply that this should be split into two levels just for the sake of it. It's only useful if it helps the code be clearer (two nicely composed functions are clearer than one giant one) and the tests easier to write and understand (not as many things to set up). This didn't end up being either of those.

I think it should be reworked in one direction or the other. Doesn't really matter if it goes back or is split up better. Just as long as it doesn't stay in a not-so-useful middle.

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -158,7 +139,7 @@ the longest line we found, and have a look at its position over time::
     longest_line = kymowidget.lines[longest_index]
 
     plt.figure(figsize=(5, 3))
-    plt.plot(np.array(longest_line.time_idx) * dt, np.array(longest_line.coordinate_idx) * pixel_size / 1000)
+    plt.plot(np.array(longest_line.seconds), np.array(longest_line.position))
```
dean0x7d, line 142: I think that you can also drop the `np.array()` casts now.

dean0x7d (changes requested):
This looks good overall and the public API is a huge improvement for users! But I just can't get over that `_track_greedy()`/`track_greedy()` split. It adds a weird internal complication. See my comment below.

lumicks/pylake/kymo.py
JoepVanlier: True. I'll remove it.

lumicks/pylake/kymotracker/detail/tracking.py
JoepVanlier: Will do.

lumicks/pylake/kymo.py
JoepVanlier: There's a second thing that could be said for a `_get_channel()` which is that it could be cached; meaning that if a user applies a kymotracking algorithm to multiple sections of the Kymograph, they all hold a reference to the same `CalibratedKymographChannel`. I could see that as a possible sanity check for concatenating groups of `KymoLines` in the future.

Anyways, I can still remove it for this PR.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: This wasn't done with the intention of making a distinction between low level and high level API. But it's true, I clearly didn't indicate well enough what I need/use these functions for.

The original design was to grab a `CalibratedKymographChannel` from whatever object the tracker widget got (through `_get_channel()`) and try to not tie the tracker directly to `Kymo`, but only use things that were in the specified interface `CalibratedKymographChannel` (to have no direct coupling between `Kymo` and the widget).

This had as extra benefit that when tracking multiple subsections of a Kymo (which is a feature), they can all hold a reference to the same `CalibratedKymographChannel`.

The alternative is that we just take a `Kymo` and create a new `CalibratedKymographChannel` for every set of lines and store both the `Kymo` and the `color_channel` in the widget and reconstruct a `CalibratedKymographChannel` every time we track a section of the image.

If you prefer the latter, I can change it around. I don't feel very strongly about it; since we could reintroduce it in the future.

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
JoepVanlier: Yeah good catch.

lumicks/pylake/kymo.py
```diff
@@ -4,6 +4,7 @@
 
 from skimage.measure import block_reduce
 from .detail.confocal import ConfocalImage
+from lumicks.pylake.kymotracker.detail.calibrated_images import CalibratedKymographChannel
```
dean0x7d, line 7: You can drop this unused import now.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Alright. It has been axed.

The price to pay was a bit more coupling between `Kymo` and the actual widget and construction of a `CalibratedKymographChannel` for every tracking operation (but this cost is negligible compared to the actual tracking). Not the end of the world, especially now that we have the factories in `Kymo` (it may be more worthwhile to turn material coming from a WIT stack into a `Kymo` now rather than have a direct route to `CalibratedKymographChannel`; but let's see about that when we get there).

lumicks/pylake/kymo.py
JoepVanlier: Truetrue.

changelog.md
```diff
@@ -6,6 +6,12 @@
 
 * Added `Kymo.downsampled_by()` for downsampling Kymographs. See [kymographs](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
 
+#### Breaking changes
+
+* The attribute `image_data` in `KymoLine` is now private.
+* Make kymotracker functions `track_greedy`, `track_lines` and `KymoWidgetGreedy` take `Kymo` and a channel (e.g. "red") as their input. The advantage of this is that now units of time (seconds) and space (microns) are propagated automatically to the tracked `KymoLine`s. See the [Cas9 kymotracking example](https://lumicks-pylake.readthedocs.io/en/latest/examples/cas9_kymotracking/cas9_kymotracking.html) or the [kymotracking tutorial](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html) for more information.
```
dean0x7d, line 12:
```suggestion
* Make kymotracker functions `track_greedy()`, `track_lines()`, and class `KymoWidgetGreedy` take `Kymo` and a channel (e.g. "red") as their input. 
  The advantage of this is that now units of time (seconds) and space (microns) are propagated automatically to the tracked `KymoLine`s. 
  See the [Cas9 kymotracking example](https://lumicks-pylake.readthedocs.io/en/latest/examples/cas9_kymotracking/cas9_kymotracking.html) or the [kymotracking tutorial](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html) for more information.
```

changelog.md
```diff
@@ -6,6 +6,12 @@
...
+* `KymoLineGroup.save` and `KymoWidgetGreedy.save_lines` no longer take `dx` and `dt` arguments. Instead, the correct time and position calibration is now passed automatically to these functions. See [kymographs](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
```
dean0x7d, line 13:
```suggestion
* `KymoLineGroup.save()` and `KymoWidgetGreedy.save_lines()` no longer take `dx` and `dt` arguments. 
  Instead, the correct time and position calibration is now passed automatically to these functions. See [kymographs](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
```

dean0x7d (approved):
Good to go! Just note the minor changelog fixups.

JoepVanlier:
> Good to go! Just note the minor changelog fixups.

Thank you for the review! :)

PR #137, opened 2021-04-15
force_calibration: don't catch and rethrow exception

JoepVanlier:
**Why this PR?**
We currently catch and lump all errors that can happen during the analytical part of the force calibration. This is not helpful when debugging issues.

This PR handles two known preconditions explicitly (and provides more useful information) and lets the others bubble up.


lpingas (approved):
Nice improvement!

dean0x7d (approved):
Great changes. Thanks.

PR #139, opened 2021-04-26
Return cropped reference channel from downsampled_like

rpauszek:
**Why this PR?**

When using `Slice.downsampled_like()` to downsample a high frequency channel to the same timestamps as a low frequency channel output from Bluelake, the first 1-2 time points cannot be constructed due to missing HF data.

This PR introduces a breaking change where `downsampled_like()` now returns both the downsampled channel and a copy of the reference LF channel that is cropped such that the timestamps of both instances are exactly the same.

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -304,7 +304,8 @@ class Slice:
             ]
         )
 
-        return self._with_data_source(TimeSeries(downsampled, timestamps))
+        new_slice = self._with_data_source(TimeSeries(downsampled, timestamps))
+        return new_slice, other_slice[new_slice._src.start : new_slice._src.stop]
 
     def plot(self, start=None, **kwargs):
         """A simple line plot to visualize the data over time
--- a/changelog.md
+++ b/changelog.md
@@ -18,6 +18,9 @@
 * `KymoLineGroup.save()` and `KymoWidgetGreedy.save_lines()` no longer take `dx` and `dt` arguments. 
   Instead, the correct time and position calibration is now passed automatically to these functions. See [kymographs](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
 
+#### Breaking changes
+* `Slice.downsampled_like()` now returns both the downsampled `Slice` and a copy of the low frequency reference `Slice` cropped such that both instances have exactly the same timestamps.
+
 ## v0.8.2 | 2021-04-30
 
 #### New features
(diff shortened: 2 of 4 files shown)


lumicks/pylake/channel.py
```diff
@@ -304,7 +304,28 @@ def downsampled_like(self, other_slice, reduce=np.mean):
...
+        def crop_slice(full_slice, start_time, stop_time):
+            overlap_indices = np.argwhere(
+                np.logical_and(
+                    full_slice.timestamps >= start_time, full_slice.timestamps <= stop_time
+                )
+            ).squeeze()
+
+            overlap_data = full_slice.data[overlap_indices]
+            overlap_timestamps = full_slice.timestamps[overlap_indices]
+            return full_slice._with_data_source(TimeSeries(overlap_data, overlap_timestamps))
```
JoepVanlier, line 321: Is there a reason you prefer this over just using the slice operator?

lumicks/pylake/channel.py
```diff
@@ -304,7 +304,28 @@ def downsampled_like(self, other_slice, reduce=np.mean):
...
+
+        time_limits = (
+            np.max([new_slice.timestamps[0], other_slice.timestamps[0]]),
+            np.min([new_slice.timestamps[-1], other_slice.timestamps[-1]]),
+        )
+
+        return crop_slice(new_slice, *time_limits), crop_slice(other_slice, *time_limits)
```
JoepVanlier, line 328: Why are you cropping the new slice? Isn't it by definition always smaller?

JoepVanlier: Would it be easier to do something like this or am I overlooking something?

```python
        start_idx = np.searchsorted(timestamps - delta_time[0], self._src.start)
        stop_idx = np.searchsorted(timestamps, self._src.stop)
        timestamps_sliced = timestamps[start_idx:stop_idx]
        delta_time = delta_time[start_idx:stop_idx]

        downsampled = np.array(
            [
                reduce(self[start:stop].data)
                for start, stop in zip(timestamps_sliced - delta_time, timestamps_sliced)
            ]
        )

        new_slice = self._with_data_source(TimeSeries(downsampled, timestamps_sliced))

        return new_slice, other_slice[timestamps[start_idx]:timestamps[stop_idx]]
```

JoepVanlier (changes requested):
I wonder if it's needed to slice the new one, and whether we can't use the slice operator to slice the second one. Seems like you'd get a simpler solution?

Maybe I'm overlooking something here though.

lumicks/pylake/channel.py
rpauszek: ugh, I apparently forgot you could slice `Slice`s ... it's much simpler now

JoepVanlier (approved):
I really like how much cleaner this is :)

I tried putting it through its paces, but couldn't break it so that's an approve from me :)

PR #140, opened 2021-04-28
Mocking confocal (MSD 2`)

JoepVanlier:
**Why this PR?**
Mocking confocal images is needed for MSD 2a. Essentially this introduces some functions that make it possible to easily mock a confocal image. It also deduplicates some of the JSON generation.


rpauszek (approved):
LGTM, seems like all my previous comments from when this was part of #133  have been addressed. It's a great start to deduplicating and cleaning up our tests!

PR #141, opened 2021-04-29
Use a dataclass for KymolineData [MSD 2c]

JoepVanlier:
This PR changes `KymoLineData` to a dataclass. It also adds some small syntactic sugar for generating a `KymoLine` (user facing) from a `KymoLineData` (backend only).

## Why this PR?
Dataclasses are nice.

Note that this PR requires dropping 3.6 support.

### DIFF
--- a/lumicks/pylake/kymotracker/detail/trace_line_2d.py
+++ b/lumicks/pylake/kymotracker/detail/trace_line_2d.py
@@ -1,5 +1,7 @@
 import enum
 import numpy as np
+from dataclasses import dataclass
+from typing import List
 from .geometry_2d import is_in_2d, is_opposite, calculate_image_geometry, get_candidate_generator
 from .scoring_functions import build_score_matrix
 
@@ -180,14 +182,26 @@ def _traverse_line_direction(
     return nodes
 
 
+@dataclass(frozen=True)
 class KymoLineData:
-    __slots__ = ["time_idx", "coordinate_idx"]
+    """Uncalibrated kymograph lines coming from low-level kymotracker functions.
 
-    def __init__(self, time_idx, coordinate_idx):
-        """Some of the kymotracker classes return numpy arrays. Appending to many small numpy
-        arrays was a bottleneck and therefore we explicitly use lists."""
-        self.time_idx = list(time_idx)
-        self.coordinate_idx = list(coordinate_idx)
+    Backend kymograph functions typically work on raw pixel data. These return lines in the
+    `KymoLineData` format. Typically, these will be wrapped with some additional metadata (in
+    the form of a `KymoLine`) before being passed to the user.
+
+    Parameters
+    ----------
+    time_idx : List[float]
+        List of time indices on the tracked image [pixels]
+    coordinate_idx : List[float]
+        List of coordinate indices on the tracked image [pixels]
+
+    Note: Some of the kymotracker functions return numpy arrays. Appending too many small numpy
+    arrays was a bottleneck and therefore we explicitly use lists."""
+
+    time_idx: List[float]
+    coordinate_idx: List[float]
 
     def append(self, time_idx, coordinate_idx):
         """Append time [pixels], coordinate pair [pixels] to the KymoLine"""
@@ -236,7 +250,7 @@ def traverse_line(
     line = np.array(indices_bwd + indices_fwd[1:])
 
     if len(line) > 0:
-        return KymoLineData(time_idx=line[:, 1], coordinate_idx=line[:, 0])
+        return KymoLineData(time_idx=line[:, 1].tolist(), coordinate_idx=line[:, 0].tolist())
 
 
 def detect_lines_from_geometry(
--- a/lumicks/pylake/kymotracker/detail/trace_line_2d.py
+++ b/lumicks/pylake/kymotracker/detail/trace_line_2d.py
@@ -182,7 +182,7 @@ def _traverse_line_direction(
     return nodes
 
 
-@dataclass(frozen=True)
+@dataclass()
 class KymoLineData:
     """Uncalibrated kymograph lines coming from low-level kymotracker functions.
 
@@ -192,21 +192,19 @@ class KymoLineData:
 
     Parameters
     ----------
-    time_idx : List[float]
+    time_idx : np.ndarray
         List of time indices on the tracked image [pixels]
-    coordinate_idx : List[float]
+    coordinate_idx : np.ndarray
         List of coordinate indices on the tracked image [pixels]
+    """
 
-    Note: Some of the kymotracker functions return numpy arrays. Appending too many small numpy
-    arrays was a bottleneck and therefore we explicitly use lists."""
-
-    time_idx: List[float]
-    coordinate_idx: List[float]
+    time_idx: np.ndarray
+    coordinate_idx: np.ndarray
 
     def append(self, time_idx, coordinate_idx):
         """Append time [pixels], coordinate pair [pixels] to the KymoLine"""
-        self.time_idx.append(time_idx)
-        self.coordinate_idx.append(coordinate_idx)
+        self.time_idx = np.append(self.time_idx, time_idx)
+        self.coordinate_idx = np.append(self.coordinate_idx, coordinate_idx)
 
     def __len__(self):
         return len(self.coordinate_idx)
@@ -250,7 +248,7 @@ def traverse_line(
     line = np.array(indices_bwd + indices_fwd[1:])
 
     if len(line) > 0:
-        return KymoLineData(time_idx=line[:, 1].tolist(), coordinate_idx=line[:, 0].tolist())
+        return KymoLineData(time_idx=line[:, 1], coordinate_idx=line[:, 0])
 
 
 def detect_lines_from_geometry(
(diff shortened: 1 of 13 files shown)


lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -180,14 +182,23 @@ def _traverse_line_direction(
     return nodes
 
 
+@dataclass(frozen=True)
 class KymoLineData:
-    __slots__ = ["time_idx", "coordinate_idx"]
+    """Data class which holds the raw data for a Kymograph line.
+
+    Parameters
```
dean0x7d, line 189: Should these be attributes? AFAIK in NumpyDoc, `Parameters` refer to things passed to functions and `Attributes` refer to members.

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -180,14 +182,23 @@ def _traverse_line_direction(
...
+
+    Note: Some of the kymotracker functions return numpy arrays. Appending too many small numpy
+    arrays was a bottleneck and therefore we explicitly use lists."""
 
-    def __init__(self, time_idx, coordinate_idx):
-        """Some of the kymotracker classes return numpy arrays. Appending to many small numpy
-        arrays was a bottleneck and therefore we explicitly use lists."""
-        self.time_idx = list(time_idx)
-        self.coordinate_idx = list(coordinate_idx)
+    __slots__ = ["time_idx", "coordinate_idx"]
```
dean0x7d, line 199: This may be a weird git rebase leftover. This is not needed any more with dataclasses.

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -180,14 +182,23 @@ def _traverse_line_direction(
...
+
+    Parameters
+    ----------
+    time_idx : List[float]
+        List of time indices on the tracked image [pixels]
+    coordinate_idx : List[float]
+        List of coordinate indices on the tracked image [pixels]
+
+    Note: Some of the kymotracker functions return numpy arrays. Appending too many small numpy
+    arrays was a bottleneck and therefore we explicitly use lists."""
```
dean0x7d, line 197: The usual Python convention is to end a docstring with `"""` by itself on the last line:
```suggestion
    Note: Some of the kymotracker functions return numpy arrays. Appending too many small numpy
    arrays was a bottleneck and therefore we explicitly use lists.
    """
```

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -236,7 +247,7 @@ def traverse_line(
     line = np.array(indices_bwd + indices_fwd[1:])
 
     if len(line) > 0:
-        return KymoLineData(time_idx=line[:, 1], coordinate_idx=line[:, 0])
+        return KymoLineData(time_idx=line[:, 1].tolist(), coordinate_idx=line[:, 0].tolist())
```
dean0x7d, line 250: Do you need to convert to a list here? Why not make `KymoLineData` store `np.ndarray` instead? It's more efficient and less to type.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -202,8 +202,8 @@ def refine_lines_centroid(lines, line_width):
     current = 0
     for line in interpolated_lines:
         line_length = len(line.time_idx)
-        line.time_idx = time_idx[current : current + line_length]
-        line.coordinate_idx = coordinate_idx[current : current + line_length]
+        line.time_idx = list(time_idx[current : current + line_length])
+        line.coordinate_idx = list(coordinate_idx[current : current + line_length])
```
dean0x7d, line 206: Same thing here.

dean0x7d (commented):
Python 3.7 and dataclasses. It's a whole new world to explore!

> Note that this PR requires dropping 3.6 support.

I feel like this should have been the headline and not a footnote. I do fully support dropping 3.6 as we've talked about before, but it's not a small thing so it should not be buried. It's also not strictly needed for the kymotracker either so it's a bit weird to have it in the middle of a chain of PRs. It's fine now, so no need reshuffle things. Just a note for the future. It's nicer to introduce big changes like this in a noticeable PR that targets `master` ("big change" as in big impact, not necessarily lots of lines changed).

lumicks/pylake/kymotracker/detail/trace_line_2d.py
JoepVanlier: I had settled on lists since `KymoLineData` is often used for appending. It's not so relevant for that specific algorithm `track_lines`, but more so for `track_greedy` where single points are appended sequentially.

Anyways, measuring is better than guessing, so I just benchmarked the performance difference on some actual Kymos, and it's negligible, so I'll change it to numpy arrays (at least we'd have a more compact memory representation then).

numpy arrays, data classes w/no slots `31.83 +- 1.1`
lists, dataclass w/ slots `30.34 +- 1.9`

I kept the changes in a separate commit so it's easier to review.

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -180,19 +182,25 @@ def _traverse_line_direction(
     return nodes
 
 
+@dataclass()
 class KymoLineData:
-    __slots__ = ["time_idx", "coordinate_idx"]
+    """Data class which holds the raw data for a Kymograph line.
```
dean0x7d, line 187: I feel like this comment is just restating the name + `@dataclass`. You might want to change it to explain why it is needed and what's its relationship with `KymoLine`.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -159,7 +163,7 @@ def sample_from_image(self, num_pixels, reduce=np.sum):
                     max(int(c) - num_pixels, 0) : min(int(c) + num_pixels + 1, y_size), int(t)
                 ]
             )
-            for t, c in zip(self.time_idx, self.coordinate_idx)
+            for t, c in zip(list(self.time_idx), list(self.coordinate_idx))
```
dean0x7d, line 166: Why is `list()` needed here? It probably shouldn't be.

lumicks/pylake/kymotracker/detail/trace_line_2d.py
dean0x7d: I think that for our purposes, `np.ndarray` is always the correct default for ints and floats. And then consider switching to something else if it turns out to be slow.

dean0x7d (approved):
```py
@dataclass
class PR:
    approved: bool = True
```

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: It's not. Removed.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -82,27 +82,31 @@ class KymoLine:
     __slots__ = ["time_idx", "coordinate_idx", "_image"]
 
     def __init__(self, time_idx, coordinate_idx, image):
-        self.time_idx = list(time_idx)
-        self.coordinate_idx = list(coordinate_idx)
+        self.time_idx = time_idx
+        self.coordinate_idx = coordinate_idx
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/tests/test_io.py
```diff
@@ -9,10 +9,10 @@ def kymolinegroup_io_data():
...
 
     test_img = CalibratedKymographChannel("test", data=test_data, time_step_ns=100e9, pixel_size=2)
-    k1 = KymoLine([1, 2, 3], [2, 3, 4], test_img)
-    k2 = KymoLine([2, 3, 4], [3, 4, 5], test_img)
-    k3 = KymoLine([3, 4, 5], [4, 5, 6], test_img)
-    k4 = KymoLine([4, 5, 6], [5, 6, 7], test_img)
+    k1 = KymoLine(np.array([1, 2, 3]), np.array([2, 3, 4]), test_img)
+    k2 = KymoLine(np.array([2, 3, 4]), np.array([3, 4, 5]), test_img)
+    k3 = KymoLine(np.array([3, 4, 5]), np.array([4, 5, 6]), test_img)
+    k4 = KymoLine(np.array([4, 5, 6]), np.array([5, 6, 7]), test_img)
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/trace_line_2d.py
JoepVanlier: True. I've added a bit more description to the docstring.

lumicks/pylake/kymotracker/tests/test_refinement.py
```diff
@@ -6,11 +6,13 @@
...
 def test_kymoline_interpolation():
-    time_idx = [1.0, 3.0, 5.0]
-    coordinate_idx = [1.0, 3.0, 3.0]
+    time_idx = np.array([1.0, 3.0, 5.0])
+    coordinate_idx = np.array([1.0, 3.0, 3.0])
     interpolated = KymoLine(time_idx, coordinate_idx, []).interpolate()
     assert np.allclose(interpolated.time_idx, [1.0, 2.0, 3.0, 4.0, 5.0])
     assert np.allclose(interpolated.coordinate_idx, [1.0, 2.0, 3.0, 3.0, 3.0])
+    assert isinstance(interpolated.time_idx, np.ndarray)
+    assert isinstance(interpolated.coordinate_idx, np.ndarray)
```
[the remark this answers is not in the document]
JoepVanlier: You are correct. We don't need it to be an array-like.

I should test the behaviour that I was trying to saveguard with this, rather than the type. I've added a small extra test, to check whether lines can still be concatenated after refinement / interpolation.

PR #142, opened 2021-04-29
release: preparing for v0.8.2

JoepVanlier:
#### Why this PR?
We want to do a release to get some of the bugfixes and new features out there.

Note that this PR puts a notice at the top of the force calibration page that this API may still be subject to change (as this depends on how well the current API integrates with BL).


tiagobonetti (approved):
`honk` (warning this in alpha honk)

PR #143, opened 2021-04-30
fitting: make test check warning explicitly

JoepVanlier:
#### Why this PR?
When running `pytest --pyargs lumicks.pylake` there is a test failure in `fitting\tests\test_stepping.py`

It seems that when invoking pytest in that way the warning fails the test despite the filterwarning decorator which is supposed to let the test pass despite the warning.

Regardless of this, it is better to explicitly use the context manager to check whether the warning fired, rather than silence it and only look at the result of the stepper.


rpauszek (approved):
LGTM

PR #144, opened 2021-05-05
black: don't let black affect the testing environment

JoepVanlier:
**Why this PR?**
Black's dependencies should be installed after running pytest, so that its dependencies don't affect the tests. Otherwise this could mask missing dependencies.

Note that I did test this on a commit which had this problem, and indeed now the test that should have failed fails.


dean0x7d (approved):

mcsmurf (approved):
Go for launch 🚀

PR #145, opened 2021-05-05
kymotracker: fix position clamping bug in sample_from_image

JoepVanlier:
**Why this PR?**
It fixes a pretty serious bug. I think the changelog describes it best:

Fixed bug in kymotracker which could result in `sample_from_image` returning erroneous values in some cases. When sampling the image, we sample pixels in a region around the traced line. This sampling procedure is constrained to stay within the image bounds. Previously, we used the incorrect axis to clamp the pixel index along the positional axis.

As a consequence, kymographs that are wider (in terms of number of pixels on the positional axis) than they are long (in terms of number of pixels along the time axis) would only have an accurate sampling in the top half of the kymograph. The lower portion of the kymograph would result in zero counts.

The fix is simple. You could clamp by the other axis. Now despite that I would like a 1 character fix, we don't actually need the upper bound clamp, since numpy handles this gracefully. The only clamp that is actually needed is the lower bound clamp (since negative indices would imply slicing from the end, which is not what we want).


lpingas (approved):
Nice fix!

PR #147, opened 2021-05-06
kymotracker: fix bug in `track_lines`

JoepVanlier:
This PR fixes a bug which could lead to bias when tracking lines using track_greedy with the rectangle tool.

Selecting which region to track used to pass that region specifically to the tracking algorithm. This algorithm does a lot of filtering, which means that all these results become biased if the distance between the edge of the ROI and the line becomes too small.

See also https://github.com/lumicks/pylake/pull/146 which fixes a similar issue for the other Kymotracking algorithm.

While the problem is similar, the fixes are different for the two algorithms (hence two PRs).

### DIFF
--- a/lumicks/pylake/kymotracker/detail/trace_line_2d.py
+++ b/lumicks/pylake/kymotracker/detail/trace_line_2d.py
@@ -342,6 +342,7 @@ def detect_lines(
     max_lines=200,
     angle_weight=10.0,
     force_dir=False,
+    roi=None,
 ):
     """Detect lines in an image, based on Steger at al, "An unbiased detector of curvilinear structures".
 
@@ -360,6 +361,9 @@ def detect_lines(
         High values push for straighter lines. Weighting occurs according to distance + angle_weight * angle difference.
     force_dir: bool
         Do not allow lines to backtrack in time.
+    rect: List[List[int, int], List[int, int]]
+        Only track a specific region of the image. ROI is specified as:
+        [[start time (pixels), start position (pixels)], [stop time (pixels), stop time (pixels)]]
     """
     # See Steger et al, "An unbiased detector of curvilinear structures. IEEE Transactions on Pattern Analysis and
     # Machine Intelligence, 20(2), pp.113–125. for a motivation of this scale.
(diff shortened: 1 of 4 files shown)


lumicks/pylake/tests/test_kymotracker.py
```diff
@@ -376,7 +376,9 @@ def test_kymotracker_integration_tests_subset():
 
 def test_kymotracker_test_bias_rect():
     """Computing the kymograph of a subset of the image should not affect the results of the
-    tracking."""
+    tracking. If this test fires, it means that kymotracking on a subset of the image does not
+    produce the same result as on the full thing for `track_greedy`.
```
dean0x7d, line 380:
```suggestion
    produce the same result as on the full thing for `track_greedy()`.
```

lumicks/pylake/tests/test_kymotracker.py
```diff
@@ -394,6 +396,26 @@ def test_kymotracker_test_bias_rect():
         assert np.allclose(t1.position, t2.position)
 
 
+def test_kymotracker_test_bias_rect_lines():
+    """Computing the kymograph of a subset of the image should not affect the results of the
+    tracking. If this test fires, it means that kymotracking on a subset of the image does not
+    produce the same result as on the full thing for `track_lines`.
```
dean0x7d, line 402:
```suggestion
    produce the same result as on the full thing for `track_lines()`.
```

lumicks/pylake/kymotracker/detail/trace_line_2d.py
```diff
@@ -372,6 +376,14 @@ def detect_lines(
     masked_derivative = inside * max_derivative
     masked_derivative[masked_derivative > 0] = 1
 
+    # Mask the maximal line derivative with the ROI. This prevents anything outside of the ROI
+    # to be a candidate for tracking.
+    if roi:
+        ((t0, p0), (t1, p1)) = roi
```
dean0x7d, line 382:
```suggestion
        (t0, p0), (t1, p1) = roi
```

dean0x7d (approved):
`()`

PR #148, opened 2021-05-11
kymo: downsample by position

JoepVanlier:
**Why this PR?**
Downsampling over the spatial coordinate was a requested feature.

**What to look out for**
There's two small tricky aspects to this feature.

1. Downsampling can change the size of the `Kymo` (because pixels for which an insufficient number of input pixels are available are omitted). This means that either we abstract the number of pixels and use a factory method like we did for temporal downsampling; or we modify the internal `_json`. I picked the former to stay consistent to the existing style. Note however, that this means that the deprecated `_json` now contains information that doesn't take into account the downsampled nature of the Kymo.
2. Downsampling in position only should not invalidate timestamps. Downsampling in time should.

Note that I have considered making downsampling by time and position separate private functions on `Kymo`, which are then called by `downsampled_by`, but writing this it introduced so much duplication that I figured it wasn't worth it.


lumicks/pylake/detail/confocal.py
```diff
@@ -256,15 +263,15 @@ def _shape(self):
...
     @property
-    def _fast_axis_metadata(self):
-        return self._json["scan volume"]["scan axes"][0]
+    def _num_pixels(self):
+        return [axes["num of pixels"] for axes in self._ordered_axes()]
 
     @property
     def fast_axis(self):
-        return "X" if self._fast_axis_metadata["axis"] == 0 else "Y"
+        return "X" if self._json["scan volume"]["scan axes"][0]["axis"] == 0 else "Y"
```
mikhas, line 284: I know why this change is in this commit, but the commit message tells me that the removal of `_fast_axis_metadata` method, together with this change, should have been its own commit. Also, the way the actual variable to check the fast axis condition is extracted looks very fragile, or do we assume the json format is part of the public API? The underscore prefix in `_json` makes it look as if it is private/undocumented API that could change at any moment.

lumicks/pylake/scan.py
```diff
@@ -48,7 +48,7 @@ def num_frames(self):
 
     @property
     def lines_per_frame(self):
-        return self._json["scan volume"]["scan axes"][1]["num of pixels"]
+        return self._num_pixels[self._scan_order[1]]
```
mikhas, line 51: (see comment about splitting commit)

lumicks/pylake/detail/confocal.py
```diff
@@ -285,14 +295,15 @@ def timestamps(self) -> np.ndarray:
...
 
     @property
     def size_um(self):
         """Returns a `List` of scan sizes in um along axes. The length of the list corresponds to
         the number of scan axes."""
         return [
-            axes["pixel size (nm)"] * axes["num of pixels"] / 1000 for axes in self._ordered_axes()
+            pixel_size * num_pixels
+            for pixel_size, num_pixels in zip(self.pixelsize_um, self._num_pixels)
         ]
```
mikhas, line 307:
```suggestion
        return map(lambda pixel_size, num_pixels: pixel_size * num_pixels, self.pixelsize_um, self._num_pixels)
```
should work as well and IMO expresses intent better than first zipping and then looping over the two pixel iterables.

JoepVanlier: Sounds good! Should still actually explicitly construct the list though, since map returns a map object (iterator).

I'll change it to 
```python
        return list(
            map(
                lambda pixel_size, num_pixels: pixel_size * num_pixels,
                self.pixelsize_um,
                self._num_pixels,
            )
        )
```

lumicks/pylake/detail/confocal.py
JoepVanlier: What I didn't like was that the `_fast_axis_metadata` was used nowhere else anymore.

What's in the `_json` depends on Bluelake. It comes straight from the file. It shouldn't change, but users shouldn't depend on it, because it would lock us into maintaining a fixed internal format.

We could change it into something like:

```python
    @property
    def _scan_axes_names(self):
        axis_labels = ("X", "Y", "Z")
        return [axis_labels[ax["axis"]] for ax in self._json["scan volume"]["scan axes"]]

    @property
    def fast_axis(self):
        return self._scan_axes_names[0]
```

But Z being the fast axis never happens (and this'd be the only place where it's used), so it seems a bit superfluous (although it may communicate the meaning a bit better).

On the other hand, there is a ticket on the backlog for exposing a few more properties that would be found in the `_json` before, and the axis scanning order is one of them. I wouldn't make the API discussion for that part of this PR, so I would suggest leaving this function private until that PR.

What do you think?

mikhas: The proposed alternative doesn't address my concern, but having axis scanning order as a property would. So let's review that whenever that other PR comes around.

JoepVanlier: Yeah, I realize that. Leave it as is for now then to not scope creep this PR?

lumicks/pylake/detail/confocal.py
```diff
@@ -256,15 +263,15 @@ def _shape(self):
...
     @property
     def pixels_per_line(self):
-        return self._fast_axis_metadata["num of pixels"]
+        return self._num_pixels[self._scan_order[0]]
 
     @property
-    def _fast_axis_metadata(self):
-        return self._json["scan volume"]["scan axes"][0]
+    def _num_pixels(self):
+        return [axes["num of pixels"] for axes in self._ordered_axes()]
```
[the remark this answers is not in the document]
JoepVanlier: Copy of what? The json dictionary? It should put a reference in the list returned in `_ordered_axes()`.

image

`for x in x` shouldn't copy and `axes["num of pixels"]` is just an int.

lumicks/pylake/detail/confocal.py
mikhas: Yup, leave as-is. I should have been more clear on that.

lumicks/pylake/detail/confocal.py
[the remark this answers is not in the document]
mikhas: OK, I was thinking that trying to access a writable field in a list element would force Python to copy the whole list element (think copy-on-write behaviour).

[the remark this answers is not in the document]
JoepVanlier: Ah okay, I don't think dicts or lists use a cow mechanism.

lumicks/pylake/scan.py
JoepVanlier: I've split them now. One specifically for the introduction of `_num_pixels`, then one that changes a few calls to use that attribute, and then the last one which gets rid of the `fast_axis` attribute.

mikhas: First commit in this PR looks wrong now :)

JoepVanlier: You mean introduce two private functions and not use them? Same. I'd preferably squash 1 and 2; but the "see comment about splitting commit" here confused me and I figured you wanted a break between introducing the function and replacing it here.

mikhas: Agree with squashing the first two commits. The third commit `confocal: remove attribute _fast_axis_metadata` is exactly what I asked for. 👍🏽

JoepVanlier: 👍

lumicks/pylake/kymo.py
```diff
@@ -261,39 +261,65 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+            The default is `np.mean`.
         """
         result = Kymo(self.name, self.file, self.start, self.stop, self._json)
 
         def image_factory(_, channel):
             data = self._image(channel)
-            return block_reduce(data, (1, time_factor), func=reduce)[
-                :, : data.shape[1] // time_factor
+            return block_reduce(data, (position_factor, time_factor), func=reduce)[
+                : data.shape[0] // position_factor, : data.shape[1] // time_factor
```
mikhas, line 287: I am trying to make sense of the data slicing here, but I am failing to parse it.  `data[ : foo, : bar ]` looks very strange to the untrained eye.  Can you perhaps explain it to me?

lumicks/pylake/kymo.py
```diff
@@ -261,39 +261,65 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+
+        def pixelcount_factory(_):
+            num_pixels = self._num_pixels
+            num_pixels[0] = num_pixels[0] // position_factor
+            return num_pixels
+
         result._image_factory = image_factory
-        result._timestamp_factory = timestamp_factory
+        result._timestamp_factory = (
+            timestamp_factory_ill_defined if time_factor != 1 else timestamp_factory
```
mikhas, line 318: I'd write it exactly the other way around, to have the happy path for `time_factor == 1` and the ill defined default for the else branch.

lumicks/pylake/tests/test_kymo.py
```diff
@@ -266,7 +266,82 @@ def test_downsampled_kymo():
...
+    kymo = generate_kymo(
+        "Mock", image, pixel_size_nm=1, start=100, dt=5, samples_per_pixel=5, line_padding=2
+    )
+    ds = np.array([[12, 12, 12], [24, 24, 24]], dtype=np.uint8)
+
+    downsampled_kymos = [
+        kymo.downsampled_by(time_factor=2, position_factor=2),
+        # Test whether sequential downsampling works out correctly as well
+        kymo.downsampled_by(position_factor=2).downsampled_by(time_factor=2),
+        kymo.downsampled_by(time_factor=2).downsampled_by(position_factor=2)
```
mikhas, line 331: I wouldn't have thought to test for function chaining. Good stuff! Perhaps could test that the original `kymo` instance is still working after downsampling and that `downsampled_by` has no side effects on the instance?

lumicks/pylake/kymo.py
```diff
@@ -261,39 +261,65 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
         ax_hist.set_ylabel("counts")
         ax_hist.set_title(self.name)
 
-    def downsampled_by(self, time_factor, reduce=np.sum):
+    def downsampled_by(
+        self, time_factor=1, position_factor=1, reduce=np.sum, reduce_timestamps=np.mean
+    ):
         """Return a copy of this Kymograph which is downsampled by `time_factor`
```
mikhas, line 267: The docstring summary here needs updating, no?

lumicks/pylake/kymo.py
JoepVanlier: It's a two dimensional matrix. `: foo` just means up to foo for that axis. We do this for both axes.

What we do here is slice to the nearest integer multiple of the axis dimension. So let's say the original had the last pixel only half filled (not ending on a full multiple of the downsampling factor) then we drop that sample in the resulting array.

Let's say we had a 5 by 5 matrix `(5, 5)`.
Then `data.shape[0] // 2` would be 2. Hence we'd only end up including row 0 and 1 and drop the last one.

lumicks/pylake/kymo.py
JoepVanlier: I had it this way because `time_factor` downsampling would be the most frequently used path. But I'm fine with turning the logic around. :)

lumicks/pylake/tests/test_kymo.py
JoepVanlier: Not a bad idea. Added as a new test!

lumicks/pylake/kymo.py
JoepVanlier: Yep, good catch!

JoepVanlier:
Just as a comment, I left the fixups as separate commits on purpose. I will merge these fixups into the respective commits where the code was introduced before merging.

mikhas (approved):
LGTM!

PR #149, opened 2021-05-12
kymo: raise NotImplemented error for slicing post downsampling

JoepVanlier:
## Why this PR?
Currently slicing after downsampling undoes the downsampling. This is undesirable.

In the future, we will implement slicing in a different way, but a good implementation of this will depend on a number of other features being present first (copy constructor, a separate function for getting start times of the lines that is preserved under downsampling). For now, it's best to leave it explicitly `NotImplemented` to prevent users from getting dependent on weird behaviour.

This PR adds a check whether the factories are the defaults prior to slicing as any specialized factories would be ignored currently.

### DIFF
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -53,6 +53,12 @@ class BaseScan(PhotonCounts, ExcitationLaserPower):
         self._timestamp_factory = _default_timestamp_factory
         self._cache = {}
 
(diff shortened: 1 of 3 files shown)


tiagobonetti (approved):
Looks good to me and the logic of not allowing users to get into odd behavior is very good.

I'm unsure how to do it but It would be good to have a reminder to remove the functions once we have a copy constructor and they become unused.

Otherwise `honk` away!

PR #150, opened 2021-05-12
force_calibration: fix plotting order model and data

JoepVanlier:
**Why this PR?**
This PR fixes a bug in the plotting order of `CalibrationResults.plot()`. Previously, when plotting after performing a force calibration, the model fit was erroneously plotted first (while the legend indicated that the model fit was plotted last). The results of the calibration itself are unchanged.

image
*Figure 1: Cool. The model looks more non-ideal than the data ?!?*

### DIFF
--- a/lumicks/pylake/force_calibration/power_spectrum_calibration.py
+++ b/lumicks/pylake/force_calibration/power_spectrum_calibration.py
@@ -87,9 +87,9 @@ class CalibrationResults:
 
     def plot(self):
         """Plot the fitted spectrum"""
-        self.ps_model_fit.plot()
-        self.ps_fitted.plot()
-        plt.legend(["Data", "Model"])
+        self.ps_fitted.plot(label="Data")
+        self.ps_model_fit.plot(label="Model")
+        plt.legend()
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -87,8 +87,8 @@ def __getitem__(self, item):
 
     def plot(self):
         """Plot the fitted spectrum"""
-        self.ps_model_fit.plot()
         self.ps_fitted.plot()
+        self.ps_model_fit.plot()
         plt.legend(["Data", "Model"])
```
dean0x7d, line 92: Whould this not work and avoid the issue from happening again?
```suggestion
        self.ps_fitted.plot(title="Data")
        self.ps_model_fit.plot(title="Model")
```

JoepVanlier: Fair point. I think it's called `label`, but yes, I reckon that'd be safer.

I've added it and argument forwarding in the `PowerSpectrum` plot function now.

dean0x7d (approved):

PR #151, opened 2021-05-19
Peak localization by Gaussian MLE with Poisson Likelihood

rpauszek:
**Why this PR?**
Adding another type of kymotrack refinement for better estimation of localization (and localization uncertainty). This procedure specifically takes into account the Poisson nature of the signal/noise in confocal images.

This PR sets up all of the algorithms for the actual optimization (backend). The user-facing API dealing with `KymoLine` and other kymotracker-specific classes will be in the next PR.

**What's included**
- Gaussian peak parameterized by amplitude, center, width (scale), and constant background offset
- MLE optimization algorithm, gaussian peak expectation, Poisson cost function, analytical Jacobian
- Options to fit frame by frame (all parameters are free) 
- or all frames in an image simultaneously
   - all parameters can be optimized independently
   - width and/or offset can be optimized for the entire track

Note: I decided to use a true gaussian function for the peak expectation (rather than a normal distribution PDF) as this significantly simplifies the equations since the width parameter is not coupled into the pre-exponential factor (amplitude).

### DIFF
--- a/lumicks/pylake/kymotracker/detail/gaussian_mle.py
+++ b/lumicks/pylake/kymotracker/detail/gaussian_mle.py
@@ -49,16 +49,20 @@ def peak_expectation_derivatives(x, amplitude, center, scale, offset):
     d_damplitude = g / amplitude
     d_dcenter = g * x_center_diff / row(scale) ** 2
 
-    fn = (lambda x: col(x.sum(axis=1))) if len(scale) == 1 else (lambda x: (x))
-    d_dscale = fn(g * x_center_diff ** 2) / scale ** 3
+    d_dscale_numerator = (lambda x: col(x.sum(axis=1))) if len(scale) == 1 else (lambda x: (x))
+    d_dscale = d_dscale_numerator(g * x_center_diff ** 2) / scale ** 3
 
     d_doffset = col(np.ones(g.shape).sum(axis=1)) if len(offset) == 1 else np.ones(g.shape)
 
     return d_damplitude, d_dcenter, d_dscale, d_doffset
 
 
-def _extract_params(params, n_frames, shared_variance):
+def _extract_params(params, n_frames, shared_variance, shared_offset):
     """Separates single array of parameters into individual components."""
+    n_var = 1 if shared_variance else n_frames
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -0,0 +1,176 @@
...
+    amplitude, center, scale, offset = _extract_params(
+        params, n_frames, shared_variance, shared_offset
+    )
+
+    tile = lambda x: np.hstack(
+        (
+            x,
+            x,
+            col(x.sum(axis=1)) if len(scale) == 1 else x,
+            col(x.sum(axis=1)) if len(offset) == 1 else x,
```
JoepVanlier, line 118: It took me a while to understand what's going on here. Some additional comments would be helpful.

Correct me if I'm wrong, but I'm not convinced that summing here is correct. You sum the expectations and photon counts for the aggregated sigmas and offsets prematurely I think. The likelihood is not linear, so you shouldn't sum results for the different frames before evaluating the likelihood for each frame.

If you test with a Jacobian with different model parameters per track you easily get relative errors in your Jacobian entries of more than 10%.

JoepVanlier: I also think it would be nice to extract the functions that map parameter derivatives out of the objective function, so that they can be tested in isolation. Things become hard to test if you need all the pieces in order to test it.

But maybe that's something for a future PR.

JoepVanlier (changes requested):
I think it's a pretty solid start but I have my doubts about the Jacobian (see comments).

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: See https://github.com/lumicks/pylake/tree/gau_loc_not_same for an example of what I mean. First commit is the test update (to use different parameters); second is the order of operation "fix". Note that this was just a quick and dirty hack, and that I'm not proposing this as an implementation of course 😅

PR #152, opened 2021-05-21
Peak localization by Gaussian MLE #2 - User API

rpauszek:
**Why this PR?**
Sets up the user-facing API to use the algorithms implemented in #151. Current implementation only exposes frame-by-frame free parameter fitting (ie, `shared_variance` and `shared_offset` are hard coded as `False`)

### DIFF
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -292,7 +292,9 @@ def refine_lines_gaussian(lines, window=5):
             limits = slice(*get_window_limits(c))
             position = full_position[limits]
             y = img.data[limits, int(t)]
-            r = run_gaussian_mle(position, y, img._pixel_size, False, False, init_center=img.to_position(c))
+            r = run_gaussian_mle(
+                position, y, img._pixel_size, False, False, init_center=img.to_position(c)
+            )
             tmp_coordinate.append(r[1] / img._pixel_size)
         gau_lines.extend(KymoLine(line.time_idx, tmp_coordinate, img))
     return gau_lines
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -8,10 +8,18 @@ from .detail.peakfinding import (
     merge_close_peaks,
     KymoPeaks,
 )
+from .detail.gaussian_mle import run_gaussian_mle
 from .kymoline import KymoLineGroup
 import numpy as np
+import warnings
 
-__all__ = ["track_greedy", "track_lines", "filter_lines", "refine_lines_centroid"]
+__all__ = [
+    "track_greedy",
+    "track_lines",
+    "filter_lines",
+    "refine_lines_centroid",
+    "refine_lines_gaussian",
+]
 
 
 def track_greedy(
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -243,3 +251,48 @@ def refine_lines_centroid(lines, line_width):
...
+    unique_frames = np.unique(window_ranges[:, 0], return_counts=True)
+    overlap_count = 0
+    for frame, count in zip(*unique_frames):
+        if count > 1:
+            idx = np.argwhere(window_ranges[:, 0] == frame)
+            ranges = window_ranges[idx.squeeze(), 1:]
+            ranges = ranges[np.argsort(ranges[:, 0])]
+            overlap_count += np.sum(ranges[1:, 0] < ranges[:-1, 1])
+    if overlap_count > 0:
+        warnings.warn("there are overlaps")
```
mikhas, line 283: So if we go so far as to trigger a warning here, what's the user supposed to do with that information? We continue with the function either way.

Would it be better if the overlapping frames were annotated in the function's returned result instead so that the user could find the problems when further analyzing the data?

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -243,3 +251,48 @@ def refine_lines_centroid(lines, line_width):
...
+
+    # optimize trace by trace, frame by frame
+    full_position = img.to_position(np.arange(img.data.shape[0]))
+    gau_lines = KymoLineGroup([])
+    for j, line in enumerate(lines):
+        # print(f"working on {j} of {len(lines)}")
```
mikhas, line 289: left over code comment

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -292,7 +292,9 @@ def refine_lines_gaussian(lines, window=5):
             limits = slice(*get_window_limits(c))
             position = full_position[limits]
             y = img.data[limits, int(t)]
-            r = run_gaussian_mle(position, y, img._pixel_size, False, False, init_center=img.to_position(c))
+            r = run_gaussian_mle(
+                position, y, img._pixel_size, False, False, init_center=img.to_position(c)
+            )
```
mikhas, line 297: coding style fixups should be separate commits IMO

lumicks/pylake/kymotracker/tests/test_refinement.py
```diff
@@ -54,3 +54,24 @@ def test_refinement_line(loc, inv_sigma=0.3):
...
+    for m in multipliers:
+        init_center = true_params[1] * m
+        init_center_idx = init_center / channel._pixel_size
+        init_tracks.append(KymoLine(np.arange(n_frames), np.full(n_frames, init_center_idx), channel))
+
+    results = ([2.604524,   2.59820109, 2.5995581 ],
+               [2.61053829, 2.58411975, 2.59177182],
+               [2.59546472, 2.59955604, 2.59679171])
+    with pytest.warns(UserWarning):
+        fitted = refine_lines_gaussian(init_tracks)
```
mikhas, line 75: Should test with pathological window sizes, too. For instance, `window=0`, `window=-1`. And where `2 * window + 1` exceeds image size.

JoepVanlier (changes requested):
Some small suggestions. Seems there were some leftovers from debugging and an error message that could be more informative.

Also, no documentation? Given that this is an API ticket, there should be some.

PR #153, opened 2021-06-07
force_calibration: tighten test tolerances

JoepVanlier:
**Why this PR?**
Diffusion constant and power spectrum tests were ineffective at catching problems, since the default absolute tolerance set by `numpy` was high enough for pretty big relative deviations to slip by (diffusion constants in volts are tiny, in the order of 1e-12). This PR tightens these tolerances to make sure the tests fire when something changes inadvertently.

Note that nothing was broken.

### DIFF
--- a/lumicks/pylake/force_calibration/tests/test_power_model.py
+++ b/lumicks/pylake/force_calibration/tests/test_power_model.py
@@ -69,7 +69,7 @@ def test_fit_analytic(
     fit = fit_analytical_lorentzian(ps.in_range(1e1, 1e4))
 
     assert np.allclose(fit.fc, corner_frequency)
-    assert np.allclose(fit.D, diffusion_constant)
+    assert np.allclose(fit.D, diffusion_constant, rtol=1e-5, atol=0)
     assert np.allclose(fit.sigma_fc, sigma_fc)
     assert np.allclose(fit.sigma_D, sigma_diffusion)
 
--- a/lumicks/pylake/force_calibration/tests/test_power_model.py
+++ b/lumicks/pylake/force_calibration/tests/test_power_model.py
@@ -68,7 +68,7 @@ def test_fit_analytic(
 
     fit = fit_analytical_lorentzian(ps.in_range(1e1, 1e4))
 
-    np.testing.assert_allclose(fit.fc, corner_frequency)
+    np.testing.assert_allclose(fit.fc, corner_frequency, rtol=1e-5)
     np.testing.assert_allclose(fit.D, diffusion_constant, rtol=1e-5, atol=0)
     np.testing.assert_allclose(fit.sigma_fc, sigma_fc)
     np.testing.assert_allclose(fit.sigma_D, sigma_diffusion)
@@ -79,4 +79,4 @@ def test_fit_analytic_curve():
     ref = [0.0396382, 0.0389208, 0.03691641, 0.03399826, 0.03061068, 0.02713453]
     fit = fit_analytical_lorentzian(ps)
     np.testing.assert_allclose(fit.ps_fit.frequency, np.arange(0, 60, 10))
-    np.testing.assert_allclose(fit.ps_fit.power, ref)
+    np.testing.assert_allclose(fit.ps_fit.power, ref, rtol=1e-5)
--- a/lumicks/pylake/force_calibration/tests/test_power_model.py
+++ b/lumicks/pylake/force_calibration/tests/test_power_model.py
@@ -4,11 +4,11 @@ from lumicks.pylake.force_calibration.power_spectrum import PowerSpectrum
 
 
 def test_friction_coefficient():
-    assert np.allclose(sphere_friction_coefficient(5, 1), 3 * np.pi * 5 * 1)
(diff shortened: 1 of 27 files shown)


mikhas (commented):
Can you please add the story behind this in numpy and Python to the commit message? Would be nice to be able to read about this problem.

JoepVanlier:
> Can you please add the story behind this in numpy and Python to the commit message? Would be nice to be able to read about this problem.

Good idea, thanks!

Added! :)

mikhas (commented):
As discussed, perhaps try monkey-patching `np.allclose` for all test cases? Would be interesting to see if the fixed version of `np.allclose` shows more breakage.

JoepVanlier:
> As discussed, perhaps try monkey-patching `np.allclose` for all test cases? Would be interesting to see if the fixed version of `np.allclose` shows more breakage.

I was looking into monkey patching it first, but I feel it's going to hurt us in the end. It seems that what we actually want is just the `assert_allclose` from `numpy`'s testing module. This has a default `atol` of zero and actually checks some interesting other characteristics as well (such as shape).

I replaced all instances of `allclose` where this seemed relevant. I left the replacements as a separate commit so you can see what broke. Not as much as I was afraid of thankfully.

Interestingly, one test was dead and wrong! There was an `np.allclose`, but no actual assert.

lumicks/pylake/tests/test_channels.py
```diff
@@ -534,7 +534,7 @@ def test_downsampling_over_subset():
     s = channel.Slice(channel.Continuous(d, 0, 10))
 
     sd = s.downsampled_over([(20, 40), (40, 60), (60, 80)])
-    np.testing.assert_allclose(sd.data, [(3+4)/2, (4+5)/2, (5+6)/2])
+    np.testing.assert_allclose(sd.data, [(3+4)/2, (5+6)/2, (7+8)/2])
     np.testing.assert_allclose(sd.timestamps, [(20+30)/2, (40+50)/2, (60+70)/2])
```
mikhas: OK, this one I don't get. There seems to be an obvious relation between `sd.data` and `sd.timestamps`, which was previously only correct for the first tuple of `(20, 40)`? Perhaps good idea to explain it in a brief comment?

lumicks/pylake/nb_widgets/tests/test_fd_selector.py
```diff
@@ -93,7 +93,7 @@ def test_selector_widget(mockevent):
     event = mockevent(selector._axes, 900, 1, rmb, False)
     selector.handle_button_event(event)
     assert selector.current_range == []
-    np.testing.assert_allclose(selector.ranges, [[]])
```
mikhas: > testing: empty range list is just an empty list

Computer science would disagree with this: The set that only contains the empty set is still a (super) set with cardinality of one :)

mikhas (approved):
LGTM!

lumicks/pylake/nb_widgets/tests/test_fd_selector.py
JoepVanlier: ghehe, yes :D

lumicks/pylake/tests/test_channels.py
JoepVanlier: I've added a small comment.

lumicks/pylake/force_calibration/tests/test_power_model.py
```diff
@@ -79,4 +79,4 @@ def test_fit_analytic_curve():
     ref = [0.0396382, 0.0389208, 0.03691641, 0.03399826, 0.03061068, 0.02713453]
     fit = fit_analytical_lorentzian(ps)
     np.testing.assert_allclose(fit.ps_fit.frequency, np.arange(0, 60, 10))
-    np.testing.assert_allclose(fit.ps_fit.power, ref)
+    np.testing.assert_allclose(fit.ps_fit.power, ref, 1e-5)
```
rpauszek, line 82: maybe add the keyword to the final argument? seems like we usually do this and it would make it clear in the code, don't have to look up the docs if you can't remember if `atol` or `rtol` comes first

lumicks/pylake/population/tests/test_mixture.py
```diff
@@ -39,13 +39,13 @@ def test_exit_flag(trace_simple):
     ef = m.exit_flag
     assert ef["converged"] == True
     assert ef["n_iter"] == 2
-    np.testing.assert_allclose(ef["lower_bound"], 0.215335)
+    np.testing.assert_allclose(ef["lower_bound"], 0.215335, 1e-5)
```
rpauszek, line 42: same as above

rpauszek (approved):
very small change would be nice, just to keep things easy for us in the long run, otherwise looks good

lumicks/pylake/population/tests/test_mixture.py
JoepVanlier: 👍

lumicks/pylake/force_calibration/tests/test_power_model.py
JoepVanlier: Yeah, good point. Done! 👍

PR #154, opened 2021-06-11
force_calibration: make `PowerSpectrum` work in `V^2/Hz`

JoepVanlier:
**Why this PR?**
The `PowerSpectrum` we use uses the convention from the Berg-Sørensen 2004 paper. It's actually only the positive part of the power spectrum _without_ multiplying it by two (as you normally would a one-sided spectrum). This means the axis actually read `0.5 V^2/Hz` (something that is even incorrectly displayed now).

This may cause confusion with users since most scientific python libraries (and a chunk of literature) define the one sided power spectrum as `2|real FFT|^2`. This PR changes the definition and takes it into account in the model and tests.

Unfortunately, this is a breaking change. But given that the Force Calibration module is still clearly marked as beta, it's probably best to get it over with now.

Note: I noticed that the guess that the initial guess function `guess_f_diode_initial_value` provides is very poor. Nevertheless, I added a test just to pin its behaviour. I'll make a ticket for investigating this a bit further. At a first glance, the code in there doesn't seem to make a whole lot of sense. It evaluates the Lorentzian at Nyquist _but_ actually, we don't fit up to Nyquist, we typically fit a range, so the frequency used as a reference to compare to the attenuated level makes no sense. In addition to that, it uses the estimates from the analytic fit as a guess for their true values, which when the diode is doing stuff are actually off by quite a lot. I'm not sure this initial guess adds much in practice.


lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -12,7 +12,7 @@ class PowerSpectrum:
     frequency : numpy.ndarray
         Frequency values for the power spectrum. [Hz]
     power : numpy.ndarray
-        Power values for the power spectrum (typically in V^2/s).
+        Power values for the power spectrum (typically in 0.5 V^2/s).
```
dean0x7d, line 15: This one was wrong in a different way before:
```suggestion
        Power values for the power spectrum (typically in 0.5 V^2/Hz).
```

changelog.md
```diff
@@ -28,6 +28,8 @@
   This means that the blurring steps involved in this algorithm become biased (since they do not get contributions from outside the selected areas, while they should).
   In the updated version, all image processing steps that depend on the image use the full image.
 * Fixed a bug in the plotting order of `CalibrationResults.plot()`. Previously, when plotting after performing a force calibration, the model fit was erroneously plotted first (while the legend indicated that the model fit was plotted last). The results of the calibration itself are unchanged.
+* Fixed incorrect docstring for `PowerSpectrum` that incorrectly defined the `power` field in the `PowerSpectrum` as `V^2/Hz` where it should have been `0.5 V^2/Hz`.
+* Fixed incorrect label on `PowerSpectrum.plot()` to reflect that it actually plots `0.5 V^2/Hz`.
```
dean0x7d, line 32: This should really be a single entry.

dean0x7d (commented):
What are the tradeoffs of adding `0.5 *` to the label vs. multiplying the result by 2? From my point of view, a unit with a non-SI scaling factor is a bit weird. But that could just be what I'm used to and a different convention makes sense here. I'm just wondering what the reasoning is.

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: > What are the tradeoffs of adding `0.5 *` to the label vs. multiplying the result by 2? From my point of view, a unit with a non-SI scaling factor is a bit weird. But that could just be what I'm used to and a different convention makes sense here. I'm just wondering what the reasoning is.

Nah, I agree with you. I see three options:

1. We update `.power` to use the more common SI convention. But this is a breaking change. It's true that this would be a release with many of those already, but what I don't like is that it has the risk of passing in user code silently. If we were to start over, this would have my preference, but unfortunately, this wasn't caught before the first release. Note that this would also mean updating a number of equations; which would then be different by a factor two from the paper. This might confuse people looking at the code, but a comment could easily mitigate that part.
If we go this route, I would maybe defer this PR to until we are ready to begin merging the active calibration stuff (because otherwise I see a lot of room for potential mistakes with factors of two, since it uses the current convention).

2. We keep the convention of the first force calibration paper and update the axis labels to correctly reflect what's in them (that's what is currently in this PR).

3. We change what we plot and multiply things by two prior to plotting. This in my opinion is the worst option, because it makes users expect that the `power` property will be what we plot.

I was thinking of doing 2 now, and 1 once we bring the two calibration scripts together again. What do you think?

JoepVanlier: > This one was wrong in a different way before:

Right. Fixing.

dean0x7d: Definitely not 3.

I'm not sure that I understand the reasoning to do both 2 and 1. I'd be fine with just 2 if that's the prevailing convention in the field. I would not want to complicate this if it also has an impact on active calibration and we'll always need to worry about the convention.

JoepVanlier: I wouldn't go as far as to call it a field-wide definition. It's used by 5 out of the 6 papers that I have. The active calibration paper (most recent) actually has the "correct factor" of two, but the implementation we have switches those equations to the old convention (that we currently have in Pylake). Note that the 5 that use the convention we use always just put `arbitrary units` on their axis.

Functions in `scipy` for instance, all use the "correct" definition (not ours).

Unfortunately, because of the first paper using an unconventional form to begin with, we'll always have to be careful about it.

Long term, my preference would be to express it in `V^2/Hz` but that would mean breaking backward compatibility and redefining `PowerSpectrum.power`. The functionality is still flagged as beta now, but it still doesn't feel right to change the definition of an existing attribute on people. We could think of a new name for `power`, but quite frankly, that's the best name for it. I guess we could deprecate `power` and add `power_density` but that feels overlong.

In either case, what we have on the axis now is incorrect. Maybe we can briefly discuss tomorrow post-standup.

dean0x7d: If that's the case and the current `0.5 V^2/Hz` convention is the uncommon one, I'd pull the bandaid off right now rather than wait to do it in the future. The feature is clearly marked as alpha in the docs and I can't imagine many people are using it right now (as opposed to just using the calibration from Bluelake). Going for the breaking change now would be as painless as it will ever be.

And since we are working on pulling this into Bluelake (not yet live), I'd rather get the correct scale from the start. We do also have to consider that this will result in a 2x PSD scale change in the force calibration window when we switch from LDAT to pylake but AFAIK the only thing that people rely on there is the shape, now the exact values.

JoepVanlier: Yeah, it'll just be a vertical shift on the graph. I indeed checked internally and people only rely on the shape.

Alright. Pulling the bandaid off it is. We will really have to be extra careful pulling in the active calibration materials too, since they still operate in `0.5 V^2/Hz`.

lumicks/pylake/force_calibration/tests/conftest.py
```diff
@@ -16,7 +17,7 @@ def lorentzian_filtered(f, fc, diffusion_constant, alpha, f_diode):
     def lorentzian_td(corner_frequency, diffusion_constant, alpha, f_diode, num_samples):
         f = np.arange(0, num_samples)
         power_spectrum = ReferenceModels.lorentzian_filtered(f, corner_frequency, diffusion_constant, alpha, f_diode)
-        data = np.fft.irfft(np.sqrt(np.abs(power_spectrum))) * (num_samples - 1) * 2
+        data = np.fft.irfft(np.sqrt(np.abs(power_spectrum * 0.5)), norm="forward")
```
dean0x7d, line 20: The `numpy` docs mention something about parts of `norm` only being available from `numpy` v1.20: https://numpy.org/doc/stable/reference/generated/numpy.fft.irfft.html

You might need to bump the minimum required version: https://github.com/lumicks/pylake/blob/667e7ff64ec8a415ade23f638e42fcb4dc3bd592/setup.py#L55

dean0x7d (approved):
Looks good to me! Just please check if you need to bump the `numpy` version requirement.

lumicks/pylake/force_calibration/tests/conftest.py
JoepVanlier: Good catch! I will update the requirement.

PR #155, opened 2021-07-01
force_calibration: store points per block with fit

JoepVanlier:
**Why this PR?**
Currently, the returned model fit contains a `num_points_per_block` of `1`. Instead, it would make a lot more sense to store the number of points per block that were used in the fitting procedure.

This PR mitigates that by storing the number of points per block used for fitting in the `num_points_per_block` attribute of the fitted power spectrum.


agom (approved):
👍

PR #157, opened 2021-07-02
calibration: freq and amp estimation for driving input

JoepVanlier:
**Why this PR?**
To be able to do active force calibration, we need to know exactly the frequency and amplitude of the oscillation with which we were driving the stage. This PR adds a backend function that can estimate this from data.

### DIFF
--- a/lumicks/pylake/force_calibration/tests/test_driving_input.py
+++ b/lumicks/pylake/force_calibration/tests/test_driving_input.py
@@ -1,0 +1,39 @@
+from lumicks.pylake.force_calibration.detail.driving_input import estimate_driving_input_parameters
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize(
+    "sine_waves, expected_amp, expected_freq, guess, f_search, n_fit",
+    [
+        [[[0.5, 36.9], [1, 50]], 0.5, 36.9, 36, 5, 2],
+        [[[0.34, 36.7], [1, 50]], 0.34, 36.7, 36, 5, 2],
+        [[[0.45, 36.554], [1, 50]], 0.45, 36.554, 36, 5, 2],
+        [[[0.5, 36.9], [1, 42.0]], 0.5, 36.9, 36, 5, 2],  # Does window exclude the bigger peak?
+        [[[0.5, 36.9], [1, 42.0]], 1.0, 42.0, 36, 10, 2],  # Bigger window includes second peak
+        [[[0.5, 36.9], [1, 41.0]], 1.0, 41.0, 36, 5, 2],  # Does it pick up on the second peak?
+        [[[0.5, 36.9], [-1, 41.0]], 1.0, 41.0, 36, 5, 2],  # Does it pick up on the second peak?
+        [[[0.5, 36.9], [1, 42.0]], 0.617576, 41.999864, 36, 10, 10],  # Fit range has both peaks
+        [[[0.5, 2.4], [0, 50]], 0.5, 2.4, 3, 5, 2],  # Does it reject offset correctly?
+    ],
+)
+def test_driving_input_estimation(sine_waves, expected_amp, expected_freq, guess, f_search, n_fit):
+    sample_rate = 78125
+    duration = 10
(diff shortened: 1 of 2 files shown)


lumicks/pylake/force_calibration/tests/test_driving_input.py
```diff
@@ -0,0 +1,32 @@
...
+
+
+@pytest.mark.parametrize(
+    "amps, frequencies, true_amp, true_freq, guess",
+    [
+        [[0.5, 1], [36.9, 50], 0.5, 36.9, 36],
+        [[0.34, 1], [36.7, 50], 0.34, 36.7, 36],
+        [[0.45, 1], [36.554, 50], 0.45, 36.554, 36],
+        [[0.5, 1], [36.9, 42.0], 0.5, 36.9, 36],  # Does window exclude the bigger peak?
+        [[0.5, 1], [36.9, 41.0], 1.0, 41.0, 36],  # Does it pick up on the second peak?
```
tiagobonetti, line 13: is not very clear why we pick up `41` but not `42` (I guess this is because the `f_search = 5`).
Maybe we should have the `f_search` be part of the parameters and have an example with the same input but changing it changes the result.

lumicks/pylake/force_calibration/tests/test_driving_input.py
```diff
@@ -0,0 +1,32 @@
+from lumicks.pylake.force_calibration.detail.driving_input import estimate_driving_input_parameters
+import numpy as np
+import pytest
+
+
+@pytest.mark.parametrize(
+    "amps, frequencies, true_amp, true_freq, guess",
+    [
+        [[0.5, 1], [36.9, 50], 0.5, 36.9, 36],
```
tiagobonetti, line 9: I find the data structure here curious.
I would have used a list of amplitude/frequency pairs, so instead of passing the "true" amplitude and frequency again, we could use an index.

``` python
        [[0.5, 36.9], [1, 50], 0, 36],
```

lumicks/pylake/force_calibration/tests/test_driving_input.py
```diff
@@ -0,0 +1,32 @@
...
+        [[0.34, 1], [36.7, 50], 0.34, 36.7, 36],
+        [[0.45, 1], [36.554, 50], 0.45, 36.554, 36],
+        [[0.5, 1], [36.9, 42.0], 0.5, 36.9, 36],  # Does window exclude the bigger peak?
+        [[0.5, 1], [36.9, 41.0], 1.0, 41.0, 36],  # Does it pick up on the second peak?
+        [[0.5, -1], [36.9, 41.0], 1.0, 41.0, 36],  # Does it pick up on the second peak?
+        [[0.5, 0], [2.4, 50], 0.5, 2.4, 3],  # Does it reject offset correctly?
+    ],
+)
+def test_driving_input_estimation(amps, frequencies, true_amp, true_freq, guess):
```
tiagobonetti, line 18: Maybe consider using `expected_amp|freq` instead of `true_amp|freq`.

tiagobonetti (approved):
This was a great PR.
I learned about some nice `numpy` and `scipy` functions and the explanation of the method is great.

I added minor suggestions that I think may improve readability a bit, but in general, this PR is ready!

`honk_amp, honk_freq = estimate_honk_parameters(honk_rate, goose_factor=10)`

lumicks/pylake/force_calibration/tests/test_driving_input.py
JoepVanlier: 👍

lumicks/pylake/force_calibration/tests/test_driving_input.py
JoepVanlier: Will do!

lumicks/pylake/force_calibration/tests/test_driving_input.py
JoepVanlier: Added `f_search` and `n_fit` as extra test input parameters and also added tests for varying both.

PR #158, opened 2021-07-05
tiff: fix deprecation warning (photometric no longer inferred)

JoepVanlier:
**Why this PR?**
`tifffile` no longer automatically infers what `photometric` type it is dealing with, so now we have to do it.


dean0x7d:
* Since which version of `tifffile` is this? Should a bug fix entry be added to the changelog?
* Can it be covered with a test?

JoepVanlier:
> * Since which version of `tifffile` is this?

Seems to be `2021.7.2` (most recent).

> * Should a bug fix entry be added to the changelog?

I'm not sure. Nothing was broken. It was just a `DeprecationWarning`.

> * Can it be covered with a test?

It is already covered. That's why it cropped up. I was working on something else and the test fired on CI.

dean0x7d:
OK, cool! You might still want to mention it in the changelog as something like "Resolved `DeprecationWarning` with `tifffile` >= 2021.7.2". So that users can know about it if they run into the same thing.

dean0x7d (approved):

PR #160, opened 2021-07-15
kymotracker: algorithm parameter units

JoepVanlier:
**Why this PR?**
We recently switched over Kymographs in a way that they can hold calibrations.

In addition, the algorithms now return traces in real units. However, the algorithm parameters were not using these physical units (and still had to be specified in pixels). This PR mitigates this.

I would argue that this should be part of the `v0.9` release to minimize the number of breaking changes in versions after this.

**Notes**
- Note that a separate ticket exists to actually put these units on the sliders. It is out of scope for this PR.
- Plan is to squash this into a single commit when merging.

Docs for your reading pleasure: https://lumicks-pylake.readthedocs.io/en/kymo_units/
Changes are in the Cas9 example and the kymotracking tutorial.

### DIFF
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -46,7 +46,7 @@ use the region from 7 to 22 micrometer. Let's have a look at what that looks lik
     >>> kymo.plot_green(aspect="auto", vmax=5)
     >>> plt.ylim([22, 7])
 
-If we zoom in a bit, we can see that our traces in this image are about 3 pixels wide, so let's set this as our
+If we zoom in a bit, we can see that our traces in this image are about 0.3 microns wide, so let's set this as our
 `line_width`. Note that we invoked `plt.colorbar()` to add a little color legend here.
 
 .. image:: kymo_zoom.png
@@ -58,12 +58,12 @@ Considering that we only want to run the kymotracker on part of the image, we al
 In this case, we track particles between 0 and 350 seconds, and 7 and 22 micron.
 Running the algorithm is easy using the function :func:`~lumicks.pylake.track_greedy`::
 
-    traces = lk.track_greedy(kymo, "green", line_width=3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
+    traces = lk.track_greedy(kymo, "green", line_width=0.3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
 
 The result of tracking is a list of kymograph traces::
 
     >>> print(len(traces))  # the number of traces found in the kymo
-    7411
+    7417
 
 Sometimes, we can have very short spurious traces. To remove these from the list of detected traces we can use
 :func:`~lumicks.pylake.filter_lines`. To omit all traces with fewer than 4 detected points, we
@@ -122,7 +122,7 @@ by invoking the following command::
 
 You can optionally also pass algorithm parameters when opening the widget::
 
-    KymoWidgetGreedy(kymo, "green", axis_aspect_ratio=2, min_length=4, pixel_threshold=3, window=6, sigma=1.4)
+    KymoWidgetGreedy(kymo, "green", axis_aspect_ratio=2, min_length=4, pixel_threshold=3, window=6, sigma=0.14)
 
 Traced lines are accessible through the `.lines` property::
 
@@ -138,7 +138,7 @@ The second algorithm present is an algorithm that works purely on signal derivat
 the image, and then performing sub-pixel accurate line detection. It can be a bit more robust to low signal levels,
 but is generally less temporally and spatially accurate due to the blurring involved::
 
-    traces = lk.track_lines(kymo, "green", line_width=3, max_lines=50)
+    traces = lk.track_lines(kymo, "green", line_width=0.3, max_lines=50)
 
 The interface is mostly the same, aside from an extra required parameter named `max_lines` which indicates the maximum
 number of lines we want to detect.
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -51,7 +51,7 @@ def track_greedy(
     channel : str
         Kymograph channel.
     line_width : float
-        Expected line width in pixels.
+        Expected line width in physical units.
     pixel_threshold : float
         Intensity threshold for the pixels. Local maxima above this intensity level will be
         designated as a line origin.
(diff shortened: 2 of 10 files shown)


docs/tutorial/kymotracking.rst
```diff
@@ -58,12 +58,12 @@ Considering that we only want to run the kymotracker on part of the image, we al
...
 Running the algorithm is easy using the function :func:`~lumicks.pylake.track_greedy`::
 
-    traces = lk.track_greedy(kymo, "green", line_width=3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
+    traces = lk.track_greedy(kymo, "green", line_width=0.3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
 
 The result of tracking is a list of kymograph traces::
 
     >>> print(len(traces))  # the number of traces found in the kymo
-    7411
+    7417
```
JoepVanlier, line 66: Note that this was already `7417` on the master branch, it was incorrect in the docs.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -107,11 +109,17 @@ def track_greedy(
...
+    velocity_scale = position_scale / kymograph_channel.line_time_seconds
+    diffusion_scale = position_scale ** 2 / kymograph_channel.line_time_seconds
     lines = points_to_line_segments(
         peaks,
-        kymo_score(vel=vel, sigma=sigma, diffusion=diffusion),
+        kymo_score(
+            vel=vel / velocity_scale,
+            sigma=sigma / position_scale if sigma else 0.5 * line_width_pixels,
+            diffusion=diffusion / diffusion_scale,
+        ),
```
rpauszek, line 122: I wonder if there should be a small comment here. It's not super apparent from first glance at the code what exactly is going on here. Of course one could reason it all through to get to what the units are for everything, but then I would argue it'd be helpful to just state it upfront to make our lives easier in the future

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -318,7 +323,7 @@ def set_show_lines(show_lines):
                 disabled=False,
                 min=1,
                 max=10,
-                tooltip="Lines below this length are filtered out",
+                tooltip="Minimum number of frames a spot has to be detected in to be considered",
```
rpauszek, line 326: 👍  I like this

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -544,17 +552,29 @@ def create_algorithm_sliders(self):
...
+        vel_calibration = position_scale / calibrated_kymo_channel.line_time_seconds
+        vel_slider = self._add_slider(
+            "Velocity",
+            "vel",
+            "How fast does the particle move?",
+            minimum=-5.0 * vel_calibration,
+            maximum=5.0 * vel_calibration,
+            step_size=0.001 * vel_calibration,
+            slider_type=ipywidgets.FloatSlider,
+        )
```
rpauszek, line 577: is this a brand new widget component? if so maybe should be mentioned in the changelog?

rpauszek (approved):
Cool, I think it'll be nice to have consistency in what's being presented to the user in plots and the actual argument units. Just a couple things on some additional commenting and docs if you agree it'd be beneficial; otherwise looks good

lumicks/pylake/nb_widgets/tests/test_kymotracker_widget.py
```diff
@@ -123,16 +134,23 @@ def test_stitch(kymograph, mockevent):
 
     assert len(kymo_widget.lines) == 2
 
+    def kymo_units(time_idx, coord_idx):
+        return kymograph.line_time_seconds * time_idx, kymograph.pixelsize_um[0] * coord_idx
```
lpingas, line 138: I find this confusing since `kymo_units` in previous test cases had a different order of parameters. Unfortunately, I don't have good suggestions...

Maybe rename it to something like `kymo_time_position`? Or break it into 2 functions and use those (`in_s` and `in_um`).

lumicks/pylake/nb_widgets/kymotracker_widgets.py
lpingas, line 326: It would be great if we also added this change to the docstrings of the public API.

lpingas (approved):
LGTM! I left some small suggestions, but none is critical.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Yeah, that's a fair point. I've moved it to its own section for clarity. The variable names should make it more clear what's happening.

```
    # Convert algorithm parameters to pixel units
    velocity_pixels = vel * kymograph_channel.line_time_seconds / position_scale
    diffusion_pixels = diffusion / (position_scale ** 2 / kymograph_channel.line_time_seconds)
    sigma_pixels = sigma / position_scale if sigma else 0.5 * line_width_pixels
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -463,7 +468,7 @@ def __init__(
             Desired aspect ratio of the viewport. Sometimes kymographs can be very long and thin. This helps you
             visualize them anyway.
         line_width : float
-            Expected width of the particles in pixels.
```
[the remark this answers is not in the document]
JoepVanlier: Added.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: This is done by a separate function named `filter_lines`. I added it there.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -51,7 +51,7 @@ def track_greedy(
     channel : str
         Kymograph channel.
     line_width : float
-        Expected line width in pixels.
```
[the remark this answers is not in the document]
rpauszek: for these, my only worry is that in the future it's likely we will support units other than microns for the position axes (base pairs is already planned and on the roadmap)

Might it be better then to keep it as `line_width` and just make sure that any variable that doesn't deal with physical units to be named with `_pixels` appended for clarity?

lumicks/pylake/nb_widgets/tests/test_kymotracker_widget.py
JoepVanlier: Good point! Decided to go with the following:

`in_um, in_s = calibrate_to_kymo(kymograph)`

And then using those.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: It is. It was found during WIT that for tubules this parameter is often quite critical. And you are right. I'll spin it into its own commit with a changelog entry.

lumicks/pylake/kymotracker/kymotracker.py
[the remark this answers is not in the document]
JoepVanlier: How about `in physical units` for all of these?

The only one that isn't is `window`. The name `window_pixels` sounded strange to me. If I had to name it from scratch, I would've probably gone with: `gap_duration_pixels`.

What name do you think is best?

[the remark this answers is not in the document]
lpingas: To clarify, we should keep it as `line_width`, and I suggest complementing the docstring.

`in physical units` is great!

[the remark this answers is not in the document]
JoepVanlier: 👍

PR #161, opened 2021-07-15
fdfit: Allow simulating the wlc model without fitting

JoepVanlier:
**Why this PR?**
It's a popular feature request to simulate a force-distance model without fitting it first.

With relatively minor modifications, the call operator can be made to take both dictionaries with parameter values and `Parameters`.

**Things to look out for**
Note that this PR changes the `keys` attribute of `Parameters` to a method. I believe having it as a property initially was a mistake, since this is not how the default dictionary behaves.

Before:
```python
import lumicks.pylake as lk
from lumicks.pylake.fitting.parameters import Params, Parameter  # These were not public by default, so users would have to know already what to do
dna = lk.inverted_odijk("DNA")
distances = [.1, .2, .3]
dna(distances, Params(**{"DNA/Lp": Parameter(50.0), "DNA/Lc": Parameter(16.0), "DNA/St": Parameter(1500.0), "kT": Parameter(4.11)}))
```

After:
```python
import lumicks.pylake as lk
dna = lk.inverted_odijk("DNA")
distances = [.1, .2, .3]
dna(distances, {"DNA/Lp": 50.0, "DNA/Lc": 16.0, "DNA/St": 1500.0, "kT": 4.11})
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 * Added Mean Square Displacement (MSD) and diffusion constant estimation to `KymoLine`. For more information, please refer to [kymotracking](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html)
 * Added `FdCurve.with_baseline_corrected_x()` to return a baseline corrected version of the FD curve if the corrected data is available. **Note: currently the baseline is only calculated for the x-component of the force channel in Bluelake. Therefore baseline corrected `FdCurve` instances use only the x-component of the force channel, unlike default `FdCurve`s which use the full magnitude of the force channel by default.**
 * Added ability to perform arithmetic on `Slice` (e.g. `(f.force1x - f.force2x) / 2`). For more information see [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#exporting-h5-files) for more information.
+* Allow simulating force model with a custom set of parameters (see tutorial section [Fd Fitting](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/fdfitting.html)).
 * Added a slider to set the algorithm parameter `velocity` to the kymotracker widget.
 
 #### Bug fixes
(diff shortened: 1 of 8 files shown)


aafkevandenberg (approved):
Thanks Joep, that will make analysis a lot easier!

changelog.md
```diff
@@ -42,6 +42,7 @@
   See the [Cas9 kymotracking example](https://lumicks-pylake.readthedocs.io/en/latest/examples/cas9_kymotracking/cas9_kymotracking.html) or the [kymotracking tutorial](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html) for more information.
 * `KymoLineGroup.save()` and `KymoWidgetGreedy.save_lines()` no longer take `dx` and `dt` arguments.
   Instead, the correct time and position calibration is now passed automatically to these functions. See [kymographs](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
+  * `Parameters.keys()` is now a member function instead of a property (used to be invoked as `parameter.keys`) to be consistent with dictionary.
```
mikhas, line 46: Put that changelog entry into the commit message as well?

docs/tutorial/fdfitting.rst
```diff
@@ -53,6 +53,16 @@ We can also obtain the parameters as a list::
...
+Simulating the model
+--------------------
+
+    We can simulate the model by passing a dictionary with parameters values::
+
+    dna = lk.odijk("DNA")
+    force = np.arange(0.1, 14, 0.1)
+    dna(force, {"DNA/Lp": 50.0, "DNA/Lc": 16.0, "DNA/St": 1500.0, "kT": 4.11})
+
+    Note how the model name is prefixed to the model specific parameters.
```
mikhas, line 65: "..., forming a key-value pair like so: `"name/parameter": value`." Doesn't hurt to add that, especially for a tutorial.

mikhas (approved):
LGTM!

changelog.md
JoepVanlier: Sure thing!

docs/tutorial/fdfitting.rst
JoepVanlier: Yeah, that's better and more explicit. Added it. Thanks!

PR #162, opened 2021-07-20
force calibration: implement exclusion of specific frequency ranges

JoepVanlier:
**Why this PR?**
For active force calibration it has to be possible to exclude specific regions from the fit. This PR adds that functionality to `PowerSpectrum`.

Typically what happens in an active force calibration procedure is that you'd have a driving input frequency (a spike in the frequency spectrum) that needs to be excluded prior to fitting the thermal part of the spectrum.

In addition, this functionality can be used to reject noise peaks that may happen at various frequencies.

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -41,9 +41,10 @@ This function returns a power_spectrum which we can plot::
 .. image:: force_calibration_blocked_spectrum.png
 
 Note that the computation of the power spectrum involves some downsampling.
-If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
 
-    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000)
+Additional parameters can be specified to change the amount of downsampling applied (`num_points_per_block`), the range over which to compute the spectrum (`fit_range`) or to exclude specific frequency ranges from the spectrum (`excluded_ranges`)::
+
+    power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=[(700, 800), (14500, 14600)])
 
 To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
 This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
--- a/lumicks/pylake/force_calibration/power_spectrum.py
+++ b/lumicks/pylake/force_calibration/power_spectrum.py
@@ -53,6 +53,27 @@ class PowerSpectrum:
 
         return ba
 
+    def _exclude_range(self, excluded_ranges):
+        """Exclude given frequency ranges from the power spectrum.
+
+        This function can be used to exclude noise peaks.
+
+        Parameters
+        ----------
+        excluded_ranges : list of tuple of float
+            List of ranges to exclude specified as a list of (frequency_min, frequency_max)."""
+        if not excluded_ranges:
+            return copy(self)
+
+        ps = copy(self)
+        indices = np.logical_and.reduce(
+            [(ps.frequency < f_min) | (ps.frequency >= f_max) for f_min, f_max in excluded_ranges]
+        )
+
+        ps.frequency = ps.frequency[indices]
+        ps.power = ps.power[indices]
+        return ps
+
     def in_range(self, frequency_min, frequency_max):
         """Returns part of the power spectrum within a given frequency range."""
         ir = copy(self)
--- a/lumicks/pylake/force_calibration/power_spectrum_calibration.py
+++ b/lumicks/pylake/force_calibration/power_spectrum_calibration.py
@@ -130,7 +130,9 @@ def guess_f_diode_initial_value(ps, guess_fc, guess_D):
         return 2 * f_nyquist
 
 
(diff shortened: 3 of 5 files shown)


lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -53,6 +53,27 @@ def downsampled_by(self, factor, reduce=np.mean):
...
+        """Exclude given frequency ranges from the power spectrum.
+
+        This function can be used to exclude noise peaks.
+
+        Parameters
+        ----------
+        exclude_ranges - `list` [`tuple` (float, float)]
+            List of ranges to exclude specified as a list of (frequency_min, frequency_max)."""
+        if not excluded_ranges:
+            return deepcopy(self)
```
dean0x7d, line 66: Is a `deepcopy` needed here? Wouldn't a shallow copy be enough? It feels like this would create a copy every time for the most common case of no excluded ranges.

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -53,6 +53,27 @@ def downsampled_by(self, factor, reduce=np.mean):
...
         return ba
 
+    def _exclude_range(self, excluded_ranges):
+        """Exclude given frequency ranges from the power spectrum.
+
+        This function can be used to exclude noise peaks.
+
+        Parameters
+        ----------
+        exclude_ranges - `list` [`tuple` (float, float)]
```
dean0x7d, line 63: Shouldn't this be `List[Tuple[float, float]]` in PEP484 notation?

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -153,6 +155,8 @@ def calculate_power_spectrum(data, sample_rate, fit_range=(1e2, 23e3), num_point
     num_points_per_block : int, optional
         The spectrum is first block averaged by this number of points per block.
         Default: 2000.
+    excluded_ranges : list [tuple (float, float)]
```
dean0x7d, line 158: Same here.

lumicks/pylake/force_calibration/tests/test_exclusions.py
```diff
@@ -0,0 +1,26 @@
...
+        [[], [1, 3, 5, 7, 9, 11], [10, 30, 50, 70, 90, 110]],  # Empty exclusion range
+        [[(3, 5)], [1, 5, 7, 9, 11], [10, 50, 70, 90, 110]],
+        [[(3, 6)], [1, 7, 9, 11], [10, 70, 90, 110]],
+        [[(3, 6), (3, 6)], [1, 7, 9, 11], [10, 70, 90, 110]],  # Same exclusion range twice
+        [[(5, 6), (3, 8)], [1, 9, 11], [10, 90, 110]],  # Overlapping range
+        [[(3, 8), (5, 6)], [1, 9, 11], [10, 90, 110]],  # Overlapping range (different order)
+        [[(3, 6), (9, 10)], [1, 7, 11], [10, 70, 110]],  # Disjoint range
+    ],
+)
+def test_exclusions(exclusion_ranges, result_frequency, result_power):
```
dean0x7d, line 18: I would sort of expect this test to appear in `test_power_spectrum.py` since it's part of the `PowerSpectrum` class.

dean0x7d (approved):
Looks good! Just:
* Please double-check the `deepcopy()`.
* Include the PR description in the commit message body. It's good to know why these changes/features are being made when look through the history.

docs/tutorial/force_calibration.rst
```diff
@@ -41,9 +41,10 @@ This function returns a power_spectrum which we can plot::
 .. image:: force_calibration_blocked_spectrum.png
 
 Note that the computation of the power spectrum involves some downsampling.
-If you wish to change the amount of downsampling applied or which range to compute the spectrum for, you can specify these as additional arguments::
 
-    lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000)
+Additional parameters can be specified to change the amount of downsampling applied (`num_points_per_block`), the range over which to compute the spectrum (`fit_range`) or to exclude specific frequency ranges from the spectrum `excluded_ranges`.
+
+    power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=[(700, 800), (14500, 14600)])
```
mcsmurf, line 47:
```suggestion
power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=[(700, 800), (14500, 14600)])
```

lumicks/pylake/force_calibration/tests/test_exclusions.py
JoepVanlier: Yeah, good point, I'll move it.

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: Good point, I'll make it a regular `copy`. It does seem overly pessimistic to always `deepcopy` it. If it was public API, there'd be more of a case for it, but it's not.

docs/tutorial/force_calibration.rst
JoepVanlier: The indent is needed for sphinx to detect it as code.

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: Actually, considering that we're using the friendlier numpydoc elsewhere, shouldn't we stick to that? So:
`list of tuple of float`. I always think that's a bit easier for new users to read than PEP484.

PR #163, opened 2021-07-21
force_calibration: add windowing to PowerSpectrum

JoepVanlier:
**Why this PR?**
For active calibration, we need to be able to average spectra of subdivisions of the data. The use case for this is that you want to average multiples of the period of the driving frequency such that the driving peak ends up in a single data point. That way, you can get a power estimate using just that bin. The test specifically tests for this.

See below how much of an error you can make when you don't ensure that.

image

### DIFF
--- a/lumicks/pylake/force_calibration/power_spectrum.py
+++ b/lumicks/pylake/force_calibration/power_spectrum.py
@@ -1,3 +1,4 @@
+import warnings
 import numpy as np
 import matplotlib.pyplot as plt
 from copy import copy
@@ -19,7 +20,7 @@ class PowerSpectrum:
         The total duration of the original data. [seconds]
     """
 
-    def __init__(self, data, sample_rate, unit="V"):
+    def __init__(self, data, sample_rate, unit="V", window_seconds=None):
         """Power spectrum
 
         Parameters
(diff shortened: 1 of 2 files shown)


lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -19,30 +20,52 @@ class PowerSpectrum:
...
         data = data - np.mean(data)
 
-        # Calculate power spectrum.
-        fft = np.fft.rfft(data)
-        self.frequency = np.fft.rfftfreq(data.size, 1.0 / sample_rate)
-        self.power = (2.0 / sample_rate) * np.square(np.abs(fft)) / data.size
+        # Calculate power spectrum for slices of data.
+        num_points_per_window = (
+            int(np.round(window_seconds * sample_rate)) if window_seconds else len(data)
+        )
```
rpauszek, line 49: should we check for negative values in `window_seconds`?

rpauszek (approved):
LGTM

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: Added!

PR #164, opened 2021-07-23
kymo: fix `plot_with_force` for kymos with an incomplete last line

JoepVanlier:
**Why this PR?**
`plot_with_force` always fails for kymographs with an incomplete last line.

The reason for this is that the last few timestamps on the last line have zero as their timestamp. This meant it was trying to downsample a range from X to 0, which made the downsampler think that there was no overlap between the kymograph and the force channel (as it checks the last timestamp of the ranges to downsample to against the first one of the channel to downsample).

I've also added a regression test for it.

There's one other issue I have with this function, which is that there's no way to have it use the LF force and that the string input for the channel is ambiguous (passing `downsampled_force1x` for channel is accepted, but it's not what would be used as a source). From the docstring it was not clear what the function expects (name of the force channel is ambiguous).

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -149,6 +149,11 @@ class Kymo(ConfocalImage):
         # downsample exactly over the scanline time range
         min_times = self.timestamps[0, :].astype(np.int64)
         max_times = self.timestamps[-1, :].astype(np.int64)
+
+        # If the last line is incomplete, drop it!
+        if max_times[-1] == 0:
+            min_times, max_times = min_times[:-1], max_times[:-1]
+
         time_ranges = [(mini, maxi) for mini, maxi in zip(min_times, max_times)]
         return force.downsampled_over(time_ranges, reduce=reduce, where="center")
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymo.py
```diff
@@ -149,6 +149,11 @@ def _downsample_channel(self, n, xy, reduce=np.mean):
         # downsample exactly over the scanline time range
         min_times = self.timestamps[0, :].astype(np.int64)
         max_times = self.timestamps[-1, :].astype(np.int64)
+
+        # If the last line is incomplete, drop it!
+        if max_times[-1] == 0:
```
mikhas, line 154: Does that mean `0` cannot be a naturally occurring value?

mikhas (approved):
LGTM!

lumicks/pylake/kymo.py
JoepVanlier: Yeah. Zero would be January 1st 1970, since they're nanoseconds since epoch :)

rpauszek (approved):
LGTM

PR #165, opened 2021-07-27
Simplify force calibration API

JoepVanlier:
**Why?**
To make integration with Bluelake easier and have a clearer separation between inputs and outputs of the calibration routine.

PR #166, opened 2021-07-29
Release v0.9

JoepVanlier:
**Why this PR?**
We would like to release v0.9. This PR makes a few small additional amendments to the changelog for clarity.

This release introduces a fair number of breaking changes. All of them are motivated in the changelog.

Note that this PR depends on #165 since that PR introduces some breaking changes that would be good to include in this release.

The dependency changes since last release are:
- `numpy` minimum version was bumped from `1.14` to `1.20`.
- `scikit-image>=0.17.2` was added
- Support for `Python` `3.6` was dropped (in favor of `DataClasses`).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,15 +1,21 @@
 # Changelog
 
-## v0.9.0 | t.b.d.
+## v0.9.0 | 2021-07-29
+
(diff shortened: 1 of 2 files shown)


mikhas (approved):
Nice to see someone taking so much care regarding the changelog!
I still think you might need an exec summary detailing the breaking changes (in API), and a brief list of new API added. Just so that the busy folks immediately see that this is a _good_ release ;)

dean0x7d (approved):
Gigantic release! Looks awesome!

mcsmurf (approved):
Great work has been done! Go for launch 🚀

PR #169, opened 2021-07-30
docs: correct error in equation used for force calibration

JoepVanlier:
**Why?**
Unfortunately, one equation in the documentation didn't get updated appropriately when we made the switch from `0.5 V^2/Hz` to `V^2/Hz`. This PR fixes that.


rpauszek (approved):
👍

PR #171, opened 2021-08-02
Make `plot_with_force` sum over entire kymo line.

JoepVanlier:
**Why this PR?**
`plot_with_force` was not summing over the full line range, but from the center of the first pixel to the center of the last pixel. It also had problems with an incomplete last pixel (always resulting in an exception). This PR changes it so that it is exact and fixes the last pixel issue.

**What to look out for**
To achieve this, I generalized `_timestamps` that allows us to provide a custom reduce function (which will also properly get cached). The only downside of this mini refactor is that it means that we cannot specify a custom downsampler for timestamps in `kymo.downsampled_by()` anymore.

Technically, that makes this a breaking change, though a breakage that I don't expect would affect many people.

The alternative is adding another factory function that is essentially mostly a copy of `_timestamps`, but with the reduce option. We could consider this to avoid the breaking change now (deprecating the reduce option first with the plan of moving to this model once we go to another major release). Personally, I'd rather just break it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,6 +1,6 @@
 # Changelog
 
-## v0.9.1 | t.b.d.
+## v0.10.0 | t.b.d.
 
 #### New features
 
@@ -10,6 +10,11 @@
 #### Bug fixes
 
 * Fixed an error in the documentation. In v0.9.0 `PowerSpectrum.power` was changed to represent power in `V^2/Hz` instead of `0.5 V^2/Hz`. However, the docs were not appropriately updated to reflect this change in the model equation that's fitted to the spectrum. This is mitigated now.
+* Fixed bug in `plot_with_force` which caused an exception on Kymographs with a partial last pixel.
+
+#### Breaking changes
+
+* Removed the option to specify a custom `reduce_timestamps` function in `kymo.downsampled_by`. The reason for the removal is that by downsampling repeatedly with different `reduce` functions for the timestamps, the data can end up in an inconsistent state. Additionally, the timestamps of the original object (read directly from the h5 file) are defined as the mean of the pixel timestamps; this definition is now consistent regardless of downsampling state.
 
 ## v0.9.0 | 2021-07-29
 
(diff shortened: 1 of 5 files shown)


mikhas:
If this is another breaking change, would that potentially risk delaying the release to follow 0.9? Or is that case where merging this quickly and having a point release (0.9.x) would be the best way forward, assuming that the new 0.9 release hasn't affected too many users yet?

JoepVanlier:
> If this is another breaking change, would that potentially risk delaying the release that is to follow 0.9? Or is that case where merging this quickly and having a point release (0.9.x) would be the best way forward, assuming that the new 0.9 release hasn't affected too many users yet?

I was wondering the same thing. I mean, we can just call it v0.10.0. We didn't change any dependencies or anything like that, so upgrading from v0.9 to v0.10 should be no problem for users.

I also don't see why anything would be delayed this time. Last time we had a lot of new breaking features that depended on each other. This time it's just this one thing (for now).

mikhas:
The discussion of possible alternatives feels heavy; many ways to skin a cat indeed. Isn't the first step then, before even reviewing this PR, to figure out which alternative is the best?

JoepVanlier:
> The discussion of possible alternatives feels heavy; many ways to skin a cat indeed. Isn't the first step then, before even reviewing this PR, to figure out which alternative is the best?

I would say the behaviour in this PR is the correct behaviour we want.

JoepVanlier:
@mikhas After our discussion I added an extra test which tests the interaction with downsampling.

mikhas (approved):
Actually I am fine with this now. Since this is a breaking change now we could consider merging other stuff first, have a 0.9.1 release, and only then merge this? Otherwise this PR, once merged, will make it harder to have a non-breaking release in between 0.9 and 0.x

rpauszek (approved):
I think it looks good. I'm ok with taking away the option to pass a custom `reduce` for downsampling, since we've never given an option for the reduce in the original timestamps construction. In  a way, I think this makes more sense so everything stays consistent

PR #174, opened 2021-08-05
Force calibration: Active calibration scaffold model

JoepVanlier:
**Why this PR?**
Because we want to have active calibration functionality in Pylake.

**What's in this PR?**
This PR contains the basic scaffold for active calibration for an ideal spectrum.

The suggested API looks something like this:

```python
    model = ActiveCalibrationModel(
        driving_data,
        force_voltage_data,
        sample_rate,
        bead_diameter,
        driving_frequency_guess,
        viscosity,
        temperature,
    )

    power_spectrum = calculate_power_spectrum(force_voltage_data, sample_rate)
    fit = fit_power_spectrum(power_spectrum, model)

    fit.plot()
    fit
```

**Out of scope, but good to keep in mind when considering architecture**
Documentation is out of scope, since this functionality is not yet public (we deliberately do not import it yet).

This PR does not cover:
- Hydrodynamic correction.
- Bias correction.
- Fixing specific parameters in the thermal calibration.

Which all affect passive and active calibration and will be part of future PRs in this vein.

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -1,5 +1,9 @@
-import math
+import numpy as np
 import scipy
+from .detail.driving_input import (
+    estimate_driving_input_parameters,
+    driving_power_peak,
+)
 from .detail.power_models import passive_power_spectrum_model, sphere_friction_coefficient
 from .power_spectrum_calibration import CalibrationParameter
 
@@ -11,8 +15,8 @@ class CalibrationModel:
     def __name__(self):
         raise NotImplementedError
 
-    def __call__(self):
-        raise NotImplementedError
+    def __call__(self, f, fc, diffusion_constant, f_diode, alpha):
+        return passive_power_spectrum_model(f, fc, diffusion_constant, f_diode, alpha)
 
     def __repr__(self):
         return f"{self.__name__()}({''.join([f'{k}={v}, ' for k, v in vars(self).items()])[:-2]})"
@@ -20,7 +24,7 @@ class CalibrationModel:
     def calibration_parameters(self):
         raise NotImplementedError
 
-    def calibration_results(self, fc, diffusion_constant):
+    def calibration_results(self, fc, diffusion_constant_volts, f_diode, alpha):
         raise NotImplementedError
 
 
@@ -61,9 +65,6 @@ class PassiveCalibrationModel(CalibrationModel):
         self.viscosity = viscosity
         self.temperature = temperature
 
-    def __call__(self, f, fc, diffusion_constant, f_diode, alpha):
-        return passive_power_spectrum_model(f, fc, diffusion_constant, f_diode, alpha)
-
     def calibration_parameters(self):
         return {
             "Bead diameter": CalibrationParameter("Bead diameter", self.bead_diameter, "um"),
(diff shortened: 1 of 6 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+            "Driving frequency (guess)": CalibrationParameter(
+                "Driving frequency (guess)", self.driving_frequency_guess, "Hz"
+            ),
+            "Sample rate": CalibrationParameter("Sample rate", self.sample_rate, "Hz"),
+            "Temperature": CalibrationParameter("Liquid temperature", self.temperature, "C"),
+            "Viscosity": CalibrationParameter("Liquid viscosity", self.viscosity, "Pa*s"),
+            "num_windows": CalibrationParameter("Number of averaged windows", self.num_windows, ""),
+        }
+
+    def _power_theory(self, f_corner):
```
tiagobonetti, line 200: May I suggest: `_ideal_power_spectrum`?

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+        reference_peak = self(self.driving_frequency, fc, diffusion_constant_volts, f_diode, alpha)
+
+        power_exp = (self.response_power_density - reference_peak) * self.frequency_resolution
+        distance_response = np.sqrt(self._power_theory(fc) / power_exp)  # m/V
+
+        omega = 2.0 * np.pi * fc
+        temperature_kelvin = scipy.constants.convert_temperature(self.temperature, "C", "K")
+        k_temperature = scipy.constants.k * temperature_kelvin
+        gamma_0 = k_temperature / (distance_response ** 2 * diffusion_constant_volts)  # kg/s^2
+        kappa = omega * gamma_0  # N/m
```
rpauszek, line 227: I think `omega` is superfluous here; considering it's only used in the assignment for `kappa` which left me searching through equations to figure out what special meaning this variable has. 

I think it'd be better to just have `kappa = 2.0 * np.pi * fc * gamma_0`. This also keeps it consistent with the definition in `PassiveCalibrationModel.calibration_results()`

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -1,5 +1,7 @@
 import math
```
rpauszek, line 1: the only place this package is used is in `PassiveCalibrationModel.calibration_results()` whereas in the `ActiveCalibrationModel` version of the same method you use numpy's implementation. My vote would be to get rid of this import and use numpy throughout

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -61,32 +63,33 @@ def __init__(self, bead_diameter, viscosity=1.002e-3, temperature=20):
...
         fc : float
             Corner frequency, in Hz.
-        diffusion_constant : float
-            Diffusion constant, in (a.u.)^2/s
+        diffusion_constant_volts : float
+            Diffusion constant, in V^2/s
+        f_diode : float
+            Diode frequency.
+        alpha : Alpha
+            Fraction of PSD signal that is instantaneous
```
rpauszek, line 85: Personally I'm ok with requiring `f_diode` and `alpha` here because models can be called with the same signature regardless of their actual type. Might be worthwhile to add a small note in the docstring though that they're not needed for passive calibration (that it's not an oversight that they're never used)

lumicks/pylake/force_calibration/detail/driving_input.py
```diff
@@ -98,3 +99,16 @@ def estimate_driving_input_parameters(
     amp = np.exp(p[2] - 0.25 * p[1] ** 2 / p[0] + 0.5 * np.log(-np.pi / p[0])) * delta_freq
 
     return amp, freq
+
+
+def response_power_density(psd_data, sample_rate, driving_frequency, num_windows, freq_window=50.0):
```
tiagobonetti, line 104: I think `response_power_density` reads a bit unclear when used.
The docstring is pretty clear, but I can think of any good suggestions...

rpauszek (commented):
Looks really good. Just some comments about clarity with variable names and docstrings. Considering the complexity of this analysis (with even more to come soon) I'd rather have overly explicit documentation

lumicks/pylake/force_calibration/tests/test_active_calibration.py
```diff
@@ -0,0 +1,80 @@
+import numpy as np
+import scipy.constants
+import pytest
+from lumicks.pylake.force_calibration.power_spectrum_calibration import (
+    calculate_power_spectrum,
+    fit_power_spectrum,
+)
+from lumicks.pylake.force_calibration.calibration_models import ActiveCalibrationModel
+from .data.simulate_calibration_data import generate_test_realisation_ideal
```
tiagobonetti, line 9: `generate_test_realisation_ideal` is such a strange name. I guess `realisation` has a specific meaning I'm unaware of?

lumicks/pylake/force_calibration/tests/test_active_calibration.py
```diff
@@ -0,0 +1,80 @@
...
+
+    np.testing.assert_allclose(fit["kappa"].value, stiffness, rtol=5e-2)
+    np.testing.assert_allclose(fit["alpha"].value, diode[0], rtol=5e-2)
+    np.testing.assert_allclose(fit["f_diode"].value, diode[1], rtol=5e-2)
+    np.testing.assert_allclose(fit["Rd"].value, pos_response_um_volt, rtol=5e-2)
+
+    response_calc = fit["Rd"].value * fit["kappa"].value * 1e3
+    np.testing.assert_allclose(fit["Rf"].value, response_calc, rtol=1e-9)
+
+    kt = scipy.constants.k * (temperature + 273.15)
```
tiagobonetti, line 72:
```suggestion
    kt = scipy.constants.k * scipy.constants.convert_temperature(temperature, "C", "K")
```

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -11,16 +13,16 @@ def __init__(self, bead_diameter):
     def __name__(self):
         raise NotImplementedError
 
-    def __call__(self):
-        raise NotImplementedError
+    def __call__(self, f, fc, diffusion_constant, f_diode, alpha):
+        return passive_power_spectrum_model(f, fc, diffusion_constant, f_diode, alpha)
```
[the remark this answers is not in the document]
JoepVanlier: What do you mean by temporary thing?

This call is used by the spectral fitter to fit the thermal calibration model. The thermal calibration part is the same for active and passive calibration.  Eventually, there are quite a few extra flags going into that model. That'd mean that the amount of duplication would increase over time. I'd prefer to have it here, because eventually, you'd end up duplicating more than just this.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+        driving_frequency_guess : float
+            Guess of the driving frequency.
+        viscosity : float, optional
+            Liquid viscosity [Pa*s].
+        temperature : float, optional
+            Liquid temperature [Celsius].
+        num_windows : int, optional
+            Number of windows to average for the uncalibrated force. Using a larger number of
+            windows potentially increases the bleed, but may be useful when the SNR is low.
+        """
```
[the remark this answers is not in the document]
JoepVanlier: Hm, but then some of the input args wouldn't be documented anymore (since they don't end up as attributes). I'll add the attributes in the top docstring though.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+        self.num_windows = num_windows
+
+        # Estimate driving input and response
+        amplitude_um, self.driving_frequency = estimate_driving_input_parameters(
+            sample_rate, driving_data, driving_frequency_guess
+        )
+        self.driving_amplitude = amplitude_um * 1e-6
+        self.response_power_density, self.frequency_resolution = response_power_density(
+            force_voltage_data, sample_rate, self.driving_frequency, num_windows
+        )
```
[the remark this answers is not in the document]
JoepVanlier: It's a good point.

While documenting it, I actually realized that these properties should probably remain private (at least for now), so I've made them private. Nevertheless, it seems they need some additional documentation.

Power density is a density. Power is the power density integrated over the frequency axis.

The maximum of the peak here is indeed a density (Volt^2 / Hz). To get the actual power you have to integrate it over the bin still. I've also renamed `frequency_resolution` to `__frequency_bin_width` since it is more descriptive. Have a look whether you think the new names plus their docstrings work for you.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+            ),
+            "Sample rate": CalibrationParameter("Sample rate", self.sample_rate, "Hz"),
+            "Temperature": CalibrationParameter("Liquid temperature", self.temperature, "C"),
+            "Viscosity": CalibrationParameter("Liquid viscosity", self.viscosity, "Pa*s"),
+            "num_windows": CalibrationParameter("Number of averaged windows", self.num_windows, ""),
+        }
+
+    def _power_theory(self, f_corner):
+        """Compute the power spectral density expected for a given driving input."""
+        return self.driving_amplitude ** 2 / (2 * (1 + (f_corner / self.driving_frequency) ** 2))
```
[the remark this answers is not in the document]
JoepVanlier: How about `_theoretical_driving_power`, since it's the driving input?

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Yeah, before I simplified stuff there were some longer equations and omega was used in several places (to keep line lengths shorter). It is indeed not needed now. Removed. :)

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Alright.

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: I added a note.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -61,32 +63,33 @@ def __init__(self, bead_diameter, viscosity=1.002e-3, temperature=20):
...
         ----------
         fc : float
             Corner frequency, in Hz.
-        diffusion_constant : float
-            Diffusion constant, in (a.u.)^2/s
+        diffusion_constant_volts : float
+            Diffusion constant, in V^2/s
+        f_diode : float
+            Diode frequency.
+        alpha : Alpha
```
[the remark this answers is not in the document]
JoepVanlier: Heh 😅

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+
+        Parameters
+        ----------
+        fc : float
+            Corner frequency, in Hz.
+        diffusion_constant_volts : float
+            Diffusion constant, in V^2/s
+        f_diode : float
+            Diode frequency.
+        alpha : Alpha
```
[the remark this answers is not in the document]
JoepVanlier: 😅

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: That's not what this is though.

This is absolute power expected from a spike (integral over a delta spike in the power spectrum).

There were two comments on this function, so it's clear it needs some better clarification on my end.

How about:

```python
    def _theoretical_driving_power(self, f_corner):
        """Compute the power expected for a given driving input.
        
        When driving the stage or trap, we expect to see a delta spike in the power density 
        spectrum. This function returns the expected power contribution of the bead motion to the
        power spectrum. It corresponds to the driven power spectrum minus the thermal power spectrum
        integrated over the frequency bin corresponding to the driving input."""
        return self.driving_amplitude ** 2 / (2 * (1 + (f_corner / self.driving_frequency) ** 2))
```

lumicks/pylake/force_calibration/detail/driving_input.py
JoepVanlier: `power_density_at_driving_input_frequency` ?

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
...
+            Fraction of PSD signal that is instantaneous
+        """
+        reference_peak = self(self.driving_frequency, fc, diffusion_constant_volts, f_diode, alpha)
+
+        power_exp = (self.response_power_density - reference_peak) * self.frequency_resolution
+        distance_response = np.sqrt(self._power_theory(fc) / power_exp)  # m/V
+
+        omega = 2.0 * np.pi * fc
+        temperature_kelvin = scipy.constants.convert_temperature(self.temperature, "C", "K")
```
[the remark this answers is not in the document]
JoepVanlier: They should produce identical results. Maybe `convert_temperature` is a little more readable.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -99,3 +102,135 @@ def calibration_results(self, fc, diffusion_constant):
             "kappa": CalibrationParameter("Trap stiffness", kappa, "pN/nm"),
             "Rf": CalibrationParameter("Force response", Rf, "pN/V"),
         }
+
+
+class ActiveCalibrationModel(CalibrationModel):
+    """Model to fit data acquired during passive calibration.
```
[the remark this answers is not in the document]
JoepVanlier: it's funny how you end up reading over things like this as the author :D

lumicks/pylake/force_calibration/detail/driving_input.py
```diff
@@ -101,7 +101,9 @@ def estimate_driving_input_parameters(
     return amp, freq
 
 
-def response_power_density(psd_data, sample_rate, driving_frequency, num_windows, freq_window=50.0):
+def power_density_driving_frequency(
```
rpauszek, line 104: I'm not sure about this new name, it took me quite a while that it should read more like "power density -and- driving frequency" instead of just one thing .... I guess maybe `power_density_and_driving_frequency` ? `input_power_and_frequency`? neither are particularly nice, but I do think it's clearer.

lumicks/pylake/force_calibration/calibration_models.py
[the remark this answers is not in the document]
rpauszek: not a huge fan of all the duplicated documentation, but I do get your point. I think this is better since at least there's consistency between the two model types now

lumicks/pylake/force_calibration/tests/test_active_calibration.py
rpauszek: maybe change it to `realization` 😉

rpauszek (approved):
looks great! just one function name that could be a bit clearer, but I could live with it

lumicks/pylake/force_calibration/calibration_models.py
tiagobonetti: This is perfectly clear now! :+1:

lumicks/pylake/force_calibration/tests/test_active_calibration.py
tiagobonetti: As Dean would say: usually the best place to find a name is in your documentation.

```py
    """Generate test data to test active calibration.
```
So maybe just `generate_active_calibration_test_data`.

Changing this is a bit out-of-scope for this PR so please feel free to move on and revisit it later if you feel is the best option.

tiagobonetti (approved):
`honk_power_density_and_frequency()`

lumicks/pylake/force_calibration/detail/driving_input.py
JoepVanlier: I discussed it with Ray a bit. We decided to go with `driving_power_peak`.

lumicks/pylake/force_calibration/tests/test_active_calibration.py
JoepVanlier: Sounds like a good name. I'll take it 👍

PR #175, opened 2021-08-09
Cropping kymographs.

JoepVanlier:
**Why?**
Users often want to exclude the beads from downstream analysis. Therefore it is useful if they can actually crop the kymo easily.

This API introduces Kymograph cropping.

There is also a plan for updating the Cas9 notebook with this functionality, but I will make a separate PR for that.

Note: This PR depends on https://github.com/lumicks/pylake/pull/171

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -39,11 +39,14 @@ class Kymo(ConfocalImage):
         End point in the relevant info wave.
     json : dict
         Dictionary containing kymograph-specific metadata.
+    position_offset : float
+        Coordinate position offset with respect to the original raw data.
     """
 
-    def __init__(self, name, file, start, stop, json):
+    def __init__(self, name, file, start, stop, json, position_offset=0):
         super().__init__(name, file, start, stop, json)
         self._line_time_factory = _default_line_time_factory
+        self._position_offset = position_offset
 
     def _has_default_factories(self):
         return (
@@ -90,7 +93,7 @@ class Kymo(ConfocalImage):
 
         start = line_timestamps[i_min]
 
-        return Kymo(self.name, self.file, start, stop, self._json)
+        return Kymo(self.name, self.file, start, stop, self._json, self._position_offset)
 
     @cachetools.cachedmethod(lambda self: self._cache)
     def _line_start_timestamps(self):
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymo.py
```diff
@@ -269,6 +272,54 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+        ----------
+        lower : float
+            Lower bound in physical units.
+        upper : float
+            Upper bound in physical units.
+        """
+        lower_pixels = int(lower / self.pixelsize_um[0])
+        upper_pixels = int(np.ceil(upper / self.pixelsize_um[0]))
+        if lower_pixels < 0 or upper_pixels < 0:
+            raise ValueError("Cropping by negative positions not allowed")
```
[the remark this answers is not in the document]
JoepVanlier: Sure. Given that there is no offset in the position we can. I'll swap it around.

lumicks/pylake/tests/test_kymo.py
```diff
@@ -444,3 +444,173 @@ def test_downsampled_slice():
...
+            [0, 12, 12, 12, 0, 6, 0],
+        ],
+        dtype=np.uint8
+    )
+
+    # Test basic functionality
+    kymo = generate_kymo(
+        "Mock", image, pixel_size_nm=2, start=100, dt=5, samples_per_pixel=5, line_padding=2
+    )
+    cropped = kymo.crop_by_distance(4e-3, 8e-3)
```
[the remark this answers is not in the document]
JoepVanlier: Added a test for this :)

lumicks/pylake/kymo.py
```diff
@@ -39,11 +39,14 @@ class Kymo(ConfocalImage):
...
         Dictionary containing kymograph-specific metadata.
+    position_offset : float
+        Coordinate position offset.
     """
 
-    def __init__(self, name, file, start, stop, json):
+    def __init__(self, name, file, start, stop, json, position_offset=0):
         super().__init__(name, file, start, stop, json)
         self._line_time_factory = _default_line_time_factory
+        self.position_offset = position_offset
```
rpauszek, line 49: should this be an internal attribute maybe? at least until we determine if we really need it?

lumicks/pylake/kymo.py
```diff
@@ -39,11 +39,14 @@ class Kymo(ConfocalImage):
         End point in the relevant info wave.
     json : dict
         Dictionary containing kymograph-specific metadata.
+    position_offset : float
+        Coordinate position offset.
```
rpauszek, line 43: I think we should be clear about what the offset is relative to. The most recent parent object that was cropped or the actual raw data?

lumicks/pylake/kymo.py
```diff
@@ -269,6 +272,54 @@ def plot_with_time_histogram(self, color_channel, pixels_per_bin=1, hist_ratio=0
...
+            return self._image(channel)[lower_pixels:upper_pixels, :]
+
+        def timestamp_factory(_, reduce):
+            return self._timestamps("timestamps", reduce)[lower_pixels:upper_pixels, :]
+
+        def line_time_factory(_):
+            return self.line_time_seconds
+
+        def pixelsize_factory(_):
+            return self.pixelsize_um
```
rpauszek, line 309: I think for things like this where you just forward the existing factory it might be cleaner to just have these as lambdas down below. It could also make a clear distinction between things that are changed or not if we consistently use lambdas for forwarded values and function definitions for changed

rpauszek (commented):
looks pretty good, just a couple comments about clarity. I'll hold off on approval so I can take a look again after #171 gets merged

lumicks/pylake/kymo.py
JoepVanlier: Alright. I've made it private for now.

lumicks/pylake/kymo.py
JoepVanlier: It's the absolute offset. I'll add it to the docstring.

lumicks/pylake/kymo.py
JoepVanlier: Sure. I'll update it to do that :)

lumicks/pylake/kymo.py
tiagobonetti: I'm unsure if the docstring changes showed up in the most recent version.

tiagobonetti (approved):
`goose.honk_by_distance()`

PR #179, opened 2021-08-18
Invalid timestamps after slicing `Kymo`

JoepVanlier:
**Why this PR?**
It fixes a critical bug in Kymograph slicing that would lead to erroneous timestamp reconstruction.

**What went wrong?**
Kymographs are encoded with the so-called infowave. It's a list containing a mapping which indicates which sample to include, where the pixel ends are etc. Reconstruction is hardcoded to use a constant number of samples per pixel. There is no way to tell where a line ends in the kymograph (other than knowing the number of pixels in the kymograph metadata).

When slicing a kymograph, we need to round the slicing to the next kymograph line to make sure reconstruction does not fail.

In PR https://github.com/lumicks/pylake/pull/124 we changed the code that handles the slicing to use the same reconstruction algorithm to obtain the timestamps. Unfortunately, this led to the following critical bug:

- The reconstruction algorithm allocates a new array to return the timestamps in. This array has a different type than the timestamps (`np.float` as opposed to `np.int64`). As a consequence, we have now incurred a small roundoff error in our timestamps (since timestamps are defined w.r.t. epoch, they are generally quite large).
- What happens next is that the timestamp returned is slightly beyond the first DAQ sample of line we intended to start at.
- Slicing from this timestamp results in an infowave that has fewer samples in the first pixel, violating our assumption of a constant number of samples per pixel.
- As a consequence, the timestamp reconstruction fails.
- As a consequence of this, the `line_time_seconds` property, which is used in plotting, kymotracking, MSD analysis is incorrect (and off by a potentially large margin!).

The errors you get in the line time are substantial and unpredictable. In the table below you can see the line time before and after slicing (both values should be the same).
```
0.1043328 vs 0.0977664
0.042393600000000004 vs 0.036479744
0.24999680000000002  vs 0.196992
0.24999680000000002  vs 0.19893760000000002
0.020070400000000002 vs 0.0120576
0.020070400000000002  vs 0.020070400000000002  (cas9 example we use for manual testing, and probably part of the reason we didn't pick up on it earlier)
```

**Who is affected?**
Everyone who sliced a Kymograph and subsequently did analysis on it since Pylake v0.8.2.

**Next steps**
While this PR fixes the most acute problem and introduces a test for it. It does not fix the precision issue at the core (yet).

This requires some additional investigation and introduces additional design questions, since `np.mean` which we use for kymograph downsampling is at risk of overflow at high sample rates (when using fixed precision).

The branch `dtype_experimental` has a fix that actually propagates the `dtypes`, but we should discuss the implications of this first.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -10,6 +10,7 @@
 
 #### Bug fixes
 
+* Fixed bug in `Kymo` which resulted in erroneous timestamps and line times after slicing the Kymograph by time (e.g. `Kymo["1s":"5s"]`). The reason for this was imprecision in the timestamp reconstruction that occurred when slicing the data. This in turn led to an erroneous reconstruction of the timestamps. **As a result, any downstream analysis that relies on the time axis of a Kymograph post-slicing cannot be trusted (MSD analysis, plotting, tracking, attributes such as `.line_time_seconds`, etc.) in pylake v0.8.2 and v0.9.0.** Only those two versions are affected. Note that regular image reconstruction was not affected. Timestamps of unsliced kymographs constructed directly from the `.h5` files are also not affected.
 * Fixed an error in the documentation. In v0.9.0 `PowerSpectrum.power` was changed to represent power in `V^2/Hz` instead of `0.5 V^2/Hz`. However, the docs were not appropriately updated to reflect this change in the model equation that's fitted to the spectrum. This is mitigated now.
 * Fixed bug in `plot_with_force` which caused an exception on Kymographs with a partial last pixel.
 
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -114,8 +114,11 @@ def line_timestamps_image(time_stamps, infowave, pixels_per_line):
     pixels_per_line : int
         The number of pixels on the fast axis of the scan.
     """
-    min_stamps = reconstruct_image(time_stamps, infowave, shape=(pixels_per_line,), reduce=np.min)
-    return min_stamps.astype(np.int64)[:, 0]
+    infowave, valid_idx = discard_zeros(infowave)
+    time_stamps = time_stamps[valid_idx]
+    pixel_start_idx = np.flatnonzero(infowave == InfowaveCode.pixel_boundary)
+    pixel_start_idx = np.concatenate(([0], pixel_start_idx + 1))
+    return time_stamps[pixel_start_idx[0:-1:pixels_per_line]]
 
 
 def seek_timestamp_next_line(infowave):
--- a/lumicks/pylake/tests/test_image.py
+++ b/lumicks/pylake/tests/test_image.py
@@ -40,11 +40,15 @@ def test_timestamps_image(num_lines, pixels_per_line, pad_size):
     pad = np.zeros(pad_size, dtype=np.int32)
     infowave = np.hstack([np.hstack([pad, line_info_wave, pad]) for _ in np.arange(num_lines)])
     start_indices = np.hstack([np.hstack([pad, line_selector, pad]) for _ in np.arange(num_lines)])
-    time = np.arange(len(infowave), dtype=np.int64) + 100
+
+    # Generate list of timestamp data.
+    # It is important to use timestamps that are large enough such that floating point
+    # round-off errors will occur if the data is converted to floating point representation.
+    time = np.arange(len(infowave), dtype=np.int64) * int(700e9) + 1623965975045144000
 
     line_stamps = line_timestamps_image(time, infowave, pixels_per_line)
     assert line_stamps.shape == (sum(start_indices), )
-    assert np.all(np.allclose(line_stamps, time[start_indices == 1]))
+    np.testing.assert_equal(line_stamps, time[start_indices == 1])
 
 
 def test_reconstruct():
--- a/lumicks/pylake/tests/test_kymo.py
+++ b/lumicks/pylake/tests/test_kymo.py
@@ -663,3 +663,47 @@ def test_incremental_offset():
     np.testing.assert_allclose(twice_cropped.line_time_seconds, kymo.line_time_seconds)
     np.testing.assert_allclose(twice_cropped.pixels_per_line, 2)
     np.testing.assert_allclose(twice_cropped._position_offset, 4e-3)
+
+
+def test_slice_timestamps():
+    """Test slicing with realistically sized timestamps (this tests against floating point errors
+    induced in kymograph reconstruction)"""
+    image = np.array(
+        [
+            [0, 12, 0, 12, 0, 6, 0],
+            [0, 0, 0, 0, 0, 6, 0],
+            [12, 0, 0, 0, 12, 6, 0],
+        ],
+        dtype=np.uint8
+    )
+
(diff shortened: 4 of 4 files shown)


lumicks/pylake/tests/test_image.py
```diff
@@ -40,11 +40,11 @@ def test_timestamps_image(num_lines, pixels_per_line, pad_size):
     pad = np.zeros(pad_size, dtype=np.int32)
     infowave = np.hstack([np.hstack([pad, line_info_wave, pad]) for _ in np.arange(num_lines)])
     start_indices = np.hstack([np.hstack([pad, line_selector, pad]) for _ in np.arange(num_lines)])
-    time = np.arange(len(infowave), dtype=np.int64) + 100
+    time = np.arange(len(infowave), dtype=np.int64) * int(700e9) + 1623965975045144000
```
mikhas, line 43: I feel you need a comment to explain why you chose to initialize `time` like that, and what kind of scenario this initialization is testing. Are you effectively testing `np.arrange()` here?

lumicks/pylake/detail/image.py
```diff
@@ -114,8 +114,11 @@ def line_timestamps_image(time_stamps, infowave, pixels_per_line):
     pixels_per_line : int
         The number of pixels on the fast axis of the scan.
     """
-    min_stamps = reconstruct_image(time_stamps, infowave, shape=(pixels_per_line,), reduce=np.min)
-    return min_stamps.astype(np.int64)[:, 0]
```
mikhas, line 118: the conversion error happens here?

lumicks/pylake/detail/image.py
```diff
@@ -114,8 +114,11 @@ def line_timestamps_image(time_stamps, infowave, pixels_per_line):
...
+    infowave, valid_idx = discard_zeros(infowave)
+    time_stamps = time_stamps[valid_idx]
+    pixel_start_idx = np.flatnonzero(infowave == InfowaveCode.pixel_boundary)
+    pixel_start_idx = np.concatenate(([0], pixel_start_idx + 1))
+    return time_stamps[pixel_start_idx[0:-1:pixels_per_line]]
```
mikhas, line 121: I am not sure I can parse this ... slices everything but the last element, with a step size of `pixels_per_line` (stride). And thats the indices we extract from `time_stamps` array?

lumicks/pylake/tests/test_image.py
JoepVanlier: No, it's making sure the timestamps are in a large enough range that roundoff errors larger than 1 ns start happening from a conversion to floating point. I'll add a comment:

```
    # It is important to use timestamps that are large enough such that floating point
    # round-off errors will occur if the data is converted to floating point representation.
```

lumicks/pylake/detail/image.py
JoepVanlier: Conversion happens in two places in `reconstruct_image`. Line `161` and `217` in `image.py`.

See the branch `dtype_experimental`.

lumicks/pylake/detail/image.py
JoepVanlier: Yes.

We find the pixel boundaries. We go to the DAQ sample immediately after the pixel boundary. We want the line starts, so we want the ones that happen every `pixels_per_line`. The last one is dropped since it would be beyond the end of the time_stamp range (last pixel of a line plus 1).

changelog.md
```diff
@@ -10,6 +10,7 @@
 
 #### Bug fixes
 
+* Fixed bug in `Kymo` which resulted erroneous timestamps and line times after slicing the Kymograph by time. The reason for this was an imprecision in the timestamp reconstruction that occurred when slicing the data. This in turn which led to an erroneous reconstruction of the timestamps. **As a result, any downstream analysis that relies on the time axis of a Kymograph post-slicing cannot be trusted (MSD analysis, plotting, tracking, etc).** This error was introduced in Pylake v0.8.2 and both v0.8.2 and v0.9.0 are affected. Note that the image reconstruction was not affected.
```
dean0x7d, line 13:
```suggestion
* Fixed bug in `Kymo` which resulted in erroneous timestamps and line times after slicing the Kymograph by time. The reason for this was imprecision in the timestamp reconstruction that occurred when slicing the data. This in turn led to an erroneous reconstruction of the timestamps. **As a result, any downstream analysis that relies on the time axis of a Kymograph post-slicing cannot be trusted (MSD analysis, plotting, tracking, etc) in pylake v0.8.2 and v0.9.0.** Only those two versions are affected. Note that regular image reconstruction was not affected.
```

lumicks/pylake/detail/image.py
dean0x7d, line 121: I think that your fix from the `dtype_experimental` branch is much better. It fixes the actual problem. This just creates yet another way to reconstruct data based on the infowave. I'm also a bit puzzled why you chose a very different reconstruction method than `reconstruct_image()`. I'd opt for our fix from the other branch that fixes the core issue in `reconstruct_image()`.

JoepVanlier: I agree that the other fix definitely has long term preference. The reason I opted for the pointfix now is because fixing `reconstruct_image` requires finding a workaround for the potential overflow issues when taking the mean of timestamps and making sure that none of the downstream algorithms are affected by a change from floating point to integer datatypes.

I went with this form because it effectively reverses #124 where this flaw was introduced. After the "good" `dtype` fix, this function should be changed back to `reconstruct_image`. In fact, after that fix, it should be possible to get rid of this function entirely and use the `reduce_timestamps` factory that's already in there for this (with argument `reduce=np.min`). That said, that means that for safety, we should test this not just at the image level, but also add the `Kymo` level. I have just added an extra test for this.

If you prefer, I can change it to be similar in structure as `reconstruct_image`.

dean0x7d: Thanks for the explanation. Makes sense to leave it as is for this PR.

dean0x7d (approved):
Great catch and quick thinking for hotfix!

changelog.md
```diff
@@ -10,6 +10,7 @@
 
 #### Bug fixes
 
+* Fixed bug in `Kymo` which resulted in erroneous timestamps and line times after slicing the Kymograph by time. The reason for this was imprecision in the timestamp reconstruction that occurred when slicing the data. This in turn led to an erroneous reconstruction of the timestamps. **As a result, any downstream analysis that relies on the time axis of a Kymograph post-slicing cannot be trusted (MSD analysis, plotting, tracking, etc) in pylake v0.8.2 and v0.9.0.** Only those two versions are affected. Note that regular image reconstruction was not affected.
```
rpauszek, line 13: We might add at the end something like:
"Timestamps of un-sliced `Kymo`s constructed directly from the .h5 files are also unaffected"
just to really nail down the point that this bug is only when you slice

rpauszek (approved):
looks good, just a quick comment on the changelog if you think it's appropriate.

I ran some extra tests on our usual test data and 2 other h5 files I found on my hard drive .... reconstruction of the images, timestamps, and line times were all identical between the original read from the h5 and a full slice 👍

PR #180, opened 2021-08-20
release: prepare for releasing v0.10.0

JoepVanlier:
**Why this PR?**
We want to get a release out that has a hotfix for the critical bug that existed in the kymograph slicing.


rpauszek (approved):
👍

PR #183, opened 2021-08-25
Kymotracker: bound dwelltime analysis

rpauszek:
**Why this PR?**
Often it is interesting to quantify the average time a protein binds to a DNA tether, as this can gives information about the kinetics and/or thermodynamics of the system. This is accomplished by fitting a list of dwelltimes (how long a single protein stays bound to the DNA) to an exponential distribution, or weighted mixture of exponential distributions.

Docs are [here](https://lumicks-pylake.readthedocs.io/en/binding_dwell_v2/tutorial/kymotracking.html#dwelltime-analysis) and details the motivation and certain caveats that are important for properly performing analysis.

The public API is a single method `KymoLineGroup.fit_binding_times()`. This returns an immutable object containing many relevant parameters and statistics of the optimization.

**Things to note**
- all of the backend is found in `lumicks.pylake.kymotracker.detail.binding_times`. The algorithms are generalized (not dependent on kymotracker input) and in the future it may make sense to move this to the `population` module as the same analysis is important for interpreting Hidden Markov Models. For now I chose to keep everything centralized with kymotracker since it's the only place it's used. 
- while the backend can handle more than 2 exponential components, only 1- and 2-component distributions are enabled currently from the public API. Often, mixture of more than 2 components is not statistically well defined and such distributions are rarely if ever reported in the literature; however we could remove this restriction if we want to leave the onus on the users.
- more complicated models exist involving mixtures of exponential/gamma/power distributions, but these are out of scope for this PR

### DIFF
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -387,3 +387,110 @@ By default, this will use the optimal number of lags (which in this case seems t
 .. image:: msdplot_100_lags.png
 
 It's not hard to see from this graph why taking too many lags results in unacceptably large variances.
+
+
+Dwelltime analysis
+------------------------
+
+The lifetime of the bound state(s) can be determined using `KymoLineGroup.fit_binding_times()`. This method defines
+the bound dwelltime as the length of each tracked line (in seconds). The lifetimes are then determined using
+Maximum Likelihood Estimation (MLE) :cite:`kaur2019dwell,woody2016memlet`. The number of exponential components
+to be used for the fit is chosen with the `n_components` argument.
+
+The likelihood :math:`\mathcal{L}` is defined for a mixture of exponential distributions as:
+
+.. math::
+
+    \mathcal{L} = \prod_j^T \left[ \frac{1}{N} \sum_i^M \frac{a_i}{\tau_i} \exp{\left( \frac{-t_j}{\tau_i} \right)} \right]
+
+where :math:`T` is the number of observed dwell times, :math:`M` is the number of exponential components, :math:`t` is time,
+:math:`\tau_i` is the lifetime of component :math:`i`, and :math:`a_i` is the fractional contribution of component :math:`i`
+under the constraint of :math:`\sum_i^M a_i = 1`. The normalization constant :math:`N` is defined as:
+
+.. math::
+
+    N = \sum_i^M a_i \left[
+    \exp{ \left( \frac{-t_{min}}{\tau_i} \right)} -
+    \exp{ \left( \frac{-t_{max}}{\tau_i} \right)}
+    \right]
+
+where :math:`t_{min}` and :math:`t_{max}` are the minimum and maximum possible observation times.
+
+The normalization constant takes into account the minimum and maximum possible observation times of
+a track. Note that when :math:`t_{min}=0` and :math:`t_{max}=\infty`, :math:`N=1`. However, for real experimental data,
+there are physical limitations on the measurement times (such as pixel integration time). Additionally, the minimum length of tracked lines
+here is dependent on the specific input parameters used for the tracking algorithm. Therefore, in order to estimate these bounds,
+the method uses the shortest track time and the length of the experiment, respectively.
+
+
+First let's try to fit the dwelltime distribution to a single exponential (the simplest case). To do this we
+simply call::
+
+    dwell_1 = traces.fit_binding_times(n_components=1)
+
+This returns a `BindingDwelltimes` object which contains information about the optimized model.
+We can visually inspect the result with::
+
+    dwell_1.plot(bin_spacing="log")
+
+.. image:: kymo_bind_dwell_1.png
+
+The `bin_spacing` argument can be either `"log"` or `"linear"` and controls the spacing of the bin edges.
+The scale of the x- and y-axes can be controlled with the optional `xscale` and `yscale` arguments; if they are not specified
+the default visualization is `lin-lin` for `bin_spacing="linear"` and `lin-log` for `bin_spacing="log"`.
+
+You can optionally pass the number of bins to be plotted as `n_bins`. Note that the number of bins
+is purely for visualization purposes; the model is optimized directly on the unbinned dwelltimes. This is the main
+advantage of the MLE method over analyses that use a least squares fitting to binned data, where the bin widths and number
+of bins can drastically affect the optimized parameters.
+
+We can clearly see that this distribution is not fit well by a single exponential decay.
+Let's now see what a double exponential distribution looks like::
+
+    dwell_2 = traces.fit_binding_times(n_components=2)
+    dwell_2.plot(bin_spacing="log")
+
+.. image:: kymo_bind_dwell_2.png
+
+Here we see that the double exponential fit visually looks better and the log likelihood is also higher than that
+for the single exponential fit. However, the log likelihood does not take into account model complexity, and will
+always increase for a model with more degrees of freedom. Instead, we can look at the Bayesian Information Criterion
+or Akaike Information Criterion to determine which model is better::
+
+    >>> print(dwell_1.bic, dwell_1.aic)
+    532.3299315589168  529.0366267341923
+
+    >>> print(dwell_2.bic, dwell_2.aic)
+    520.4562630650156  510.5763485908421
+
+These information criterion values weight the log likelihood with the model complexity, and as such are more useful for
+model selection. In general, the model with the lowest value is optimal. We can see that both values are lower for the double exponential model,
+indicating that it is a better fit to the data.
+
+We can see this effect if we purposely overfit the data. The following plot shows the result of fitting simulated data (randomly sampled
+from a single exponential distribution) with either a one- or two-component model. In the figure legends we see that the log likelihood
+decreases slightly for the two-component model because of the larger degrees of freedom. However, the BIC for the one-component model is
+indeed lower, as expected:
+
+.. image:: kymo_bic_compare.png
+
+Going back to our experimental data, we can next attempt to estimate confidence intervals (CI) for the parameters using bootstrapping.
+Here, a random dataset with the same size as the original is sampled (with replacement) from the original dataset. This sampled dataset
+is then fit using the MLE method, just as for the original dataset. The fit results in a new estimate for the model parameters.
+This process is repeated many times, and the distribution of the resulting parameters can be analyzed to estimate certain statistics about the them::
+
+    dwell_2.run_bootstrap(iterations=1000)
+    dwell_2.bootstrap.plot(alpha=0.05)
+
+.. image:: kymo_bind_bootstrap_2.png
+
+Here we see the distributions of the bootstrapped parameters. The vertical lines indicate the
+means of the distributions, while the red area indicates the estimated confidence intervals. The `alpha` argument determines
+the CI that is estimated as `100*(1-alpha)` % CI; in this case we're showing the estimate for the 95% CI. The values for the
+lower and upper bounds are the `100*(alpha/2)` and `100*(1-alpha/2)` percentiles of the distributions.
+
+Note, however, that while the means correspond well with the optimized model parameters, the distributions are not symmetric.
+In such a case, the simple method of using percentiles as CI values may not be appropriate. For more advanced analysis,
+the distribution values are directly available through the properties `BindingDwelltimes.bootstrap.amplitude_distributions` and
+`BindingDwelltimes.bootstrap.lifetime_distributions` which return the data as a `numpy` array with
+shape `[# components, # bootstrap samples]`.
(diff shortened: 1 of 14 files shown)


lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,377 @@
...
+
+    @property
+    def lifetimes(self):
+        """Lifetime parameter (in seconds) of each model component."""
+        return self._parameters[self.n_components :]
+
+    @property
+    def aic(self):
+        """Akaike Information Criterion."""
+        k = 2 * self.n_components  # number of parameters
```
JoepVanlier, line 188: As far as I know, the Akaike Information Criterion has as an assumption that the parameter estimation problem is not subject to constraints. This one is.

I'm wondering if the number of parameters is correct here. I think every active constraint reduces it by one parameter (since one can rephrase the model in an unconstrained problem with one fewer parameter in this case), but this would have to be verified.

Do people typically use aic for model selection in these models?

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,377 @@
...
+        if spacing == "log":
+            scale = np.logspace
+            limits = (np.log10(self.dwelltimes_sec.min()), np.log10(self.dwelltimes_sec.max()))
+        elif spacing == "linear":
+            scale = np.linspace
+            limits = (self.dwelltimes_sec.min(), self.dwelltimes_sec.max())
+        else:
+            raise ValueError("spacing must be either 'log' or 'linear'")
+
+        bins = scale(*limits, n_bins)
```
JoepVanlier, line 226: nice

JoepVanlier (commented):
I haven't finished my review yet, but just a small comment / question on the AIC (I want to play with this to get a feel for the API when I get the chance).

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,377 @@
...
+        maximum observation time in seconds, generally the length of the experiment
+    """
+    amplitudes = amplitudes[:, np.newaxis]
+    lifetimes = lifetimes[:, np.newaxis]
+    t = t[np.newaxis, :]
+
+    norm_factor = np.log(amplitudes) + np.log(
+        np.exp(-min_observation_time / lifetimes) - np.exp(-max_observation_time / lifetimes)
+    )
+    norm_factor = -logsumexp(norm_factor, axis=0)
```
JoepVanlier, line 299: I'd prefer the name `minus_log_norm_factor`, since you log-transformed it and took the negative. Personally, I might even prefer to just have this be `log_norm_factor` and then use the minus in the expression below.

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,377 @@
...
+
+    components = norm_factor + np.log(amplitudes) - np.log(lifetimes) - t / lifetimes
```
JoepVanlier, line 301: Can just be returned directly without the need for the intermediate variable.

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,377 @@
...
+        t=t,
+        min_observation_time=min_observation_time,
+        max_observation_time=max_observation_time,
+    )
+
+    if initial_guess is None:
+        initial_guess_amplitudes = np.ones(n_components) / n_components
+        initial_guess_lifetimes = np.mean(t) * np.arange(1, n_components + 1)
+        initial_guess = np.hstack([initial_guess_amplitudes, initial_guess_lifetimes])
+
```
JoepVanlier, line 365: Should we throw is if there is an observation below `min_observation_time` or above `max_observation_time`?

JoepVanlier (changes requested):
I went over the rest of it. Mostly looks good. The only thing I have my doubts about is the AIC/BIC implementation.

I quite like how clean it is.

docs/tutorial/kymotracking.rst
```diff
@@ -387,3 +387,100 @@ By default, this will use the optimal number of lags (which in this case seems t
...
+
+First let's try to fit the dwelltime distribution to a single exponential (the simplest case). To do this we
+simply call::
+
+    dwell_1 = traces.fit_binding_times(n_components=1)
+
+This returns a `BindingDwelltimes` object which contains information about the optimized model.
+We can visually inspect the result with::
+
+    dwell_1.plot(spacing="log")
```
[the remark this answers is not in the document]
rpauszek: In the literature I've seen examples of `linlin` `loglin` `linlog` and `loglog`; seems to depend on the author and specific experiment/model. I've added `xscale` and `yscale` as optional arguments and updated the default behavior if these are not supplied by the user

lumicks/pylake/kymotracker/detail/binding_times.py
rpauszek: In general when called from the public API this won't be the case because the limits are set using `min` and `max` of the data; but added anyway, is probably useful for internal purposes or when other more flexible public API becomes available

aafkevandenberg (commented):
Hi Ray, looking good! Just added some small comments. Will continue later today.

docs/tutorial/kymotracking.rst
```diff
@@ -387,3 +387,100 @@ By default, this will use the optimal number of lags (which in this case seems t
...
 It's not hard to see from this graph why taking too many lags results in unacceptably large variances.
+
+
+Dwelltime analysis
+------------------------
+
+The lifetime of the bound state(s) can be determined using `KymoLineGroup.fit_binding_times()`. This method defines
+the bound dwelltime as the length of each tracked line (in seconds) and optimizes the parameters of an
+exponential mixture distribution on these observations by Maximum Likelihood Estimation (MLE)
+:cite:`kaur2019dwell,woody2016memlet`.
```
aafkevandenberg, line 398: Could be simplified, eg: The lifetime of the bound state(s) can be determined using `KymoLineGroup.fit_binding_times()`. This method defines
the bound dwelltime as the length of each tracked line (in seconds). The lifetimes are then determined  using Maximum Likelihood Estimation (MLE) :cite:`kaur2019dwell,woody2016memlet`.  The user can choose the number of exponentials (components) that should to be used for the fit.

docs/tutorial/kymotracking.rst
```diff
@@ -387,3 +387,100 @@ By default, this will use the optimal number of lags (which in this case seems t
...
+    525.7495678897401  512.5763485908421
+
+These information criterion values weight the log likelihood with the model complexity, and as such are more useful for
+model selection. In general, the model with the lowest value is optimal. We can see that both values are lower for the double exponential model,
+indicating that it is a better fit to the data.
+
+We can next attempt to estimate confidence intervals (CI) for the parameters using a bootstrapping technique. Here, a random
+dataset with the same size as the original is sampled (with replacement) from the original dataset. The model is then
+trained on this bootstrapped data sample. This process is repeated many times, and the distribution of the resulting
+parameters can be analyzed to estimate certain statistics about the them::
```
aafkevandenberg, line 478: Use layman's explanation, eg:
We can next attempt to estimate confidence intervals (CI) for the parameters using bootstrapping. Here, a random
dataset with the same size as the original is sampled (with replacement) from the original dataset. This sampled dataset is then fit using the MLE method, just as for the orignal data set. The fit results in a new estimate for the model parameters. This process is repeated many times, and the distribution of the resulting
parameters can be analyzed to estimate certain statistics about them::

docs/tutorial/kymotracking.rst
```diff
@@ -387,3 +387,100 @@ By default, this will use the optimal number of lags (which in this case seems t
...
+
+    dwell_2.run_bootstrap(iterations=1000)
+    dwell_2.bootstrap.inspect_component(1, alpha=0.05)
+
+.. image:: kymo_bind_bootstrap_2_1.png
+
+Here we see the distributions of the bootstrapped parameters for the second component. The vertical line indicates
+means of the distributions, while the red area indicates the estimated confidence intervals. The `alpha` argument determines
```
aafkevandenberg, line 486: The vertical lines indicate the means of the distributions, while the red areas indicate the estimated confidence intervals.

docs/tutorial/kymotracking.rst
```diff
@@ -387,3 +387,100 @@ By default, this will use the optimal number of lags (which in this case seems t
...
+We can next attempt to estimate confidence intervals (CI) for the parameters using a bootstrapping technique. Here, a random
+dataset with the same size as the original is sampled (with replacement) from the original dataset. The model is then
+trained on this bootstrapped data sample. This process is repeated many times, and the distribution of the resulting
+parameters can be analyzed to estimate certain statistics about the them::
+
+    dwell_2.run_bootstrap(iterations=1000)
+    dwell_2.bootstrap.inspect_component(1, alpha=0.05)
+
+.. image:: kymo_bind_bootstrap_2_1.png
+
```
aafkevandenberg, line 484: I see you added the parameter names to many of the plots, would be good to do that here too, in the plot title.

aafkevandenberg (approved):

lumicks/pylake/kymotracker/detail/binding_times.py
rpauszek: yes, this is fixed now; sorry forgot to mention it

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -388,3 +389,25 @@ def save(self, filename, delimiter=";", sampling_width=None):
...
+            number of components in the model. Currently only values of {1, 2} are supported.
+        """
+
+        if n_components not in (1, 2):
+            raise ValueError("Only 1- and 2-component exponential distributions are currently supported.")
+
+        image = self[0]._image
+        dwelltimes_sec = np.hstack([line.seconds[-1] - line.seconds[0] for line in self])
+
+        min_observation_time = np.min(dwelltimes_sec)
```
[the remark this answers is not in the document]
rpauszek: This is still true, but the caveat (which took me a while to realize) is that the minimum possible time for kymotracking isn't controlled by the scan time, but rather the parameters set for the tracking algorithm itself. I added a note to the documentation to make this point explicit

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,118 @@
...
+    Parameters
+    ----------
+    amplitudes : array_like
+        fractional amplitude parameters for each component
+    lifetimes : array_like
+        lifetime parameters for each component in seconds
+    t : array_like
+        dwelltime observations in seconds
+    min_observation_time : float
+        minimum observation time in seconds, generally the line/frame time of the experiment
```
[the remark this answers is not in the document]
rpauszek: this is going to be very dependent on where the data is coming from (especially in the future with different types of dwelltime analysis) so I just removed the comment about line/frame times

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,118 @@
...
+        initial_guess = np.hstack([initial_guess_amplitudes, initial_guess_lifetimes])
+
+    bounds = (
+        *[(np.finfo(float).eps, 1) for _ in range(n_components)],
+        *[(min_observation_time * 0.1, max_observation_time * 1.1) for _ in range(n_components)],
+    )
+    constraints = {"type": "eq", "fun": lambda x, n: 1 - sum(x[:n]), "args": [n_components]}
+    result = minimize(
+        cost_fun, initial_guess, method="SLSQP", bounds=bounds, constraints=constraints
+    )
```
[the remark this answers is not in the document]
rpauszek: SLSQP is the only one that supports constraints :smirk:

lumicks/pylake/kymotracker/detail/binding_times.py
```diff
@@ -0,0 +1,118 @@
...
+
+    if initial_guess is None:
+        initial_guess_amplitudes = np.ones(n_components) / n_components
+        initial_guess_lifetimes = np.mean(t) * np.arange(1, n_components + 1)
+        initial_guess = np.hstack([initial_guess_amplitudes, initial_guess_lifetimes])
+
+    bounds = (
+        *[(np.finfo(float).eps, 1) for _ in range(n_components)],
+        *[(min_observation_time * 0.1, max_observation_time * 1.1) for _ in range(n_components)],
```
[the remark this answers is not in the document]
rpauszek: Statistically it is possible to estimate lifetimes shorter than the minimum observed time (although the estimation will probably not be super accurate/precise). In general though the estimated parameters should be between the min/max observation times. I chose these bounds mainly to prevent collapse to zero, which could cause numerical problems

lumicks/pylake/kymotracker/kymoline.py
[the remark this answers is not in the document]
aafkevandenberg: I agree, the tracking algorithm (combined with the linetime) sets the minimum possible time. Because the tracking algorithm will never mark a single pixel as a line. Good to make this explicit.

PR #184, opened 2021-08-25
Import kymolines with Kymo instance

rpauszek:
**Why this PR?**
Currently, in order to import kymolines from a saved .csv you first need to load a `Kymo` instance, manually construct a `CalibratedKymographChannel` from this and the channel name, and finally feed this into the import function. It would be easier to just hand the kymo and channel name directly to the import function and construct the calibrated image internally.

Note, this is an internal function for the time being as the .csv format is likely to change; however in the future when it is made public it cannot rely on the internal `CalibratedKymographChannel` class

### DIFF
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -1,4 +1,5 @@
 from lumicks.pylake.kymotracker.detail.msd_estimation import *
+from lumicks.pylake.kymotracker.detail.calibrated_images import CalibratedKymographChannel
 
 
 def export_kymolinegroup_to_csv(filename, kymoline_group, delimiter, sampling_width):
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -71,6 +74,8 @@ def import_kymolinegroup_from_csv(filename, image, delimiter=";"):
 
     indices = data[:, 0]
     lines = np.unique(indices)
+
+    image = CalibratedKymographChannel.from_kymo(kymo, channel)
```
mikhas, line 78: If this is all it takes to construct a `CalibratedKymographChannel` then the current API feels better? As in: less parameters and concrete types is better than a bunch of parameters that could just as well be mapped to that concrete type.

lumicks/pylake/kymotracker/tests/test_io.py
```diff
@@ -44,13 +45,20 @@ def read_txt(testfile, delimiter):
...
+        dt=dt,
+        samples_per_pixel=5,
+        line_padding=3
+    )
 
     # Test round trip through the API
     testfile = f"{tmpdir_factory.mktemp('pylake')}/test.csv"
     lines.save(testfile, delimiter, sampling_width)
-    read_file = import_kymolinegroup_from_csv(testfile, test_img, delimiter=delimiter)
+    read_file = import_kymolinegroup_from_csv(testfile, kymo, "red", delimiter=delimiter)
```
mikhas, line 61: So what you are saying is that this new API would make testing easier?

mikhas (commented):
I probably lack the context to understand why this is an improvement.

lumicks/pylake/kymotracker/kymoline.py
rpauszek: In the future once we've settled on the format for the exported .csv the import function will become public. `CalibratedKymographChannel` will always remain private (it was introduced to be able to handle calibration for kymotracker, but users should never construct this manually).

So in the future, the new API is needed as `Kymo` is public. Implementing now is mostly an improvement for our internal use

mikhas: OK, now the PR makes sense to me: Function will become public but still uses an internal type in its signature, and this needs to be fixed.

rpauszek: edited the PR description for clarity

mikhas (approved):
LGTM!

PR #185, opened 2021-08-25
Fix CorrelatedStack export bugs

rpauszek:
**Why this PR?**
This PR addresses two bugs in `CorrelatedStack.export_tiff()`
- previously, attempting to export grayscale images (such as IRM) would throw an exception because the alignment matrices metadata is not there
- exporting changes the alignment matrices metadata keys from `Channel X alignment` to `Applied channel X alignment` to indicate that the data no longer corresponds to the raw data recorded in Bluelake. However, these fields were always changed regardless if the alignment was actually applied to the data or not.

In the process of fixing these, the mock test data was also updated to bring it more in line with what the real files look like.

### DIFF
--- a/lumicks/pylake/detail/widefield.py
+++ b/lumicks/pylake/detail/widefield.py
@@ -19,11 +19,14 @@ class TiffFrame:
 
     def __init__(self, page, align):
         self._src = page
-        self._description = ImageDescription(page)
-        self._align = align
+        self._description = ImageDescription(page, align)
+
+    @property
+    def _align(self):
+        return self._description._align
 
     def _align_image(self):
-        """ reconstruct image using alignment matrices from Bluelake; return aligned image as a NumPy array"""
+        """reconstruct image using alignment matrices from Bluelake; return aligned image as a NumPy array"""
 
         if not self._description:
             warnings.warn("File does not contain metadata. Only raw data is available")
@@ -66,7 +69,7 @@ class TiffFrame:
 
     @property
     def is_rgb(self):
-        return self._src.tags["SamplesPerPixel"].value == 3
+        return self._description.is_rgb
 
     def _get_plot_data(self, channel="rgb", vmax=None):
         """return data an numpy array, appropriate for use by `imshow`
@@ -129,7 +132,9 @@ class TiffStack:
 
 
 class ImageDescription:
-    def __init__(self, src):
+    def __init__(self, src, align):
+        self.is_rgb = src.tags["SamplesPerPixel"].value == 3
+        self._align = align
         try:
             self.json = json.loads(src.description)
         except json.decoder.JSONDecodeError:
@@ -191,6 +196,7 @@ class ImageDescription:
     def for_export(self):
         out = copy(self.json)
         if self:
-            for j in range(3):
-                out[f"Applied channel {j} alignment"] = out.pop(f"Channel {j} alignment")
+            if self.is_rgb and self._align:
+                for j in range(3):
+                    out[f"Applied channel {j} alignment"] = out.pop(f"Channel {j} alignment")
         return json.dumps(out, indent=4)
--- a/lumicks/pylake/tests/data/mock_widefield.py
+++ b/lumicks/pylake/tests/data/mock_widefield.py
@@ -134,7 +134,7 @@ def make_wt_description(version, bit_depth, m_red, m_blue, offsets):
 
 
 def make_alignment_image_data(spots, red_warp_parameters, blue_warp_parameters, bit_depth,
-                              offsets=None, version=1):
+                              offsets=None, camera="wt", version=1):
 
     spots = np.array(spots).T # [2 x N]
     m_red, red_spots = apply_transform(spots, offsets=offsets, **red_warp_parameters)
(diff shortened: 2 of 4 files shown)


lumicks/pylake/detail/widefield.py
```diff
@@ -129,7 +132,9 @@ def num_frames(self):
 
 
 class ImageDescription:
-    def __init__(self, src):
+    def __init__(self, src, align):
+        self.is_rgb = src.tags["SamplesPerPixel"].value == 3
+        self._align = align
```
JoepVanlier, line 137: I wonder if `_want_align` may be a better name, since we don't know at this point whether alignment is even possible (there's no check for the presence of alignment matrices yet).

What I'd otherwise find a bit confusing is for example a single color image, where this property is `True` and a three color image without alignment matrices where this property is also `True` even though it can clearly not be aligned.

JoepVanlier (changes requested):
Looks good to me.

One thing I'm wondering about is the way metadata is handled upon export. I think converting from aligned to applied aligned is good.

I wonder however whether we should keep warning on missing alignment matrices when exporting as it may be confusing to users. Right now, exporting, then importing then exporting again produces the following warning.

```
UserWarning: File does not contain alignment matrices. Only raw data is available
  warnings.warn("File does not contain alignment matrices. Only raw data is available
```

This may confuse someone, because now they might think that they've just exported unaligned images, which is not the case.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -340,20 +340,26 @@ def test_export(rgb_tiff_file, rgb_tiff_file_multi, gray_tiff_file, gray_tiff_fi
...
-
-        with tifffile.TiffFile(str(filename)) as tif0, tifffile.TiffFile(savename) as tif:
-            assert len(tif0.pages) == len(tif.pages)
-            assert tif0.pages[0].software != tif.pages[0].software
-            assert "pylake" in tif.pages[0].software
-            assert "Applied channel 0 alignment" in tif.pages[0].description
-            assert "Channel 0 alignment" not in tif.pages[0].description
-            for page0, page in zip(tif0.pages, tif.pages):
-                assert page0.tags["DateTime"].value == page.tags["DateTime"].value
+        for align in (True, False):
```
tiagobonetti, line 343: Maybe we can avoid the nested loops by using [`itertools.product`](https://docs.python.org/3/library/itertools.html#itertools.product)

```python
for filename, align in itertools.product((rgb_tiff_file, rgb_tiff_file_multi, gray_tiff_file, gray_tiff_file_multi), (True, False)):
```

lumicks/pylake/tests/data/mock_widefield.py
```diff
@@ -146,17 +146,21 @@ def make_alignment_image_data(spots, red_warp_parameters, blue_warp_parameters,
...
+        # by definition, grayscale images only have 1 channel
+        reference_image = reference_image[:,:,1]
+        warped_image = warped_image[:,:,1]
+    else:
+        raise ValueError("camera argument must be 'wt' or 'irm'")
 
     return reference_image, warped_image, description, bit_depth
 
 
 def write_tiff_file(reference_image, warped_image, description, bit_depth, n_frames, filename):
```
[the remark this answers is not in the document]
rpauszek: `bit_depth` is still needed because that field isn't in the `v1` form of the metadata but is needed in the tests for some things.

Also, `reference_image` is used in the tests (for example `test_image_reconstruction_rgb`)

lumicks/pylake/tests/data/mock_widefield.py
```diff
@@ -146,17 +146,21 @@ def make_alignment_image_data(spots, red_warp_parameters, blue_warp_parameters,
 
     reference_image = np.repeat(green_image[:,:,np.newaxis], 3, axis=2)
     warped_image = np.stack((red_image, green_image, blue_image), axis=2).squeeze()
-    description = make_wt_description(version, bit_depth, m_red, m_blue, offsets)
+    if camera == "wt":
+        description = make_wt_description(version, bit_depth, m_red, m_blue, offsets)
```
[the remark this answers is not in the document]
rpauszek: from what I can tell based on the different test images I have, the only difference between brightfield and IRM in the meta data is the values of the `camera` field and in `v2` brightfield doesn't have `exposure sync available` and `exposure sync enabled`. Right now we don't have any specific functionality that supports anything other than IRM/WT other than opening and plotting the images, do you think it's worth it to add tests?

lumicks/pylake/detail/widefield.py
rpauszek: indeed, especially now that we support reading of images that were previously aligned and then exported from pylake, we actually can't fully specify the state with a simple boolean. Additonally, yes the warnings can be confusing because they don't necessarily match the state. What we really need is something like `align_requested : bool` and `alignment_status : enum` where the status could be `ready`, `already applied`, `missing` etc ... 

I played around with this a bit, but it's actually quite a bit of additions/refactoring that needs to be done. I'd like to open this up as a 2nd PR based on this one

lumicks/pylake/tests/data/mock_widefield.py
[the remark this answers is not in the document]
JoepVanlier: I'm talking about `write_tiff_file` specifically. These fields are not used in the function. The only place I see that function used is `test_correlated_stack`. I assume they're there because of the old format and directly being able to unpack the alignment data with `*`. Maybe rename the args to `_` if they're unused to indicate that we know about it.

lumicks/pylake/tests/data/mock_widefield.py
```diff
@@ -160,6 +160,7 @@ def make_alignment_image_data(spots, red_warp_parameters, blue_warp_parameters,
 
 
 def write_tiff_file(reference_image, warped_image, description, bit_depth, n_frames, filename):
+    # single channel data [x, y] array, RGB is [x, y, c] array
```
tiagobonetti, line 163:
```suggestion
    # We use the dimension of image data to evaluate the number of color channels
```

lumicks/pylake/detail/widefield.py
JoepVanlier: Alright. Sounds fair.

JoepVanlier (approved):

tiagobonetti (approved):
`honk`

PR #187, opened 2021-08-27
Calibrate kymographs to base pairs

rpauszek:
**Why this PR?**

Often it is more interesting to know the position of binding events in base pairs along the DNA tether rather than in microns.

**Things to note:**
- calibration is forwarded to all downstream analyses (plotting, kymotracking, MSD analysis)
- length in microns is still available (since this information is in the original h5 file and some attributes such as `pixelsize_um` are public). I think this is reasonable since you might want to know some information in um and other info in bp
- I chose to use a small dataclass to store the calibration information. This allows access to the current unit, formatted labels for plotting, and we can extend with other functionality in the future as needed without introducing tons of attributes to `Kymo`. Additionally in the future we will need to support variable calibration (depending on the position of the beads for instance) and it is nice to have this functionality abstracted away from the main class

docs are [here](https://lumicks-pylake.readthedocs.io/en/calib_bp/tutorial/kymographs.html#kymo-data-and-details)

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -1,6 +1,7 @@
 import numpy as np
 import warnings
 import cachetools
+from dataclasses import dataclass
 
 from skimage.measure import block_reduce
 from .detail.confocal import ConfocalImage
@@ -41,12 +42,15 @@ class Kymo(ConfocalImage):
         Dictionary containing kymograph-specific metadata.
     position_offset : float
         Coordinate position offset with respect to the original raw data.
+    calibration : PositionCalibration
+        Class defining calibration from microns to desired position units.
     """
 
-    def __init__(self, name, file, start, stop, json, position_offset=0):
+    def __init__(self, name, file, start, stop, json, position_offset=0, calibration=None):
         super().__init__(name, file, start, stop, json)
         self._line_time_factory = _default_line_time_factory
         self._position_offset = position_offset
+        self._calibration = PositionCalibration() if calibration is None else calibration
 
     def _has_default_factories(self):
         return (
@@ -93,7 +97,9 @@ class Kymo(ConfocalImage):
 
         start = line_timestamps[i_min]
 
-        return Kymo(self.name, self.file, start, stop, self._json, self._position_offset)
+        return Kymo(
+            self.name, self.file, start, stop, self._json, self._position_offset, self._calibration
+        )
 
     @cachetools.cachedmethod(lambda self: self._cache)
     def _line_start_timestamps(self):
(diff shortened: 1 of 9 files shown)


lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -37,8 +40,8 @@ def store_column(column_title, format_string, new_data):
     store_column("time (pixels)", "%.18e", times_idx)
     store_column("coordinate (pixels)", "%.18e", coords_idx)
 
-    store_column("time", "%.18e", seconds)
-    store_column("position", "%.18e", position)
+    store_column(f"time ({time_units})", "%.18e", seconds)
+    store_column(f"position ({position_units})", "%.18e", position)
```
JoepVanlier, line 44: Changing this is a breaking change (data format change). Should be in the changelog.

I totally support the change though. We should always put units there.

lumicks/pylake/kymo.py
```diff
@@ -380,6 +408,35 @@ def pixelcount_factory(_):
...
+            self.stop,
+            self._json,
+            self._position_offset,
+            calibration,
+        )
+        result._image_factory = self._image_factory
+        result._timestamp_factory = self._timestamp_factory
+        result._line_time_factory = self._line_time_factory
+        result._pixelsize_factory = self._pixelsize_factory
+        result._pixelcount_factory = self._pixelcount_factory
```
JoepVanlier, line 436: Not suggesting to piggyback it on this PR, but feels like we're a bit overdue for a copy method.

lumicks/pylake/kymo.py
```diff
@@ -380,6 +408,35 @@ def pixelcount_factory(_):
         result._pixelcount_factory = pixelcount_factory
         return result
 
+    def calibrate_to_kbp(self, length_kbp):
+        """Calibrate from microns to other units.
+
+        Parameters
+        ----------
+        length : int
+            length of the kymo in kilobase pairs
```
JoepVanlier, line 417: Is this the most convenient way of specifying the calibration factor?

It seems convenient when cropping. Inconvenient when not cropping (because then you have to extrapolate from the bead to bead region to the whole kymo, since the kbps reflect the distance between the beads).

Also, `int`? Shouldn't kbp's be `float` or `bp` (`int`)?

JoepVanlier (changes requested):
Cool stuff! I'm sure this functionality will be appreciated 👍 

I'm missing a changelog entry. Other than that I only have some minor comments / questions.

lumicks/pylake/kymo.py
rpauszek: we have a ticket open for that, I pushed it up a bit in the backlog

lumicks/pylake/kymo.py
rpauszek: Do you have something else in mind that would make more sense?

For the current implementation, to me, it really only makes sense to calibrate if you crop out the beads. Otherwise you would have to set the origin and end point at the bead edges since the units don't make sense in these regions anyway. This was the reason I made the small `PositionCalibration` class, which I envisioned would be for "simple" calibrations like this.

For more involved calibrations (such as using landmarks) we could have another class to handle that but that exposes the same API to be used internally by `Kymo`

mikhas: I find it surprising that a variable called `length` should _not_ be `int`.

rpauszek: CS-wise ok sure because you're usually talking about length of an array. But for a physical measurement, length is almost always a `float`. The only reason it was `int` for base pairs is because you can't have a fraction of a nucleotide

JoepVanlier: @rpauszek Fair enough. That makes sense.
@mikhas So are you 1 or 2 meters tall? :D

JoepVanlier (approved):
Looks good to me! 👍

PR #188, opened 2021-08-30
Force calibration: Hydrodynamically correct power spectra

JoepVanlier:
**Why this PR?**
To provide the option for our users to use hydrodynamically correct power spectra.

**What to look out for?**
The literature has different forms for these hydrodynamic spectra. We chose to implement them according to:

```Tolić-Nørrelykke, S. F., Schäffer, E., Howard, J., Pavone, F. S., Jülicher, F., & Flyvbjerg, H. (2006). Calibration of optical tweezers with positional detection in the back focal plane. Review of scientific instruments, 77(10), 103101.```

Reason being that at low frequencies this one agrees with Faxen's law best, and for active calibration low frequencies are critical.

If you want to play with some simulated spectra then see `/force_calibration/tests/test_hydro.m`. The test `test_integration_passive_calibration_hydrodynamics` should give you a pretty good idea how to simulate (active) calibration data and fit to it.

I had to find a way to deal with allowing use of the "bulk" model and the near surface model. I've chosen to have the model go with the bulk model when the distance from the surface parameter is set to `None` and setting the default to this. Alternatively, we could also opt with `np.inf` (which may be more semantically correct, but would require users to import `numpy` which I thought was a bit less friendly).

I do realize the function has quite a few parameters, but unfortunately, the hydrodynamic model does have a lot of parameters. Luckily, most of the time, you really only need to be setting the important one (the `distance_to_surface`).

Note: I disabled `black` on some equations, and those lines may be overlong, but I found them unreadable after `black` treatment and splitting them up into many intermediate variables makes comparing them to the paper a lot harder later. I hope that's ok.

**What is out of scope?**
Documentation is still out of scope for this PR.

### CODE REVIEW

File: lumicks/pylake/force_calibration/detail/hydrodynamics.py [hunk 1]
```diff
@@ -0,0 +1,193 @@
...
+        return real_stokes, imag_stokes
+
+    epsilon = (2 * distance_to_surface - bead_radius) * sqrt_freq_ratio / bead_radius
+    r_over_l = bead_radius / distance_to_surface
+
+    # Effect of the surface (depends on depth) e.g. denominator from Eq D6 in [1].
+    # fmt: off
+    real_surf = 1 - 9 / 16 * r_over_l * (1 - sqrt_freq_ratio / 3 - 4 / 3 * (1 - np.exp(-epsilon) * np.cos(epsilon)))
+    imag_surf = - 9 / 16 * r_over_l * (sqrt_freq_ratio / 3 + 2 / 9 * freq_ratio + 4 / 3 * np.exp(-epsilon) * np.sin(epsilon))
+    # fmt: on
```
rpauszek, line 87: these forms of the equations are from Berg Eqn 34, which is indeed `[1]` in the `calibration_models.py` reference list.  Additionally, those equations already define `Re or Im {gamma/gamma_0}` so the division below isn't needed(?)

Is this intentional to be mixing equations from 2 different manuscripts, since everything else if from Tolic `[6]`?

lumicks/pylake/force_calibration/tests/test_hydro.py
```diff
@@ -0,0 +1,241 @@
...
+    [
+        [500e-6, 4800, 1.16e-8, 0.52e-6, 997.0, 1060.0, 25e-9, 3.420441139e-11],
+        [250e-6, 4800, 1.16e-8, 0.52e-6, 1030.0, 1060.0, 100e-9, 6.586892852e-12],
+        [500e-6, 4800, 1.26e-8, 0.52e-6, 997.0, 1060.0, 100e-9, 2.636236948e-11],
+        [500e-6, 2445, 1.978234e-8, 1.04e-6, 997.0, 1060.0, 100e-9, 1.178556597e-10],
+        [500e-6, 1165, 4.2694e-8, 2.2e-6, 997.0, 1060.0, 100e-9, 5.576540852e-10],
+        [500e-6, 545.57, 7.557e-8, 4.4e-6, 997.0, 1060.0, 100e-9, 2.467139096e-09],
+    ],
+)
+def test_hydro_spectra(
```
rpauszek, line 229: is there a reason this has the same name as the previous test function?

lumicks/pylake/force_calibration/detail/power_models.py
```diff
@@ -163,3 +163,21 @@ def sphere_friction_coefficient(eta, d):
...
     return 3.0 * math.pi * eta * d
+
+
+def theoretical_driving_power_lorentzian(driving_frequency, driving_amplitude, fc):
+    """Compute the power expected for a given driving input.
+
+    When driving the stage or trap, we expect to see a delta spike in the power density
+    spectrum. This function returns the expected power contribution of the bead motion to the
+    power spectrum. It corresponds to the driven power spectrum minus the thermal power spectrum
+    integrated over the frequency bin corresponding to the driving input.
```
rpauszek, line 174: really nice to have a docstring for this now even though it's a simple equation 👍

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -198,6 +293,24 @@ def __init__(
...
+                rho_bead=self.rho_bead,
+                distance_to_surface=self.distance_to_surface,
+            )
+        else:
+            self._theoretical_driving_power_model = partial(
+                theoretical_driving_power_lorentzian,
+                driving_frequency=self.driving_frequency,
+                driving_amplitude=self.driving_amplitude,
+            )
+
```
rpauszek, line 313: nice!

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -235,19 +346,29 @@ def calibration_results(self, fc, diffusion_constant_volts, f_diode, alpha):
...
+        # Equation 12 from [6]
         distance_response = np.sqrt(self._theoretical_driving_power(fc) / power_exp)  # m/V
 
+        # Equation 16 from [6]
         temperature_kelvin = scipy.constants.convert_temperature(self.temperature, "C", "K")
         k_temperature = scipy.constants.k * temperature_kelvin
-        gamma_0 = k_temperature / (distance_response ** 2 * diffusion_constant_volts)  # kg/s^2
-        kappa = 2.0 * np.pi * fc * gamma_0  # N/m
+        measured_drag_coeff = k_temperature / (distance_response ** 2 * diffusion_constant_volts)
```
rpauszek, line 358: I think it's good to leave the unit comment in

rpauszek (changes requested):
Good stuff! I really like the way things are set up, nicely separated into easily digestible parts. Most of the comments are about docstrings for clarity, but there is a concern about the equations used for the complex drag coefficient. I may just be misunderstanding something, but worth double checking.

Also when I checked out the branch there seems to be an empty file at `test.data.generate_hydro_data.py`

lumicks/pylake/force_calibration/detail/hydrodynamics.py
```diff
@@ -0,0 +1,193 @@
...
+    bead_radius : float
+        Bead radius [m]
+    rho_sample : float
+        Sample mass density [kg/m^3]
+    distance_to_surface : float
+        Distance from bead center to nearest surface [m]
+
+    [1] Tolić-Nørrelykke S. F, Schäffer E, Howard J, Pavone F. S, Jülicher F and Flyvbjerg, H.
+    Calibration of optical tweezers with positional detection in the back focal plane, Review of
+    scientific instruments 77, 103101 (2006).
```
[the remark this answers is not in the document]
JoepVanlier: Yeah... should I just name this one `[6]` in this docstring then? Or do we put it somewhere else?

lumicks/pylake/force_calibration/detail/hydrodynamics.py
JoepVanlier: They're just the equations from Tolic rewritten to look like the ones in Berg.

Welp, here goes nothing:

Starting from the denominator in Tolic.

image

Note that the notation is pretty strange here. It almost looks like they're doing 2 * i * f / 9 * f_nu, but the 9 and f_nu are actually together part of the denominator like `(9*fv)`. If you don't consider them together, the model blows up.

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + 2 * i * f / (9 * fv) - (4/3) * (1 - exp(-(1-i) * (2*l-R)/delta))}

Let's group some terms:

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-(1-i) * (2*l-R)/delta))}

Write c = (2 * l - R) / delta to not have to write down so much.

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-(1-i) * c))}

Multiply out the terms in the exponential:

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-c + i * c))}

Split the complex part from the exponential:

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-c) * exp(i * c))}

Rewrite the complex exponential as a cosine and sine term:

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-c) * (cos(c) + i * sin(c)))}

Multiply out the terms:

1 - (9/16) * (R/l) * {1 - [(1 - i)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-c) * cos(c) - exp(-c) * i * sin(c))}

We have multiplied everything out now, so Real and Imag is simply a matter of collecting the real and imaginary parts.

Real = 1 - (9/16) * (R/l) * {1 - [(1)/3] sqrt(f/fv) + (2 / 9) * i * (f / fv) - (4/3) * (1 - exp(-c) * cos(c) - exp(-c) * i * sin(c))}

Real = 1 - (9/16) * (R/l) * {1 - (1 / 3) * sqrt(f/fv) - (4/3) * (1 - exp(-c) * cos(c))}

Imag = - (9/16) * (R/l) * {- [(-1)/3] sqrt(f/fv) + (2 / 9) * (f / fv) - (4/3) * (- exp(-c) * sin(c))}

Imag = - (9/16) * (R/l) * {(1/3) sqrt(f/fv) + (2 / 9) * (f / fv) + (4/3) * (exp(-c) * sin(c))}

Just to verify that the final term corresponds to the denominator in Tolic, you can run the following in `Sympy`:

```
import sympy as sym

R, l, f, fv, c = sym.symbols('R, l, f, fv, c', real=True, positive=True)

fulleq = 1 - sym.Rational(9, 16) * (R/l) * (1 - (sym.Rational(1,3) * (1 - sym.I)) * sym.sqrt(f/fv) + sym.Rational(2, 9) * sym.I * f / fv - sym.Rational(4, 3) * (1 - sym.exp(-(1 - sym.I) * c)))

manual_real = 1 - sym.Rational(9, 16) * (R/l) * (1 - sym.Rational(1,3) * sym.sqrt(f/fv) - sym.Rational(4, 3) * (1 - sym.exp(-c) * sym.cos(c)))
manual_imag = - sym.Rational(9, 16) * (R/l) * (sym.Rational(1,3) * sym.sqrt(f/fv) + sym.Rational(2, 9) * (f / fv) + sym.Rational(4, 3) * (sym.exp(-c) * sym.sin(c)))

print(sym.re(fulleq) - manual_real)
print(sym.im(fulleq) - manual_imag)
```

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -187,18 +170,16 @@ def __init__(
         driving_frequency_guess : float
             Guess of the driving frequency.
         viscosity : float, optional
-            Liquid viscosity [Pa*s].
+            Liquid viscosity [Pa*s]. Only used when hydrodynamic corrections are enabled.
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, you are absolutely correct. Fixing!

lumicks/pylake/force_calibration/detail/hydrodynamics.py
```diff
@@ -0,0 +1,193 @@
...
+        radius of the bead [m]
+    rho_bead : float
+        density of the bead [kg/m^3]
+
+    Note that the mass term has lead to some confusion in literature. [6] mentions specifically
+    in reference 30 (used as a footnote) that this term only includes the bead mass whereas [1]
+    contains a typographical error. Note that in the code that implements [6] (ref [3]), they did
+    include it however. We decided to follow the paper and omit the second term. The effects of
+    this difference were very small for all simulations tested.
+        m_star = bead_mass + (2.0 * np.pi * rho_sample * radius ** 3.0) / 3.0"""
```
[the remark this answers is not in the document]
JoepVanlier: Right, yeah. Reading it again, it does read a bit confusing. I've tried rephrasing it. I left it as a separate fixup commit.

lumicks/pylake/force_calibration/detail/hydrodynamics.py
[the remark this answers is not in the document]
JoepVanlier: I've put the full reference list at the top and just cited consistently with the other file.

lumicks/pylake/force_calibration/tests/test_hydro.py
JoepVanlier: No, that's an error. Fixing.

lumicks/pylake/force_calibration/detail/power_models.py
JoepVanlier: :)

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: It felt a bit like duplicating the unit that's right below. Also the old one was wrong. It's `kg/s`.

lumicks/pylake/force_calibration/detail/hydrodynamics.py
rpauszek: thanks for the detailed derivation 🤣 double checked by hand just to triple verify. I wonder if it might be worthwhile putting a small comment, something on the order of `"D6 from [6] re-written with Euler's formula to approximate the form used in [1]"` just so any user that's looking through the code (or us in the future) don't have the same initial reaction I did.

lumicks/pylake/force_calibration/calibration_models.py
rpauszek: fair enough

rpauszek (approved):
looks great! just a small comment about equation `D6` indicating that we chose to re-write it using Euler's formula

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -193,12 +257,28 @@ def __init__(
         num_windows : int, optional
             Number of windows to average for the uncalibrated force. Using a larger number of
             windows potentially increases the bleed, but may be useful when the SNR is low.
+        hydrodynamically_correct : bool, optional
+            Enable hydrodynamically correct model.
+        distance_to_surface : float, optional
+            Distance from bead center to the surface [m]
+            Only used when using hydrodynamically correct model. None uses the approximation valid
+            for deep in bulk.
```
lpingas, line 265: What do you think about adding an assertion to avoid having `distance_to_surface < bead_radius`?

JoepVanlier: Yeah, that's a good point. I've added it as an assertion (+ tests).

lpingas (approved):
This is great work! The API is looking good, and I also liked the simpler calibration models hierarchy.

lumicks/pylake/force_calibration/detail/hydrodynamics.py
JoepVanlier: Yeah, fair point. I've added a comment about it.

PR #189, opened 2021-09-02
force_calibration: better validation driving input

JoepVanlier:
**Why this PR?**
When doing active calibration, we have to find the input magnitude and frequency from the nanostage (or in the future trap) signal.  To obtain this we choose a small region of the Gaussian windowed frequency spectrum and find the largest peak in that region. 

After that, we regress a polynomial to an even smaller range (typically three points) of the log magnitude which provides us with the peak amplitude and frequency.

The best accuracy is obtained if the peak is actually in this region, but this relies on the search range being chosen appropriately in the first step.

This PR adds some validation on whether the peak was indeed in the search range, and whether the found magnitudes correspond to a peak rather than a dip.

If either of these conditions is not met, we throw because our peak estimate may be less accurate than optimal.

Note: I noticed this because I was seeing small biases in my calibrations rather than errors, which underlines the danger of this not being an explicit exception.


rpauszek (approved):
👍 LGTM

PR #191, opened 2021-09-08
force_calibration: check bead diameter > 1e-6

JoepVanlier:
**Why this PR?**
Using bead diameters less or equal to zero produces non-sensical calibrations (NaNs).

The field for bead diameter in Bluelake defaults to zero, so this can be confusing. With this fix it now raises an exception when zero is passed to the calibration routine.

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -69,6 +69,12 @@ class PassiveCalibrationModel:
         rho_sample=None,
         rho_bead=1060.0,
     ):
+        if bead_diameter < 1e-2:
+            raise ValueError(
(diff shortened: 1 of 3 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -69,6 +69,12 @@ def __init__(
         rho_sample=None,
         rho_bead=1060.0,
     ):
+        if bead_diameter < 1e-6:
```
dean0x7d, line 72: I think that we anchor this a bit closer to real bead size limits: make it 1e-2? I doubt that anyone will be using sub-nanometer beads.

JoepVanlier: Yeah, fair point. I've changed it to 1e-2.

PR #192, opened 2021-09-10
Trust region

JoepVanlier:
## Why this PR?
In several numerical tests, we see that this optimizer is better at optimizing.

Considering that the optimizer we were using does not support bounds or parameter scaling natively, and this one does, it seems like a good additional improvement.

- The confidence interval for alpha was previously a questionable approximation (assuming linearity), whereas now it is just as advertised on the tin.
- Note that we also get rid of the unreliable initial guess estimate for `f_diode`, which produces erroneous results most of the time (usually estimates way beyond Nyquist). Instead we just use 14 kHz which is a reasonable guess for the diodes we use. Based on the tests it doesn't seem like it's terribly sensitive to this initial guess as long as it is not way above the top frequency of the fitted range.
- A bound was added to the optimization problem to make sure that `f_diode` does not drift beyond the Nyquist frequency, which could sometimes happen because of the poor initial estimate. The optimizer has difficulty recovering from this.

## What to look out for
Note that point 2 and 3 effectively force the sample rate to be > 28 kHz or the optimizer will throw an `x0 is infeasible` error. I see two options here. Either:

A. We assert on the condition `power_spectrum.sample_rate` > 28 kHz or ,

B. We put the initial guess of `f_diode` to `min(f_diode, ps.sample_rate / 2)`.

## A note about the tests
Note that I had to remove a few tests in test_fit_settings since they were relying on fitting with max_function_evaluations set to 1 working. Note that the behaviours tested in these tests is covered by other tests and they were superfluous.

- Fitting a purely analytic model is tested separately in `test_fit_analytic`.
- The `analytic_fit_range` argument is analyzed now by deliberately choosing it such that the fit fails (before it was tested by deliberately jinxing the fit).
- Finally, the following did not actually test anything because it was missing an assert:
```python
    # If we set the termination tolerance very high, even a totally unconverged fit is accepted.
    psc.fit_power_spectrum(
        power_spectrum=power_spectrum, model=model, ftol=10, max_function_evals=1
    )
```


rpauszek (approved):
I think it looks great! Definitely appreciate the detailed introduction, but I'm missing the Discussion and Methods sections 😉 

As for the lack of the legacy option, I agree this is the way to go. As you point out, users can always use an older version (in a virtual env even) if they want to check the validity of the BL calibration; I don't think this is something that is done often anyways. And it is nice to get rid of the alpha hack and use an algorithm that handles bounds naturally.

PR #193, opened 2021-09-15
Support `None` value parameters in calibration results repr

rpauszek:
Why this PR?

Currently a calibration parameter with value `None` throws a `TypeError` when a `CalibrationResults` object is printed. Now it will just be printed as a blank string

### DIFF
--- a/lumicks/pylake/force_calibration/power_spectrum_calibration.py
+++ b/lumicks/pylake/force_calibration/power_spectrum_calibration.py
@@ -84,7 +84,9 @@ class CalibrationResults:
                 [
                     key,
                     f"{param.description}{f' ({param.unit})' if param.unit else ''}",
-                    param.value if isinstance(param.value, str) else f"{param.value:.6g}",
+                    param.value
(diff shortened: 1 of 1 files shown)


JoepVanlier (approved):
Looks good to me.

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -84,7 +84,9 @@ def generate_table(entries):
                 [
                     key,
                     f"{param.description}{f' ({param.unit})' if param.unit else ''}",
-                    param.value if isinstance(param.value, str) else f"{param.value:.6g}",
+                    param.value
+                    if isinstance(param.value, str)
+                    else ("" if param.value is None else f"{param.value:.6g}"),
```
tiagobonetti, line 89: The nested ternaries look confusing, but I have no better suggestion.
Probably pattern matching would be the best option, but we would need python 3.10.

tiagobonetti (approved):
```python
honk if goose is not None else None
```

PR #196, opened 2021-09-24
Document hydrodynamic correction

JoepVanlier:
**Why this PR?**
Because it'd be nice to have the hydrodynamic correction documented prior to release.

Please see a built version of the docs here:  https://lumicks-pylake.readthedocs.io/en/document_hydro_only/tutorial/force_calibration.html

Note that Active Calibration is still out of scope for this PR (still private functionality).


docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,86 @@ We can plot the calibration by calling::
...
+This is illustrated in the figure below.
+Here we see the same power spectrum (acquired on a fast detector) fitted with three different models.
+
+.. image:: fast_hydro.png
+
+Here we can see that the fit with the Lorentzian model with the diode filtering effect seems to fit the data quite well.
+However, when we enable the `fast_sensor` flag, we see that the Lorentzian model doesn't actually describe the data.
+With the diode model disabled, we obtain a very biased fit.
+Enabling the hydrodynamic corrections, we can see that this describes the power spectrum much better.
+Note that for this particular data set the Lorentzian with diode model and hydrodynamically correct model without diode model result in very similar stiffness estimates, no general statement can be made about the bias the former approach occurs in different conditions.
```
tiagobonetti, line 203: This sentence is a bit confusing, let me see if it is correct:

It means using the passive model with `hydrodynamically_correct=True` and `fast_sensor=False` may actually give reasonable results (like the ones in the example) but we don't guarantee anything?

tiagobonetti (commented):
> While these models may look daunting ...

Understatement of the year.

I've read through the whole thing and it reads pretty well.
But I think I'm a bit biased from hearing about this frequently, so maybe we should run this through BD and/or Demolab?

docs/tutorial/force_calibration.rst
JoepVanlier: Hm, no, that's not what I meant with that phrase. Let me rephrase that and see if this is better:

_If we compare the different fits, we can see that the Lorentzian model with diode effect (`fast_sensor=False`) gives similar stiffness estimates as the hydrodynamically correct model without the diode effect. While this is true for this particular dataset, no general statement can be made about the bias of fitting a Lorentzian rather than the hydrodynamically correct power spectrum. If low bias is desired, one should use the hydrodynamically correct model. On regular sensors, it is best to fit the hydrodynamically correct model with the diode model enabled._

But you raise a good additional point. In this case, fitting `hydrodynamically_correct=True` with `fast_sensor=False` actually _also_ gave a good result (it estimated `alpha` to one, effectively disabling the diode effect during the fitting procedure). Again though, I would not recommend that procedure either if you are dealing with a fast sensor.

JoepVanlier:
> > While these models may look daunting ...
> 
> Understatement of the year.
> 
> I've read through the whole thing and it reads pretty well.
> But I think I'm a bit biased from hearing about this frequently, so maybe we should run this through BD and/or Demolab?

Agreed. I'll tag Aafke in the list of reviewers. I'll bring it up with the CST+ group this afternoon as well.

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+
+While the idealized Lorentzian model discussed in the previous section provides a reasonable first description of the physical power spectrum, better models for this spectrum exist.
+
+The idealized model is based on the assumption that the drag force is only proportional to the bead's velocity.
+This assumption is realistic when the bead moves at a constant speed with respect to the fluid.
+When oscillating a bead however, this frictional force becomes frequency dependent.
+
+Equations have been derived by other authors which provide frequency dependent drag coefficients :cite:`tolic2006calibration`.
+These equations can be used to fit a hydrodynamically correct power spectrum:
+
```
aafkevandenberg, line 133: 'The following equation accounts for a frequency dependent drag coefficient  :cite:`tolic2006calibration`:'

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+    f_{m, 0} = \frac{\gamma_0}{2 \pi m}
+
+with :math:`m` the mass of the bead and :math:`\gamma` the frequency dependent drag coefficient.
+Assuming a measurement deep in bulk, these are given by:
+
+.. math::
+
+    \gamma_{stokes} = \gamma_0 \left(1 + (1 - i)\sqrt{\frac{f}{f_{\nu}}} - \frac{2}{9}\frac{f}{f_{\nu}} i\right)
+
```
aafkevandenberg, line 156: First you say \gamma, then \gamma_{stokes}.
It is obvious from the text that these are the same, but to be mathematically accurate, I would either use just one of those, or say \gamma = \gamma_{stokes}

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+Where :math:`f_{\nu}` is given by :math:`4 \nu/(\pi d^2)` with :math:`\nu` the kinematic viscosity.
+
+This approximation is reasonable, when the bead is far from the surface.
+When approaching the surface, the drag experienced by the bead depends on the distance between the center of the bead and the nearest surface of the flow cell.
+An approximate expression for the frequency dependent drag coefficient is then given by :cite:`tolic2006calibration`:
+
+.. math::
+
+    \gamma(f, R/l) = \frac{\gamma_{stokes}(f)}{1 - \frac{9}{16}\frac{R}{l}\left(1 - \left((1 - i)/3\right)\sqrt{\frac{f}{f_{\nu}}} + \frac{2}{9}\frac{f}{f_{\nu}}i - \frac{4}{3}(1 - e^{-(1-i)(2l-R)/\delta})\right)}
+
```
aafkevandenberg, line 166: I would give this \gamma a different name, as it is different from the one above

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+    force_model = lk.PassiveCalibrationModel(bead_diameter, hydrodynamically_correct=True, rho_sample=999, rho_bead=1060.0, distance_to_surface=1e-6)
+
+Note that when `rho_sample` and `rho_bead` are omitted, values for water and polystyrene are used for the sample and bead density respectively.
+
+Additionally, when the parameter `distance_to_surface` is omitted, a simpler model is used which assumes the experiment was performed deep in bulk (neglecting the increased drag induced by the nearby surface).
+
+Fast Sensors
+------------
+
+Fast detectors provide a higher bandwidth, meaning that they do not need the effect of the diode in the model.
```
aafkevandenberg, line 187: At this point, the reader will have forgotten about the diode, so I would repeat what 'the effect of the diode' is. 
And how exactly is this implemented, do you omit the term (1-alpha^2)/(1+(f/fdiode)^2) ?

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+
+.. math::
+
+    f_{c, 0} = \frac{\kappa}{2 \pi \gamma_0}
+
+and
+
+.. math::
+
+    f_{m, 0} = \frac{\gamma_0}{2 \pi m}
```
aafkevandenberg, line 148: explain briefly in words what f_m is

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+Note that when `rho_sample` and `rho_bead` are omitted, values for water and polystyrene are used for the sample and bead density respectively.
+
+Additionally, when the parameter `distance_to_surface` is omitted, a simpler model is used which assumes the experiment was performed deep in bulk (neglecting the increased drag induced by the nearby surface).
+
+Fast Sensors
+------------
+
+Fast detectors provide a higher bandwidth, meaning that they do not need the effect of the diode in the model.
+
+We can omit the diode response model by passing `fast_sensor=True` to the `CalibrationModel`::
```
aafkevandenberg, line 189: When is a sensor fast? Can you give a relation between the bandwidth of the sensor and eg the corner frequency in the power spectrum to say what is fast?

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+
+This approximation is reasonable, when the bead is far from the surface.
+When approaching the surface, the drag experienced by the bead depends on the distance between the center of the bead and the nearest surface of the flow cell.
+An approximate expression for the frequency dependent drag coefficient is then given by :cite:`tolic2006calibration`:
+
+.. math::
+
+    \gamma(f, R/l) = \frac{\gamma_{stokes}(f)}{1 - \frac{9}{16}\frac{R}{l}\left(1 - \left((1 - i)/3\right)\sqrt{\frac{f}{f_{\nu}}} + \frac{2}{9}\frac{f}{f_{\nu}}i - \frac{4}{3}(1 - e^{-(1-i)(2l-R)/\delta})\right)}
+
+where :math:`\delta = R \sqrt{\frac{f_{\nu}}{f}}` and :math:`R` corresponds to the bead radius, :math:`l` to the distance from the bead center to the nearest surface.
```
aafkevandenberg, line 167: explain in words what \delta is

docs/tutorial/force_calibration.rst
```diff
@@ -53,7 +53,9 @@ This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated
 
     P(f) = \frac{D}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha)
 
-where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the fitted cutoff. The second term :math:`g` takes into account the slower response of the position detection system and is given by:
+where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the fitted cutoff.
```
aafkevandenberg, line 56: I am more familiar with the term  'corner frequency' than with 'fitted cutoff'. Maybe good to say cutoff/corner frequency or something, such that people with different backgrounds understand the meaning of this term.

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+This is illustrated in the figure below.
+Here we see the same power spectrum (acquired on a fast detector) fitted with three different models.
+
+.. image:: fast_hydro.png
+
+Here we can see that the fit with the Lorentzian model with the diode filtering effect seems to fit the data quite well.
+However, when we enable the `fast_sensor` flag, we see that the Lorentzian model doesn't actually describe the data.
+With the diode model disabled, we obtain a very biased fit.
+Enabling the hydrodynamic corrections, we can see that this describes the power spectrum much better.
+If we compare the different fits, we can see that the Lorentzian model with diode effect (fast_sensor=False) gives similar stiffness estimates as the hydrodynamically correct model without the diode effect.
```
aafkevandenberg, line 203: Could you explain in 1 sentence why the model with the Lorentzian model with diode effect actually gives a good fit?

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+When oscillating a bead however, this frictional force becomes frequency dependent.
+
+Equations have been derived by other authors which provide frequency dependent drag coefficients :cite:`tolic2006calibration`.
+These equations can be used to fit a hydrodynamically correct power spectrum:
+
+.. math::
+
+    P_{hydro}(f) = \frac{D Re(\gamma / \gamma_0)}{\pi^2 \left(\left(f_{c,0} + f Im(\gamma/\gamma_0) - f^2/f_{m, 0}\right)^2 + \left(f Re(\gamma / \gamma_0)\right)^2\right)}
+
+where
```
aafkevandenberg, line 138: where the cutoff/corner frequency is given by

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
...
+    f_{m, 0} = \frac{\gamma_0}{2 \pi m}
+
+with :math:`m` the mass of the bead and :math:`\gamma` the frequency dependent drag coefficient.
+Assuming a measurement deep in bulk, these are given by:
+
+.. math::
+
+    \gamma_{stokes} = \gamma_0 \left(1 + (1 - i)\sqrt{\frac{f}{f_{\nu}}} - \frac{2}{9}\frac{f}{f_{\nu}} i\right)
+
+Where :math:`f_{\nu}` is given by :math:`4 \nu/(\pi d^2)` with :math:`\nu` the kinematic viscosity.
```
aafkevandenberg, line 157: What is the definition in words for f_{\nu}? This frequency determines when a hydrodynamic correction becomes important. Could you explain in a few words why that is? Do you get turbulence when you go above this frequency?

aafkevandenberg (changes requested):

docs/tutorial/force_calibration.rst
JoepVanlier: Amended.

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,89 @@ We can plot the calibration by calling::
     calibration.plot()
 
 .. image:: force_calibration_fit.png
+
+Hydrodynamics
```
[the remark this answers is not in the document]
JoepVanlier: I'd prefer to use `Hydrodynamically correct model` then since the models are not nested, and correction may imply that they are nested.

docs/tutorial/force_calibration.rst
JoepVanlier: Yes. It just omits the diode model entirely. I've amended it to this to make it more clear:

```
Up to now, we've always fitted a physical spectrum multiplied by a filtering effect.
This filtering effect arose because a fraction of the light is not measured instantaneously by the detector.  
Fast detectors respond much faster, meaning that they do not need to model the slower response time of the diode in the model.
We can omit the diode response model by passing `fast_sensor=True` to the `CalibrationModel`, this removes the diode part from the model entirely
```

docs/tutorial/force_calibration.rst
JoepVanlier: Yes. Good point! :)
Amended!

docs/tutorial/force_calibration.rst
JoepVanlier: I've rephrased it as such:
```
Fast detectors respond much faster resulting in no visible filtering effect by the detector in the frequency range we are fitting, meaning that they do not need to model the slower response time of the diode in the model.
```
Most users will likely not have a fast sensor.

In the future, the hope is to read this from Bluelake and avoid users from having to set this flag themselves.

docs/tutorial/force_calibration.rst
JoepVanlier: I added the following passage:

```
Here :math:`f_{m, 0}` parameterizes the time it takes for friction to dissipate the kinetic energy of the bead with :math:`m` the mass of the bead.
```

docs/tutorial/force_calibration.rst
JoepVanlier: I added the following passage:
```
where :math:`\delta = R \sqrt{\frac{f_{\nu}}{f}}` represents the aforementioned penetration depth and :math:`R` corresponds to the bead radius, :math:`l` to the distance from the bead center to the nearest surface.
```

docs/tutorial/force_calibration.rst
JoepVanlier: Yeah you raise a fair point about all the funky `gamma`'s. I tried to keep it in line with the paper as much as possible, to not confuse people who want to compare this stuff to the paper.

The `gamma`'s you mention are only the same deep in bulk. The final `gamma` is the "real" one. All the others are approximations indicated by their subscript (0 being zero frequency / frequency independent drag, stokes being stokes drag only, and the final gamma being the actual one).

`gamma_0` is the drag coefficient for zero frequency and without any surface effect.

`gamma_{stokes}` is the frequency dependent `gamma`.

`gamma` is the full drag coefficient. This includes frequency dependence and a height effect.

If you are deep in bulk then `gamma` = `gamma_{stokes}`. 

I've given it some more thought and changed it to the following to clarify that a bit more:

```
Assuming a measurement deep in bulk, then :math:`\gamma` = :math:`\gamma_{stokes}` which is given by:
```

Then later as the more complex effects are introduced, it is narrowed down to the final `gamma`:
```
This approximation is reasonable, when the bead is far from the surface.
When approaching the surface, the drag experienced by the bead depends on the distance between the center of the bead and the nearest surface of the flow cell.
An approximate expression for the frequency dependent drag coefficient is then given by :cite:`tolic2006calibration`:
```

Do you think that'd be sufficiently clear?

docs/tutorial/force_calibration.rst
JoepVanlier: Is two okay? 😅 

```
As we can see in the comparisons above with the hydrodynamics on and off, including hydrodynamics attenuates higher frequencies (an effect similar to a low pass filter).
In the case of the Lorentzian with a diode model, the fitting procedure has used the diode model to fit some of this high frequency attenuation.
```

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
JoepVanlier: I'd personally prefer not to deviate from the paper here because it may be confusing for people comparing it to the papers.

Including the previous block, do you think it is sufficiently clearly explained in the docs now?

Or do you know of a good way to make it more clear in the accompanying text.

docs/tutorial/force_calibration.rst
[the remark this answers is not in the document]
aafkevandenberg: 'Hydronamically correct model' sounds perfect! I just didn't like 'Hydrodynamics', as it was too general.

docs/tutorial/force_calibration.rst
aafkevandenberg: Ah, I understand it much better now. I didn't realize the gamma's with the subscripts were all approximations. The current version is sufficiently clear.

docs/tutorial/force_calibration.rst
aafkevandenberg: I like the way it is now

docs/tutorial/force_calibration.rst
aafkevandenberg: Ah yes, I am familiar with the term penetration depth, I think you can leave out the definition. The way it is written now, it looks like the formula at the end of the sentence ( 4 \nu/(\pi d^2) is the penetration depth. 

How about: 

Where :math:`f_{\nu}` is the frequency at which the penetration depth in the liquid of the bead’s linear harmonic motion equals the radius of the bead, 
`f_{\nu}=4 \nu/(\pi d^2)`
with \nu the kinematic viscosity

aafkevandenberg (approved):

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,94 @@ We can plot the calibration by calling::
...
+
+This approximation is reasonable, when the bead is far from the surface.
+When approaching the surface, the drag experienced by the bead depends on the distance between the center of the bead and the nearest surface of the flow cell.
+An approximate expression for the frequency dependent drag coefficient is then given by :cite:`tolic2006calibration`:
+
+.. math::
+
+    \gamma(f, R/l) = \frac{\gamma_{stokes}(f)}{1 - \frac{9}{16}\frac{R}{l}\left(1 - \left((1 - i)/3\right)\sqrt{\frac{f}{f_{\nu}}} + \frac{2}{9}\frac{f}{f_{\nu}}i - \frac{4}{3}(1 - e^{-(1-i)(2l-R)/\delta})\right)}
+
+Where :math:`\delta = R \sqrt{\frac{f_{\nu}}{f}}` represents the aforementioned penetration depth and :math:`R` corresponds to the bead radius, :math:`l` to the distance from the bead center to the nearest surface.
```
aafkevandenberg, line 169: ....depth,  :math:`R` corresponds to the bead radius _and_ :math:`l` to the.....

docs/tutorial/force_calibration.rst
```diff
@@ -104,3 +118,94 @@ We can plot the calibration by calling::
...
+Note that when `rho_sample` and `rho_bead` are omitted, values for water and polystyrene are used for the sample and bead density respectively.
+
+Additionally, when the parameter `distance_to_surface` is omitted, a simpler model is used which assumes the experiment was performed deep in bulk (neglecting the increased drag induced by the nearby surface).
+
+Fast Sensors
+------------
+
+Up to now, we've always fitted a physical spectrum multiplied by a filtering effect.
+This filtering effect arose because a fraction of the light is not measured instantaneously by the detector.
+Fast detectors respond much faster resulting in no visible filtering effect by the detector in the frequency range we are fitting, meaning that they do not need to model the slower response time of the diode in the model.
```
aafkevandenberg, line 191: meaning that they do not need to model the slower response time of the diode in the model.

I assume the last words, 'in the model' can be omitted?

docs/tutorial/force_calibration.rst
JoepVanlier: Ah yes, that sounds much better! Thanks!

docs/tutorial/force_calibration.rst
JoepVanlier: Sounds good to me! :)

PR #198, opened 2021-09-29
force calibration: add residual plot

JoepVanlier:
**Why this PR?**
This functionality adds the ability to quickly make a residual plot of the fit. These can be helpful to diagnose issues with the models as it is much easier to see model deviations on these. I find myself writing this code frequently, so I figured it'd be good to just have it in as a feature.

While fit 1 and 3 look great at a glance, the third actually still has quite some deviance if we look at the actual model residuals.

Fit:
image

Residual plot:
image


mcsmurf (approved):
Code looks solid, great idea for visual diagnosis! 🚀

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -78,6 +78,26 @@ def plot(self):
         self.ps_model.plot(label="Model")
         plt.legend()
 
+    def plot_residual(self):
```
mcsmurf, line 81: nitpick: Could this be `plot_spectrum_residuals(self)` instead? Would save extra googling what residuals we're talking about 😅

PR #202, opened 2021-10-13
Force calibration: Drag correction

JoepVanlier:
**Why this PR?**
Both for distances very close to the surface as well as axial force, the hydrodynamic model is less than suitable.
In these cases it is recommended to just use a simple Lorentzian, but take the distance to the nearest surface into account through the drag coefficient.

The effective drag coefficient at a particular distance from the surface is given by:
`gamma(l) = gamma_bulk * height_factor(l)`

This PR adds this option.

**More info**
When the hydrodynamic flag is set to off, but a height is specified, it will apply a correct correction on the drag force.
Note that the correction is different for axial and lateral force! For this an extra flag was added to the calibration models named `axial`.

Note that in the future, we will rely on carrying over drag coefficients from lateral to axial calibrations. The way I set up the models in this PR, this is all handled by the calibration models themselves. When doing any sort of calibration, the model will always return "bulk" values. Then the models only responsibility is to forward correct (and they don't need to know where the bulk value came from). This decouples the from and target model.

Implicitly, it would amount to the same as calculating a local drag coefficient for lateral, uncorrecting it with Faxen's law and then correcting the result with Brenner's law. While we explicitly test the lateral -> axial behaviour in a test, we do not document this yet, since a good API for this procedure is still t.b.d.

I have also added a test to specifically test the transfer of drag coefficients from lateral to axial calibration.

**Recommended reading (if you are interested)**: 
Schäffer, E., Nørrelykke, S. F., & Howard, J. (2007). Surface forces and drag coefficients of microspheres near a plane surface measured with optical tweezers. Langmuir, 23(7), 3654-3665.
https://publications.mpi-cbg.de/Sch%C3%A4ffer_2007_894.pdf

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -188,6 +188,53 @@ Note that when `rho_sample` and `rho_bead` are omitted, values for water and pol
 
 Additionally, when the parameter `distance_to_surface` is omitted, a simpler model is used which assumes the experiment was performed deep in bulk (neglecting the increased drag induced by the nearby surface).
 
+Faxen's law
+-----------
+
+The hydrodynamically correct model presented in the previous section works well when the bead center is at least 1.5 times the radius above the surface.
+
+When going closer, the drag effect becomes stronger than the frequency dependent effects and better models to approximate the local drag exist.
+
+For lateral calibration, the following approximation is typically used :cite:`schaffer2007surface`:
+
+.. math::
+
+    \gamma_\mathrm{faxen}(R/l) = \frac{\gamma_0}{
+        1 - \frac{9R}{16l} + \frac{1R^3}{8l^3} - \frac{45R^4}{256l^4} - \frac{1R^5}{16l^5}
+    }
+
+We can use this model by setting `hydrodynamically_correct` to `False`, while still providing a distance to the surface::
+
+    force_model = lk.PassiveCalibrationModel(bead_diameter, hydrodynamically_correct=False, distance_to_surface=1e-6)
+
+Note that `pylake` always returns the bulk drag coefficient :math:`\gamma_0`.
+
+Axial Calibration
+-----------------
+
+For calibration in the axial direction, no hydrodynamically correct theory exists.
+In this case, one should use a Lorentzian with a specific correction term :cite:`schaffer2007surface`:
+
+.. math::
+
+    \gamma_\mathrm{axial}(R/l) = \frac{\gamma_0}{
+        1.0
+        - \frac{9R}{8l}
+        + \frac{1R^3}{2l^3}
+        - \frac{57R^4}{100l^4}
+        + \frac{1R^5}{5l^5}
+        + \frac{7R^{11}}{200l^{11}}
+        - \frac{1R^{12}}{25l^{12}}
+    }
+
+This model deviates less than 0.1% from Brenner's exact formula for :math:`l/R >= 1.1` and less than 0.3% over the entire range of :math:`l` :cite:`schaffer2007surface`:.
+
+This model can be used in Pylake by specifying `axial=True`::
+
+    force_model = lk.PassiveCalibrationModel(bead_diameter, distance_to_surface=1e-6, axial=True)
+
+Note that no hydrodynamically correct model is available for axial calibration.
+
 Fast Sensors
 ------------
 
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -593,10 +593,12 @@ class ActiveCalibrationModel(PassiveCalibrationModel):
             "kappa": CalibrationParameter("Trap stiffness", kappa * 1e3, "pN/nm"),
             "Rf": CalibrationParameter("Force response", force_response * 1e12, "pN/V"),
             "gamma_0": CalibrationParameter(
-                "Theoretical drag coefficient", self.drag_coeff, "kg/s"
+                "Theoretical bulk drag coefficient", self.drag_coeff, "kg/s"
             ),
             "gamma_ex": CalibrationParameter(
-                "Measured drag coefficient", measured_drag_coeff, "kg/s"
+                "Measured bulk drag coefficient",
+                measured_drag_coeff / self._drag_correction_factor,
+                "kg/s",
             ),
             **self._format_passive_result(
                 fc,
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -6,6 +6,7 @@ from .detail.driving_input import (
     estimate_driving_input_parameters,
     driving_power_peak,
 )
+from .detail.drag_models import faxen_factor, brenner_axial
 from .detail.power_models import (
     passive_power_spectrum_model,
     sphere_friction_coefficient,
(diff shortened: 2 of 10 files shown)


lumicks/pylake/force_calibration/tests/test_active_calibration.py
```diff
@@ -139,3 +139,63 @@ def test_bias_correction():
 
     assert fit_biased.params["Bias correction"].value is False
     assert fit_debiased.params["Bias correction"].value is True
+
+
+def test_faxen_correction_active():
+    """Active calibration should barely be affected by surface corrections for the drag coefficient.
+    However, the interpretation of gamma_ex, which may be carried over to the other calibration
+    *is* important, so this should be covered by a specific test."""
```
dean0x7d, line 147: The test docstrings are excellent! Great to have the purpose of the test explained!

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -158,13 +159,15 @@ class PassiveCalibrationModel:
         Enable hydrodynamic correction.
     distance_to_surface : float, optional
         Distance from bead center to the surface [um].
-        Currently not supported for models without hydrodynamics.
+        None uses the approximation valid for deep in bulk.
```
dean0x7d, line 162: It took me a while to understand this sentence. Not sure I have a good suggestion to rephrase it, but maybe you do. (It's also good to use backticks to distinguish `None` as a keyword.)

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -228,18 +241,28 @@ def __init__(
                 else self.distance_to_surface * 1e-6,  # um => m
             )
         else:
-            if self.distance_to_surface:
-                raise NotImplementedError(
-                    "Using a distance to the surface is currently not supported for models without "
-                    "hydrodynamic effects."
+            if distance_to_surface:
+                args = [distance_to_surface * 1e-6, bead_diameter * 1e-6 / 2.0]
```
dean0x7d, line 245: Tuples are a better default for immutable data. They better convey the intent and are also faster (not that performance matters here, but it's a better default).
```suggestion
                args = (distance_to_surface * 1e-6, bead_diameter * 1e-6 / 2.0)
```

dean0x7d (approved):
This looks pretty good to me! I only have a couple of minor comments below. And I assume that you'd want to add something to the changelog.

lumicks/pylake/force_calibration/tests/test_hydro.py
```diff
@@ -434,7 +435,33 @@ def test_distance_to_surface_input(integration_test_parameters):
...
     ActiveCalibrationModel(distance_to_surface=None, hydrodynamically_correct=True, **pars)
 
+
+def test_invalid_model():
     with pytest.raises(NotImplementedError):
-        ActiveCalibrationModel(
-            distance_to_surface=0.5 * 1.51, hydrodynamically_correct=False, **pars
-        )
+        PassiveCalibrationModel(1.0, hydrodynamically_correct=True, axial=True)
+
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, good point. I tried to make a habit out of doing this. Not sure why I overlooked it this time. Thanks for pointing it out.

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Yeah, fair point. I'll rephrase it to:
```python

File: lumicks/pylake/force_calibration/calibration_models.py (hunk 3 above)
Comment (JoepVanlier) [reply lumicks/pylake/force_calibration/calibration_models.py:245]: Good point!

File: lumicks/pylake/force_calibration/calibration_models.py [hunk 5]
```diff
@@ -158,13 +159,16 @@ class PassiveCalibrationModel:
         Enable hydrodynamic correction.
     distance_to_surface : float, optional
         Distance from bead center to the surface [um].
-        Currently not supported for models without hydrodynamics.
+        When specifying `None`, the model will use an approximation which is only suitable for
+        measurements performed deep in bulk.
```
lpingas, line 163: Now that this is supported without hydro, the check for `hydrodynamically_correct` on line 192 has to be removed.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -209,6 +215,14 @@ def __init__(
...
+                raise NotImplementedError(
+                    "No hydrodynamically correct axial force model is currently available."
+                )
+
+            # The hydrodynamically correct model already accounts for the effect of a nearby wall.
+            # Here the drag coefficient in the model represents the bulk drag coefficient.
+            self._drag_correction_factor = 1
             # This model is only valid up to l/R < 1.5 [6] so throw in case that is violated.
             if distance_to_surface and distance_to_surface / (self.bead_diameter / 2) < 1.5:
                 raise ValueError(
```
lpingas, line 228: Not really related to this PR, but I noticed during testing that this message could be a little bit clearer:
It raises
``` 
"This model is only valid for distances to the surface ..."
```
but it could be something like
```
"A hydrodynamically correct model is only valid for distances to the surface ..."
```

In this way, the user can figure out that hydro can be disabled if they are working so close to the surface.

dean0x7d: The error message could also include explicit troubleshooting advice: "If you are working very close to the surface, turn off the hydrodynamic correction."

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Damn! Good catch!

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Good ideas. I've combined the two suggestions into:
```python
                    "The hydrodynamically correct model is only valid for distances to the surface "
                    "larger than 1.5 times the bead radius. For distances closer to the surface, "
                    "turn off the hydrodynamically correct model."
```

rpauszek (approved):
Cool, looks great. just 2 very minor comments about documentation

docs/tutorial/force_calibration.rst
```diff
@@ -186,6 +186,49 @@ Note that when `rho_sample` and `rho_bead` are omitted, values for water and pol
...
+        + \frac{1R^5}{5l^5}
+        + \frac{7R^{11}}{200l^{11}}
+        - \frac{1R^{12}}{25l^{12}}
+    }
+
+This model deviates less than 0.1% from Brenner's exact formula for :math:`l/R >= 1.1` and less than 0.3% over the entire range of :math:`l` :cite:`schaffer2007surface`:.
+
+This model can be used in Pylake by specifying `axial=True`::
+
+    force_model = lk.PassiveCalibrationModel(bead_diameter, hydrodynamically_correct=False, distance_to_surface=1e-6, axial=True)
```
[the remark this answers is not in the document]
JoepVanlier: Right. Good point. I'll leave `hydrodynamically_correct=False` out entirely. It's not relevant here.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -228,18 +238,28 @@ def __init__(
...
         physical_spectrum = self._passive_power_spectrum_model(f, fc, diffusion_constant)
         return physical_spectrum * self._filter(f, *filter_params)
 
+    @property
+    def _drag(self):
+        """Returns the corrected drag coefficient
+
+        Note that when using the hydrodynamic model, no correction is necessary.
+        """
+        return self.drag_coeff * self._drag_correction_factor
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, it's a good point. In ten years we'll be thankful for that comment. I added the following:
```
Note that the hydrodynamic model already has a surface dependent drag coefficient baked in and therefore no correction is necessary (in this case self._drag_correction_factor is set to 1).
```

PR #204, opened 2021-10-18
Deprecate `CorrelatedStack.raw`

rpauszek:
**Why this PR?**

`CorrelatedStack.raw` exposes a list of internal `widefield.TiffFrame` instances. Additionally, the output is dependent on the number of frames in the stack (returns a `list` if `num_frames > 1` or a single `TiffFrame` if `num_frames == 1`). 

 Likely the main use for this property is in constructing a `np.array` of the frame data, which will be implemented as a public method in an upcoming PR. Any other uses for the output of this property in the future should also be implemented as a public method/property directly on `CorrelatedStack` or (for internal use) can still be accessed by `[stack._get_frame(j) for j in range(stack.num_frames)]`.

### DIFF
--- a/lumicks/pylake/tests/test_correlated_stack.py
+++ b/lumicks/pylake/tests/test_correlated_stack.py
@@ -105,11 +105,11 @@ def test_correlation(shape):
     with pytest.raises(AssertionError):
         cc["40ns":"70ns"].downsampled_over(stack[0:1].timestamps)
 
-    assert (stack[0].raw.start == 10)
-    assert (stack[1].raw.start == 20)
-    assert (stack[1:3][0].raw.start == 20)
-    assert (stack[1:3].raw[0].start == 20)
-    assert (stack[1:3].raw[1].start == 30)
+    assert (stack[0]._get_frame(0).start == 10)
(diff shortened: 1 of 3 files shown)


lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -105,11 +105,11 @@ def test_correlation(shape):
...
-    assert (stack[0].raw.start == 10)
-    assert (stack[1].raw.start == 20)
-    assert (stack[1:3][0].raw.start == 20)
-    assert (stack[1:3].raw[0].start == 20)
-    assert (stack[1:3].raw[1].start == 30)
+    assert (stack[0]._get_frame(0).start == 10)
+    assert (stack[1]._get_frame(0).start == 20)
+    assert (stack[1:3]._get_frame(0).start == 20)
+    assert (stack[1:3]._get_frame(0).start == 20)
+    assert (stack[1:3]._get_frame(1).start == 30)
```
mikhas, line 112: The deprecation message talks about using the public API, but `_something` is deemed private API, no? Shouldn't the tests also use the public API?

PR #208, opened 2021-10-27
force_calibration: document and release active calibration

JoepVanlier:
**Why this PR?**
This PR documents active calibration and makes it part of the public API.

Docs build here:
https://lumicks-pylake.readthedocs.io/en/active_calibration_docs/tutorial/force_calibration.html#active-calibration

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -47,49 +47,51 @@ Additional parameters can be specified to change the amount of downsampling appl
     power_spectrum = lk.calculate_power_spectrum(volts, sample_rate=force_slice.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=[(700, 800), (14500, 14600)])
 
 To fit the passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+Passive calibration is also often referred to as thermal calibration.
 This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
 
 .. math::
 
-    P(f) = \frac{D}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha)
+    P(f) = \frac{D_\mathrm{measured}}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_\mathrm{diode}, \alpha) \quad \mathrm{[V^2/Hz]}
 
-where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the corner frequency.
+where :math:`D_\mathrm{measured}` corresponds to the diffusion constant (in `volts`), :math:`f` the frequency and :math:`f_c` the corner frequency.
 The first part of the equation is known as the Lorentzian model and represents the part of the spectrum that originates from the physical motion of the bead.
 The second term :math:`g` takes into account the slower response of the position detection system and is given by:
 
 .. math::
 
-    g(f, f_{diode}, \alpha) = \alpha^2 + \frac{1 - \alpha ^ 2}{1 + (f / f_{diode})^2}
+    g(f, f_\mathrm{diode}, \alpha) = \alpha^2 + \frac{1 - \alpha ^ 2}{1 + (f / f_\mathrm{diode})^2}
 
-Here :math:`\alpha` corresponds to the fraction of the signal response that is instantaneous, while :math:`f_{diode}` characterizes the frequency response of the diode. Note that not all sensor types require this second term.
+Here :math:`\alpha` corresponds to the fraction of the signal response that is instantaneous, while :math:`f_\mathrm{diode}` characterizes the frequency response of the diode. Note that not all sensor types require this second term.
 
 To convert the parameters obtained from this spectral fit to a trap stiffness, the following is computed:
 
 .. math::
 
-    \kappa = 2 \pi \gamma_0 f_c
+    \kappa = 2 \pi \gamma_0 f_c \quad \mathrm{[N/m]}
 
 Here :math:`\kappa` then represents the final trap stiffness.
 
-We can calibrate the position by considering the diffusion of the bead in squared microns per second:
+We can calibrate the position by considering the diffusion of the bead:
 
 .. math::
 
-    D = \frac{k_B T}{\gamma_0}
+    D_\mathrm{physical} = \frac{k_B T}{\gamma_0} \quad \mathrm{[m^2/s]}
 
-Here :math:`k_B` refers to the Boltzmann constant and :math:`T` reflects the temperature in Kelvin. Comparing this to its measured counterpart in volts squared per second provides us with the desired calibration factor:
-
-.. math::
-
-    \beta = \sqrt{\frac{D_{physical}}{D_{volts}}}
-
-Both of these quantities depend on the parameter :math:`\gamma_0`, which corresponds to the friction coefficient of a sphere and is given by:
+Here :math:`k_B` is the Boltzmann constant and :math:`T` is the local temperature in Kelvin. Comparing this to its measured counterpart in volts squared per second provides us with the desired calibration factor:
 
 .. math::
 
-    \gamma_0 = 3 \pi \eta d
+    R_d = \sqrt{\frac{D_\mathrm{physical}}{D_\mathrm{measured}}} \quad \mathrm{[m/V]}
 
-where :math:`\eta` corresponds to the dynamic viscosity and :math:`d` represents the bead diameter.
+Both of these quantities depend on the parameter :math:`\gamma_0`, which corresponds to the drag coefficient of a sphere and is given by:
+
+.. math::
+
+    \gamma_0 = 3 \pi \eta d \quad \mathrm{[kg/s]}
+
+where :math:`\eta` corresponds to the dynamic viscosity [Pa*s] and :math:`d` is the bead diameter [m].
+Note that in `pylake` we actually use `microns` to specify the bead diameter for convenience.
 
 We use the bead diameter found in the calibration performed in Bluelake.
 You can optionally also provide a viscosity (in Pa/s) and temperature (in degrees Celsius)::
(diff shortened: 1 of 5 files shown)


docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+
+    W_{experiment} = \left(P^{volt}(f_{drive}) - P^{volt}_{thermal}(f_{drive})\right) \Delta f
+
+Here the thermal contribution that needs to be subtracted is obtained from fitting the thermal part of the spectrum using the passive calibration procedure from before.
+The desired positional calibration is then:
+
+.. math::
+
+    \beta = \sqrt{\frac{W_{theory}}{W_{experiment}}}
+
```
aafkevandenberg, line 264: The expressions of W_theory and W_experiment now look identical, as all the variables have exactly the same name. So it looks as if this ratio is always 1. Maybe explaining the superscript 'volt' will make more clear what this expression means exactly.

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+To check the determined frequency, we can look at the determined driving frequency::
+
+    >>> active_model.driving_frequency
+    36.95
+
+We can now use this model to fit the power spectrum::
+
+    calibration = lk.fit_power_spectrum(power_spectrum, active_model)
+    calibration
+
```
aafkevandenberg, line 295: Nice! I like how this complex computation now looks so simple.

aafkevandenberg (changes requested):
Hi Joep, looks really nice. I like how Pylake makes complex analyses look simple. Just added some comments to make the explanation more explicit.

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+Sometimes, more accuracy is required.
+One way to achieve this is with active force calibration.
+By providing a known perturbation to the bead, active calibration uses fewer assumptions than passive calibration.
+
+When performing passive calibration, we rely on knowledge of the drag coefficient to obtain a trap stiffness and positional calibration.
+This theoretical drag coefficient depends on parameters that are only known with limited precision.
+These parameters are the diameter of the bead and the viscosity.
+This viscosity in turn depends strongly on the local temperature around the bead, which is typically also poorly known.
+
```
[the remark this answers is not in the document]
JoepVanlier: Paragraph 1 and 3 I fully agree.

For paragraph two I would rephrase it very slightly, because the drag coefficient is not estimated (which is actually what you indirectly do in active), but assumed, which is actually the problem.

What do you think of the following for paragraph 2?
```
When performing passive calibration, we base our calculations on a theoretical drag coefficient.
This theoretical drag coefficient depends on parameters that are only known with limited precision: the diameter of the bead and the viscosity.
This viscosity in turn depends strongly on the local temperature around the bead, which is typically poorly known.
```

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+When performing active calibration, either the nanostage or trap is oscillated with a specific amplitude.
+This results in a visible driving peak at that particular frequency in the power spectrum.
+If we choose the windowing to include a full number of periods of the driving input, then the resulting driving peak will end up in a single data point in the frequency spectrum.
+
```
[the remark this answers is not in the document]
JoepVanlier: Correct yes.

I think your phrasing is a lot more clear than mine. The only thing that I would change is that it's a peak at that frequency rather than that frequency itself. What do you think of the following:
```When the power spectrum is computed from an integer number of oscillations, the driving peak is visible at a single data point at :math:`f_{drive}`.```

In practice, what we actually do is chop the time series up into blocks that are an integer number of oscillations and average the spectra resulting from those (otherwise you get a bias from the fact that the errors on power spectral densities are not normally distributed).

File: docs/tutorial/force_calibration.rst [hunk 5]
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+.. math::
+
+    P_{thermal}(f) = \frac{D}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)}
+
+    P_{active}(f) = \frac{A^2}{2\left(1 + \frac{f_c^2}{f_{drive}^2}\right)} \delta(f - f_{drive})
+
+    P(f) = P_{thermal}(f) + P_{active}(f)
+
+Since we know the driving amplitude, we know how the bead reacts to the driving motion and we can observe this response in the PSD, we can use this relation to determine the positional calibration.
+
```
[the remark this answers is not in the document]
JoepVanlier: Indeed! Amended! :)

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+.. math::
+
+    W_{theory} = \frac{A^2}{2\left(1 + \frac{f_c^2}{f_{drive}^2}\right)}
+
+Subtracting the thermal part of the spectrum, we can determine the same quantity experimentally.
+
+.. math::
+
+    W_{experiment} = \left(P^{volt}(f_{drive}) - P^{volt}_{thermal}(f_{drive})\right) \Delta f
+
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,98 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+    \beta = \sqrt{\frac{W_{theory}}{W_{experiment}}}
+
+Note how this time around, we did not rely on assumptions on the viscosity of the medium or the bead size.
+
+As a side effect of this calibration, we actually obtain an experimental estimate of the drag coefficient:
+
+.. math::
+
+    \gamma_{measured} = \frac{k_B T}{\beta^2 D^{volt}}
+
```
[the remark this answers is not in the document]
JoepVanlier: Good call!

docs/tutorial/force_calibration.rst
JoepVanlier: I could add a superscript to the `W`s that says `volt` or `physical`.

I'm actually beginning to doubt a little whether it shouldn't be `sensor` vs `physical`, because the integrated PSD is actually in volts squared, and calling it Volt may be a little confusing. Then again, after the previous change, it should be clear what the superscript means now.

What do you think?

docs/tutorial/force_calibration.rst
JoepVanlier: That's the idea! :)

Hopefully we can keep it that way as we get to more and more complex methods :D

docs/tutorial/force_calibration.rst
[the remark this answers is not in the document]
aafkevandenberg: Sounds good!

And on a second though, I believe the first sentence of the 3rd paragraph should be: During active calibration, the trap or nanostage _is_ oscillated sinusoidally.

docs/tutorial/force_calibration.rst
aafkevandenberg: I actually find volt vs physical confusing, as Volt is a proper unit, but physical is not.
It would be clearest to give the units explicitly. How about using the labels `sensor` and `physical `and then specify the specific units in the text?

docs/tutorial/force_calibration.rst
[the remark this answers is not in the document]
JoepVanlier: Ah yes, good catch! Fixing...

aafkevandenberg (approved):

docs/tutorial/force_calibration.rst
JoepVanlier: I agree. Poor choice on my part there.

How about `measured` and `physical` and we add units to the equations most relevant to the positional calibration.

I've just pushed a version with those changes.

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,102 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+During active calibration, the trap or nanostage is oscillated sinusoidally.
+These oscillations result in a driving peak in the force spectrum.
+Using power spectral analysis, the force can then be calibrated without prior knowledge of the drag coefficient.
+
+When the power spectrum is computed from an integer number of oscillations, the driving peak is visible at a single data point at :math:`f_{drive}`.
+
+.. image:: driving_input.png
+
+The physical spectrum is then given by a thermal part (like before) and an active part:
```
rpauszek, line 234: this is the first time in the docs that we refer to the passive Lorentzian as "thermal". It might be good to mention this term up above when talking about the thermal/passive fit just for continuity and context

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,102 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+As a side effect of this calibration, we actually obtain an experimental estimate of the drag coefficient:
+
+.. math::
+
+    \gamma_{measured} = \frac{k_B T}{R_d^2 D_{measured}} \quad \mathrm{[kg/s]}
+
+How to do active calibration in Pylake
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Using `Pylake`, the procedure to use active calibration is not very different from passive calibration.
```
rpauszek, line 277: `pylake` should be lowercase since we're referring to the package (code-formatted)

docs/tutorial/force_calibration.rst
```diff
@@ -212,3 +212,102 @@ If we compare the different fits, we can see that the Lorentzian model with diod
...
+
+.. math::
+
+    P_{hydro, active}(f) = \delta \left(f - f_{drive}\right) \frac{\left(A f_{drive} \left|\gamma / \gamma_0\right|\right)^2}{2 \left(\left(f_{c,0} + f Im(\gamma/\gamma_0) - f^2/f_{m, 0}\right)^2 + \left(f Re(\gamma / \gamma_0)\right)^2\right)}
+
+We can also include a distance to the surface like before.
+This results in the expression for `\gamma` becoming dependent on the distance to the surface.
+This uses the same expression as listed in the section on the :ref:`hydrodynamically correct model<Hydrodynamically correct model>`.
+One thing to note is that when using the hydrodynamically correct model, the equation for the drag _does_ include the viscosity and bead diameter.
+However, they now appear in a term which already amounts to a small correction therefore the impact of any errors in these is reduced :cite:`tolic2006calibration`.
```
rpauszek, line 313: I might split this into 2 paragraphs with the 2nd paragraph starting at `One thing to note...`. Otherwise, to me, this all just looks like a giant block of text and the note might get overlooked

rpauszek (approved):
Awesome documentation, definitely distills down a very complicated analysis to understandable bits

Just some minor comments on some of the wording. I'll also add a commit for some formatting (getting rid of italics for some labels) and you can choose whether or not it looks better

docs/tutorial/force_calibration.rst
```diff
@@ -51,9 +51,9 @@ This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated
 
 .. math::
 
-    P(f) = \frac{D}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha)
+    P(f) = \frac{D_{measured}}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_{diode}, \alpha) \quad \mathrm{[V^2/Hz]}
 
-where :math:`D` corresponds to the diffusion constant, :math:`f` the frequency and :math:`f_c` the corner frequency.
+where :math:`D_{measured}` corresponds to the diffusion constant (in `Volts`), :math:`f` the frequency and :math:`f_c` the corner frequency.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
```diff
@@ -67,27 +67,27 @@ To convert the parameters obtained from this spectral fit to a trap stiffness, t
...
 
-We can calibrate the position by considering the diffusion of the bead in squared microns per second:
+We can calibrate the position by considering the diffusion of the bead:
 
 .. math::
 
-    D = \frac{k_B T}{\gamma_0}
+    D_{physical} = \frac{k_B T}{\gamma_0} \quad \mathrm{[m^2/s]}
 
 Here :math:`k_B` refers to the Boltzmann constant and :math:`T` reflects the temperature in Kelvin. Comparing this to its measured counterpart in volts squared per second provides us with the desired calibration factor:
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
```diff
@@ -67,27 +67,27 @@ To convert the parameters obtained from this spectral fit to a trap stiffness, t
...
 
 .. math::
 
-    \beta = \sqrt{\frac{D_{physical}}{D_{volts}}}
+    R_d = \sqrt{\frac{D_{physical}}{D_{measured}}} \quad \mathrm{[m/V]}
 
 Both of these quantities depend on the parameter :math:`\gamma_0`, which corresponds to the friction coefficient of a sphere and is given by:
```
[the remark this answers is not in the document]
JoepVanlier: Ah yes. Renamed to drag. I think my mind was probably still thinking about frictional force. It is definitely the drag coefficient yes.

docs/tutorial/force_calibration.rst
JoepVanlier: Good point. I've added `Passive calibration is also often referred to as thermal calibration.` to the section above.

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
JoepVanlier: 👍

docs/tutorial/force_calibration.rst
```diff
@@ -67,27 +67,27 @@ To convert the parameters obtained from this spectral fit to a trap stiffness, t
...
 
 .. math::
 
-    \gamma_0 = 3 \pi \eta d
+    \gamma_0 = 3 \pi \eta d \quad \mathrm{[kg/s]}
 
 where :math:`\eta` corresponds to the dynamic viscosity and :math:`d` represents the bead diameter.
```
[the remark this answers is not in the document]
JoepVanlier: Added!

PR #209, opened 2021-11-02
Prevent conversion of `timestamps` to floating point during image reconstruction.

JoepVanlier:
**Why this PR?**
Currently, `reconstruct_image` converts `Scan` and `Kymo` timestamps to a floating point representation. This is undesirable.

After the fix, `reconstruct_image` will provide as output the same type as the input.

To prevent changing two things at once, we enforce the type on image reconstruction to remain `float` (since certain downstream algorithms have issues with fixed precision types).

Note that I have hardened the tests for tests involving `timestamps` as well by providing a more realistic offset. I made these two separate commits, so I would recommend going commit by commit.

### DIFF
--- a/docs/tutorial/file.rst
+++ b/docs/tutorial/file.rst
@@ -105,6 +105,14 @@ The channels have a few convenient methods, like `.plot()` which make it easy to
     plt.plot(f1x_timestamps, f1x_data)
 
 The `timestamps` attribute returns absolute values in nanoseconds.
+Note that `timestamps` cannot be converted to floating point without losing precision::
+
+    >>> t = f1x_timestamps[0]
+    >>> roundtrip_t = np.int64(np.float64(t))
+    >>> print(t - roundtrip_t)
+    24
+
+The reason for this is that timestamps exceed the maximum integer value representable by the mantissa.
 The relative time values in seconds can also be accessed directly::
 
     f1x_seconds = file.force1x.seconds
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -8,12 +8,16 @@ from .mixin import ExcitationLaserPower
 from .image import reconstruct_image_sum, reconstruct_image, save_tiff, ImageMetadata
 
 
-def _timestamp_mean(a, axis):
-    """Compute mean of consecutive timestamps.
+def _contiguous_timestamp_mean(a, axis):
+    """Compute mean of contiguous timestamps.
 
-    Note: This function should only be used for consecutive timestamps. It is used to avoid
+    Note: This function should only be used for contiguous timestamps. It is used to avoid
     incurring an integer overflow when computing the mean of a series of timestamps."""
-    return (np.min(a, axis=axis) + np.max(a, axis=axis)) // 2  # We lose at most half a nanosecond
+    second_derivative = np.diff(a, n=2, axis=axis)
+    assert np.all(
+        second_derivative == 0
+    ), "This function should only be used for contiguous timestamps"
+    return (np.take(a, 0, axis=axis) + np.max(a, axis=axis)) // 2  # Lose at most half a nanosecond
 
 
 def _default_image_factory(self: "ConfocalImage", color):
@@ -22,7 +26,7 @@ def _default_image_factory(self: "ConfocalImage", color):
     return self._to_spatial(raw_image)
 
 
-def _default_timestamp_factory(self: "ConfocalImage", reduce=_timestamp_mean):
+def _default_timestamp_factory(self: "ConfocalImage", reduce=_contiguous_timestamp_mean):
     # Uses the timestamps from the first non-zero-sized photon channel
     for color in ("red", "green", "blue"):
         channel_data = getattr(self, f"{color}_photon_count").timestamps
@@ -228,7 +232,7 @@ class ConfocalImage(BaseScan):
         return self._image_factory(self, channel)
 
     @cachetools.cachedmethod(lambda self: self._cache)
-    def _timestamps(self, channel, reduce=_timestamp_mean):
+    def _timestamps(self, channel, reduce=_contiguous_timestamp_mean):
         assert channel == "timestamps"
         return self._timestamp_factory(self, reduce)
 
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -18,7 +18,7 @@ def _timestamp_mean(a, axis):
 
 def _default_image_factory(self: "ConfocalImage", color):
     channel_data = getattr(self, f"{color}_photon_count").data
-    raw_image = reconstruct_image_sum(channel_data, self.infowave.data, self._shape)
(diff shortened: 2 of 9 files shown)


docs/tutorial/file.rst
```diff
@@ -105,6 +105,13 @@ The channels have a few convenient methods, like `.plot()` which make it easy to
     plt.plot(f1x_timestamps, f1x_data)
 
 The `timestamps` attribute returns absolute values in nanoseconds.
+Note that `timestamps` cannot be converted to floating point without losing precision::
+
+    >>> t = f1x_timestamps[0]
+    >>> roundtrip_t = np.int64(np.double(t))
```
dean0x7d, line 111: It's better to use `float64` since it's not platform-dependent. And it's a better showcase for the example: 64-bit integer vs 64-bit float. You may also want to mention that the precision loss is caused by the limited size of the mantissa.
```suggestion
    >>> roundtrip_t = np.int64(np.float64(t))
```

lumicks/pylake/detail/confocal.py
```diff
@@ -8,20 +8,29 @@
 from .image import reconstruct_image_sum, reconstruct_image, save_tiff, ImageMetadata
 
 
+def _timestamp_mean(a, axis):
+    """Compute mean of consecutive timestamps.
+
+    Note: This function should only be used for consecutive timestamps. It is used to avoid
+    incurring an integer overflow when computing the mean of a series of timestamps."""
+    return (np.min(a, axis=axis) + np.max(a, axis=axis)) // 2  # We lose at most half a nanosecond
```
dean0x7d, line 16: You should definitely add an `assert` that checks that precondition. Otherwise, it's going to silently produce wrong results. I'm not sure about the performance impact of such an extra check. The simple version would be something like:
```py
diff = np.diff(a)
assert np.all(diff[0] == diff), "This function should only be used for consecutive timestamps"
```

dean0x7d (commented):
It's wonderful to have this fixed and so well covered with tests! Just see my concern about the precondition in `_timestamp_mean()`.

docs/tutorial/file.rst
JoepVanlier: Done and added:

`The reason for this is that timestamps exceed the maximum integer value representable by the mantissa.`

Unless you can think of a more user-friendly way of describing that.

lumicks/pylake/detail/confocal.py
dean0x7d: Actually, I just realized that the consecutive timestamp limitation would be dropped if you implement the mean like this:
```py
def _timestamp_mean(a):
    min = np.min(a)
    return min + np.sum(a - min) // a.size
```
or
```py
def _timestamp_mean(a):
    return min + np.mean(a - min).astype(np.int64)
```
This gets around the precision issue by reducing the range in the critical operation. Provided that `sum(a - min)` doesn't overflow.

JoepVanlier: Unfortunately, it's messier than that and that condition is too strict. Our test data doesn't even fulfill that stricter condition actually.

Within each pixel (along the averaged axis) that condition has to apply, but that doesn't mean that the difference between the `timestamps` has to be the same for every pixel.

A test that tests this condition would look something like this:
```python
second_derivative = np.diff(a, n=2, axis=axis)
assert np.all(second_derivative == 0), "This function should only be used for consecutive timestamps"
```
But unfortunately, that one also doesn't work, because `block_reduce` passes two axes when reducing along the positional axis. Anyways, I'll think about it. It might be possible to use the condition above, but change `block_reduce` for something else.

Either that, or come up with an alternative that doesn't rely on that assumption. I'll have a look tomorrow.

dean0x7d: See my comment just above. I think that an unconditional mean could be done. You can even protect against overflow like this:
```py
delta = np.max(a) - np.min(a)
if int(delta) * int(a.size) > 2**63:
   # It will overflow
# No overflow
return min + np.sum(a - min) // a.size
```
The basic Python `int` supports big integers so this works are long as you cast from the numpy `np.int64` to `int`.

At that point, it would even be possible to make it work in any arbitrary case since the mean could be done in multiple smaller ranges:
```
return min + np.sum(a[:n] - min) // n + np.sum(a[n:] - min) // n
```

JoepVanlier: I think we were posting at the same time and my git hadn't notified me yet :)

Anyways, unfortunately, all of this gets very messy in practice. I'll try to put the constraints into words, so that it's logged for the future.

When called from `reconstruct_image`, this customized mean function gets something in the following structure (bluelake can end a `Kymo` or `Scan` mid pixel, hence the zeros):

```python
ts1  ts2  ts3  <gap>
ts4  ts5  ts6  <gap>
ts7  ts8  ts10 <gap>
ts11 0    0    <gap>
```

By gap I mean that between `ts3` and `ts4` there's a gap in time. The reduce function can also be called from `downsampled_by`, but it will be averaging over a different axis.

Initially I considered subtracting the minimal timestamp and going that route as well. Although, there I initially just subtracted the minimum of the entire `Kymo` rather than the individual pixel (which still put us at risk for overflow). Doing it like this on a per pixel basis may be sufficient for any real world use case actually. I had a quick check and for a high frequency system (2.5 MHz), overflow would occur when the pixel time is around 60 seconds. For a normal setup (78125 Hz), it would overflow at a pixel time of 343 seconds. Those times seem extremely long for pixel times.

The final pixel will require special handling, since it may be partial.

Also, wouldn't the overflow protected suggestion be subject to roundoff at every sub-block? Meaning that the blocksize (`n`) would impact the results? If we were to go with that solution, wouldn't it make more sense to then to sum the sub-sums obtained with `np.sum` in a python `(big)int` and then divide at the very end to make sure you only incur the roundoff once? I think that solution could get fairly complicated given that the axis over which we apply the reduce function may change.

When you consider that it will have to be generalized to work along an arbitrary axis on matrices instead of a single list, the check becomes a bunch of list comprehensions. And there are a lot of corner cases to test.


I'm not sure the complexity is warranted here. I prefer your first suggestion of testing and asserting if the condition is not met instead, since we do explicitly not allow downsampling over non-constant rate `timestamps` in the API (and this is a design choice). Because in those conditions the question even arises, what does the sample rate weighted `timestamp` average even mean at that point. Such a `mean` could end up in a location that's no longer in the original sampled region (if you have a long gap). With the simple version plus check, we'd also catch that case. We just have to make clear in the function name that this is really only intended for contiguous timestamps so that it isn't misused.

The second issue is that Bluelake can end the last pixel arbitrarily mid-pixel (this actually produces incorrect results in the current implementation as well (a timestamp lower than the starting timestamp) and it wouldn't with the `(start + max) / 2`).

The second issue is that Bluelake can end the last pixel arbitrarily mid-pixel (this actually produces incorrect results in the current implementation as well (a timestamp lower than the starting timestamp) and it wouldn't with the `(start + max) / 2`).

I would propose simply testing whether the second derivative along the averaged axis is zero for all but the last pixel. If there's a change of samplerate in the middle, this will throw (as it should). If there is one at the end, it will pass. I added an explicit test for this behaviour.

I've pushed a new proposal with a check. What do you think?

PR #210, opened 2021-11-08
Get `CorrelatedStack` data as `np.ndarray`

rpauszek:
**Why this PR?**
Sometimes you just want the pure data to work on. This adds a convenience method so that users don't have to fiddle around with the internal source data classes.

[docs are here](https://lumicks-pylake.readthedocs.io/en/wit_get_data/tutorial/correlatedstacks.html)

### DIFF
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -24,6 +24,14 @@ from Bluelake if available. This functionality can be turned off with the option
 
     stack2 = lk.CorrelatedStack("example2.tiff", align=False)
 
+You can obtain the image stack data as a `numpy` array using the `get_image()` method::
+
+    red_data = stack.get_image(channel="red") # shape = [n_frames, y_pixels, x_pixels]
+    rgb_data = stack.get_image(channel="rgb") # shape = [n_frames, y_pixels, x_pixels, 3 channels]
+
+If the `channel` argument is not provided, the default behavior is `"rgb"` for 3-color images. For single-color
+images, this argument is ignored as there is only one channel available.
+
 Quite often, it is interesting to correlate events on the camera's to `channel` data.
 To quickly explore the correlation between images in a `CorrelatedStack` and channel data
 you can use the following function::
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -115,6 +115,23 @@ class CorrelatedStack:
         data = self.src.with_roi(np.array([x_min, x_max, y_min, y_max]))
         return self.from_dataset(data, self.name, self.start_idx, self.stop_idx)
 
+    def get_image(self, channel="rgb"):
+        """Get image data for the full stack as an `np.ndarray`.
+
+        Parameters
+        ----------
+        channel : {'red', 'green', 'blue', 'rgb'}
+            The color channel of the requested data.
+            For single-color data, this argument is ignored.
+        """
+        if self.src._description.is_rgb:
+            channel_indices = {"red": 0, "green": 1, "blue": 2, "rgb": slice(None)}
+            slc = (slice(None), slice(None), channel_indices[channel])
+        else:
+            slc = (slice(None),)
+
+        return np.stack([frame.data[slc] for frame in self], axis=0).squeeze()
+
     def plot(self, frame=0, channel="rgb", show_title=True, **kwargs):
         """Plot image from image stack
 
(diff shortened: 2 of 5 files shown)


docs/tutorial/correlatedstacks.rst
```diff
@@ -24,6 +24,14 @@ from Bluelake if available. This functionality can be turned off with the option
 
     stack2 = lk.CorrelatedStack("example2.tiff", align=False)
 
+You can obtain the image stack data as a `numpy` array using the `get_image()` method::
+
+    red_data = stack.get_image(channel="red") # shape = [y_pixels, x_pixels, n_frames]
+    rgb_data = stack.get_image(channel="rgb") # shape = [y_pixels, x_pixels, 3 channels, n_frames]
```
JoepVanlier, line 30: Is there a reason to deviate from `Scan` here? Multiframe scans are stored as `n_frames`, `pixel`, `pixel`, `colors`.

lumicks/pylake/detail/widefield.py
```diff
@@ -193,6 +193,11 @@ class ImageDescription:
     def __init__(self, tiff_file, align_requested):
         first_page = tiff_file.pages[0]
         self.is_rgb = first_page.tags["SamplesPerPixel"].value == 3
+        self._cmap = (
+            {"red": 0, "green": 1, "blue": 2, "rgb": slice(None), None: slice(None)}
+            if self.is_rgb
+            else {None: slice(None)}
```
JoepVanlier, line 199: Can't this just be an empty dict?

Or even just `None`? (Considering you have the if-statement down there anyway and alignment matrices only exist in multi-color images?).

lumicks/pylake/correlated_stack.py
```diff
@@ -115,6 +115,26 @@ def crop_by_pixels(self, x_min, x_max, y_min, y_max):
...
 
+    def get_image(self, channel=None):
+        """Get image data for the full stack as an `np.ndarray`.
+
+        Parameters
+        ----------
+        channel : {'red', 'green', 'blue', 'rgb', `None`}
+            The color channel of the requested data.
+            For single-color data, this argument is ignored.
+            For RGB data the default value `None` corresponds to 'rgb'.
```
JoepVanlier, line 126: Can't we just make the default `"rgb"` and ignore the channel argument when it's a single color stack?

It doesn't seem like you have error handling for requesting `rgb` when there is no `rgb` anyhow?

Then you also don't need the `None` entry anymore in the `_cmap.`

JoepVanlier (commented):
Cool!

It's good to have this back up to parity.

Please also update the deprecation notice. Now it says the following:
`All operations on these objects should be handled through the `CorrelatedStack` public API`
But it would be good to give specific instructions now that this function is available.

docs/tutorial/correlatedstacks.rst
rpauszek: the reason is that I meant to make sure it was in line with `Scan` but somehow permuted through every other combination of axes before finally getting it right.....

lumicks/pylake/detail/widefield.py
rpauszek: agreed. and having the axes in the correct order, the reasons for making a more complicated handling basically went away, so we're back to the original code for this

lumicks/pylake/correlated_stack.py
```diff
@@ -115,6 +115,23 @@ def crop_by_pixels(self, x_min, x_max, y_min, y_max):
...
+            The color channel of the requested data.
+            For single-color data, this argument is ignored.
+        """
+        if self.src._description.is_rgb:
+            channel_indices = {"red": 0, "green": 1, "blue": 2, "rgb": slice(None)}
+            slc = (slice(None), slice(None), channel_indices[channel])
+        else:
+            slc = (slice(None),)
+
+        return np.stack([frame.data[slc] for frame in self], axis=0)
```
JoepVanlier, line 133: Correct me if I'm wrong, but I think you need to squeeze this one to get rid of the extraenous axis (if you want to make it fully consistent with `Scan`)?

I found it when trying to slice the stack. Might also be good to add a test for this:
```
np.testing.assert_array_equal(np.squeeze(stack[0].get_image()), rgb_data)
```

JoepVanlier (commented):
Mostly looks good to me. Just one minor thing remaining when it comes to compatibility with the `Scan` API.

JoepVanlier (approved):
LGTM! 👍

PR #211, opened 2021-11-08
Bugfix: WIT color-aligned data is returned but not requested

rpauszek:
**Why this PR?**
While prepping figures to add to the `CorrelatedStack` docs, I noticed a bug in which alignment was not requested (`CorrelatedStack(filename, align=False)`) yet the aligned data was still returned. This PR fixes the bug and adds a regression test


JoepVanlier (approved):
Good catch and nice fix! 👍

PR #214, opened 2021-11-09
calibration: use viscosity of water as default

JoepVanlier:
**Why this PR?**
Viscosity is a sensitive calibration parameter. Typically, users do experiments in water.

When no viscosity is specified, we will now use the viscosity of water at that particular temperature as the default.

I decided to go with the simplified model from [1]. The advantage of this model is that it covers a wider range (the Peterman model [2] mentions it should only be used >= 20 C). [3] is the reference Stijn used in his code (but not in his report, where he used [2]).

[[1] Huber, M. L., Perkins, R. A., Laesecke, A., Friend, D. G., Sengers, J. V.,Assael, M. J., & Miyagawa, K. (2009). New international formulation for the viscosity of H2O. Journal of Physical and Chemical Reference Data, 38(2), 101-125.](http://ikee.lib.auth.gr/record/218228/files/Assael3.pdf)
[[2] Peterman, E. J., Gittes, F., & Schmidt, C. F. (2003). Laser-induced heating in optical traps. Biophysical journal, 84(2), 1308-1316.](https://www.sciencedirect.com/science/article/pii/S0006349503749467/pdfft?md5=b928f2468df54a1e79a4bddcc01784f9&pid=1-s2.0-S0006349503749467-main.pdf)
[3] http://ddbonline.ddbst.de/VogelCalculation/VogelCalculationCGI.exe?component=Water&tunit=K&punit=mPa*s&TemperaturesEdit=&calculate=Calculate

See below for the three other models:
image

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -22,6 +22,34 @@ from dataclasses import dataclass
 from typing import Callable
 
 
+def viscosity_of_water(temperature):
+    """Computes the viscosity of water in [Pa*s] at a particular temperature.
+
+    These equations come from section 3.7 from [7].
+
+    [7] Huber, M. L., Perkins, R. A., Laesecke, A., Friend, D. G., Sengers, J. V.,
+    Assael, M. J., & Miyagawa, K. (2009). New international formulation for the viscosity of H2O.
+    Journal of Physical and Chemical Reference Data, 38(2), 101-125.
+
+    Parameters
+    ----------
+    temperature : array_like
+        Temperature [Celsius]. Should be between -20°C and 110°C
+
+    Returns
+    -------
+    array_like
+        Viscosity of water [Pa*s]
+    """
+    if not np.all(np.logical_and(temperature >= -20, temperature < 110)):
+        raise ValueError("Function for viscosity of water is only valid for -20°C <= T < 110°C")
+
+    temp_tilde = np.tile((temperature + 273.15) / 300, (4, 1)).T
+    ai = np.array([280.68, 511.45, 61.131, 0.45903])
+    bi = np.array([-1.9, -7.7, -19.6, -40.0])
+    return np.sum(ai * temp_tilde ** bi, axis=1) * 1e-6
+
+
 @dataclass
 class Param:
     name: str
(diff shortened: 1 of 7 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -22,6 +22,29 @@
...
+
+    [7] Huber, M. L., Perkins, R. A., Laesecke, A., Friend, D. G., Sengers, J. V.,
+    Assael, M. J., & Miyagawa, K. (2009). New international formulation for the viscosity of H2O.
+    Journal of Physical and Chemical Reference Data, 38(2), 101-125.
+
+    Parameters
+    ----------
+    temperature : float
+        Temperature [Celsius]. Should be between -20°C and 110°C
+    """
```
lpingas, line 38: I recommend adding the return value unit somewhere (Pa*s)

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -22,6 +22,29 @@
...
+
+    These equations come from section 3.7 from [7].
+
+    [7] Huber, M. L., Perkins, R. A., Laesecke, A., Friend, D. G., Sengers, J. V.,
+    Assael, M. J., & Miyagawa, K. (2009). New international formulation for the viscosity of H2O.
+    Journal of Physical and Chemical Reference Data, 38(2), 101-125.
+
+    Parameters
+    ----------
+    temperature : float
```
lpingas, line 36: We should document that the function also works with numpy arrays.

lumicks/pylake/force_calibration/tests/test_power_spectrum_calibration.py
```diff
@@ -309,3 +310,31 @@ def test_faxen_correction():
     np.testing.assert_allclose(fit.results["Rf"].value, 52.35949300703634)
     # Not affected since this is gamma bulk
     np.testing.assert_allclose(fit.results["gamma_0"].value, 1.0678273429551705e-08)
+
+
+@pytest.mark.parametrize("temperatures", [np.arange(-30, -10), np.arange(100, 130), -21, 111])
```
lpingas, line 315: 110 is a bit more interesting than 111 in this case : D

lpingas (approved):
Nice! This is good functionality.

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Good point. I'll rename it `array_like`.

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Added a return field.

PR #215, opened 2021-11-09
force calibration: small private API for carrying over drag coefficient

JoepVanlier:
**Why this PR?**
- Add a small private API for using drag from lateral experiment, and make sure the field name changes so it can be seen in the results.
- The name of the field where the experimental drag can be found is stored as an attribute so we don't have to hardcode the fieldname in both Pylake and Bluelake.
- Fixes a bug in `viscosity_of_water`. This function should return a 0D array for a 0D input.

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -41,13 +41,14 @@ def viscosity_of_water(temperature):
     array_like
         Viscosity of water [Pa*s]
     """
+    temperature = np.asarray(temperature)
     if not np.all(np.logical_and(temperature >= -20, temperature < 110)):
         raise ValueError("Function for viscosity of water is only valid for -20°C <= T < 110°C")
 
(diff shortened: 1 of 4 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -262,6 +264,17 @@ def _drag(self):
         """
         return self.drag_coeff * self._drag_correction_factor
 
+    def _set_drag(
+        self,
+        drag,
+        new_name="gamma_0_lateral",
```
lpingas, line 270: Nitpick: `gamma_0_lateral` implies that we are using `gamma_0` from the lateral force calibration, while we would actually use  `gamma_ex`. I wonder if we shouldn't call it `gamma_ex_lateral`

JoepVanlier: Good point. Updating it.

lpingas (approved):
LGTM!

PR #216, opened 2021-11-10
Update to newer version of `black`

JoepVanlier:
**Why this PR?**
The version of black we use used `typed-ast`. Typed-ast currently has an open issue on 3.9.8 which is causing black to break; see here: https://github.com/python/typed_ast/issues/169

The latest version of black doesn't rely on the library with the issue anymore and just works. Note that the latest version of black made a few touchups to some docstrings. I have committed these in a separate commit as `black`.

Note that after this gets merged, you will have to update your local install of `black` to `21.10b0`.

### DIFF
--- a/.github/workflows/pylake_test.yml
+++ b/.github/workflows/pylake_test.yml
@@ -28,5 +28,5 @@ jobs:
         python pylake_test.py
     - name: Black
       run: |
-        pip install black==20.8b1
+        pip install black==21.10b0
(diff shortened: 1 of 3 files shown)


dean0x7d (approved):
Nice quick fix! It seems that the new version of `black` brings only minimal changes. I quite like that it's aware of docstrings now. I'll make a similar change in our C++ projects so that we stay consistent.

PR #217, opened 2021-11-15
kymotracker: allow changing slider ranges in the Kymotracker widget

JoepVanlier:
**Why this PR?**
It's been requested to allow changing the ranges of the sliders in the kymotracker widget.

The modification in this PR allows you to override the slider ranges when opening the widget.

rangetest

Docs build here: https://lumicks-pylake.readthedocs.io/en/kymo_slider_ranges/tutorial/kymotracking.html#using-the-kymotracker-widget

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -456,6 +456,7 @@ class KymoWidgetGreedy(KymoWidget):
         min_length=3,
         use_widgets=True,
         output_filename="kymotracks.txt",
+        slider_ranges={},
         **kwargs,
     ):
         """Create a widget for performing kymotracking.
@@ -499,6 +500,10 @@ class KymoWidgetGreedy(KymoWidget):
             Add interactive widgets for interacting with algorithm parameters.
         output_filename : str, optional
             Filename to save to and load from.
+        slider_ranges : dict of list, optional
+            Dictionary with custom ranges for selected parameter sliders. Ranges should be in the
+            following format: (lower bound, upper bound).
+            Valid options are: "window", "pixel_threshold", "line_width", "sigma" and "vel".
         """
         algorithm = track_greedy
         calibrated_kymo_channel = CalibratedKymographChannel.from_kymo(kymo, channel)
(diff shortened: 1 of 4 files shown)


lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -516,6 +521,36 @@ def __init__(
...
+                    f"(lower bound, upper bound)."
+                )
+
+            if slider_range[1] < slider_range[0]:
+                raise ValueError(
+                    f"Lower bound should be lower than upper bound for parameter {key}"
+                )
+
+            self._slider_ranges[key] = slider_range
+
```
aafkevandenberg, line 559: Maybe add a `raise ValueError` for giving a range of negative values for the line width, line length, pixel threshold, window and sigma.

aafkevandenberg (approved):
Awesome!

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: Added an exception for it.

mikhas: I'd drop it, for being too strict. The current class of parameters might only make sense as positive integral types, but what if new parameters were to be added? Would this restriction automatically apply to those as well?

Technically, you'd have to specify the concrete type of each parameter, if you wanted to do it correctly.

Also, what's the harm of _allowing_ negative slider values for those parameters? Sure, you'd have buggy sliders, but those are still user input to the algorithm itself (the sliders only offer convenience to enter those parameters), and it's the responsibility of the algorithm to check the input and the parameters, IMO.

mikhas (approved):
Code looks fine. Feature seems useful, too.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
aafkevandenberg: It does make sense that the algorithm should check for the validity of the input, not the sliders.

(Btw, in the current version the velocity is a parameter that can be negative).

JoepVanlier: The list of which parameters should be non-negative was explicitly set in the last commit though? Velocity wasn't part of it.
```non_negative = ["window", "pixel_threshold", "line_width", "sigma", "min_length"]```

I can drop the last commit (and not enforce positivity). Currently, the downstream algorithm does not enforce positivity. I can make a ticket for that, but I wouldn't piggyback that on this PR.

Personally, I don't feel strongly about it either way. The reason I checked for increasing bounds is because otherwise it breaks _before_ it gets to the slider (it's a widget requirement) with a cryptic error message. Hence it felt like a requirement for the input to this function.

So final verdict, enforce positive sliders where applicable, yay or nay?

aafkevandenberg: I don't feel strongly about it either, my answer is 'may'.

You could also change it into a warning :P 

Btw, I mentioned the velocity, to make clear that not all of the parameters are currently nonnegative. I was responding to the sentence 'but what if new parameters were to be added? Would this restriction automatically apply to those as well?'

JoepVanlier: Alright. I've dropped the positive constraint for now and made a ticket for handling this at the algorithm level in the future.

PR #219, opened 2021-11-18
calibration: store driving peak parameters

JoepVanlier:
**Why?**
It's nice to have a way to access all the output from the calibration procedure in one and the same data structure. Up to now, you had to keep both the `CalibrationModel` and `FitResult` around, since you needed results from both.

Note: I piggybacked a tiny fix on this for a matplotlib `DeprecationWarning` that showed up in the tests.


mikhas:
How does changing the result affect API in Pythonland? I assume because it's a dictionary, you are free to add but aren't allowed to remove stuff from the result?

lpingas (approved):
Looks great!

JoepVanlier:
> How does changing the result affect API in Pythonland? I assume because it's a dictionary, you are free to add but aren't allowed to remove stuff from the result?

What do you mean exactly? The result is just a public output parameter to communicate results to the user. So yes, while adding is easy, removing would likely have to be paired with a deprecation notice first.

PR #220, opened 2021-11-25
Wrap affine transformation logic in a small dataclass

rpauszek:
**Why this PR?**
As I was working on the feature to define the location of the tether between two beads (and rotate the image such that the tether is horizontal) it became clear that it is cleaner to centralize all of the code having to do with affine transformations. In addition, when the tether feature is implemented, we can simply multiply the color correction and rotation matrices to do a single linear interpolation to get to the final data.

Note: I added a `tests` directory to the `details` module as this module is growing and we have a bunch of code in there that either isn't tested at the low level or is tested in the main tests.

### DIFF
--- a/lumicks/pylake/detail/widefield.py
+++ b/lumicks/pylake/detail/widefield.py
@@ -6,7 +6,7 @@ import tifffile
 import warnings
 import enum
 from copy import copy
-from dataclasses import dataclass
+from dataclasses import dataclass, field
 
 
 class TiffFrame:
@@ -32,19 +32,9 @@ class TiffFrame:
 
     def _align_image(self):
         """reconstruct image using alignment matrices from Bluelake; return aligned image as a NumPy array"""
-        align_mats = [self._description.alignment_matrix(color) for color in ("red", "blue")]
-
         img = self._page.asarray()
-        rows, cols, _ = img.shape
-        for mat, channel in zip(align_mats, (0, 2)):
-            img[:, :, channel] = cv2.warpAffine(
-                img[:, :, channel],
-                mat,
-                (cols, rows),
-                flags=(cv2.INTER_LINEAR | cv2.WARP_INVERSE_MAP),
-                borderMode=cv2.BORDER_CONSTANT,
-                borderValue=0,
-            )
+        for channel, mat in self._description._alignment_matrices.items():
+            img[:, :, channel] = mat.warp_image(img[:, :, channel])
         return img
 
     @property
@@ -186,8 +176,6 @@ class ImageDescription:
         details of color channel alignment.
     json : dict
         dictionary of parsed json string held in frame ImageDescription tag.
-    _cmap : dict
-        dictionary of color : channel index pairs
     """
 
     def __init__(self, tiff_file, align_requested):
@@ -199,19 +187,27 @@ class ImageDescription:
             self.json = json.loads(first_page.description)
         except json.decoder.JSONDecodeError:
             self.json = {}
-            self._cmap = {}
 
         # if metadata is missing, set default values
         if len(self.json) == 0:
             self._alignment = Alignment(align_requested, AlignmentStatus.empty, False)
             return
 
-        # update format if necessary
-        self._cmap = {"red": 0, "green": 1, "blue": 2}
+        # update json fields if necessary
         if "Alignment red channel" in self.json:
-            for color, j in self._cmap.items():
+            for j, color in enumerate(("red", "green", "blue")):
                 self.json[f"Channel {j} alignment"] = self.json.pop(f"Alignment {color} channel")
                 self.json[f"Channel {j} detection wavelength (nm)"] = "N/A"
+
+        if "Channel 0 alignment" in self.json:
+            self._alignment_matrices = {
+                channel: TransformMatrix.from_alignment(
+                    self._raw_alignment_matrix(channel), *self.offsets
+                ).invert()
+                for channel in range(3)
+            }
+        else:
+            self._alignment_matrices = {}
 
         # check alignment status
         if not self.is_rgb:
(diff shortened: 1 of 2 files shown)


lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,121 @@ def crop(self, roi):
...
+        translation[0, -1] = -x_offset
+        translation[1, -1] = -y_offset
+        # apply the original transform to the translated image.
+        # it only needs to be resized from a 2x3 to a 3x3 matrix
+        original = np.vstack((alignment, [0, 0, 1]))
+        # translate the image back to the original origin.
+        # takes into account both the offset and the scaling performed by the first step
+        back_translation = np.eye(3)
+        back_translation[0, -1] = original[0, 0] * x_offset
+        back_translation[1, -1] = original[1, 1] * y_offset
```
JoepVanlier, line 344: This strikes me as odd. Why are we taking into account only scaling and not other parts of the affine transform?

Also of note: no tests fail if I remove the multiplication by these scaling factors, which makes me a bit worried about the testing for this particular function.

lumicks/pylake/detail/widefield.py
```diff
@@ -32,19 +32,9 @@ def _is_aligned(self):
...
-            img[:, :, channel] = cv2.warpAffine(
-                img[:, :, channel],
-                mat,
-                (cols, rows),
-                flags=(cv2.INTER_LINEAR | cv2.WARP_INVERSE_MAP),
-                borderMode=cv2.BORDER_CONSTANT,
-                borderValue=0,
-            )
+        for channel, mat in self._description._alignment_matrices.items():
+            img[:, :, channel] = mat.warp_image(img[:, :, channel])
```
JoepVanlier, line 37: I really like that this is abstracted away now.

lumicks/pylake/detail/widefield.py
```diff
@@ -199,20 +187,28 @@ def __init__(self, tiff_file, align_requested):
...
         if len(self.json) == 0:
             self._alignment = Alignment(align_requested, AlignmentStatus.empty, False)
             return
 
-        # update format if necessary
-        self._cmap = {"red": 0, "green": 1, "blue": 2}
+        # update json fields if necessary
         if "Alignment red channel" in self.json:
-            for color, j in self._cmap.items():
+            for j, color in enumerate(("red", "green", "blue")):
```
JoepVanlier, line 198: 👍

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,121 @@ def crop(self, roi):
...
+        data: np.ndarray
+            image data
+        interp_mode: cv2.InterpolationFlags
+            interpolation type
+        """
+
+        if np.all(np.equal(self.matrix, np.eye(3)[:2, :])):
+            return data
+
+        data = np.atleast_3d(data)
```
JoepVanlier, line 386: Nice! I didn't know about that function!

JoepVanlier (changes requested):
It'll be nice to have a class that wraps some of these geometric behaviours and I think it was a good call to split this off into a separate PR.

I made a first pass through the code and have a few requests / questions.

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,121 @@ def crop(self, roi):
...
+    def __mul__(self, mat):
+        """Perform matrix multiplication such that `self * mat == np.matmul(self, mat)`."""
+        assert isinstance(mat, TransformMatrix), "Operands must be of type `TransformMatrix`."
+        m1 = np.vstack((self.matrix, [0, 0, 1]))
+        m2 = np.vstack((mat.matrix, [0, 0, 1]))
+        return TransformMatrix(np.matmul(m1, m2)[:2])
+
+    @classmethod
+    def with_offset(cls, alignment, x_offset, y_offset):
+        """Recalculate matrix with offset. Same implementation as Bluelake.
```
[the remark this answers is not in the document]
dean0x7d: I'd vote for `TransformMatrix(m).with_offset(x, y)`. Seems like the most logical and Python API to me.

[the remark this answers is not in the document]
JoepVanlier: I originally suggested that as well, but that was before I looked closer and noticed that this operation does not do what I would expect from that name. I would prefer to be alerted that this is a nonstandard operation. `with_offset` makes me think it's just a translation of the operating point, which this is not.

The following produces a translation, which is not what I would expect as a user of this API:
```
rot = TransformMatrix.rotation(10, (10, 5))
irot = TransformMatrix.rotation(-10, (10, 5))
TransformMatrix.with_offset(rot.matrix, 10, 10) * TransformMatrix.with_offset(irot.matrix, 10, 10)
```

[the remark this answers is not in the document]
dean0x7d: I see. Then the dedicated `classmethod` makes sense. I'd suggest simply `from_alignment()` to avoid being too verbose. The offset parameters are clear in the signature and could be defaulted.

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,121 @@ def crop(self, roi):
...
+        rotate = np.eye(3)
+        rotate[:2, :2] = [[np.cos(theta), -np.sin(theta)], [np.sin(theta), np.cos(theta)]]
+
+        translate_to_origin = np.eye(3)
+        translate_to_origin[:2, -1] = center
+
+        # multiply to get overall transformation matrix
+        return cls(np.matmul(translate_to_origin, np.matmul(rotate, translate_to_center))[:2, :])
+
+    def warp_image(self, data, inverse=True, interp_mode=cv2.INTER_LINEAR):
```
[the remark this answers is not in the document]
rpauszek: 👍 added an `invert()` method and got rid of the `interp_mode` altogether since we do a single linear interpolation now to take care of both color channel alignment and rotation about the tether

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,121 @@ def crop(self, roi):
...
+        translate_to_center[:2, -1] = -np.array(center)
+
+        rotate = np.eye(3)
+        rotate[:2, :2] = [[np.cos(theta), -np.sin(theta)], [np.sin(theta), np.cos(theta)]]
+
+        translate_to_origin = np.eye(3)
+        translate_to_origin[:2, -1] = center
+
+        # multiply to get overall transformation matrix
+        return cls(np.matmul(translate_to_origin, np.matmul(rotate, translate_to_center))[:2, :])
```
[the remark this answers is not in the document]
rpauszek: oops, missed that one. good catch! much nicer. implementation now uses degrees for the angle just as `cv2`

lumicks/pylake/detail/widefield.py
JoepVanlier: Alright.

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,102 @@ def crop(self, roi):
...
+                (cols, rows),
+                flags=cv2.INTER_LINEAR,
+                borderMode=cv2.BORDER_CONSTANT,
+                borderValue=0,
+            )
+            for channel in range(channels)
+        ]
+        return np.stack(image, axis=2).squeeze()
+
+    def warp_coordinates(self, coordinates, inverse=True):
```
JoepVanlier, line 388: Leftover argument `inverse`, maybe rebase mishap?

JoepVanlier (approved):
Looks good to me. Yay transforms! Please fix up that last unused argument and it's good to go I think.

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,102 @@ def crop(self, roi):
...
+        """Apply affine (inverse) transformation to coordinate points.
```
JoepVanlier, line 389: `(inverse)` can also be removed from the docstring.

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +299,102 @@ def crop(self, roi):
...
+        ----------
+        theta: float
+            angle of rotation
+        center: np.ndarray
+            (x, y) point, center of rotation
+        """
+        return cls(cv2.getRotationMatrix2D(center, theta, scale=1.0))
+
+    def warp_image(self, data):
+        """Apply affine (inverse) transformation to image data.
```
JoepVanlier, line 362: Remove `(inverse)` from the docstring :)

PR #221, opened 2021-11-25
docs: update installation instructions

onnodb:
Remove the instruction to run `python -m ipykernel install --user` from the
installation instructions, since this interferes with kernel detection in other
virtual/conda environments.

Instead, add a FAQ/troubleshooting section, explaining the situations in which
running this command may be necessary.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -53,10 +53,6 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
 
     conda install lumicks.pylake
 
-#. To make sure Jupyter notebook can find the ipykernel, we invoke::
-
-    python -m ipykernel install --user
-
 #. It should be possible to open a jupyter notebook in this environment by calling::
 
     jupyter notebook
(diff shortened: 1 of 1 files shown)


JoepVanlier (approved):
Looks good to me. Thanks for picking this up! One small request though. Unfortunately I cannot comment on lines that were not edited, but further down we have some instructions for people using conda from China that still have the same instructions.

The following passage can be removed entirely.
```
On Windows, you should also make sure that Jupyter notebook can connect to the ipython kernel by invoking:

conda install pywin32
python -m ipykernel install --user
```

The `pywin32` issue seems to have been resolved as well (I just tried a fresh conda w/pip install in a clean windows VM). There used to be an issue where an incompatible version of `pywin32` ended up being installed (resulting in a DLL load error).

onnodb:
> further down we have some instructions for people using conda from China that still have the same instructions

Ah, my bad, I should've done a global search for further occurrences of `ipykernel`. Fixed now.

Regarding `pywin32`: I still see issues with that package quite regularly, and in my case the problem always seems to fix itself by running:

    python %CONDA_ENV%\Scripts\pywin32_postinstall.py -install

From my understanding so far, `pywin32` needs some DLLs in specific places, and this install command doesn't always seem to run correctly during a conda/pip install, for whatever reason. Running it again explicitly typically solves it. (See AutoFAT issue [AUTO-679](https://lumicks.atlassian.net/browse/AUTO-679) for more background, if you're curious.)

JoepVanlier:
> > further down we have some instructions for people using conda from China that still have the same instructions
> 
> Ah, my bad, I should've done a global search for further occurrences of `ipykernel`. Fixed now.
> 
> Regarding `pywin32`: I still see issues with that package quite regularly, and in my case the problem always seems to fix itself by running:
> 
> ```
> python %CONDA_ENV%\Scripts\pywin32_postinstall.py -install
> ```
> 
> From my understanding so far, `pywin32` needs some DLLs in specific places, and this install command doesn't always seem to run correctly during a conda/pip install, for whatever reason. Running it again explicitly typically solves it. (See AutoFAT issue [AUTO-679](https://lumicks.atlassian.net/browse/AUTO-679) for more background, if you're curious.)

Thanks!

Yeah, I think it came up via support once. Back then I could reproduce it with a VM (installing `anaconda`, then installing `pip` in a new environment and then installing `pylake` and firing up the notebook), but now that procedure seems to work correctly.

Still, it makes me think it that it should sooner be relegated to a specific FAQ question if we want to mention it rather than be a default installation instruction for this particular installation type. Perhaps a separate question in the FAQ would make sense. Perhaps something along the line of:
I'm trying to use `pylake` in a Jupyter notebook and I encounter `ImportError: DLL load failed while importing win32api: The specified procedure could not be found`.

Do you know whether that post install trick always works? I've only ever tried force re-installing it explicitly with the last package manager used in that environment, and it always seems to have solved it (I assume it called that post_install python file as part of the installation procedure).

onnodb:
> Still, it makes me think it that it should sooner be relegated to a specific FAQ question if we want to mention it rather than be a default installation instruction for this particular installation type. 

I agree on that one!

> Perhaps a separate question in the FAQ would make sense. Perhaps something along the line of: I'm trying to use `pylake` in a Jupyter notebook and I encounter `ImportError: DLL load failed while importing win32api: The specified procedure could not be found`.

Yes, that's a good idea. I've pushsed a commit with an entry for this.

> Do you know whether that post install trick always works?

I can't guarantee it, but I've only ever seen cases in which it worked to resolve the exception you mentioned above. To the extent where I made it part of the AutoFAT installation package for development environments, just to get it over with :)

JoepVanlier:
> > Still, it makes me think it that it should sooner be relegated to a specific FAQ question if we want to mention it rather than be a default installation instruction for this particular installation type.
> 
> I agree on that one!
> 
> > Perhaps a separate question in the FAQ would make sense. Perhaps something along the line of: I'm trying to use `pylake` in a Jupyter notebook and I encounter `ImportError: DLL load failed while importing win32api: The specified procedure could not be found`.
> 
> Yes, that's a good idea. I've pushsed a commit with an entry for this.

Looks good.

> > Do you know whether that post install trick always works?
> 
> I can't guarantee it, but I've only ever seen cases in which it worked to resolve the exception you mentioned above. To the extent where I made it part of the AutoFAT installation package for development environments, just to get it over with :)

Given that I am unable to reproduce the original issue on my setup and that you've already been testing this fix as part of AutoFAT, I agree that this is the best way to have this in the FAQ for now at least.

PR #222, opened 2021-11-26
Add function to define the coordinates of the tether between beads

rpauszek:
**Why this PR?**
This adds functionality to define the location of the tether between beads, which will be needed in order to construct WIT kymographs in the future.

Additionally, when the tether coordinates are set, the returned image data is rotated such that the tether is horizontal. Rotation is combined with the color channel alignment if applicable, such that only one linear interpolation is needed to get to the final image

### DIFF
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -17,9 +17,22 @@ You can quickly plot an individual frame using the `plot()` method::
 
 .. image:: correlatedstack_aligned.png
 
+To define the location of the tether between beads, supply the `(x, y)` pixel coordinates of the end points
+to the `define_tether()` method::
+
+    stack = stack.define_tether((126, 193), (341, 200))
+    stack.plot()
+    stack.plot_tether(lw=0.7)
+
+.. image:: correlatedstack_tether.png
+
+Note, after defining a tether location the image is rotated such that the tether is horizontal in the field of view.
+You can also plot the overlay of the tether location using `plot_tether(**kwargs)`, which also accepts keyword
+arguments that are passed to `plt.plot()`.
+
 You can also spatially crop to select a smaller region of interest::
 
-    stack_roi = stack.crop_by_pixels(60, 460, 100, 200)  # pixel coordinates as x_max, x_min, y_max, y_min
+    stack_roi = stack.crop_by_pixels(45, 420, 150, 245)  # pixel coordinates as x_min, x_max, y_min, y_max
     stack_roi.plot()  # note: the default channel is "rgb"
 
 .. image:: correlatedstack_cropped.png
@@ -30,7 +43,7 @@ can become corrupted due to interpolation artifacts.
 You can also plot only a single color channel. Note that here we pass some additional formatting arguments, which are
 forwarded to `plt.imshow()`::
 
-    stack_roi.plot(channel="red", cmap="magma", vmin=250, vmax=450)
+    stack_roi.plot(channel="red", cmap="magma", vmin=550, vmax=800)
 
 .. image:: correlatedstack_red.png
 
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -115,6 +115,20 @@ class CorrelatedStack:
         data = self.src.with_roi(np.array([x_min, x_max, y_min, y_max]))
         return self.from_dataset(data, self.name, self.start_idx, self.stop_idx)
 
+    def define_tether(self, point1, point2):
+        """Returns a copy of the stack rotated such that the tether defined by `point_1` and
+        `point_2` is horizontal.
+
+        Parameters
+        ----------
+        point_1 : (float, float)
+            (x, y) coordinates of the tether start point
+        point_2 : (float, float)
+            (x, y) coordinates of the tether end point
+        """
+        data = self.src.with_tether((point1, point2))
+        return self.from_dataset(data, self.name, self.start_idx, self.stop_idx)
+
     def get_image(self, channel="rgb"):
         """Get image data for the full stack as an `np.ndarray`.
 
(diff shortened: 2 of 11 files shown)


docs/tutorial/correlatedstacks.rst
```diff
@@ -17,9 +17,22 @@ You can quickly plot an individual frame using the `plot()` method::
...
+.. image:: correlatedstack_tether.png
+
+Note, after defining a tether location the image is rotated such that the tether is horizontal in the field of view.
+You can also plot the overlay of the tether location using `plot_tether(**kwargs)`, which also accepts keyword
+arguments that are passed to `plt.plot()`.
+
 You can also spatially crop to select a smaller region of interest::
 
-    stack_roi = stack.crop_by_pixels(60, 460, 100, 200)  # pixel coordinates as x_max, x_min, y_max, y_min
+    stack_roi = stack.crop_by_pixels(45, 420, 150, 245)  # pixel coordinates as x_max, x_min, y_max, y_min
```
dean0x7d, line 35:
```suggestion
    stack_roi = stack.crop_by_pixels(45, 420, 150, 245)  # pixel coordinates as x_min, x_max, y_min, y_max
```

lumicks/pylake/correlated_stack.py
```diff
@@ -115,6 +115,25 @@ def crop_by_pixels(self, x_min, x_max, y_min, y_max):
...
+        Parameters
+        ----------
+        point_1 : (float, float)
+            (x, y) coordinates of the tether start point
+        point_2 : (float, float)
+            (x, y) coordinates of the tether end point
+        """
+        if self.src._tether:
+            raise AttributeError(
+                "A tether is already defined; re-defining the coordinates is not allowed."
```
dean0x7d, line 131: It would be good to mention why it's not allowed.

lumicks/pylake/detail/tests/test_widefield.py
```diff
@@ -122,3 +122,13 @@ def test_transform_multiplication_identity():
...
     np.testing.assert_equal(mat_a.matrix, (mat_c * mat_a).matrix)
     np.testing.assert_equal(mat_a.matrix, (mat_d * mat_a).matrix)
+
+
+def test_tether():
+    point_1 = (1.46446609, 1.46446609)
+    point_2 = (8.53553391, 8.53553391)
+
+    # test coordinates of tether after rotation
+    tether = widefield.Tether((point_1, point_2))
```
dean0x7d, line 132: You should also test the `points is None` path.

dean0x7d (approved):
Nice feature! LGTM

lumicks/pylake/detail/tests/test_widefield.py
```diff
@@ -0,0 +1,34 @@
+import pytest
+import numpy as np
+from lumicks.pylake.detail import widefield
+
+
+def test_transform_defult():
```
mikhas: typo?
~~~suggestion
def test_transform_default():
~~~

lumicks/pylake/correlated_stack.py
rpauszek: well ... I managed to work out how to properly handle any permutation of cropping/tether definition so this will no longer throw. I've requested re-review just for the last commit because enough changed since your initial approval to warrant a quick look through (all changes in the final commit)

lumicks/pylake/detail/widefield.py
```diff
@@ -327,3 +298,73 @@ def crop(self, roi):
...
+        # translate the image back to the original origin.
+        # takes into account both the offset and the scaling performed by the first step
+        back_translation = np.eye(3)
+        back_translation[0, -1] = original[0, 0] * x_offset
+        back_translation[1, -1] = original[1, 1] * y_offset
+        # concatenate the transforms by multiplying their matrices and ignore the unused 3rd row
+        return cls(np.dot(back_translation, np.dot(original, translation))[:2, :])
+
+    @classmethod
+    def rotation(cls, theta, center):
```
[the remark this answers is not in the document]
rpauszek: In general I would agree, but this naming more closely matches the mathematical convention. Here we're instantiating an instance that represents a *rotation matrix*, although the actual rotation isn't applied to anything until you call one of the `warp_...()` methods ... so I think in this case the noun form is more appropriate

lumicks/pylake/correlated_stack.py
```diff
@@ -115,6 +115,20 @@ def crop_by_pixels(self, x_min, x_max, y_min, y_max):
         data = self.src.with_roi(np.array([x_min, x_max, y_min, y_max]))
         return self.from_dataset(data, self.name, self.start_idx, self.stop_idx)
 
+    def define_tether(self, point_1, point_2):
```
dean0x7d, line 118: Nitpick: I think numbering without underscores is a bit more common:
```suggestion
    def define_tether(self, point1, point2):
```

lumicks/pylake/correlated_stack.py
```diff
@@ -162,6 +176,24 @@ def plot(self, frame=0, channel="rgb", show_title=True, **kwargs):
...
 
+    def plot_tether(self, **kwargs):
+        """Plot a line at the tether position.
+
+        Parameters
+        ----------
+        **kwargs
+            Forwarded to :fun:`matplotlib.pyplot.plot`.
+
+        """
```
dean0x7d, line 187:
```suggestion
            Forwarded to :fun:`matplotlib.pyplot.plot`.
        """
```

lumicks/pylake/detail/widefield.py
```diff
@@ -398,3 +469,57 @@ def warp_coordinates(self, coordinates):
...
+            list of (x,y) coordinates for the tether end points
+            in the coordinate system of the current processed image
+
+        Attributes
+        ----------
+        offsets: np.ndarray
+            (x,y) coordinate offsets from the raw image origin
+        _ends: list
+            list of (x,y) coordinates for the tether end points
+            in the coordinate system of the raw image data
```
dean0x7d, line 492: It's a bit weird to document private attributes. I think there 2 options:
* Remove this attribute from the docstring and only keep the information in the class as a comment.
* Since this entire class is an implementation detail, you could make the attribute public but name it `raw_ends` if it needs to be exposed in some way.

dean0x7d (approved):
> @rpauszek 
> I managed to work out how to properly handle any permutation of cropping/tether definition so this will no longer throw. I've requested re-review just for the last commit because enough changed since your initial approval to warrant a quick look through (all changes in the final commit)

Nice! I went over it again. Nice generalization. I only have a couple of nitpicks.

PR #224, opened 2021-12-01
force_calibration: validate calibration params

JoepVanlier:
Added more input validation for force calibration. We now force bead and sample density to be more than 100 kg/m^3 when specified. Temperature should stay between 5 and 90 degrees Celsius and viscosity should be bigger than 0.0003 (viscosity of water at 90 degrees Celsius).

***Why this PR?**
Reason we wish to validate some more input parameters is because certain invalid parameters produce garbage results (`viscosity = 0`, `bead_density = 0` etc). Also, by pinning them to at least ballpark values we reduce the chances of user error (for example passing Kelvins instead of Celsius, or passing viscosity or density in an incorrect unit).


lpingas (approved):
Looks great!

PR #225, opened 2021-12-02
force_calibration: ensure zero parameter values are rejected

JoepVanlier:
**Why this PR?**
This fixes a bug which allowed calibrations with a parameter set to zero to go through.
This fix ensures that they lead to an exception.


lpingas (approved):
👍

PR #226, opened 2021-12-02
fd fitting: add units to (in)dependent variable

JoepVanlier:
**Why this PR?**
When subtracting an independent offset, this new parameter is not given a unit.

```python
dna_handles = lk.inverted_odijk("dna_handles").subtract_independent_offset() + lk.force_offset("dna_handles") 
fit = lk.FdFit(dna_handles)
fit[dna_handles].add_data("test", [1], [1])
fit
```

Old:
image

New:
image

Note: Eventually, we may want to carry over the unit handling to pint, which is why I've explicitly made these private properties for now.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v0.11.1 | t.b.d.
+
+#### Bug fixes
+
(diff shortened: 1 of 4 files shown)


tiagobonetti (approved):
L`honk`TM.

PR #227, opened 2021-12-03
`Scan`: added `frame_timestamp_ranges()`

JoepVanlier:
**Why this PR?**
This PR adds a function to obtain the start and stop timestamp of individual `Scan` frames. This was a user request.

Note that the name `frame_timestamp_ranges` is a bit verbose, but `timestamps` is already taken for the per pixel timestamps. `time_ranges` doesn't communicate that we are talking about `timestamps` here and `frame_timestamps` doesn't imply well enough that the returned quantity is a range. Please feel free to suggest a better name. Especially since the plan is to add an alias on `CorrelatedStack` with the same name (so that the API for getting these downsampled series is the same on both sides).

Another PR will build on top of this to provide users with a small widget to quickly correlate these. This second PR will expose the same widget as we already have for `CorrelatedStack`.

Docs build here: https://lumicks-pylake.readthedocs.io/en/ds_scan/
Note that it is still a little spartan to warrant a whole section, but in the next PR the widget will be there too.

image


mwilkowski (approved):
solid

PR #228, opened 2021-12-03
Update API docs for `Slice.downsampled_like()`

rpauszek:
**Why this PR?**
We got a support question about the returned items from this function. Now it returns a tuple of `Slice` objects for the downsampled data and a cropped version of the original LF data with the same time stamps. Previously, only the downsampled `Slice` was returned.  This change is already documented in the tutorial, but here we add a description to the API page for clarity.

https://lumicks-pylake.readthedocs.io/en/ds_like_docs/_api/lumicks.pylake.channel.Slice.html#lumicks.pylake.channel.Slice.downsampled_like


JoepVanlier (approved):
LGTM! 🚢

PR #229, opened 2021-12-03
scan: add `plot_correlated` for scans

JoepVanlier:
**Why this PR?**
Allow users to also interactively explore the correlation between `Scan` objects and channel data.

**What to look out for?**
I also changed some of the default scaling behaviour. Reason being that the old behaviour opens figures too wide for realistically sized scans. Now I force the relative widths to be `1 : aspect_ratio` of the image, but I'd happily take any suggestions for better alternatives.

I considered piggybacking forwarding of `**kwargs` to this, similarly how we do it for `.plot()`, but decided against it, because I think we need to think about a better API for image adjustments than just forwarding to the very limited options on `imshow`.

Example (this was wider than the notebook):
scan

Here are some examples of the widget in use now:
wit_new
On a cropped TIFF stack (RGB)

wit_uncropped
On an uncropped TIFF stack (RGB)

scan_new
On a scan

image
On a scan (Green only)

### DIFF
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -1,6 +1,5 @@
 import numpy as np
 import os
-import json
 import tifffile
 import warnings
 from deprecated.sphinx import deprecated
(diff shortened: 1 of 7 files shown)


lumicks/pylake/nb_widgets/correlated_plot.py
```diff
@@ -0,0 +1,81 @@
...
+
+    if len(downsampled.timestamps) < len(frame_timestamps):
+        warnings.warn("Only subset of time range available for selected channel")
+
+    plot_data = get_plot_data(frame)
+    aspect_ratio = plot_data.shape[0] / np.max([plot_data.shape])
+
+    fig, (ax1, ax2) = plt.subplots(1, 2, figsize=plt.figaspect(aspect_ratio / 2))
+    t0 = downsampled.timestamps[0]
+    t, y = (downsampled.timestamps - t0) / 1e9, downsampled.data
```
rpauszek, line 53: can we use `downsampled.seconds` to define `t` here?

lumicks/pylake/correlated_stack.py
```diff
@@ -184,6 +186,8 @@ def plot_correlated(self, channel_slice, frame=0, reduce=np.mean, channel="rgb")
         channel : 'rgb', 'red', 'green', 'blue', None; optional
             Channel to plot for RGB images (None defaults to 'rgb')
             Not used for grayscale images
+        figure_scale : float
+            Scaling of the plot.
```
rpauszek, line 190: It's not super clear to me what this actually is. Is it the aspect ratio of the entire figure? Or the split ratio between the two plots? Might need a better docstring, maybe even a different parameter name

rpauszek (commented):
Good stuff! Really nice to have this functionality available for `Scan` also. I didn't approve just yet because it'd be nice to have some clarification on the new `fig_scale` argument, but other than that everything looks great!

lumicks/pylake/nb_widgets/correlated_plot.py
JoepVanlier: We certainly can. Fixing.

lumicks/pylake/correlated_stack.py
JoepVanlier: Rephrased docstring to:
```
        figure_scale : float
            Scaling of the figure width and height. Values greater than one increase the size of the
            figure.
```

rpauszek (approved):
noice

PR #231, opened 2021-12-09
force_calibration: add extra integration test hydro

JoepVanlier:
**Why this PR?**
This case was not specifically covered but provides a good integration test / sanity check for the surface handling.

For small beads, the results (Rf, Rd, and kappa) with or without the hydrodynamically correct model should be the same near a surface.

Since internally, the effect of the surface is handled very differently through all the calibration options (hydro on/off, active on/off), this provides an integration test that actually tests whether all the parts are functioning correctly.

### DIFF
--- a/lumicks/pylake/force_calibration/tests/test_hydro.py
+++ b/lumicks/pylake/force_calibration/tests/test_hydro.py
@@ -468,3 +468,79 @@ def test_brenner(bead_radius, result):
     np.testing.assert_allclose(
         brenner_axial(bead_radius + np.arange(40, 250, 80) * 1e-9, bead_radius), result
     )
+
+
+def test_near_surface_consistency():
+    """For small beads, the results (Rf, Rd, and kappa) with or without the hydrodynamically correct
+    model should be the same near a surface.
+
+    Since internally, the effect of the surface is handled very differently in all the calibration
+    options (hydro on/off, active on/off), this provides an integration test that actually tests
+    whether all the parts are functioning correctly.
+
+    AC on,  Hydro off - Drag is measured and used directly.
+    AC on,  Hydro on  - Drag is measured and used directly.
+    AC off, Hydro off - Surface effect on drag coefficient is modelled with Faxen's law.
+    AC off, Hydro on  - Surface effect on drag coefficient is in equation describing the spectrum.
+    """
+    bead_diameter = 0.5
+    shared_pars = {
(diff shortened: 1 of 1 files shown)


lumicks/pylake/force_calibration/tests/test_hydro.py
```diff
@@ -468,3 +468,80 @@ def test_brenner(bead_radius, result):
...
+        "sample_rate": 78125,
+        "driving_frequency_guess": 32,
+    }
+
+    def fit_spectrum(active, hydro):
+        model = (
+            PassiveCalibrationModel(**shared_pars, hydrodynamically_correct=hydro)
+            if active
+            else ActiveCalibrationModel(
+                **active_pars, **shared_pars, hydrodynamically_correct=hydro
```
rpauszek, line 521: shouldn't this be the other way around?

lumicks/pylake/force_calibration/tests/test_hydro.py
```diff
@@ -468,3 +468,80 @@ def test_brenner(bead_radius, result):
...
+        "Rf": sim_pars["stiffness"] * sim_pars["pos_response_um_volt"] * 1e3,
+    }
+
+    for active_calibration in (True, False):
+        for hydrodynamic_model in (True, False):
+            fit = fit_spectrum(active_calibration, hydrodynamic_model)
+
+            # Note that the corner frequency of the hydrodynamic model is natively specified in
+            # bulk, while the regular model has its corner frequency specified at the current
+            # height.
```
rpauszek, line 538: I think you mean `naively specified` in the first line? instead of `natively`

rpauszek (approved):
LGTM, just see my note on the ternary expression. It still goes through all the permutations, just isn't doing what you'd expect based on the input parameters

lumicks/pylake/force_calibration/tests/test_hydro.py
JoepVanlier: Nah, I meant `natively`, but maybe that was an English mishap. I'll omit the word entirely, because it doesn't add much.

lumicks/pylake/force_calibration/tests/test_hydro.py
JoepVanlier: Damn. Yes, of course. Sorry about that.

PR #232, opened 2021-12-09
force_feedback: support marker metadata

JoepVanlier:
**Why this PR?**
Users need to be able to access specific metadata associated with markers.
Basically it's a `json` attached to a key named `payload` on a marker.

Private API for now since this is for a single user in particular for now.

### DIFF
--- a/lumicks/pylake/tests/test_file.py
+++ b/lumicks/pylake/tests/test_file.py
@@ -3,6 +3,7 @@ from lumicks import pylake
 import pytest
 from textwrap import dedent
 from lumicks.pylake.detail.h5_helper import write_h5
+from .data.mock_json import force_feedback_dict
 
 
 def test_scans(h5_file):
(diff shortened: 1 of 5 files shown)


lumicks/pylake/tests/test_file.py
```diff
@@ -102,6 +103,17 @@ def test_marker(h5_file):
         assert np.isclose(f.markers["test_marker2"].stop, 300)
 
 
+def test_marker_metadata(h5_file):
+    f = pylake.File.from_h5py(h5_file)
+
+    if f.format_version == 2:
+        assert np.isclose(f.markers["force feedback"].start, 200)
+        assert np.isclose(f.markers["force feedback"].stop, 300)
```
rpauszek, line 111: ultimate nit-picky comment here, but I'd move these 2 lines to the end of this block .... keep all of the `markers["force feedback"]` grouped together

rpauszek (approved):
feel free to un-friend me over the nit-picky comment 😬

lumicks/pylake/tests/test_file.py
JoepVanlier: sure :)

PR #233, opened 2021-12-14
fdfit: move fdfit tests into its own subdirectory

JoepVanlier:
**Why this PR?**
All these tests lived in a single test file at the root level. I split them up to make it easier to navigate them.

**Notes**
- The first commit changes nothing to the tests themselves (only moves them). I simply categorized them and placed them in the fd fitting folder where they should be.
- Note that one test was dead, because it was incorrectly named (`integration_test_parameter_linkage`). I fixed this by renaming it.
- I also ran `black` on them for formatting and did a few tiny fixups to make sure black didn't split out some lines over many.

I also did some small fixups (in a separate commit) which fixes some issues flagged by the linter (unused variables, variables being shadowed and some assignments that had no effect (usually inside a context manager waiting for an exception)). Note that unfortunately, https://youtrack.jetbrains.com/issue/PY-39449 the latest PyCharm handles deleted arguments incorrectly.


rpauszek (approved):
no more 1200 line test file 🙌

PR #234, opened 2021-12-14
fdfit: fix standard errors in the presence of fixed parameters

JoepVanlier:
**Why this PR?**
Currently, Pylake evaluates the covariance matrix for the unconstrained fitting problem (not taking into account possible fixed parameters). This leads to overly pessimistic standard errors (too big).

This PR fixes that issue.

### DIFF
--- a/lumicks/pylake/fitting/tests/test_stderr.py
+++ b/lumicks/pylake/fitting/tests/test_stderr.py
@@ -4,24 +4,51 @@ import pytest
 import numpy as np
 
 
-def test_no_jac():
+def test_data():
     x = np.arange(10)
     y = np.array(
         [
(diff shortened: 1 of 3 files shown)


lumicks/pylake/fitting/tests/test_stderr.py
```diff
@@ -4,24 +4,52 @@
...
+            124.25500144427338,
         ]
     )
+    return x, y
+
+
+def linear_func():
+    def linear(x, a, b):
+        f = a * x + b
+        return f
```
rpauszek, line 29: just return directly, don't need to define `f`

rpauszek (approved):
LGTM

PR #235, opened 2021-12-16
widgets: increase tolerance connect lines

JoepVanlier:
**Why this PR?**
Previously, we were determining which line tip was clicked by converting to pixel coordinates. This connection tolerance can be surprisingly finicky in `jupyter notebook` but not `lab`. It seems that internally it renders to a higher resolution image.

This PR changes that handling to handle whether the click was close enough by considering the distance relative to the axes data limits.


rpauszek (approved):
👍

PR #238, opened 2021-12-17
Replace `CorrelatedStack.timestamps` with `frame_timestamp_ranges` (via deprecation)

rpauszek:
**Why this PR?**

The values of `CorrelatedStack.timestamps` was inconsistent with those from `Scan.timestamps`. The frame start/stop timestamp pairs were returned for the former while individual pixel timestamps are returned for the latter.  Instead, `Scan.frame_timestamp_ranges` (added in #229) is in line with the values available for camera based images. With this change, the two APIs are now consistent


JoepVanlier (changes requested):
Hey, looks good. A few requests though:

- You should also update the `.timestamps` on line `37` in `correlated_stack.py`.
- Same thing in `channel.py` on line `177` (`downsampled_over`).

JoepVanlier (approved):
LGTM! 👍

PR #239, opened 2021-12-20
Add cropping to WIT image editor

rpauszek:
**Why this PR?**
drag to crop is nice. Docs are [here](https://lumicks-pylake.readthedocs.io/en/interactive_crop/tutorial/nbwidgets.html#image-stack-editor)

*Note*: while writing the tests for this, I noticed that our mock event for rectangle selection had the coordinates backwards; it was receiving `(y, x)` pairs instead of `(x, y)` pairs. I piggybacked the fix for this here as I needed that even for tests anyway.

### DIFF
--- a/lumicks/pylake/nb_widgets/image_editing.py
+++ b/lumicks/pylake/nb_widgets/image_editing.py
@@ -2,6 +2,7 @@ import numpy as np
 from dataclasses import dataclass, field
 import matplotlib.pyplot as plt
 from matplotlib.axes import Axes
+from matplotlib.widgets import RectangleSelector
 
 
 class ImageStackAxes(Axes):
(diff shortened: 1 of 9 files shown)


lumicks/pylake/nb_widgets/image_editing.py
```diff
@@ -141,6 +165,13 @@ def update_plot(self, tether_coordinates):
             setter(lims)
         self.get_figure().canvas.draw()
 
+    def get_edited_image(self):
+        return (
+            self.current_image
```
JoepVanlier, line 170: Does anyone still use `current_image`? Should it still be public? Otherwise it might be confusing because now there are two non-private images.

JoepVanlier (commented):
Tested it. Works well! :)

A few minor comments.

JoepVanlier (approved):
LGTM :)

PR #240, opened 2021-12-22
Histogram kymotrack binding events

rpauszek:
**Why this PR?**
Requested convenience function from users. I split this up into an internal backend function that actually calculates the data needed to construct the histogram (for use in other places) and a public function for actually generating the plot.

The original ticket requested a histogram of only the initial binding events (the first position of each track); however it was straightforward to also include histogramming all of the positions of each track, which is also a useful metric, so I wrapped it in this PR.

Docs are [here](https://lumicks-pylake.readthedocs.io/en/hist_bind_events/tutorial/kymotracking.html#plotting-binding-histograms)

### DIFF
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -224,6 +224,39 @@ Here `num_pixels` is the number of pixels to sum on either side of the trace.
 .. image:: kymo_sumcounts.png
 
 
+Plotting binding histograms
+---------------------------
+
+We can easily plot some histograms of the binding events located with the kymotracker. Simply use::
+
+    plt.figure()
+    traces.plot_binding_histogram(kind="binding")
+
+.. image:: kymo_bind_histogram_1.png
+
+Here, the `kind="binding"` argument indicates that we only wish to analyze the initial binding events (the first
+position of each track). We can optionally supply a `bins` argument, which is forwarded to `np.histogram()`.
+For instance, we can increase the number of bins from 10 (the default) to 50::
+
+    plt.figure()
+    traces.plot_binding_histogram("binding", bins=50)
+
+.. image:: kymo_bind_histogram_2.png
+
+When an integer is supplied to the `bins` argument, the full position range is used to calculate the bin edges (this is
+equivalent to using `np.histogram(data, bins=n, range=(0, max_position))`). This facilitates comparison of histograms calculated
+from different kymographs, as the absolute x-scale is dependent on the kymograph acquisition options, rather than the positions
+of the tracked lines. Alternatively, it is possible to supply a custom array of bin edges, as demonstrated below::
+
+    plt.figure()
+    traces.plot_binding_histogram("kind=all", bins=np.linspace(12, 18, 75), fc="#dcdcdc", ec="tab:blue")
+
+.. image:: kymo_bind_histogram_3.png
+
+Notice that here we use `kind="all"` to include all of the bound positions for each track. This snippet also demonstrates
+how we can pass keyword arguments (forwarded to `plt.bar()`) to format the histogram.
+
+
 Exporting kymograph traces
 --------------------------
 
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -416,3 +416,51 @@ class KymoLineGroup:
         return _kinetic_mle_optimize(
             n_components, dwelltimes_sec, min_observation_time, max_observation_time
         )
+
+    def _histogram_binding_events(self, kind, bins=10):
+        """Make histogram of bound events.
+
+        Parameters
+        ----------
+        kind : {'binding', 'all'}
+            Type of events to count. 'binding' counts only the first position of each track whereas
+            'all' counts positions of all time points in each track.
+        bins : int or sequence of scalars
+            Definition of the histogram bins; passed to `np.histogram()`. If `int`, defines the total
+            number of bins. If sequence of scalars, defines the edges of the bins.
+        """
+        if kind == "all":
+            slc = slice(None)
+        elif kind == "binding":
+            slc = slice(0, 1)
+        else:
+            raise ValueError(f"`kind` argument '{kind}' must be 'all' or 'binding'.")
+        events = np.hstack([track.position[slc] for track in self])
+
+        image = self[0]._image
+        pos_range = (0, image._pixel_size * image.data.shape[0])
+        return np.histogram(events, bins=bins, range=pos_range)
+
+    def plot_binding_histogram(self, kind, bins=10, **kwargs):
+        """Plot histogram of bound events.
+
+        Parameters
+        ----------
+        kind : {'binding', 'all'}
+            Type of events to count. 'binding' counts only the first position of each track whereas
+            'all' counts positions of all time points in each track.
+        bins : int or sequence of scalars
+            Definition of the histogram bins; passed to `np.histogram()`. When an integer is supplied
+            to the `bins` argument, the full position range is used to calculate the bin edges (this
+            is equivalent to using `np.histogram(data, bins=n, range=(0, max_position))`). If a
+            sequence of scalars is provided, they directly define the edges of the bins.
+        **kwargs
+            Keyword arguments forwarded to `plt.bar()`.
+        """
+        import matplotlib.pyplot as plt
+
+        counts, edges = self._histogram_binding_events(kind, bins)
+        widths = np.diff(edges)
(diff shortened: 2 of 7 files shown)


docs/tutorial/kymotracking.rst
```diff
@@ -224,6 +224,35 @@ Here `num_pixels` is the number of pixels to sum on either side of the trace.
...
+For instance, we can increase the number of bins from 10 (the default) to 50::
+
+    plt.figure()
+    traces.plot_binding_histogram("binding", bins=50)
+
+.. image:: kymo_bind_histogram_2.png
+
+Finally, we can use `kind="all"` to include all of the bound positions for each track. This snippet also demonstrates
+another way to define the histogram bins, by supplying an array of bin edges directly. We can also pass
+keyword arguments (forwarded to `plt.bar()`) to format the histogram::
```
mcsmurf, line 248:
```suggestion
Finally, we can use `kind="all"` to include all of the bound positions for each track. This snippet also demonstrates another way to define the histogram bins, by supplying an array of bin edges directly. We can also pass
keyword arguments (forwarded to `plt.bar()`) to format the histogram::
```

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -416,3 +416,49 @@ def fit_binding_times(self, n_components):
...
+        kind : {'binding', 'all'}
+            Type of events to count. 'binding' counts only the first position of each track whereas
+            'all' counts positions of all time points in each track.
+        bins : int or sequence of scalars
+            Definition of the histogram bins; passed to `np.histogram()`. If `int`, defines the total
+            number of bins. If sequence of scalars, defines the edges of the bins.
+        **kwargs
+            Keyword arguments forwarded to `plt.bar()`.
+        """
+        import matplotlib.pyplot as plt
```
mcsmurf, line 459: I'd probably move imports to the top, as explained [here](https://stackoverflow.com/a/9614212/3154670)

lumicks/pylake/kymotracker/tests/test_kymoline.py
```diff
@@ -229,3 +229,34 @@ def test_kymoline_msd_plot(max_lag, x_data, y_data):
...
+    channel = CalibratedKymographChannel("test_data", np.array([[]]), 1e9, 1)
+
+    k1 = KymoLine(np.array([1, 2, 3]), np.array([2.5, 3.5, 4.5]), channel)
+    k2 = KymoLine(np.array([2, 3, 4]), np.array([3.5, 4.5, 5.5]), channel)
+    k3 = KymoLine(np.array([3, 4, 5]), np.array([4.5, 5.5, 6.5]), channel)
+    k4 = KymoLine(np.array([4, 5, 6]), np.array([5.5, 6.5, 7.5]), channel)
+
+    lines = KymoLineGroup([k1, k2, k3, k4])
+
+    counts, centers, width = lines._histogram_binding_events("binding")
```
mcsmurf, line 244: I'd annotate these sections for additional coherence. Something like
```suggestion
   # Counting only the first position of each track with the default number of bins
    counts, centers, width = lines._histogram_binding_events("binding")
```

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -416,3 +416,49 @@ def fit_binding_times(self, n_components):
...
+
+        Parameters
+        ----------
+        kind : {'binding', 'all'}
+            Type of events to count. 'binding' counts only the first position of each track whereas
+            'all' counts positions of all time points in each track.
+        bins : int or sequence of scalars
+            Definition of the histogram bins; passed to `np.histogram()`. If `int`, defines the total
+            number of bins. If sequence of scalars, defines the edges of the bins.
+        """
```
mcsmurf, line 431: This sounds like there are two functions packed into one. We could maybe split this into `_histogram_count_first_positions()` and a `_histogram_count_all_points()` if that makes sense.

mcsmurf (commented):
Looks pretty good! I left some comments.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -416,3 +416,49 @@ def fit_binding_times(self, n_components):
...
+        if kind == "all":
+            slc = slice(None)
+        elif kind == "binding":
+            slc = slice(0, 1)
+        else:
+            raise ValueError(f"`kind` argument '{kind}' must be 'all' or 'binding'.")
+        binds = np.hstack([track.position[slc] for track in self])
+
+        counts, edges = np.histogram(binds, bins=bins)
```
JoepVanlier, line 440: We already discussed this briefly, but just so it isn't forgotten.

It would be good to specify the `range` parameter to make sure the `extents` correspond to the kymograph's `extents`. This will make the binning with a set number of bins reproducible, no matter where the lines are actually tracked.

This can be useful if we're for example looking at a few slices of the same kymo, and want to compare binding histograms. If we don't specify the `range`, the bin edges will depend on the lines tracked (using the minimum and maximum of the found line positions). Ideally they shouldn't, so histograms for a particular kymo are directly comparable.

In the future, we could consider maybe an optional argument for the range behaviour (`full` (default), `auto` or a manually specified range for instance), but for now this `full` behaviour is probably sufficient.

JoepVanlier (changes requested):
Mostly looks good to me. I only have one request regarding the `range` parameter.

lumicks/pylake/kymotracker/kymoline.py
rpauszek: I'd agree, although generally throughout the library we keep `matplotlib` imports within the functions where they are needed. This is because plotting is only required for a select number of convenience functions and so only imported specifically when needed. This probably doesn't lead to a huge boost in import time for `pylake` in general, but here, for consistency, I used the same pattern we usually do

lumicks/pylake/kymotracker/kymoline.py
rpauszek: I had a chat about this with Tiago to try to get some insight about why this might be a better option; for something like C++ I can see that there might be compiler optimizations or other reasons to split this into separate functions. But for python I don't think it makes sense. We'd end up with two functions that are identical except for a single line, so I'd rather take the de-duplication and deal with a bit more logic within the single function.

That being said, I'm not particularly fond of using a string as an argument to determine the type of histogram calculated; but this is a fairly common pattern in python so I guess it's ok - I don't think there's really a better option, especially considering that this might become a public function in the future. I would've also preferred something like `type` instead of `kind`, but `type` is a core function. I'm open to other ideas for the argument name though

lumicks/pylake/kymotracker/kymoline.py
rpauszek: good point! I also updated the documentation to specifically address this point

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -416,3 +416,49 @@ def fit_binding_times(self, n_components):
...
+        else:
+            raise ValueError(f"`kind` argument '{kind}' must be 'all' or 'binding'.")
+        events = np.hstack([track.position[slc] for track in self])
+
+        image = self[0]._image
+        pos_range = (0, image._pixel_size * image.data.shape[0])
+        return np.histogram(events, bins=bins, range=pos_range)
+
+    def plot_binding_histogram(self, kind, bins=10, **kwargs):
+        """Make histogram of bound events.
```
JoepVanlier, line 445: Make or plot?

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -416,3 +416,49 @@ def fit_binding_times(self, n_components):
...
+
+        Parameters
+        ----------
+        kind : {'binding', 'all'}
+            Type of events to count. 'binding' counts only the first position of each track whereas
+            'all' counts positions of all time points in each track.
+        bins : int or sequence of scalars
+            Definition of the histogram bins; passed to `np.histogram()`. If `int`, defines the total
+            number of bins. If sequence of scalars, defines the edges of the bins.
```
JoepVanlier, line 454: You could add the little addition you made to the docs here as well, for people who prefer reading from those. Something like:

```suggestion
            Definition of the histogram bins; passed to `np.histogram()`. When an integer is supplied
            to the `bins` argument, the full position range is used to calculate the bin edges (this 
            is equivalent to using `np.histogram(data, bins=n, range=(0, max_position))`). If a 
            sequence of scalars is provided, they directly define the edges of the bins.
```

JoepVanlier (approved):
LGTM. Two minor optional nitpicks. Up to you whether you want to add them or not.

PR #241, opened 2021-12-28
Cleanup tests

rpauszek:
**Why this PR?**
We've added a lot of new functionality to `pylake`; however, our tests need a bit of spring cleaning. In particular, certain files are getting too long to manage. Additionally, the use of the single `h5_file` fixture over many tests has led to problems in the past where the fixture needed to be changed and many tests needed to be updated.

So I propose the following refactors:
* Move long test files into subdirectories. These can then be split into separate files to group different types of tests
* Decouple `h5_file` as much as possible. Many of the tests that currently rely on this can instead be handled by newer mock classes

**Notes**
* I haven't touched any of the assertions (except see next point). All of the refactoring is moving things around and re-structuring where the data is generated (mock classes when possible, new file fixtures when needed)
* When I moved tests to subdirectories, it broke some of the plotting tests. These currently assert on the axes limits, which seems to be fragile and not particularly informative (since the limits are dependent on the configuration of `mpl`; we don't specify the limits ourselves.) So I've re-written the tests such that they spin up the plot, grab the line/image artists, and then assert on the artist data. I think this is a safer way to go about things.
* Each of the subdirectories have their own `conftest.py` to collect fixtures specific to those specific tests. Accordingly, I changed the `scope` from `"session"` to `"module"` to enforce that these fixtures should not be used elsewhere.


lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -6,303 +6,300 @@
...
+    # np.testing.assert_allclose(np.sort(plt.xlim()), [-0.5, 3.5], atol=0.05)
+
+    # # The following two assertions are fragile (they seem to break when the test file is moved
+    # # to a different directory). These have been replaced with tests asserting on the actual
+    # # image data.
+    # np.testing.assert_allclose(plt.xlim(), [-0.515625, 3.609375])
+    # np.testing.assert_allclose(np.sort(plt.ylim()), [0, 0.05])
+    image = plt.gca().get_images()[0]
+    np.testing.assert_allclose(image.get_array(), kymo.red_image)
+    # np.testing.assert_allclose(image.get_extent(), [0, 0.197 * 3, 0.191 * 4, 0])
```
JoepVanlier, line 155: While I agree that maybe testing the limits is not a good idea, we do explicitly set extents. So may be good to indeed check these.

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -6,303 +6,300 @@
...
+    # # The following assertion fails because of unequal line times in the test data. These
+    # # unequal line times are not typical for BL data. Kymo nowadays assumes equal line times
+    # # which is why the old version of this test fails.
+    # np.testing.assert_allclose(np.sort(plt.xlim()), [-0.5, 3.5], atol=0.05)
+
+    # # The following two assertions are fragile (they seem to break when the test file is moved
+    # # to a different directory). These have been replaced with tests asserting on the actual
+    # # image data.
+    # np.testing.assert_allclose(plt.xlim(), [-0.515625, 3.609375])
+    # np.testing.assert_allclose(np.sort(plt.ylim()), [0, 0.05])
```
JoepVanlier, line 152: I'm fine with dropping the old tests, but there should be some test that verifies that what we show on the axes is correct.

lumicks/pylake/tests/test_imaging_confocal/conftest.py
```diff
@@ -0,0 +1,211 @@
...
+        dt,
+        [0],
+        [5],
+        [10.0],
+        infowave=reference_infowave,
+        red_photon_counts=reference_counts,
+        green_photon_counts=reference_counts,
+        blue_photon_counts=reference_counts,
+    )
+    kymos["damaged"] = Kymo("fast Y slow X", mock_file, start - 62500000, stop, metadata)
```
JoepVanlier, line 82: Key name could indicate what's broken about it. Or indeed as Dean mentioned, a separate fixture with a docstring.

Also the name of the `Kymo` itself is not consistent with the name of the key which could be confusing if it fires?

lumicks/pylake/tests/test_imaging_confocal/conftest.py
```diff
@@ -0,0 +1,211 @@
...
+        [1, 0],
+        [4, 5],
+        [191.0, 197.0],
+        infowave=reference_infowave,
+        red_photon_counts=reference_counts,
+        green_photon_counts=reference_counts,
+        blue_photon_counts=reference_counts,
+    )
+    middle = scans["fast Y slow X"].red_photon_count.timestamps[5]
+    scans["sted bug"] = Scan("fast Y slow X", mock_file, middle - 62400000, stop, metadata)
```
JoepVanlier, line 174: Naming inconsistency between key and `Scan` name.

JoepVanlier (changes requested):
Great work man.

I had a closer look specifically at the number of tests before and after this PR (since this reduced). But it seems that it was just counting a bunch of the test iterations that use `MockFile_v1` vs `MockFile_v2` as separate tests. Since some of those now directly use a `Kymo` it indeed removed some superfluous "tests" that are found during discovery but actually did nothing. As far as I can tell, everything seems to be there.

I also had a look at the coverage between your branch and `main` and this is unchanged.

I have a few requests, but nothing major.

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
rpauszek: yes I agree. I had written that line that's commented out (oops) incorrectly. It's fixed now

lumicks/pylake/tests/test_imaging_confocal/conftest.py
rpauszek: good catch, copy/paste error

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
rpauszek: same issue as above, handled correctly now

JoepVanlier (approved):
Looks good to me. :shipit:

PR #242, opened 2022-01-04
Add copy method for confocal classes

rpauszek:
**Why this PR?**
Specifically for `Kymo` we ended up with 4 methods which require copying `self` and modifying only a few attributes resulting in duplication of the rather long `__init__()` call. This way is cleaner.

Note: inside `BaseScan.__copy__()` we call the `__init__()` method such that the cache is not propagated to the copied instances.


JoepVanlier (approved):
LGTM!

PR #244, opened 2022-01-06
Deprecate `plot_{color}` methods for confocal classes

rpauszek:
**Why this PR?**
Currently the confocal classes (`PointScan`, `Kymo`, and `Scan`) have 4 separate methods for plotting each of the color(s). However, more recent (and more complex) plotting features use the signature `.plot(channel="color")`. Here, we suggest deprecating the old methods and aliasing to a new `BaseScan.plot()` method. This also allowed for some simplification of the existing code.

I also included an optional `axes` argument where you can supply an `mpl.Axes` instance to plot in. If this argument is not supplied, the default functionality is exactly the same as before (using the current axes or creating an instance if there is not one current). This simplifies construction of more complex compound plots and widgets by allowing use of the `plot()` method rather than duplicating the plotting code. In the future, I think we should include this argument in some of our other base plotting functions.


lumicks/pylake/kymo.py
```diff
@@ -155,12 +165,13 @@ def _plot(self, image, **kwargs):
             # pixel center aligned with mean time per line
             extent=[-0.5 * linetime, duration - 0.5 * linetime, size_calibrated, 0],
             aspect=(image.shape[0] / image.shape[1]) * (duration / size_calibrated),
+            cmap=linear_colormaps[channel],
```
JoepVanlier, line 168: Is there a reason you prefer pulling this into both functions, instead of having it at the `_plot_color` level like we did before?

I have a mild preference for having it in the other location, since changing it once will change it in both places in that case.

lumicks/pylake/detail/confocal.py
```diff
@@ -13,6 +13,7 @@
     "red": LinearSegmentedColormap.from_list("red", colors=[(0, 0, 0), (1, 0, 0)]),
     "green": LinearSegmentedColormap.from_list("green", colors=[(0, 0, 0), (0, 1, 0)]),
     "blue": LinearSegmentedColormap.from_list("blue", colors=[(0, 0, 0), (0, 0, 1)]),
+    "rgb": None,
```
JoepVanlier, line 16: I like that this simplifies the logic elsewhere.

lumicks/pylake/scan.py
```diff
@@ -159,13 +155,15 @@ def _to_spatial(self, data):
...
+    def _plot(self, channel, axes, frame=1, image_handle=None, **kwargs):
+        """Plot a scan frame for requested color channel(s).
 
         Parameters
         ----------
-        image : array_like
-            Image data
+        channel : {'red', 'green', 'blue', 'rgb'}
+            Color channel to plot.
+        axes : mpl.axes.Axes or None
```
JoepVanlier, line 165: I don't think this can be optional? Since you use it.

lumicks/pylake/kymo.py
```diff
@@ -206,8 +217,7 @@ def set_aspect_ratio(axis, ar):
         _, (ax1, ax2) = plt.subplots(2, 1, sharex="all")
 
         # plot kymo
-        plt.sca(ax1)
-        getattr(self, f"plot_{color_channel}")()
+        self.plot(channel=color_channel, axes=ax1)
```
JoepVanlier, line 220: nice fixup :)

JoepVanlier (changes requested):
Looks fine to me. Just a few minor comments.

One thing that I'm wondering about though, is whether we should consider something similar for the images, since now it seems like we have two different API styles for plotting and getting the images out.

JoepVanlier (approved):
:shipit:

PR #245, opened 2022-01-06
Interactive cropping and calibration for kymos

rpauszek:
**Why this PR?**
Interactive plots are nice. The interactive plot works essentially the same as the recently added one for widefield data (#236 and #239)

I also re-arranged some of the [docs](https://lumicks-pylake.readthedocs.io/en/interactive_kymo/tutorial/kymographs.html) and added images, as these docs were pretty barren before.


JoepVanlier (approved):
Looks nice :)

I really like that you took the time to improve the docs as well.

Only have two minor requests really.

PR #247, opened 2022-01-07
docs: small fixes to msd documentation

aafkevandenberg:
The example `kymolines[0].msd(mag_lag = 5)`, has been changed to `kymolines[0].msd(max_lag = 5)`

The docs mentioned that 'it may be tempting to use a large number of MSDs', but this should be 'a large number of lags'


JoepVanlier (approved):
Looks good to me! Thanks! :)

PR #249, opened 2022-01-07
Fixup `CorrelatedStack` slicing behavior

rpauszek:
**Why this PR?**
I noticed some weird behavior when attempting to slice a stack in reverse (`stack[5:2]`) which ended up giving a new instance with `stack.num_frames < 0`, which obviously doesn't make sense. Since there is no "empty set" version for `CorrelatedStack` as would be the default behavior for this slicing in python, we decided to just throw. Additionally, we now throw when the resulting slice is empty (`stack[5:5]`).

I also took this opportunity to split up the tests a bit, in line with the recent overhaul #241. The last 2 commits are just moving the original test file, and then copy/paste some tests into new files for manageability.


tiagobonetti (commented):
Looks good to me. 

I'm not sure this is a breaking change since it only breaks things that make no sense anyway, but we probably need some notes about it for the changelog.

lumicks/pylake/tests/test_correlated_stack.py
```diff
@@ -41,7 +41,8 @@ def test_correlated_stack(shape):
     with pytest.raises(IndexError):
         stack[3:5][2]
 
-    assert(stack[2:5][3:5].num_frames == 0)
+    with pytest.raises(NotImplementedError, match="Slice is empty"):
+        assert(stack[2:5][3:5].num_frames == 0)
```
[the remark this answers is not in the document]
rpauszek: Good point, my initial instinct was not to delete any tests. Instead, I added another specific test for iterative slicing

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -63,6 +60,44 @@ def test_correlated_stack(shape):
...
+    compare_frames([3], stack0[3])  # single frame
+    compare_frames([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], stack0[:])  # all frames
+    compare_frames([3, 4, 5], stack0[3:6])  # normal slice
+    compare_frames([0, 1, 2], stack0[:3]) #  from beginning
+    compare_frames([6, 7, 8, 9], stack0[6:]) #  until end
+    compare_frames([0, 1, 2, 3, 4, 5], stack0[:-4]) #  until negative index
+    compare_frames([5, 6, 7], stack0[5:-2]) #  mixed sign indices
+    compare_frames([6, 7], stack0[-4:-2]) #  negative indices slice
+    compare_frames([3, 4], stack0[2:6][1:3]) #  iterative slicing
+
```
JoepVanlier, line 87: I like that we now have all these slicing tests.

I'd suggest three extra cases:
```python
    compare_frames([1, 2, 3, 4, 5, 6, 7, 8, 9], stack0[1:1000])  # test clamping past the end
    compare_frames([0, 1, 2, 3, 4, 5, 6, 7, 8], stack0[-1000:10])  # test clamping past the beginning
    compare_frames([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], stack0[-1000:1000])  # test clamping both dirs
```

JoepVanlier (changes requested):
I didn't review this one in detail, but just two minor things I noticed.

I like that you added tests for the slicing that actually test the data.

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -63,6 +60,44 @@ def test_correlated_stack(shape):
...
+    compare_frames([9], stack0[-1])  # last frame
+    compare_frames([3], stack0[3])  # single frame
+    compare_frames([0, 1, 2, 3, 4, 5, 6, 7, 8, 9], stack0[:])  # all frames
+    compare_frames([3, 4, 5], stack0[3:6])  # normal slice
+    compare_frames([0, 1, 2], stack0[:3]) #  from beginning
+    compare_frames([6, 7, 8, 9], stack0[6:]) #  until end
+    compare_frames([0, 1, 2, 3, 4, 5], stack0[:-4]) #  until negative index
+    compare_frames([5, 6, 7], stack0[5:-2]) #  mixed sign indices
+    compare_frames([6, 7], stack0[-4:-2]) #  negative indices slice
+    compare_frames([3, 4], stack0[2:6][1:3]) #  iterative slicing
```
JoepVanlier, line 86: Would be nice to stick to PEP8 commenting style like the first 5.

JoepVanlier (approved):
LGTM!

PR #250, opened 2022-01-10
Slice `Scan` objects by frame indices

rpauszek:
**Why this PR?**
This is a basic functionality that we've been putting off for a while, but the recent introduction of `Scan.frame_timestamp_ranges` makes it fairly simple to implement. Behavior is identical to that implemented for `CorrelatedStack`.

[Docs here](https://lumicks-pylake.readthedocs.io/en/scan_slicing/tutorial/images.html)

### DIFF
--- a/docs/tutorial/images.rst
+++ b/docs/tutorial/images.rst
@@ -29,13 +29,31 @@ Or just pick a single one::
 Scan data and details
 ---------------------
 
-You can access the raw image data::
+You can access the raw image data directly. For a `Scan` with only a single frame::
 
     rgb = scan.rgb_image  # matrix with `shape == (h, w, 3)`
     blue = scan.blue_image  # single color so `shape == (h, w)`
 
     # Plot manually
     plt.imshow(rgb)
+
+For scans with multiple frames::
+
+    # returned data has `shape == (n_frames, h, w, 3)`
+    rgb = multiframe_scan.rgb_image
+    # returned data has `shape == (n_frames, h, w)`
+    blue = multiframe_scan.blue_image
+
+    # Manually plot the RGB image of the first frame.
+    plt.imshow(rgb[0, :, :, :])
+
+We can also slice out a subset of frames from an image stack::
+
+    sliced_scan = multiframe_scan[5:10]
+
+This will return a new `Scan` containing data equivalent to:
+
+    multiframe_scan.rgb_image[5:10, :, :, :]
 
 The images contain pixel data where each pixel represents summed photon counts.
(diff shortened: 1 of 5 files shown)


docs/tutorial/images.rst
```diff
@@ -29,14 +29,32 @@ Or just pick a single one::
...
+    # Manually plot the RGB image of the first frame.
+    plt.imshow(rgb[0, :, :, :])
+
+We can also slice out a subset of frames from an image stack::
+
+    sliced_scan = multiframe_scan[5:10]
+
+This will return a new `Scan` containing data equivalent to:
+
+    multiframe_scan.rgb_image[5:10, :, :, :]
```
JoepVanlier, line 56: Depending on which PR goes in first, you may want to update these to the new API.

lumicks/pylake/scan.py
```diff
@@ -36,7 +36,22 @@ def __repr__(self):
...
     def __getitem__(self, item):
-        raise NotImplementedError("Indexing and slicing are not implemented for scans")
+        """All indexing is in frames"""
+        ts_ranges = self.frame_timestamp_ranges(exclude=False)
+
+        if isinstance(item, slice):
+            if item.step is not None:
+                raise IndexError("Slice steps are not supported")
+            start = ts_ranges[0 if item.start is None else item.start][0]
+            stop = ts_ranges[self.num_frames - 1 if item.stop is None else item.stop - 1][-1]
```
JoepVanlier, line 46: This logic looks a bit complicated and isn't really necessary?

It also misses a few corner cases. For example:
`scan[-more_than_the_frame_count:]` or `scan[:more_than_the_frame_count]`

Could just let Python handle it?
```python
sliced_ts_ranges = ts_ranges[item]
if not sliced_ts_ranges:
  raise NotImplementedError("Slice is empty.")

start, stop = sliced_ts_ranges[0][0], sliced_ts_ranges[-1][-1]
```

Quickly trying it passes all tests, but maybe I'm overlooking something?

JoepVanlier (changes requested):
Lookin' good. This'll be very convenient to have :)

I have one suggestion regarding how the slicing is handled.

lumicks/pylake/scan.py
rpauszek: good catch! Implemented this now and took advantage of the new copy functionality for confocal classes. Also added some additional assertions to the tests to verify `num_frames` is correct

JoepVanlier (approved):
LGTM! 🚢

PR #252, opened 2022-01-11
fdfit: allow using custom plot label

JoepVanlier:
**Why this PR?**
Currently, when plotting a fit made with the FdFitter, we put the data name as default label. If a user tries to override this, they get an exception (since the label is provided both by us and the `**kwargs` then.

This PR adds the option to provide a `label` argument.

image


rpauszek (approved):
:shipit:

PR #253, opened 2022-01-12
Convenient calibrations

JoepVanlier:
**Why this PR?**
It can be convenient to just have a single function call for performing a calibration and have the backend worry about how to hook everything up in the low level API.

This allows some ease of use when writing notebooks where you generate dictionaries with calibration parameters or when you quickly want to switch between active and passive calibration and compare results.

I tried my best to order and group the parameters in a sensible way. I do enforce keyworded arguments, so that we could potentially iterate on the order of the optional ones if feedback arises.

A note on the testing. I tested that the non-default parameters came through correctly. I did not do a full spectrum simulation, because those are a bit costly to generate, and there's no added value considering that all the backend functions are separately tested for correctness.

From:
```python
model = lk.ActiveCalibrationModel(driving_data, force_voltage_data, sample_rate, **model_params)
ps = lk.calculate_power_spectrum(force_voltage_data, sample_rate, **spectral_pars)
fit = lk.fit_power_spectrum(ps, model)
```
To:
```python
fit = lk.force_calibration(force_voltage_data, driving_data=driving_data, sample_rate=sample_rate, **model_params, **spectral_pars)
```

Docs build here: https://lumicks-pylake.readthedocs.io/en/convenient_calibrations/tutorial/force_calibration.html#more-convenient-calibration


lpingas (approved):
It looks great!

PR #254, opened 2022-01-13
kymowidget: add undo/redo functionality

JoepVanlier:
### Why this PR?
This was fairly highly requested functionality for the kymotracker widget.

### What to look out for?
Since the tracking steps typically take long and the actual data being stored is very very small (just some coordinates), I opted for a model where a history stack of the entire state of all the lines is stored. This keeps things implementationally quite simple.

However, since the `KymoLineGroup` is not immutable, there were two places where I was forced to make an explicit copy before changing it. I have added a comment on the lines where this happens.

We could alternatively consider going for a model where the `KymoLineGroup` is immutable. This would simplify the code on the widget side, but induce unnecessary copies elsewhere. It would also be a breaking change.

undoredo

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -1,12 +1,14 @@
 import time
 import numpy as np
 import matplotlib.pyplot as plt
+from copy import copy
 from matplotlib.widgets import RectangleSelector
 from lumicks.pylake.kymotracker.detail.calibrated_images import CalibratedKymographChannel
 from lumicks.pylake.kymotracker.kymotracker import track_greedy
 from lumicks.pylake import filter_lines, refine_lines_centroid
 from lumicks.pylake.nb_widgets.detail.mouse import MouseDragCallback
 from lumicks.pylake.kymotracker.kymoline import KymoLineGroup, import_kymolinegroup_from_csv
+from lumicks.pylake.nb_widgets.detail.undostack import UndoStack
 
 
 class KymoWidget:
@@ -61,7 +63,7 @@ class KymoWidget:
             if axis_aspect_ratio
             else None
         )
-        self.lines = KymoLineGroup([])
+        self._lines_history = UndoStack(KymoLineGroup([]))
         self.plotted_lines = []
         self.min_length = min_length
         self._min_length_range = min_length_range
@@ -85,6 +87,14 @@ class KymoWidget:
         self.show(use_widgets=use_widgets, **kwargs)
 
     @property
+    def lines(self):
+        return self._lines_history.state
+
+    @lines.setter
+    def lines(self, new_lines):
+        self._lines_history.state = new_lines
+
+    @property
     def _line_width_pixels(self):
         calibrated_image = CalibratedKymographChannel.from_kymo(self._kymo, self._channel)
         return np.ceil(self.algorithm_parameters["line_width"] / calibrated_image._pixel_size)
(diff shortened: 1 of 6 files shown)


lumicks/pylake/nb_widgets/detail/undo_redo.py
```diff
@@ -0,0 +1,24 @@
+class UndoableState:
```
lpingas, line 1: I would suggest a different name... 

Maybe `StateHistory`? I've also seen people calling this type of thing `Memento`.

lumicks/pylake/nb_widgets/detail/undo_redo.py
```diff
@@ -0,0 +1,24 @@
...
+    """Keeps a state with an undo and redo list."""
+
+    def __init__(self, initial_state):
+        self.state = initial_state
```
lpingas, line 5: I believe `state` should not be public to keep the class invariant. I suggest going with `._state` and with maybe a property and setter?

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -84,6 +86,14 @@ def __init__(
 
         self.show(use_widgets=use_widgets, **kwargs)
 
+    @property
+    def lines(self):
+        return self.lines_history.state
+
+    @lines.setter
+    def lines(self, new_lines):
+        self.lines_history.set_state(new_lines)
```
lpingas, line 95: 👌

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -95,6 +105,7 @@ def track_kymo(self, click, release):
         Removes lines in a region, and traces new ones."""
         p1 = [click.xdata, click.ydata]
         p2 = [release.xdata, release.ydata]
+        self.lines = copy(self.lines)  # Explicit copy to maintain history, since we modify in-place
```
lpingas, line 108: I would argue that `copy` is needed within `UndoableState` since its purpose is defeated if it doesn't hold separate copies.

lpingas (commented):
This is wonderful work! I really liked how you solved it and I left some comments for you to consider.

lumicks/pylake/nb_widgets/detail/undo_redo.py
mikhas: UndoStack

JoepVanlier: I like `UndoStack`

lumicks/pylake/nb_widgets/detail/undo_redo.py
JoepVanlier: Right. As we discussed on meet, I'll pull it in.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: Yeah, that sounds reasonable. For an immutable object, we could just have a very thin copy constructor.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -61,7 +63,7 @@ def __init__(
             if axis_aspect_ratio
             else None
         )
-        self.lines = KymoLineGroup([])
+        self.lines_history = UndoStack(KymoLineGroup([]))
```
rpauszek, line 66: I would make this an internal attribute to be on the safe side

rpauszek (approved):
Awesome functionality! I think this final version has a nice balance between letting the history ensure copy behavior internally, and not having extraneous one-liner functions.

Just a small comment about the history attribute in the widget, but otherwise good to go

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: Good point! I'll amend it!

lpingas (approved):
Looks great!

PR #255, opened 2022-01-13
kymo: plot with low frequency force channels

JoepVanlier:
### Why this PR?
This PR combines two small tickets. The first is to document `kymo.plot_with_force`. The second is to use low frequency force as a fallback when high frequency force is unavailable (this was a requested feature).

I've piggybacked forwarding custom arguments to `imshow` as well, since without this feature, most kymographs will simply be too dark.


changelog.md
```diff
@@ -13,6 +13,7 @@
 * Added functionality to slice `Scan` objects by frame indices. See [Confocal images](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/images.html).
 * Added convenience function which allows users to perform a force calibration procedure with a single function call `calibrate_force()`. See [force calibration](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/force_calibration.html#more-convenient-calibration).
 * Added `Kymo.crop_and_calibrate()` for interactive cropping of a kymograph. If the optional `tether_length_kbp` argument is supplied, the resulting kymograph will be automatically calibrated to this length. Check out the [documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html#calibrating-to-base-pairs) for more information.
+* Added option to use `Kymo.plot_with_force()` when only low frequency force data is available.
```
mcsmurf, line 16: Since the user cannot explicitly opt-in to the low frequency force data:
```suggestion
* Added the function `Kymo.plot_with_force()` as a fallback when only low-frequency force data is available.
```

mcsmurf (approved):
Great feature! Go for launch :rocket:

changelog.md
JoepVanlier: I added the fallback, but the original function was already there. How about:
```
Added fallback to the function `Kymo.plot_with_force()` when only low-frequency force data is available.
```

PR #256, opened 2022-01-14
docs: use `sphinx_lfs_content`

JoepVanlier:
**Why this PR?**
Because of the size of the images on `readthedocs` version 1.5 of the python package `git-lfs` (not to be confused with actual `git-lfs`) started failing because it was pulling them as one big request. The requests exceeded some limit and returned HTTP error 413.

That package does have a newer version which introduced splitting requests (see https://github.com/liberapay/git-lfs-fetch.py/pull/7). However, this version requires auth tokens now (https://github.com/liberapay/git-lfs-fetch.py/pull/5).

This is why I propose switching to https://github.com/ssciwr/sphinx_lfs_content which maintains a simple workaround for installing `git-lfs`.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -33,14 +33,6 @@ version = ".".join(pylake.__version__.split(".")[:2])
 # The full version, including alpha/beta/rc tags
 release = pylake.__version__
 
-# Make sure we have git-lfs support on rtd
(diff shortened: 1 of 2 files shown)


dean0x7d (approved):
Excellent! It's great that there's now a better way to do this.

PR #257, opened 2022-01-25
Cleanup image plotting

rpauszek:
**Why this PR?**
We're striving to have consistency between confocal and widefield image classes. This PR fixes up a number of small things having to do with plotting in `Scan` and `CorrelatedStack`. I've tried to keep the different tasks separated in different commits, but most can be squashed before merging.

Changes to docs [here](https://lumicks-pylake.readthedocs.io/en/image_plot_consistency/tutorial/images.html#plotting-and-exporting)

- `Scan.plot()` now takes `frame=0` as the first frame in the stack. Previously the first frame was referred to by `frame=1` which feels very un-pythonic.  *This is breaking, so we will wait to merge this PR until `v0.11.1` is released*
- Add a title factory to be used for all image plotting. This generates a title indicating the currently plotted frame and ensures consistency across all plots.
    - previously the title string was generated in each individual function leading to inconsistencies 
    -  first frame is `[frame 1 / N]`  -- previously some plots had this, others had `[frame 0 / N]`
    - last frame is `[frame N / N]`  -- previously some plots had this, others had `[frame N-1 / N]`
-  added the `axes` argument to `CorrelatedStack.plot()` for consistency with confocal 
- requesting a frame outside the available range now throws; previously for `Scan` this value was clipped to the range while `CorrelatedStack` threw
- added tests for `CorrelatedStack` plotting


lumicks/pylake/correlated_stack.py
```diff
@@ -151,7 +152,7 @@ def get_image(self, channel="rgb"):
 
         return np.stack([frame.data[slc] for frame in self], axis=0).squeeze()
 
-    def plot(self, frame=0, channel="rgb", show_title=True, **kwargs):
+    def plot(self, frame=0, channel="rgb", show_title=True, axes=None, **kwargs):
```
dean0x7d, line 155: Is the `axes` argument needed? Users can already override it using `plt.sca()`. At least I always felt like this was easier for third-party functions (ones that are not built into matplotlib). At least easier compared to always passing around the `axes` arguments.

For example, for builitin `pyplot` functions these two are equivalent:
```py
ax = ...

# class API
ax.plot(...)

# stateful API
plt.sca(ax)
plt.plot(...)
```

rpauszek: This was something I put in for the confocal plotting (and now here for camera images) mostly for our benefit. Indeed you can use `plt.sca()` but with these compound plots and interactive plots/widgets we've been building it makes our code a lot simpler/cleaner if you can pass around the axes instance directly. For instance if you look [here](https://github.com/lumicks/pylake/pull/244#discussion_r781923759) I find the new syntax much cleaner

That being said, that's why I made it optional, so if the user doesn't supply it, it just gets `plt.gca()` so essentially nothing has changed from the user standpoint

dean0x7d: Ah, if you find the syntax more practical, don't mind my comment. I tend to prefer to minimize the args/API area to avoid extra testing. But if it makes the API more practical, go for it.

JoepVanlier (approved):
LGTM.

PR #258, opened 2022-01-26
kymotracker: generalize Gaussian peak fitting to overlapping peaks

JoepVanlier:
**Why this PR?**
When particles get close in a kymograph, we currently have no algorithm to refine the positions of the individual fluorophores.
To enable this, we would like to add the ability to fit particle locations simultaneously in the same fit (taking into account the overlap in their point spread functions).

This PR is the first step in that, which is generalizing the backend function.

Since this PR does not introduce any front end changes, there are no relevant docs yet.

The next PR will build on this one and enable its use through `refine_lines_gaussian`.


lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -80,40 +80,45 @@ def normal_pdf_1d(x, center, sigma):
...
+        The peak parameters are: photon counts, center (in microns) and sigma (in microns).
+    fixed_background : float or None
+        Fixed background value in photons per pixel.
     pixel_size : float
         Pixel size in um.
     """
-    return total_photons * pixel_size * normal_pdf_1d(x, center, sigma) + background
+    signal = np.zeros(x.shape)
+    for total_photons, center, sigma in params[: n_components * 3].reshape(-1, 3):
+        signal += total_photons * pixel_size * normal_pdf_1d(x, center, sigma)
```
rpauszek, line 103: `normal_pdf_1d` can be vectorized pretty straightforwardly. Broadcast `x` to a row shape `[1, x.size]` and each parameter to a column shape `[n_peaks, 1]`. The result will then be an array with the pdf for each peak as rows. I only know this off hand because I've done it before for HMM code where I had to optimize whereever I possibly could.

I think if we make this change we can skip the `for` loop here and in `peak_expectation_1d_derivatives`

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -141,30 +146,38 @@ def gaussian_mle_1d(
     if fixed_background is not None and fixed_background <= 0:
         raise ValueError("Fixed background should be larger than zero.")
 
+    initial_position = x[np.argmax(photon_count)] if initial_position is None else initial_position
+    initial_position = np.atleast_1d(initial_position)
```
rpauszek, line 150: I think it'd be cleaner if we just combine these two. Still would end up as multiple lines after `black` but it always feels icky to me to assign a variable and then immediately reassign

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -141,30 +146,38 @@ def gaussian_mle_1d(
...
+
     expectation_fun = partial(
         peak_expectation_1d,
         x,
-        **({"background": fixed_background} if fixed_background is not None else {}),
+        fixed_background=fixed_background,
```
rpauszek, line 155: noice

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -141,30 +146,38 @@ def gaussian_mle_1d(
...
+        bounds.append((np.finfo(float).eps, None))
+
     result = _mle_optimize(initial_guess, expectation_fun, derivatives_fun, photon_count, bounds)
+
+    # Pack the results
+    background = result.x[-1] if fixed_background is None else fixed_background
+    result = tuple(
+        (*param, background) for param in result.x[: initial_position.size * 3].reshape(-1, 3)
+    )
     return result
```
rpauszek, line 183: I think it makes sense to just return the tuple directly, it's clear from the comment what's going on

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -342,7 +342,7 @@ def refine_lines_gaussian(
                 initial_sigma=initial_sigma,
                 fixed_background=fixed_background,
             )
-            tmp_positions.append(r.x[1] / image._pixel_size)
+            tmp_positions.append(r[0][1] / image._pixel_size)
```
rpauszek, line 345: why is only the first peak stored?

rpauszek (commented):
Nice work, it'll be great to have this feature.

I left a comment about vectorizing the PDF calculation, I think it'd be good to do even if it makes the code a tad bit messier (because you need to broadcast arrays to specific shapes) but considering this is already a pretty slow algo I think it'd be worth it

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: I just pushed a commit that vectorizes the code, but it's slower when I measure it.
```
Vectorized:     3.749 +- 0.0418
Non-vectorized: 3.403 +- 0.0570
```
I want to try reordering the parameters, to see if that has any sort of meaningful impact (right now I need a swapaxes and a reshape which I could possibly get around if I reorder the parameter values).

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Sure, no problem.

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Ah yes. Done.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Because this is feature parity and only the one peak case (see PR description).

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Trying to do it over a different axes was worse.
```
Vectorized: 4.093789029121399 +- 0.1270268842186797
```

JoepVanlier: And the last option, going over the same axis as the first but avoiding the `swapaxes` and `reshape`:
```
Vectorized: 3.523 +- 0.0817
```

JoepVanlier: The last option is a bit faster when dealing with multiple peaks (previous tests were for single peak):
```python
Vectorized: 1.2794423580169678 +- 0.05959272465348172
Non-vectorized: 1.5789220094680787 +- 0.13292848948017963
```
Long story short, I'd go with vectorized option. And specifically with option 3, with the parameters grouped per type (avoiding the swap and reshape).

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -146,38 +159,50 @@ def gaussian_mle_1d(
...
-    for init_pos in initial_position:
-        initial_guess.extend([amp_estimate / initial_position.size, init_pos, initial_sigma])
-        bounds.extend([(0.01, None), (np.min(x), np.max(x)), (pixel_size, 10 * pixel_size)])
+
+    num_pars = num_peaks * 3 + (0 if fixed_background else 1)
+    initial_guess = np.empty(num_pars)
+    initial_guess[:num_peaks] = amp_estimate / initial_position.size
+    initial_guess[num_peaks : 2 * num_peaks] = initial_position
+    initial_guess[2 * num_peaks :] = initial_sigma
+
```
rpauszek, line 195: maybe we could add a small comment above all this to just explicitly state the order of parameters `# amplitudes, centers, widths, background offset`. I think, for these long mixed parameter lists, it just helps to quickly be able to keep track of things without needing to decipher the code everytime you look at it

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -179,6 +179,12 @@ def gaussian_mle_1d(
         num_peaks=num_peaks,
     )
 
+    # We want to preserve the original order, so we sort, so that we can also sort afterwards
+    # and then apply the inverse sort.
+    idx = np.argsort(initial_position)
+    initial_position = initial_position[idx]
+    reverse_idx = np.argsort(idx)
```
rpauszek, line 186: I find this comment confusing. Maybe break it up into 2 distinct thoughts/lines:
```
# Sort the positions for consistency during optimization
# Keep a set of reverse sort indices so the original order can be preserved on output
```

rpauszek (approved):
Nice work, thanks for checking on the best way to vectorize. That exercise was a lot more work than I had anticipated, but I think the final payoff is much more readable code.

Just a couple notes on docstring/comments for extra clarity, but looks great!

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -90,35 +90,48 @@ def peak_expectation_1d(x, params, fixed_background, pixel_size, n_components):
         Position data at which the function is to be evaluated.
     params : array-like
         Model parameters.
-        Model parameters should be provided as an array of parameters for each peak in the fit
-        followed by an optional offset parameter (in photons per pixel).
-        The peak parameters are: photon counts, center (in microns) and sigma (in microns).
+        Model parameters should be provided as an array of parameters. The parameters are:
+        photon counts for all peaks, center (in microns) for all peaks and sigma (in microns) for
+        all peaks followed by an optional offset parameter (in photons per pixel).
```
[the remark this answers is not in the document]
JoepVanlier: Good catch. I fully agree. This was an oversight on my part!

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Yeah, that's fair.

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Good point. Done!

PR #259, opened 2022-01-31
gaussian_mle: make refine gaussian take into account multiple peaks in close proximity

JoepVanlier:
**Why this PR?**
Because we wish to correctly deal with overlapping lines in `refine_lines_gaussian()`. If two lines get close, the refinement will now do a simultaneous fit for these lines.

Notes:
Part 1 of this PR can be found in https://github.com/lumicks/pylake/pull/258
This PR also fixes a bug in `refine_lines_gaussian()` which incorrectly rounded the pixel position (`rint`) instead of flooring it. For pixels with a subpixel position larger than half the pixel, this resulted in shifting the window by one pixel in the positive direction. Typically, this would have little or no effect since the majority of the peak should still be covered.

image
Blue is when taking a big window and using `"ignore"` (blue) instead of `"fit_multiple"` (red)

*from Ray: I pushed 3 commits at the end to try to clean up the logic in `refine_lines_gaussian()` for detecting overlap. If approved these should all be squashed, but I wanted to keep a history of what I actually did because it's not straightforwardly obvious going between the two versions. The original code was good, but I tried to re-arrange/rename some things to hopefully make it clearer.*

### DIFF
--- a/lumicks/pylake/kymotracker/detail/gaussian_mle.py
+++ b/lumicks/pylake/kymotracker/detail/gaussian_mle.py
@@ -3,6 +3,33 @@ from functools import partial
 from scipy.optimize import minimize
 
 
+def overlapping_pixels(coordinates, width):
+    """Determine which fitting regions are expected to overlap.
+
+    Returns a list of indices grouped by overlapping coordinates within a window 2 * width.
+
+    This function constructs a list of lists. Each inner list contains the indices of the
+    coordinates which are separated by < 2 * width. The outer list is sorted by spatial position.
+
+    Parameters
+    ----------
+    coordinates : array_like
+        coordinates
+    width : float
+        width in either direction
+    """
+    if not len(coordinates):
+        return []
+
+    coordinates = np.asarray(coordinates)
+    indices = np.argsort(coordinates)
+    sorted_coordinates = coordinates[indices]
+    differences = np.diff(sorted_coordinates)
+    groups = np.split(indices, np.flatnonzero(differences > 2 * width) + 1)
+
+    return groups
+
+
 def poisson_log_likelihood(params, expectation_fun, photon_count):
     """Calculates the log likelihood of Poisson distributed values.
 
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -293,7 +293,8 @@ def refine_lines_gaussian(
     overlap_strategy : {'ignore', 'skip'}
         How to deal with frames in which the fitting window of two `KymoLine`'s overlap.
 
-        - 'ignore' : do nothing, fit the frame as-is.
+        - 'multiple' : fit the peaks simultaneously.
+        - 'ignore' : do nothing, fit the frame as-is (ignoring overlaps).
         - 'skip' : skip optimization of the frame; remove from returned `KymoLine`.
     initial_sigma : float
         Initial guess for the `sigma` parameter.
@@ -301,7 +302,7 @@ def refine_lines_gaussian(
         Fixed background parameter in photons per second.
         When supplied, the background is not estimated but fixed at this value.
     """
-    assert overlap_strategy in ("ignore", "skip")
+    assert overlap_strategy in ("ignore", "skip", "multiple")
     if refine_missing_frames:
         lines = [line.interpolate() for line in lines]
 
@@ -326,7 +327,11 @@ def refine_lines_gaussian(
             int(lines[line_index].coordinate_idx[idx]) for (line_index, idx) in frame
         ]
 
-        groups = overlapping_pixels(pixel_coordinates, window)
+        groups = (
+            [[idx] for idx in range(len(pixel_coordinates))]
+            if overlap_strategy == "ignore"
+            else overlapping_pixels(pixel_coordinates, window)
+        )
         for group in groups:
             if len(group) > 1 and overlap_strategy == "skip":
                 overlap_count += 1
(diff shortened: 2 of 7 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -295,15 +295,16 @@ def refine_lines_gaussian(
     overlap_strategy : {'ignore', 'skip'}
         How to deal with frames in which the fitting window of two `KymoLine`'s overlap.
 
-        - 'ignore' : do nothing, fit the frame as-is.
+        - 'fit_multiple' : fit the peaks simultaneously.
```
rpauszek, line 298: I'm not sure about `"fit_multiple"`, for arguments that users are required to supply I prefer to keep to short names and if possible without special characters (`_`) .... maybe it's just a weird pet peeve of mine. 

At the same time I can't really come up with something better. First thing that comes to mind is `"overlap"` but I don't know if that's descriptive enough. Maybe just `"multiple"`?

lumicks/pylake/kymotracker/tests/test_refinement.py
```diff
@@ -158,6 +158,24 @@ def test_gaussian_refinement_fixed_background(kymogroups_2lines, fit_mode):
     )
 
 
+def test_gaussian_proximity(kymogroups_close_lines):
```
rpauszek, line 161: maybe `test_gaussian_refinement_overlap()`? for consistency with other test names?

rpauszek (commented):
Looks good. As per my comment in the main PR description, I added a couple commits to try to de-duplicate or pre-compute some things before the loop that determines if there is an overlap. There's really no good/clean way to do this, but I think we can settle on some version that's relatively clear

lumicks/pylake/kymotracker/detail/overlap_sweep.py
```diff
@@ -0,0 +1,34 @@
+import numpy as np
```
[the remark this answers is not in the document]
JoepVanlier: Merged it into `gaussian_mle`.

lumicks/pylake/kymotracker/detail/overlap_sweep.py
```diff
@@ -0,0 +1,34 @@
...
+    sorted_coordinates = coordinates[indices]
+    groups = []
+    group = [indices[0]]
+    differences = np.diff(sorted_coordinates)
+    for idx, diff in zip(indices[1:], differences):
+        if diff > 2 * width:
+            groups.append(group)
+            group = [idx]
+        else:
+            group.append(idx)
```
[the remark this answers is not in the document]
JoepVanlier: Added.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -10,6 +11,7 @@
     KymoPeaks,
 )
 from .kymoline import KymoLineGroup
+from copy import deepcopy
```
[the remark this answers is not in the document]
JoepVanlier: Dropped it.

lumicks/pylake/kymotracker/tests/test_sweep.py
```diff
@@ -0,0 +1,43 @@
...
+    if len(result) != len(output):
+        raise RuntimeError(
+            f"Input: {positions} / {width}, Expected: {output}, Got {overlapping_pixels(positions, width)}"
+        )
+
+    for p, q in zip(result, output):
+        if p != q:
+            raise RuntimeError(
+                f"Input: {positions} / {width}, Expected: {output}, Got {overlapping_pixels(positions, width)}"
+            )
```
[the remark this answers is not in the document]
JoepVanlier: No there isn't. I'll switch it over to an assert.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: `multiple` it is!

lumicks/pylake/kymotracker/tests/test_refinement.py
JoepVanlier: 👍

dean0x7d (approved):
LGTM

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -3,6 +3,33 @@
...
+        width in either direction
+    """
+    if not len(coordinates):
+        return []
+
+    coordinates = np.asarray(coordinates)
+    indices = np.argsort(coordinates)
+    sorted_coordinates = coordinates[indices]
+    differences = np.diff(sorted_coordinates)
+    groups = np.split(indices, np.flatnonzero(differences > 2 * width) + 1)
```
dean0x7d, line 28: Damn nice!

lumicks/pylake/kymotracker/detail/gaussian_mle.py
```diff
@@ -3,6 +3,41 @@
...
+    groups = []
+    group = [indices[0]]
+    differences = np.diff(sorted_coordinates)
+    for idx, diff in zip(indices[1:], differences):
+        if diff > 2 * width:
+            groups.append(group)  # no overlap detected; commit current group
+            group = [idx]  # re-initialize with current index
+        else:
+            group.append(idx)
+    groups.append(group)  # overlap detected; add current index to current group
```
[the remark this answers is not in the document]
JoepVanlier: Thanks for pointing that out. I wasn't aware of that function and it's a good one to know about.

While considering using that I figured the function could be vectorized more easily actually (but you already saw that :D). 

This also got rid of the weird edge case with the initial point assignment.

lumicks/pylake/kymotracker/detail/gaussian_mle.py
JoepVanlier: Thank you :)

PR #260, opened 2022-02-02
Add background subtraction to kymotracking example

aafkevandenberg:
I added an example on how to select the background and how to subtract it from the line intensities after tracking.

### DIFF
--- a/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
+++ b/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
@@ -67,6 +67,13 @@ of the colormap. Any photon count higher than that will be clipped to the maxima
 What we can observe in this data is that as more force is applied, we get an increased binding activity. Let’s see
 if we can put the kymotracker to some good use and quantify these.
 
+Computing the background
+------------------
+First, we select a small region without traces to determine the background signal::
+
+    background = kymo["100s":"200s"].crop_by_distance(28, 31)
+    green_background_per_pixel = np.mean(background.green_image)
+
 Downsampling the kymograph
 --------------------------
 
@@ -150,11 +157,13 @@ the longest line we found, and have a look at its position over time::
 
 .. image:: kymo_position_over_time.png
 
-We can use such a line to sample the photon counts in the image. If we want to sum the photon count in a pixel region
-around the line from -3 to 3, we can achieve this by::
+The line coordinates can be used to sample the photon counts in the image. The example below demonstrates how to obtain the sum of the photon counts in a pixel region around the line from -3 to 3 (a line with a width of 7 pixels). The background per pixel as computed earlier is subtracted from the photon counts. Since the kymograph was downsampled by a factor 2 after computing the background, the background per pixel is multiplied by 2::
 
+    window = 3
+    bg_corrected = longest_line.sample_from_image(window) - (2 * window + 1) * 2 * green_background_per_pixel
+    
     plt.figure()
-    plt.plot(longest_line.seconds, longest_line.sample_from_image(3))
+    plt.plot(longest_line.seconds, bg_corrected)
     plt.ylabel('Photon count')
     plt.xlabel('Time [s]')
     plt.title('Photon counts along the longest line')


docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -67,6 +67,13 @@ of the colormap. Any photon count higher than that will be clipped to the maxima
 What we can observe in this data is that as more force is applied, we get an increased binding activity. Let’s see
 if we can put the kymotracker to some good use and quantify these.
 
+Compute background
+------------------
+First, we select a small region without traces to determine the background signal::
+
+    background=kymo['100s':'200s'].crop_by_distance(28,31)
+    green_background_ppixel=np.mean(background.green_image)
```
mikhas, line 75: Is is really `ppixel`, or just `pixel`? Also, the document apparently uses active voice for section titles. So `Compute the background` perhaps?

aafkevandenberg: I really meant to say 'per pixel'. And yes, it makes sense to use the same voice as in other section titles.

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -150,11 +157,11 @@ the longest line we found, and have a look at its position over time::
...
 .. image:: kymo_position_over_time.png
 
-We can use such a line to sample the photon counts in the image. If we want to sum the photon count in a pixel region
-around the line from -3 to 3, we can achieve this by::
+We can use the line coordinates to sample the photon counts in the image. If we want to sum the photon count in a pixel region
+around the line from -3 to 3 (a line with a width of 7 pixels) and subtract the background, we can achieve this by::
 
     plt.figure()
-    plt.plot(longest_line.seconds, longest_line.sample_from_image(3))
+    plt.plot(longest_line.seconds, longest_line.sample_from_image(3)-7*green_background_ppixel)
```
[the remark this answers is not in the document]
JoepVanlier: One could consider making the effect of the window width explicit by doing something like
```python
window_size = 3
plt.plot(
    longest_line.seconds,
    longest_line.sample_from_image(window_size) - (2 * window_size + 1) * green_background_ppixel,
)
```
But then it might make sense to first assign the sampled array to not have the actual plot statement be so long.
```python
window = 3
bg_corrected = longest_line.sample_from_image(window) - (2 * window + 1) * green_background_pixel
plt.plot(longest_line.seconds, bg_corrected)
```

JoepVanlier (changes requested):
Looks good! I only have two small suggestions regarding formatting.

The only thing I personally don't like so much is how the `7` is hardcoded. One could make the origin a little more explicit by writing `2 * window + 1`.

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
JoepVanlier: Maybe `per_pixel` then?

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -150,11 +157,11 @@ the longest line we found, and have a look at its position over time::
 
 .. image:: kymo_position_over_time.png
 
-We can use such a line to sample the photon counts in the image. If we want to sum the photon count in a pixel region
-around the line from -3 to 3, we can achieve this by::
+We can use the line coordinates to sample the photon counts in the image. If we want to sum the photon count in a pixel region
+around the line from -3 to 3 (a line with a width of 7 pixels) and subtract the background, we can achieve this by::
```
mcsmurf, line 161: Not sure about the line break, but I'd word it something like this:
```suggestion
The line coordinates can be used to sample the photon counts in the image. The example below demonstrates how to obtain the sum of the photon count in a pixel region and subtract the background around the line from -3 to 3 (a line with a width of 7 pixels). This is achieved by::
```

JoepVanlier (approved):
LGTM! 👍

JoepVanlier (approved):
Good catch about the factor two. I had completely looked over that earlier 😮 

Looks good now though. I tested it, just to be sure.

:shipit:

PR #261, opened 2022-02-04
ci: fix tifffile.imsave deprecation warning

JoepVanlier:
**Why this PR?**
Fixes a `DeprecationWarning` regarding `TiffFile`.

`imwrite` was introduced in 2018.11.6 (https://github.com/cgohlke/tifffile/blob/master/CHANGES.rst).
Unfortunately `TiffWriter.save` was introduced in `2020.9.30` and will require a version bump.


lpingas (approved):
LGTM! Before merging, consider adding an entry on release notes since you are bumping a dependency.

PR #263, opened 2022-02-09
Touchdown Method pt 1

JoepVanlier:
## Why this PR?
We need a way to find the surface position and focal shift. This PR implements that.

## What to look out for?
The API expects data to be in numpy array, and to be somewhat downsampled before entry (approx 150 Hz is fine). I was doubting whether to make this API call do that for the user or not. I opted for not, but it'd be easy to change. I also doubted whether to take `Slice` or numpy arrays.

There are no docs since it is private API for now.

I recommend reviewing commit by commit. Some commits have descriptions as to the what and why.

image
image

### DIFF
--- a/lumicks/pylake/force_calibration/touchdown.py
+++ b/lumicks/pylake/force_calibration/touchdown.py
@@ -192,6 +192,10 @@ def touchdown(
         Degree of the polynomial to use for the background intensity when fitting the intensity
         pattern.
     """
+    if len(nanostage) != len(axial_force):
+        min_length = min(len(nanostage), len(axial_force))
+        nanostage, axial_force = nanostage[:min_length], axial_force[:min_length]
+
     # Find the surface position
     piecewise_parameters = fit_piecewise_linear(nanostage, axial_force)
     surface_position = piecewise_parameters[0]
--- a/lumicks/pylake/force_calibration/touchdown.py
+++ b/lumicks/pylake/force_calibration/touchdown.py
@@ -1,0 +1,33 @@
+import numpy as np
+from scipy.optimize import curve_fit
+
+
+def piecewise_linear(x, x0, y0, k1, k2):
+    return np.piecewise(x, [x < x0], [lambda x: k1 * (x - x0) + y0, lambda x: k2 * (x - x0) + y0])
+
+
+def fit_piecewise_linear(x, y):
+    """Fits a two-segment piecewise linear function and returns the parameters.
+
+    Parameters:
+        x: array-like
+            independent variable
+        y: array-like
+            dependent variable
+    """
+    # We should be able to handle cases where x moves in positive or negative direction
+    difference = x[-1] - x[0]
+    center = x[0] + difference / 2
+    mid_point = np.nonzero(x > center)[0][0] if difference > 0 else np.nonzero(x < center)[0][0]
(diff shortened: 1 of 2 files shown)


lumicks/pylake/force_calibration/touchdown.py
```diff
@@ -0,0 +1,33 @@
...
+    center = x[0] + difference / 2
+    mid_point = np.nonzero(x > center)[0][0] if difference > 0 else np.nonzero(x < center)[0][0]
+    x_start, y_start, x_mid, y_mid, x_end, y_end = (
+        (x[0], y[0], x[mid_point], y[mid_point], x[-1], y[-1])
+        if difference > 0
+        else (x[-1], y[-1], x[mid_point], y[mid_point], x[0], y[0])
+    )
+
+    k1_est = (y_mid - y_start) / (x_mid - x_start)
+    k2_est = (y_end - y_mid) / (x_end - x_mid)
```
rpauszek, line 93: maybe here use `slope1_est` instead of `k1_est` to be a bit more descriptive

lumicks/pylake/force_calibration/touchdown.py
```diff
@@ -31,3 +31,36 @@ def fit_piecewise_linear(x, y):
...
+        amp * sin(a * x + phase) + other
+
+    can we written as:
+
+        amp * sin(phase) * cos(a * x) + amp * cos(phase) * sin(a * x) + other
+
+    By absorbing the sin and cosine into the amplitude, we change variables from amplitude and
+    phase, to sine amplitude and cosine amplitude. This means the problem of fitting a sine plus
+    polynomial is essentially a linear one in all variables except the frequency (our parameter of
+    interest). Doing this allows us to estimate this using only one estimated variable."""
```
rpauszek, line 115: I think adding the parameters to the docstring would be useful here

lumicks/pylake/force_calibration/touchdown.py
```diff
@@ -2,6 +2,68 @@
...
+    nanostage_z_position,
+    surface_position,
+    displacement_sensitivity,
+    intensity_amplitude,
+    intensity_phase_shift,
+    intensity_decay_length,
+    scattering_polynomial_coeffs,
+    focal_shift,
+    nonlinear_shift,
+):
```
rpauszek, line 19: should we maybe move this to `tests.data.somewhere`? Since it seems like it's only used in the tests?

rpauszek (approved):
Looks good to me, really cool trig stuff to decrease the nonlinear parameters! Just a couple minor comments

lumicks/pylake/force_calibration/touchdown.py
JoepVanlier: 👍

lumicks/pylake/force_calibration/touchdown.py
JoepVanlier: Good point. Not sure why I forgot them. Added!

lumicks/pylake/force_calibration/touchdown.py
rpauszek: yeah, you're right it'll likely end up back here anyway. keeping it here is fine

PR #264, opened 2022-02-11
touchdown: improve robustness sine fit

JoepVanlier:
### Why this PR?
One of the challenges with the previous approach is that for the optimization to be successful, we need to be close to the actual optimum. The reason for this is that the log likelihood of this problem is very flat (the amplitude collapses when the frequency becomes more wrong) once you go away from the correct frequency (see figure below).

image

I noticed a lack of robustness when providing estimates that were even moderately far from the optimum. This means that the strategy employed previously (minimizing for each value of a range) is likely quite suboptimal, as many of these optimizations will either just fail or go to the same minimum.

I propose a small change, opting for a quick scan followed by a bounded optimization. This yields more robust results even when providing a rather rough range.

### DIFF
--- a/lumicks/pylake/force_calibration/tests/test_touchdown.py
+++ b/lumicks/pylake/force_calibration/tests/test_touchdown.py
@@ -35,23 +35,23 @@ def test_piecewise_linear_fit(direction, surface, slope1, slope2, offset):
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tiagobonetti (approved):
This looks really good.
I managed to read the comments and understand most of what the new code does.

`honk`

PR #265, opened 2022-02-11
Kymotracker: allow manual connection between any nodes in tracks

rpauszek:
**Why this PR?**
Request from multiple users. Sometime you want to connect two nearby tracks that are slightly overlapped in time (by a couple frames). Previously this was impossible as we only supported connecting between the track ends. Now you can connect between any two nodes; the ends of the tracks are trimmed to these points and then concatenated

tracker_connect_demo
.

### DIFF
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -349,6 +349,26 @@ class KymoLineGroup:
         self._src[self._src.index(starting_line)] = starting_line + ending_line
         self._src.remove(ending_line)
 
+    def _merge_lines(self, starting_line, starting_node, ending_line, ending_node):
+        starting_node = int(starting_node) + 1
+        ending_node = int(ending_node)
+
+        first_half = KymoLine(
+            starting_line.time_idx[:starting_node],
+            starting_line.coordinate_idx[:starting_node],
+            starting_line._image,
+        )
+
+        last_half = KymoLine(
+            ending_line.time_idx[ending_node:],
+            ending_line.coordinate_idx[ending_node:],
+            ending_line._image,
+        )
+
+        self._src[self._src.index(starting_line)] = first_half + last_half
+        if starting_line != ending_line:
+            self._src.remove(ending_line)
+
     def __len__(self):
         return len(self._src)
 
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -153,52 +153,76 @@ class KymoWidget:
 
     def _connect_line_callback(self):
         cutoff_radius = 0.05  # We use a connection cutoff of 5% of the axis ranges
-        line_to_extend = None
+        clicked_line_info = None
         plotted_line = None
+        nodes = None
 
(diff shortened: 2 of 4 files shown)


lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -349,6 +349,25 @@ def _concatenate_lines(self, starting_line, ending_line):
...
+        first_half = KymoLine(
+            starting_line.time_idx[:starting_node],
+            starting_line.coordinate_idx[:starting_node],
+            starting_line._image,
+        )
+
+        last_half = KymoLine(
+            ending_line.time_idx[ending_node:],
+            ending_line.coordinate_idx[ending_node:],
+            ending_line._image,
```
JoepVanlier, line 365: This is more a workflow question.

Shouldn't there be two, three or four lines (depending on whether you clicked endpoints or not) here since now we are merging in the middle? The mental model that I have of this operation is that basically, what you do is you cut both lines at the point where you click and release, and then connect the new ends tip to tip. Otherwise you risk losing otherwise good traces (see below).

oops

This was never an issue when merging only from tip to tip.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -2,6 +2,7 @@
 import numpy as np
 import matplotlib.pyplot as plt
 from copy import copy
+import itertools
```
JoepVanlier, line 5: Are you using this?

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -153,52 +154,76 @@ def _get_scale(self):
...
         def initiate_line(event):
-            nonlocal line_to_extend
+            nonlocal nodes, clicked_line_info
             if len(self.lines) == 0:
                 return
 
-            tips = np.array([np.array([l.seconds[-1], l.position[-1]]) for l in self.lines])
-            distance, idx = get_nearest(tips, np.array([event.x, event.y]), self._get_scale())
+            nodes = get_node_info()
+            distance, idx = get_nearest(event.x, event.y)
```
JoepVanlier, line 188: nicer :)

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Second related question is, what if the start and end line are the same? I could imagine what I attempt in the gif below to be useful:
self_connect

JoepVanlier (commented):
Codewise it looks good to me.

I have a question on whether this is the expected operation however. I know we specified "remove points that are between the lines", but I'm second guessing that a little bit now that I see it in action.

I'm wondering whether split line 1 (if not end), split line 2 (if not start), connect new tips might be more sensible behavior. Then again, I'm not sure, so I would be okay with merging as is as well (assuming the self case is repaired).

Let me know what you think.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
rpauszek: nope, not anymore

JoepVanlier (approved):
Looks good to me!

PR #267, opened 2022-02-21
Add backend for making kymolines binding profile histogram

rpauszek:
**Why this PR?**
One of the requested output figures for kymotracked lines is a waterfall plot of binding histograms along the spatial axis for `n` time bins along the kymo. This PR adds an internal method on `KymoLineGroup` to calculate the data needed to make the plot (right side of the figure below).

Kernel density estimate is used to calculate a smoothed histogram, similar to what has been published in literature

image

### DIFF
--- a/lumicks/pylake/kymotracker/kymoline.py
+++ b/lumicks/pylake/kymotracker/kymoline.py
@@ -1,4 +1,5 @@
 from copy import copy
+from sklearn.neighbors import KernelDensity
 from lumicks.pylake.kymotracker.detail.msd_estimation import *
 from lumicks.pylake.kymotracker.detail.calibrated_images import CalibratedKymographChannel
 from lumicks.pylake.kymotracker.detail.binding_times import _kinetic_mle_optimize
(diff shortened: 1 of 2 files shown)


lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -488,3 +489,37 @@ def plot_binding_histogram(self, kind, bins=10, **kwargs):
...
+            Requested number of time bins.
+        bandwidth : float
+            KDE bandwidth; units are in the physical spatial units of the kymograph.
+        n_position_points : int
+            Length of the returned density array(s).
+        """
+
+        n_rows, n_frames = self[0]._image.data.shape
+        position_max = n_rows * self[0]._image._pixel_size
+        bin_size = n_frames // n_time_bins
```
JoepVanlier, line 511: This fails when `n_time_bins > n_frames`, produces a division by zero when computing the bin edges in the next step.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -488,3 +489,37 @@ def plot_binding_histogram(self, kind, bins=10, **kwargs):
...
+
+        frames = np.hstack([line.time_idx for line in self])
+        positions = np.hstack([line.position for line in self])
+        bin_labels = np.digitize(frames, bin_edges, right=False)
+
+        x = np.linspace(0, position_max, n_position_points)[:, np.newaxis]
+        densities = []
+        for bin_index in np.arange(n_time_bins) + 1:
+            binned_positions = positions[bin_labels == bin_index][:, np.newaxis]
+            kde = KernelDensity(kernel="gaussian", bandwidth=bandwidth).fit(binned_positions)
```
JoepVanlier, line 522: This fails when a time bin doesn't contain any lines. Please also add a test for this case.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -488,3 +489,37 @@ def plot_binding_histogram(self, kind, bins=10, **kwargs):
...
+            densities.append(np.exp(kde.score_samples(x)))
```
JoepVanlier, line 523: This fails cryptically when `n_position_points` is zero. Should probably assert that it's bigger than zero (or one?).

JoepVanlier (changes requested):
There are still a few corner cases that would be nice to squish, but what's there looks very nice :)

lumicks/pylake/kymotracker/kymoline.py
rpauszek: good catch, re-throw with a better error message and added a test

lumicks/pylake/kymotracker/kymoline.py
rpauszek: another good catch, default to zero density (since this is a likely scenario) and added test

lumicks/pylake/kymotracker/kymoline.py
rpauszek: third good catch (I feel like there should be some kind of baseball comment here), handled in the same `try` block as above and rethrow with a better error message; test added.

We'll want to keep the single bin case where you just histogram the whole kymo (requested feature)

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -488,3 +489,51 @@ def plot_binding_histogram(self, kind, bins=10, **kwargs):
...
+        positions = np.hstack([line.position for line in self])
+        bin_labels = np.digitize(frames, bin_edges, right=False)
+
+        x = np.linspace(0, position_max, n_position_points)[:, np.newaxis]
+        densities = []
+        for bin_index in np.arange(n_time_bins) + 1:
+            binned_positions = positions[bin_labels == bin_index][:, np.newaxis]
+            try:
+                kde = KernelDensity(kernel="gaussian", bandwidth=bandwidth).fit(binned_positions)
+                densities.append(np.exp(kde.score_samples(x)))
```
JoepVanlier, line 532: `n_position_points` being 0 still gives a cryptic error.
image

JoepVanlier (approved):
One last thing to note, but otherwise looks fine to me.

Issue #266, opened 2022-02-19
ImportError on Python 3.10


b-dawes:
The pylake package currently depends on tabulate 0.8.6 (at least on conda). However, this version of tabulate is [incompatible with python 3.10](https://github.com/astanin/python-tabulate/issues/104) (the reason is rather silly, they were comparing the python version number as a string so "10"<"3").  Because of this, importing either lumicks.pylake or tabulate on python 3.10 throws an ImportError. 

[The fix](https://github.com/astanin/python-tabulate/commit/24625986d76b50abea7cdce45f62861e72cff2b6) was implemented in tabulate 0.8.8 (current release is 0.8.9).

JoepVanlier:
Thanks for bringing this to our attention (and the detailed report of the issue)! We'll bump the requirement in the next release.

rpauszek:
`pylake v0.11.1` has now been released and adds support for `python 3.10`

PR #268, opened 2022-02-21
Spring cleaning and Python 3.10

JoepVanlier:
**Why this PR?**
This PR adds Python `3.10` support to CI. It also bumps the `tabulate` dependency which prevented use of Python 3.10 (see https://github.com/lumicks/pylake/issues/266).

Since we were generating a few warnings when running the tests, I decided to delve into these a little bit and clean them up. Therefore this PR also consists of a few small fixups. I have described the reason for each minor fix in the relevant commit message.

Note that we still generate some warnings from matplotlib renaming an argument in the constructor of the `RectangleSelector`. We will fix that on the next breaking release by bumping the `mpl` dependency.

The fix in `crop_and_rotate()` hasn't actually been in a release yet, so it doesn't warrant a changelog entry. The other fixes merely affect tests.

### DIFF
--- a/.github/workflows/pylake_test.yml
+++ b/.github/workflows/pylake_test.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: ["3.7", "3.10"]
 
     steps:
     - uses: actions/checkout@v2
--- a/lumicks/pylake/force_calibration/tests/test_convenience.py
+++ b/lumicks/pylake/force_calibration/tests/test_convenience.py
@@ -50,11 +50,17 @@ def test_active_force_calibration_active(reference_models):
     """This test tests merely whether values were handed to the low level API correctly, not whether
     the calibration results are correct."""
     data, f_sample = reference_models.lorentzian_td(4000, 1, 0.4, 14000, 78125)
+    oscillation = np.sin(77 * 2 * np.pi * np.arange(data.size) / 78125)
+
+    # We need a driving peak in the power spectrum, otherwise we obtain a negative value for the
+    # peak power which leads to a downstream warning.
+    data += oscillation
+
     fit = calibrate_force(
         data,
         2,
         9,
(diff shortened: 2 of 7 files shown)


.github/workflows/pylake_test.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python-version: [3.7, 3.8, 3.9]
+        python-version: [3.7, "3.10"]
```
dean0x7d, line 11: For conceptual symmetry :)
```suggestion
        python-version: ["3.7", "3.10"]
```

setup.py
```diff
@@ -56,7 +56,7 @@ def read(filename):
         "scipy>=1.1, <2",
         "matplotlib>=2.2",
         "tifffile>=2020.9.30",
-        "tabulate==0.8.6",
+        "tabulate>=0.8.8, <0.8.10",
```
dean0x7d, line 59: Why specifically `<0.8.10`? Usually, the top is set at a semver breaking change boundary, e.g. `<0.9`.

lumicks/pylake/force_calibration/tests/test_convenience.py
```diff
@@ -50,11 +50,13 @@ def test_active_force_calibration_active(reference_models):
     """This test tests merely whether values were handed to the low level API correctly, not whether
     the calibration results are correct."""
     data, f_sample = reference_models.lorentzian_td(4000, 1, 0.4, 14000, 78125)
+    oscillation = np.sin(77 * 2 * np.pi * np.arange(data.size) / 78125)
+    data += oscillation
```
dean0x7d, line 54: I'd suggest copy/pasting the commit message here as a comment since this is not immediately obvious.

JoepVanlier: Good point. Will do.

setup.py
JoepVanlier: Should indeed just have gone with `<0.9`.

dean0x7d (approved):
Looks great 👌

PR #269, opened 2022-02-21
Catch some touchdown failure modes

JoepVanlier:
### Why this PR?
This PR is intended to catch a few of the failure cases that we observe when applying the touchdown procedure.

It adds three things (images of the failure modes are shown below).
- It sets the minimum for the focal shift to `0.7` (valid focal shifts for our objectives are between `0.7` and `1.0`.
- It ensures that there is sufficient data to compute the focal shift. At least two periods of the interference pattern are required.
image
The reason is that despite the fit shown here looking very good, the computed focal shift is actually complete nonsense. The focal shift model is given as a background polynomial plus a sine wave contribution. When you have only a partial period of the sine wave, the polynomial simply captures all the data (resulting in a sine wave amplitude of zero and an arbitrary frequency and phase).
- It validates that the piecewise linear fit is significantly better than a single linear fit. This guards against the case where we do not actually see the change in slope expected from the touchdown.
image
image

### DIFF
--- a/lumicks/pylake/force_calibration/touchdown.py
+++ b/lumicks/pylake/force_calibration/touchdown.py
@@ -2,6 +2,7 @@ import numpy as np
 from scipy.optimize import curve_fit, minimize_scalar
 from dataclasses import dataclass
 import matplotlib.pyplot as plt
+import warnings
 
 
 def mack_model(
@@ -178,7 +179,12 @@ class TouchdownResult:
 
     def plot(self, legend=True):
         plt.plot(self.nanostage_position, self.axial_force, label="Axial force")
-        plt.plot(self.interference_nanostage, self.interference_force, label="Interference fit")
+        plt.plot(
+            self.interference_nanostage,
+            self.interference_force,
+            label=f"Interference fit{'' if self.focal_shift else ' (failed)'}",
+            linestyle="--" if self.focal_shift is None else "-",
+        )
         plt.plot(self.nanostage_position, self.surface_fit, label="Piecewise linear fit")
         plt.axvline(self.surface_position, label="Determined surface position")
         if legend:
@@ -236,6 +242,17 @@ def touchdown(
     )
 
     focal_shift = par * expected_wavelength
+    used_range = np.max(stage_trimmed) - np.min(stage_trimmed)
+
+    if used_range < (2 * expected_wavelength / focal_shift):
+        warnings.warn(
+            RuntimeWarning(
+                "Insufficient data available to reliably fit touchdown curve. We need at least two "
+                "oscillations to reliably fit the interference pattern."
+            )
+        )
+        focal_shift = None
+
     return TouchdownResult(
         surface_position=surface_position,
         focal_shift=focal_shift,
--- a/lumicks/pylake/force_calibration/touchdown.py
+++ b/lumicks/pylake/force_calibration/touchdown.py
@@ -1,5 +1,6 @@
 import numpy as np
 from scipy.optimize import curve_fit, minimize_scalar
+from scipy import stats
 from dataclasses import dataclass
 import matplotlib.pyplot as plt
 import warnings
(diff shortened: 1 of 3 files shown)


michaelbugiel:
An easy to implement additional test is to check if the found surface position is between the minimum and maximum position.
Also I think that more, in such special cases, the fitting will fail rather than fitting a nonsense fit. Can we also check for this? Or is this rather unlikely or are they better ways?
Different question: Where are the 2 last figures from? Look very "special" to m...

JoepVanlier:
> An easy to implement additional test is to check if the found surface position is between the minimum and maximum position.

True. But I think this is already caught by the F-test that's currently in place. Reason being that if the surface position it outside hte range of the nanostage, we would not see an improvement from the additional fitting parameter and hence that check would fail.

> Also I think that more, in such special cases, the fitting will fail rather than fitting a nonsense fit. Can we also check for this? Or is this rather unlikely or are they better ways?

We'd need to define fail a bit more precisely. Reaching some maximum number of iterations is too arbitrary a condition and not something that happens in the current implementation. The way the algorithm currently works is that it solves the linear subproblem (polynomial, amplitude and phase) in a single step and only optimizes over the one parameter which is actually non-linear (frequency). This means it doesn't fail. Since this is really fast, we can spend the time this frees up to simply scan the focal shift. We could demand a certain minimum R^2, but I would like to leave this for a future PR / improvement.

lumicks/pylake/force_calibration/touchdown.py
```diff
@@ -70,6 +72,15 @@ def piecewise_linear(x, x0, y0, k1, k2):
     return np.piecewise(x, [x < x0], [lambda x: k1 * (x - x0) + y0, lambda x: k2 * (x - x0) + y0])
 
 
+def f_test(sse_restricted, sse_unrestricted, num_data, num_pars_difference, num_pars_unrestricted):
+    """Determine p-value whether the bigger model describes the variance significantly better"""
+    nominator = (sse_restricted - sse_unrestricted) / num_pars_difference
+    denominator = sse_unrestricted / (num_data - num_pars_unrestricted)
+    f_statistic = nominator / denominator
```
mcsmurf, line 79: This fails sometimes, not sure if it's just my environment: 
`pylake/force_calibration/touchdown.py:79: RuntimeWarning: divide by zero encountered in double_scalars
    f_statistic = nominator / denominator`
    
```suggestion
    if denominator == 0:
        RuntimeWarning("Could not determine variance due to invalid model data".)
        denominator = 1
        
    f_statistic = nominator / denominator
```

mcsmurf (commented):
LGTM! Another pair of eyes closer to the domain wouldn't hurt though :slightly_smiling_face:

lumicks/pylake/force_calibration/touchdown.py
JoepVanlier: That's a fair point. The reason this happens is because the simulated data in the tests is noise free. It will never happen for real data. It would also happen if <= 4 data points were provided. I will add a warning for it, and filter that warning specifically in the tests that are affected.

mcsmurf: Sounds reasonable - certainly if it does not normally occur outside testing.

lumicks/pylake/force_calibration/touchdown.py
```diff
@@ -70,6 +72,24 @@ def piecewise_linear(x, x0, y0, k1, k2):
...
 
+def f_test(sse_restricted, sse_unrestricted, num_data, num_pars_difference, num_pars_unrestricted):
+    """Determine p-value whether the bigger model describes the variance significantly better"""
+    nominator = (sse_restricted - sse_unrestricted) / num_pars_difference
+    denominator = sse_unrestricted / (num_data - num_pars_unrestricted)
+    if denominator == 0:
+        warnings.warn(
+            RuntimeWarning(
+                "Denominator in F-Test is zero. "
+                "Is it possible you used noise free data or fewer than 4 data points?"
```
mcsmurf, line 83:
```suggestion
                "This may be caused by using noise-free data or fewer than 4 data points"
```

mcsmurf (approved):
We have touched down! Go for launch :rocket:

PR #274, opened 2022-03-02
confocal/widefield: set plot color ranges

JoepVanlier:
**Why this PR?**
Contrast enhancement for confocal and widefield images has been requested frequently.

Previously, you could set a `vmin` and `vmax` which were forwarded to `imshow`. The annoying part of this is that while this works for single color channels, it did not work for composite RGB images. This PR adds the ability to specify color ranges in a unified way for single channels and RGB images, and to specify them (optionally) in percentiles (which is particularly useful for widefield images).

Docs build here: https://lumicks-pylake.readthedocs.io/en/contrast/tutorial/images.html

contrast_enhance

single_channel

image

image

I tried to make sure that the contrast handling is the same for widefield and confocal. Note that because of the way `mpl.pyplot.imshow` works, RGB images need to be handled differently than single color images. There is a good reason for this. When providing a single color image, we do not want to change the actual data (because this modifies downstream properties of the plot (colorbar, tooltip etc). With RGB images, this is unavoidable however.

**What to look out for?**
I debated whether to remove the kind of funky `vmax` handling for widefield images (since it's not "just" forwarded to `imshow`, but it used to adjust the data as well, but only for widefield). Personally, I think that now that it is redundant we should probably get rid of that explicit handling. I'm just not sure whether it should be part of this PR.


lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -531,3 +532,118 @@ def check_result(stack, ref_stack, bit_depth):
     offset_correct_points = [np.array(pp) - (25, 25) for pp in correct_points]
     stack = stack.define_tether(*offset_correct_points)
     check_result(stack, ref_stack, 16)
+
+
+@cleanup
```
rpauszek, line 537: how would you feel about pulling all the plotting tests out into a separate file? I think there's enough that it makes sense, and this one is getting pretty long

If you don't wanna piggyback it on this PR that's fine, but I think it's something we should do in general for all the image class tests

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -972,3 +973,60 @@ def test_plot_with_lf_force():
 
     with pytest.raises(RuntimeError, match="Desired force channel 1x not available in h5 file"):
         kymo.plot_with_force("1x", "red")
+
+
+@cleanup
```
rpauszek, line 978: same thing here with separating out the plotting tests? `test_kymo_plotting.py`?

lumicks/pylake/correlated_stack.py
```diff
@@ -299,11 +307,18 @@ def plot_correlated(
         plot_correlated(
             channel_slice,
             frame_timestamps,
-            lambda frame_idx: self._get_frame(frame_idx)._get_plot_data(channel),
+            lambda frame_idx: self._get_frame(frame_idx)._get_plot_data(
+                channel, adjustment=adjustment
+            ),
```
rpauszek, line 312: maybe define this before the function call?

lumicks/pylake/correlated_stack.py
```diff
@@ -299,11 +307,18 @@ def plot_correlated(
...
             title_factory,
             frame,
             reduce,
             figure_scale=figure_scale,
+            post_update=None
+            if adjustment is None
+            else lambda image_handle, image: adjustment._update_limits(
+                image_handle, image, channel
+            ),
```
rpauszek, line 321: this is some funky formatting, I think it'd be better to just define it before the function call

- or if you go with my previous suggestion the conditional drops out anyway

rpauszek (changes requested):
This PR is a beast, but really awesome functionality! Everyone will be super happy about this.

I've left some suggestions for cleaning up and/or simplifying things a bit, but in general everything looks great and works perfectly with any experimental data I've tried it on

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,97 @@
...
+        """Color adjustment
+
+        Parameters
+        ----------
+        minimum : array_like
+            1 or 3-component list or array of numbers that provide a lower limit for the color
+            mapping.
+        maximum : array_like
+            1 or 3-component list or array of numbers that provide an upper limit for the color
+            mapping.
```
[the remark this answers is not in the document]
JoepVanlier: Much better, thank you!

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,97 @@
...
+            # Plot force 1x with this stack and adjust the contrast by specifying an absolute upper
+            # and lower bound for the intensities.
+            absolute_adjustment = lk.ColorAdjustment([50, 30, 20], [1000, 195, 95])
+            stack.plot_correlated(file.force1x, channel="rgb", adjustment=absolute_adjustment)
+        """
+        minimum, maximum = np.atleast_1d(minimum), np.atleast_1d(maximum)
+        self.mode = mode
+        for bound in (minimum, maximum):
+            if not (len(bound) == 3 or len(bound) == 1):
+                raise ValueError("Color value bounds should be of length 1 or 3")
```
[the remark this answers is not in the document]
JoepVanlier: Not nitpicky, I like it! :)

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,97 @@
...
+        if self.mode == "absolute":
+            minimum, maximum = self.minimum, self.maximum
+        else:
+            bounds = np.array(
+                [
+                    np.percentile(img, [mini, maxi])
+                    for img, mini, maxi in zip(np.moveaxis(image, 2, 0), self.minimum, self.maximum)
+                ]
+            )
+            minimum, maximum = (b for b in np.moveaxis(bounds, 1, 0))
```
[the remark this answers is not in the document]
JoepVanlier: Right. I think this was a leftover from when I had it do all the frames at first. I'll change it, thanks!

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,97 @@
...
+
+        nominator = maximum - minimum
+        nominator[nominator == 0] = 1.0  # prevent div by zero
```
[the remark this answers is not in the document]
JoepVanlier: Whoops! 😊

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,97 @@
...
+            Raw image data.
+        channel : str
+            Channel that's being plotted (e.g. "red" or "rgb").
+        """
+        idx = {"red": 0, "green": 1, "blue": 2, "rgb": 0}[channel]
+        minimum, maximum = self.minimum[idx], self.maximum[idx]
+        if self.mode == "absolute":
+            image_handle.set_clim(minimum, maximum)
+        else:
+            image_handle.set_clim(*np.percentile(image, [self.minimum[idx], self.maximum[idx]]))
```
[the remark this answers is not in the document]
JoepVanlier: Done!

lumicks/pylake/tests/test_image.py
```diff
@@ -193,3 +194,55 @@ def size_test(x, axis):
...
+
+@pytest.mark.parametrize(
+    "minimum, maximum, ref_minimum, ref_maximum",
+    [
+        (1.0, 5.0, [1.0, 1.0, 1.0], [5.0, 5.0, 5.0]),
+        ([1.0, 2.0, 3.0], [3.0, 4.0, 5.0], [1.0, 2.0, 3.0], [3.0, 4.0, 5.0]),
+        ([1.0], [3.0, 4.0, 5.0], [1.0, 1.0, 1.0], [3.0, 4.0, 5.0]),
+        ([1.0, 2.0, 3.0], [5.0], [1.0, 2.0, 3.0], [5.0, 5.0, 5.0]),
+        ([1.0, 2.0, 3.0], 5.0, [1.0, 2.0, 3.0], [5.0, 5.0, 5.0]),
+    ],
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/tests/test_image.py
```diff
@@ -193,3 +194,55 @@ def size_test(x, axis):
...
+def tet_percentile_ranges(minimum, maximum, ref_minimum, ref_maximum):
+    ref_img = np.array([np.reshape(np.arange(10), (2, 5)) for i in np.arange(3)]).swapaxes(0, 2)
+
+    ca = ColorAdjustment(minimum, maximum, mode="percentile")
+    np.testing.assert_allclose(ca.minimum, ref_minimum)
+    np.testing.assert_allclose(ca.maximum, ref_maximum)
+
+    np.testing.assert_allclose(
+        ca._get_data_rgb(ref_img), np.clip((ref_img - ca.minimum) / (ca.maximum - ca.minimum), 0, 1)
+    )
```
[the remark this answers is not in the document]
JoepVanlier: Good point. I think there was even an error in this test actually (see the commits I just pushed) 😮 

After fixing it, the deduplication wouldn't be that much (since the number of cases in the lines part would double with the mode flag). Personally I have a mild preference for having them separate since they test different things.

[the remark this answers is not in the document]
rpauszek: cool, I'm fine keeping them separate. Maybe then move `test_invalid_range()` above or below to keep these two tests grouped

lumicks/pylake/detail/confocal.py
```diff
@@ -310,12 +310,27 @@ def _timestamps(self, channel, reduce=timestamp_mean):
...
+        frame_image = image if frame is None else image[frame]
+
         if channel == "rgb":
-            image = image / np.max(image)
-        return image
+            frame_image = (
+                frame_image / np.max(image)
+                if adjustment is None
+                else adjustment._get_data_rgb(frame_image)
+            )
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, sounds nicer.

lumicks/pylake/correlated_stack.py
```diff
@@ -177,8 +177,13 @@ def plot(self, frame=0, channel="rgb", show_title=True, axes=None, **kwargs):
         default_kwargs = dict(cmap="gray", vmax=None)
         kwargs = {**default_kwargs, **kwargs}
 
-        image = self._get_frame(frame)._get_plot_data(channel, vmax=kwargs["vmax"])
-        axes.imshow(image, **kwargs)
+        image = self._get_frame(frame)._get_plot_data(
+            channel, vmax=kwargs["vmax"], adjustment=adjustment
+        )
```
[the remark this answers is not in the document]
JoepVanlier: Unfortunately, it's not that simple because the `vmax` is implemented kind of weirdly in `_get_plot_data`.

Because of the way it's implemented, `vmax` for widefield is expressed in the space of `(self.data / (2**self.bit_depth - 1))`.

Personally, I'd vote for dropping it (raise with an error to use `ColorAdjustment` instead).

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
JoepVanlier: I can do a PR for that immediately after this one. The reason I'd prefer to not do it immediately is to not get into a mess with fixups and moves at the same time.

lumicks/pylake/correlated_stack.py
JoepVanlier: Done and done.

lumicks/pylake/tests/test_image.py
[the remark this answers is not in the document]
JoepVanlier: Right. Makes sense! Swapped 'em. :)

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,111 @@
...
+            return
+
+        for bound in (minimum, maximum):
+            if len(bound) not in (1, 3):
+                raise ValueError("Color value bounds should be of length 1 or 3")
+
+        if mode not in ("percentile", "absolute", "nothing"):
+            raise ValueError("Mode nust be percentile or absolute")
+
+        self.minimum, self.maximum = minimum * np.ones((3,)), maximum * np.ones((3,))
```
mcsmurf, line 61: I found this easier to reason about, feel free to discard. Essentially favors the power of `falsy` by using `None` for `self.mode` instead of  "nothing". I left the class method as it helps to understand the API, though theoretically that could also go.

Lastly used `np.multiply` to replace `*` since I kept getting this :sweat_smile: 

py_err


```suggestion
        for bound in (minimum, maximum):
            if len(bound) not in (1, 3):
                raise ValueError("Color value bounds should be of length 1 or 3")

        if mode not in ("percentile", "absolute", "nothing"):
            raise ValueError("Mode nust be percentile or absolute")
            
        self.mode = None if mode == "nothing" else mode
        if not self.mode:
            return
            
        self.minimum, self.maximum = np.multiply(minimum, np.ones((3,))), np.multiply(maximum, np.ones((3,)))
```

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def _get_data_rgb(self, image):
+        """Scale RGB data for plotting.
+
+        Parameters
+        ----------
+        image : array_like
+            Raw image data.
+        """
+        if self.mode == "nothing":
```
mcsmurf, line 71:
```suggestion
        if not self.mode:
```

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,111 @@
...
+        Parameters
+        ----------
+        image_handle : matplotlib.image.AxesImage
+            Image handle to apply color limits to
+        image : array_like
+            Raw image data.
+        channel : str
+            Channel that's being plotted (e.g. "red" or "rgb").
+        """
+        if self.mode == "nothing":
```
mcsmurf, line 100:
```suggestion
        if not self.mode:
```

lumicks/pylake/adjustments.py
rpauszek: I like the idea of using `None` internally here, but if you keep the assignment at the top (lines 57-59 in your suggestion), I think you shouldn't have to use `np.multiply`, which I really don't like

lumicks/pylake/adjustments.py
```diff
@@ -0,0 +1,111 @@
...
+        self.mode = mode
+        if self.mode == "nothing":
+            return
+
+        for bound in (minimum, maximum):
+            if len(bound) not in (1, 3):
+                raise ValueError("Color value bounds should be of length 1 or 3")
+
+        if mode not in ("percentile", "absolute", "nothing"):
+            raise ValueError("Mode nust be percentile or absolute")
```
rpauszek, line 59: I guess we don't have to test for `"nothing"` here if we return above

lumicks/pylake/correlated_stack.py
```diff
@@ -291,14 +308,21 @@ def plot_correlated(
         title_factory = lambda frame: make_image_title(self, frame, show_name=False)
         frame_timestamps = self.frame_timestamp_ranges
 
+        def frame_grabber(frame_idx):
+            return self._get_frame(frame_idx)._get_plot_data(channel, adjustment=adjustment)
+
+        def post_update(image_handle, image):
+            return adjustment._update_limits(image_handle, image, channel)
```
rpauszek, line 315: Can this be wrapped into the base `plot_correlated` function somehow? Because right now both here and for `Scan` you have the exact same `post_udpate` defined.

lumicks/pylake/adjustments.py
mcsmurf: Do tell, why do you not like `np.multiply`? Not vouching for it per se just curious. I imagined it was not too bad since the docs do say it is "Equivalent to x1 * x2 in terms of array [broadcasting](https://numpy.org/doc/stable/user/basics.broadcasting.html)."

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -61,8 +61,7 @@ data looks like::
 
 .. image:: kymo_force.png
 
-Note how we specified a `vmax` for the image. This argument reflects the photon count that corresponds to the maximum
-of the colormap. Any photon count higher than that will be clipped to the maximal color.
+Note how we specified a color adjustment for the kymograph. Any photon count outside the range provided to :class:`~lumicks.pylake.ColorAdjustment` will be clipped to the nearest color.
```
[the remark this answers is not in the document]
mcsmurf: Makes sense to do an overhaul of the docs. I imagined it would take more time/effort to do it all at once - hence this "chipping off one at a time" approach. If the preferred way is the "tutorial voice" that's alright as well :ok_hand:

lumicks/pylake/adjustments.py
rpauszek: It just seems very un-pythonic to me ... actually I didn't even know that `np.multiply()` existed. To me, using the operator is a lot cleaner. It's also not quite clear to me why you'd get the error above, but I didn't try out your code explicitly

lumicks/pylake/adjustments.py
JoepVanlier: 👍

lumicks/pylake/adjustments.py
JoepVanlier: 👍

lumicks/pylake/adjustments.py
JoepVanlier: 👍

lumicks/pylake/adjustments.py
JoepVanlier: Alright, I'll switch it to `None` for internal use. I'd personally still prefer the early out though.

lumicks/pylake/correlated_stack.py
JoepVanlier: I did this intentionally. I don't want `plot_correlated` to be dependent on implementational details of `ColorAdjustment`. `plot_correlated` may be used for other things in the future as well. I don't really think deduplicating 1 line of code is worth coupling them?

rpauszek: Yeah ok, fair enough. In that case you might want to consider pulling out the `post_update` in `Scan.plot_correlated()` to match what you do here. Defining that lambda in the function call as an argument looks a bit strange the way it's formatted, but really that's just being super nitpicky so feel free to just keep it as is

rpauszek (approved):
looks amazing, can't wait to be able to use this feature!

lumicks/pylake/correlated_stack.py
JoepVanlier: Okay. I'll pull it out :)

PR #277, opened 2022-03-11
Generalize dwelltime analysis and make public

rpauszek:
**Why this PR?**
The idea was always to have public dwelltime analysis available at some point, but the initial feature was written specifically for kymotracking bound lifetime analysis. Now, there is a request to release a notebook to Harbor which requires state-specific dwelltime analysis for unfolding experiments. So it now makes sense to have this feature generalized and public so the notebook doesn't have to dig into the private API. This also helps to unlock some analyses for protein folding experiments.

**Notes:**
- There are quite a few commits which I kept separate for clarity during review. Many will be squashed on final merge
- I moved the original class from the `kymotracker` subpackage to the `population` subpackage as it's no longer specific to kymotracking analysis.
- Classes are renamed to be more general. This isn't a breaking change as the classes were previously only exposed through `KymoLineGroup.fit_binding_times()` and there was a specific note in the docstring to not construct these manually.
- The `DwelltimeModel` constructor was refactored and `.pdf()` method was added to more closely match the API of `GaussianMixtureModel` (trying to keep all model classes as similar in API as possible for the `population` package.
- All public functionality remains unchanged
- Optimization algorithm functionality is untouched
- `.plot()` was renamed to `.hist()` to more closely match the actual functionality and naming convention in the GMM class. `.plot()` is not needed for this type of data, so we can alias via a deprecation.

Docs are [here](https://lumicks-pylake.readthedocs.io/en/public_dwelltimes/)


mikhas (commented):
Most changes in this PR are straight-forward, that's really nice. Great work on the docs, as always.

docs/tutorial/kymotracking.rst
```diff
@@ -452,109 +452,31 @@ Dwelltime analysis
...
+    dwell2 = traces.fit_binding_times(n_components=2)
+    print(dwell2.lifetimes)  # list of bound lifetimes
+    print(dwell2.amplitudes)  # list of fractional amplitudes for each component
 
-Here we see the distributions of the bootstrapped parameters. The vertical lines indicate the
-means of the distributions, while the red area indicates the estimated confidence intervals. The `alpha` argument determines
-the CI that is estimated as `100*(1-alpha)` % CI; in this case we're showing the estimate for the 95% CI. The values for the
-lower and upper bounds are the `100*(alpha/2)` and `100*(1-alpha/2)` percentiles of the distributions.
+For a detailed description of the optimization method and available attributes/methods see the Dwelltime Analysis section
+in :doc:`Population Dynamics </tutorial/population_dynamics>`.
```
JoepVanlier, line 478: Nice to crossreference like this!

lumicks/pylake/population/dwelltime.py
```diff
@@ -38,15 +39,18 @@ def _sample_distributions(self, optimized, iterations):
...
+        min_observation_time, max_observation_time = optimized._observation_limits
         for itr in range(iterations):
             sample = np.random.choice(optimized.dwelltimes_sec, size=n_data, replace=True)
-            result = _kinetic_mle_optimize(
+            result, _ = _exponential_mle_optimize(
                 optimized.n_components,
                 sample,
-                *optimized.observation_limits,
+                min_observation_time,
+                max_observation_time,
```
JoepVanlier, line 49: nitpick: but I thought the unpack was just as clean, since you don't use these times anywhere else?

lumicks/pylake/population/dwelltime.py
```diff
@@ -145,37 +149,52 @@ def plot_axes(data, key, component, use_index):
...
-    dwelltimes_sec: np.ndarray = field(repr=False)
-    observation_limits: list = field(repr=False)
-    _parameters: np.ndarray = field(repr=False)
-    log_likelihood: float
-    bootstrap: BindingDwelltimesBootstrap = field(
-        default_factory=BindingDwelltimesBootstrap, init=False, repr=False
-    )
+    def __init__(
+        self,
+        dwelltimes_sec,
```
JoepVanlier, line 173: Does it have to be seconds?

lumicks/pylake/population/dwelltime.py
```diff
@@ -145,37 +147,31 @@ def plot_axes(data, key, component, use_index):
...
-    bootstrap: BindingDwelltimesBootstrap = field(
-        default_factory=BindingDwelltimesBootstrap, init=False, repr=False
-    )
+        self.n_components = n_components
+        self.dwelltimes_sec = dwelltimes_sec
+        self._observation_limits = (min_observation_time, max_observation_time)
+        self._parameters, self._log_likelihood = _exponential_mle_optimize(
+            n_components, dwelltimes_sec, min_observation_time, max_observation_time
+        )
+        self.bootstrap = BindingDwelltimesBootstrap()
```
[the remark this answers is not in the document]
JoepVanlier: It's easier to reason about this class if the settings don't change. If the settings can't change, and the only purpose of this class is to do this analysis, why not optimize immediately? Saves users an extra call to a `.optimize()` or the like?

The really heavy part (the bootstrap) is not always done, as far as I can see.

lumicks/pylake/population/dwelltime.py
```diff
@@ -204,6 +204,14 @@ def bic(self):
     def calculate_bootstrap(self, iterations=500):
         self.bootstrap._sample_distributions(self, iterations)
 
+    def pdf(self, x):
+        """Probability Distribution Function (states as rows)."""
```
JoepVanlier, line 230: Since it's public, describe the input parameter?

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -422,13 +422,23 @@ def save(self, filename, delimiter=";", sampling_width=None):
         """
         export_kymolinegroup_to_csv(filename, self._src, delimiter, sampling_width)
 
-    def fit_binding_times(self, n_components, *, exclude_ambiguous_dwells=True):
+    def fit_binding_times(
+        self, n_components, *, exclude_ambiguous_dwells=True, tol=None, max_iter=None
```
JoepVanlier, line 426: `tol` and `max_iter` are not actually forwarded.

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -422,13 +422,23 @@ def save(self, filename, delimiter=";", sampling_width=None):
...
         ----------
         n_components : int
-            number of components in the model. Currently only values of {1, 2} are supported.
+            Number of components in the model. Currently only values of {1, 2} are supported.
+        exclude_ambiguous_dwells : bool
+            Determines whether to exclude dwelltimes which are not exactly determined. If `True`, tracks which
+            start in the first frame or end in the last frame of the kymograph are not used in the analysis,
+            since the exact start/stop times of the binding event are not definitively known.
+        tol : float
+            The tolerance for optimization convergence.
```
JoepVanlier, line 439: This is stated too ambiguously, what tolerance is specified here?

lumicks/pylake/population/dwelltime.py
```diff
@@ -145,37 +149,52 @@ def plot_axes(data, key, component, use_index):
...
-    bootstrap : BindingDwelltimesBootstrap
-        object containing information about the bootstrapping analysis.
+    n_components : int
+        number of components in the model.
+    min_observation_time : float
+        minimum experimental observation time
+    max_observation_time : float
+        maximum experimental observation time.
+    tol : float
+        The tolerance for optimization convergence.
```
JoepVanlier, line 166: Same as before, which tolerance is this?

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: Maybe add a test for this as well?

JoepVanlier (changes requested):
Lookin' good!

I only have some small requests. :)

lumicks/pylake/population/dwelltime.py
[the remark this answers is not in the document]
rpauszek: Indeed, what Joep said. And correct, the bootstrap (which can take several seconds depending on how many iterations you want) is only called on demand through `DwelltimeModel.calculate_bootstrap()`

lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -1,21 +1,21 @@
...
-from lumicks.pylake.population.dwelltime import _exponential_mle_optimize
+from lumicks.pylake.population.dwelltime import DwelltimeModel
 
 
 @pytest.mark.filterwarnings("ignore:Values in x were outside bounds")
 def test_likelihood(exponential_data):
     # single exponential data
     dataset = exponential_data["dataset_1exp"]
-    fit = _exponential_mle_optimize(1, dataset["data"], *dataset["parameters"].observation_limits)
+    fit = DwelltimeModel(dataset["data"], 1, *dataset["parameters"].observation_limits)
```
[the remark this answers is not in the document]
rpauszek: this is one that I mentioned in the main note will be squashed before the final merge

docs/tutorial/population_dynamics.rst
```diff
@@ -82,3 +79,108 @@ two different states. The result is the creation of a new state that does not ar
...
+Dwelltime analysis
+------------------
+
+The lifetimes of bound states can be estimated by fitting to an Exponential distribution. The `DwelltimeModel` class can be used
+to optimize the model parameters for an array of determined dwelltimes::
+
+    dwell_1 = lk.DwelltimeModel(dwelltimes_seconds, n_components=1)
+
+The model is optimized using Maximum Likelihood Estimation (MLE) :cite:`kaur2019dwell,woody2016memlet`. The advantage of this method
+is that it does not require binning the data. The number of exponential components to be used for the fit is chosen with the `n_components` argument.
```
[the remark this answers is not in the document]
rpauszek: this is typical phrasing, in this context "binning" can be used as a verb

lumicks/pylake/population/dwelltime.py
rpauszek: I guess not. I originally named it like that to hint that if you want the lifetime in seconds, the input needs to be in units of seconds. but I'll just beef up the docstring

lumicks/pylake/kymotracker/kymoline.py
rpauszek: added a note that these are forwarded to `scipy.minimize()`

lumicks/pylake/kymotracker/kymoline.py
rpauszek: good catch

lumicks/pylake/population/dwelltime.py
```diff
@@ -145,37 +147,58 @@ def plot_axes(data, key, component, use_index):
...
-        log likelihood of the trained model
-    bootstrap : BindingDwelltimesBootstrap
-        object containing information about the bootstrapping analysis.
+    min_observation_time : float
+        minimum experimental observation time
+    max_observation_time : float
+        maximum experimental observation time.
+    tol : float
+        The tolerance for optimization convergence. This parameter is forwarded as the `ftol` argument
+        to `scipy.minimize(method="L-BFGS-B")`.
```
JoepVanlier, line 168: Don't we use `SLSQP`?

Also, maybe just call  it `ftol` if that's the argument it is forwarded to. While I'm not a huge fan of their naming convention, it's one less mental translation that people have to make.

JoepVanlier (changes requested):

lumicks/pylake/population/dwelltime.py
rpauszek: oops...

I get what you mean about `ftol`, but I would argue it's more important to keep our syntax consistent rather than staying in line with the syntax used by the optimizers in the backend. This way, people using `pylake` only ever have to remember `tol` instead of trying to remember which of our functions use `tol` vs `ftol` etc... This is also why here I used our `max_iter` rather than the `scipy` `maxiter` arg

JoepVanlier: Fair enough!

JoepVanlier (approved):
LGTM! 👍

mcsmurf (approved):
Nicely done! Works as advertised (the code at least :slightly_smiling_face: ). Go for launch :rocket:

lumicks/pylake/population/dwelltime.py
```diff
@@ -365,8 +422,8 @@ def exponential_mixture_log_likelihood(params, t, min_observation_time, max_obse
     return -np.sum(log_likelihood)
 
 
-def _kinetic_mle_optimize(
-    n_components, t, min_observation_time, max_observation_time, initial_guess=None
+def _exponential_mle_optimize(
+    n_components, t, min_observation_time, max_observation_time, initial_guess=None, options={}
```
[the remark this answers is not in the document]
rpauszek: I actually didn't know that 😆 this is just how I've always done it

lumicks/pylake/population/dwelltime.py
```diff
@@ -145,37 +147,58 @@ def plot_axes(data, key, component, use_index):
...
-    _parameters: np.ndarray = field(repr=False)
-    log_likelihood: float
-    bootstrap: BindingDwelltimesBootstrap = field(
-        default_factory=BindingDwelltimesBootstrap, init=False, repr=False
-    )
+    def __init__(
+        self,
+        dwelltimes,
+        n_components=1,
+        *,
```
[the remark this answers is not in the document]
JoepVanlier: mandatory keyworded arguments after the `*` ;)

PR #278, opened 2022-03-14
Fix kymotracker aspect ratio bug

rpauszek:
**Why this PR?**
While working on analysis with kymos acquired under different pixel sizes, I noticed that sometimes the image in the kymotracker did not make any sense with the requested `axis_aspect_ratio` argument.  Finally I found that the aspect ratio was actually being calculated as `height/width` rather than `width/height`. This is fixed now, and a specific test added.

*In the following, the bug in the images aren't very apparent, but look at the printed values at the bottom of each figure*

Before:

image

After:

image

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -35,7 +35,8 @@ class KymoWidget:
             Kymograph channel to use.
         axis_aspect_ratio : float
             Desired aspect ratio of the viewport. Sometimes kymographs can be very long and thin.
-            This helps you visualize them.
+            This helps you visualize them. The aspect ratio is defined in physical spatial and
+            temporal units (rather than pixels).
         min_length : int
             Minimum length of a trace. Traces shorter than this are discarded.
         use_widgets : bool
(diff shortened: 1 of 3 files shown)


lumicks/pylake/nb_widgets/tests/test_kymotracker_widget.py
```diff
@@ -45,6 +45,16 @@ def test_invalid_algorithm_parameter(kymograph):
         kymo_widget.track_all()
 
 
+@cleanup
+def test_aspect_ratio(kymograph, region_select):
+    for requested_aspect in (2, 3, 5):
+        kymo_widget = KymoWidgetGreedy(kymograph, "red", use_widgets=False, axis_aspect_ratio=requested_aspect)
+        ax = kymo_widget._axes
+        aspect = ax.get_xlim()[1] / ax.get_ylim()[0]
+        print(aspect)
```
JoepVanlier, line 54: leftover print statement? :)

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -57,8 +57,8 @@ def __init__(
         self.axis_aspect_ratio = (
             min(
                 axis_aspect_ratio,
-                calibrated_image.to_position(calibrated_image.data.shape[1])
-                / calibrated_image.to_seconds(calibrated_image.data.shape[0]),
+                calibrated_image.to_seconds(calibrated_image.data.shape[1])
+                / calibrated_image.to_position(calibrated_image.data.shape[0]),
```
JoepVanlier, line 61: Might be nice to put in the docstring that the aspect ratio is defined in the units of the kymo (rather than data or screen pixels).

JoepVanlier (approved):
Nice catch! A few small things I noticed.

PR #282, opened 2022-03-24
force calibration: allowing fixing both diode frequency and relaxation factor

JoepVanlier:
**Why this PR?**
This is in preparation of the pre-characterized diodes. The plan is to measure both the diode frequency and relaxation factor for our diodes. This PR allows fixing either or both (previously one could only fix `f_diode`).

It also deduplicates some parameter descriptions by using the diode model the fixed model is derived from to obtain the parameter descriptions.


agom (approved):
LGTM!

PR #283, opened 2022-03-28
docs: update sphinx

JoepVanlier:
**Why?**
We were getting build failures on RtD. See: http://github.com/readthedocs/readthedocs.org/issues/9037
That link actually mentions that we could have also solved it by pinning `jinja`. I did try this initially, but unfortunately, installing `pylake` afterwards bumps it again and makes the build fail. I figured it's a good idea to just update `sphinx` and get us up to date.

I also took the liberty of clearing out all the warnings Sphinx was producing. Reason being that  it's difficult to see real ones when they're drowned in noise.

Docs build can be found here: https://lumicks-pylake.readthedocs.io/en/update_sphinx/index.html

### DIFF
--- a/.github/workflows/pylake_test.yml
+++ b/.github/workflows/pylake_test.yml
@@ -28,5 +28,5 @@ jobs:
         python pylake_test.py
     - name: Black
       run: |
-        pip install black==22.1.0
+        pip install black==22.3.0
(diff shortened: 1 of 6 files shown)


dean0x7d (approved):
Nice fix and cleanup!

PR #285, opened 2022-03-30
channel: preserve calibration if valid

JoepVanlier:
**Why this PR?**
When adding a scalar value, subtracting a scalar value or negating a channel the force calibration is still valid and should be forwarded to the constructed `Slice`. This PR forwards the appropriate calibration items to the processed slice.

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -84,13 +84,19 @@ class Slice:
     def __neg__(self):
         if isinstance(self._src, TimeTags):
             raise NotImplementedError("This operation is not supported for TimeTag data")
-        return Slice(self._src._with_data(-self.data))
+        return Slice(self._src._with_data(-self.data), calibration=self._calibration)
 
     def __add__(self, other):
-        return Slice(self._src._with_data(self.data + self._unpack_other(other)))
+        return Slice(
+            self._src._with_data(self.data + self._unpack_other(other)),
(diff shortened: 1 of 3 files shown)


lumicks/pylake/channel.py
```diff
@@ -84,13 +84,19 @@ def _unpack_other(self, other):
...
             raise NotImplementedError("This operation is not supported for TimeTag data")
-        return Slice(self._src._with_data(-self.data))
+        return Slice(self._src._with_data(-self.data), calibration=self._calibration)
 
     def __add__(self, other):
-        return Slice(self._src._with_data(self.data + self._unpack_other(other)))
+        return Slice(
+            self._src._with_data(self.data + self._unpack_other(other)),
+            calibration=self._calibration if np.isscalar(other) else None,
+        )
```
mcsmurf, line 93: I don't see `__radd__(self, other)` though it's also supposed to preserve calibration. Intentional?

JoepVanlier: https://github.com/lumicks/pylake/blob/1b51ad75519f81e8bfc532ea8f6f56ca5cb2abc6/lumicks/pylake/channel.py#L121 :)

mcsmurf: Ah, the ol' commutative property :ok_hand:

mcsmurf (approved):
LGTM! :rocket:

PR #286, opened 2022-03-30
channel: mask slices

JoepVanlier:
**Why this PR?**
A fairly frequent operation is to mask some `Slice` data. Currently, we always have to import `TimeSeries` manually and basically construct a new slice when we need to do this. Technically, there is no way for users to do this at the moment.

This PR introduces an extra option to the `__getitem__` dunder that allows slicing by a logical boolean numpy array.

(While working on implementing piezo tracking in Pylake, I encountered a few things that were missing from the `Slice` API. Rather than appending them on an already sizable PR, I've decided to spin them out and put them up for review separately).

Note: Currently, this doesn't accept a `List` of booleans.  We could consider extending the feature to also allow this.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 * Allow applying color intensity adjustments on images using `lk.ColorAdjustment()`. See [Confocal images](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/images.html#correlating-scans) and [Correlated stacks](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/correlatedstacks.html#correlated-stacks) for more information.
 * Added `DwelltimeModel` to fit dwelltimes to an exponential (mixture) distribution. For more information, see the tutorials section on [Population Dynamics](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/population_dynamics.html)
 * Allow slicing directly with an object with a `.start` and `.stop` property. See [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#markers) for more information.
+* Allow boolean array indexing on `Slice` (e.g. `file.force1x[file.force1x.data > 5]`. See [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#boolean-array-indexing) for more information.
 
 #### Bug fixes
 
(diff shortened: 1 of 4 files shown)


docs/tutorial/file.rst
```diff
@@ -166,6 +166,21 @@ Plotting is typically performed with the origin of the plot set to the timestamp
...
+
+    mask = file.force1x.data <= 5
+    masked = file.force1x[mask]
+
+Multiple criteria can be combined by using `numpy`'s logic operators.
+For example, restricting the forces between 2 and 5 can be accomplished as follows::
+
+    mask = np.logical_and(file.force1x.data > 2, file.force1x.data < 5)
+    masked = file.force1x[mask]
+
```
dean0x7d, line 183: It would be good to follow `numpy` naming conventions here and call this boolean indexing: https://numpy.org/doc/stable/user/basics.indexing.html#boolean-array-indexing This is because masks and masked arrays have a very different meaning in `numpy`: https://numpy.org/doc/stable/reference/maskedarray.html And I think it's good for us to stay close to `numpy` conventions as closely as possible.

changelog.md
```diff
@@ -9,6 +9,7 @@
 * Added `DwelltimeModel` to fit dwelltimes to an exponential (mixture) distribution. For more information, see the tutorials section on [Population Dynamics](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/population_dynamics.html)
 * Allow slicing directly with an object with a `.start` and `.stop` property. See [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#markers) for more information.
 * When doing arithmetic on `Slice`, propagate the calibration if it is still valid.
+* Allow masking a `Slice` by a boolean numpy array (e.g. `file.force1x[file.force1x.data > 5]`. See [files and channels](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/file.html#markers) for more information.
```
dean0x7d, line 12: The URL is pointing to markers instead of the new section.

dean0x7d (commented):
Nice convenience feature! The code looks good to me. I just think that the terminology in the docs should be aligned with `numpy`.

changelog.md
JoepVanlier: Good catch. Fixed!

dean0x7d (approved):
`shipit = pull_requests[pull_requests == approved]`

PR #287, opened 2022-04-01
touchdown: include exponential decay in touchdown

JoepVanlier:
**Why?**
As you move away from the surface, the interference pattern decays over time. Including this decay provides a far better fit for longer touchdowns.

image


dean0x7d (approved):
LGTM! (Sorry for the late review. I must have missed it last week.)

PR #288, opened 2022-04-05
fitting: Add ssDNA and dsDNA wrapper functions

mcsmurf:
### Why?
This allows the construction of models for ssDNA and dsDNA using initial guesses based on literature. These are based on the odijk and freely jointed chain implementations.

### DIFF
--- a/lumicks/pylake/fitting/models.py
+++ b/lumicks/pylake/fitting/models.py
@@ -259,6 +259,51 @@ def odijk(name):
     )
 
 
+def dsdna_odijk(name, dna_length_kbp, um_per_kbp=0.34, temperature=24.53608821):
+    """Odijk's Extensible Worm-Like Chain model with distance as the dependent variable, using user-specified kilobase-pairs (useful for 10 pN < F < 30 pN).
+
+    References:
+      1. T. Odijk, Stiff Chains and Filaments under Tension, Macromolecules
+         28, 7016-7018 (1995).
+      2. M. D. Wang, H. Yin, R. Landick, J. Gelles, S. M. Block, Stretching
+         DNA with optical tweezers., Biophysical journal 72, 1335-46 (1997).
+      3. Manning, G. S. (2006). The persistence length of DNA is reached from
+         the persistence length of its null isomer through an internal
+         electrostatic stretching force. Biophysical journal, 91(10),
+         3607-3616.
+      4. Liu, J. H., Xi, K., Zhang, X., Bao, L., Zhang, X., & Tan, Z. J.
+         (2019). Structural flexibility of DNA-RNA hybrid duplex: stretching
+         and twist-stretch coupling. Biophysical journal, 117(1), 74-86.
+      5. Herrero-Galán, E., Fuentes-Perez, M. E., Carrasco, C., Valpuesta,
+         J. M., Carrascosa, J. L., Moreno-Herrero, F., & Arias-Gonzalez, J. R.
+         (2013). Mechanical identities of RNA and DNA double helices unveiled
+         at the single-molecule level. Journal of the American Chemical Society,
+         135(1), 122-131.
+     6.  Saenger, W. (1984). Principles of Nucleic Acid Structure. Springer. New York, Berlin, Heidelberg.
+
+    Parameters
+    ----------
+    name : str
+        Name for the model. This name will be prefixed to the model parameter names.
+    dna_length_kbp: integer
+        The length of the dna in tether/construct measured in kilobase-pairs
+    um_per_kbp: float
+        The number of kilobase pairs evaluating to 1 um. This is used to convert the length in kbp to um as applied in the fit function [6].
+    temperature: float
+        The temperature in celsius. This is used to calculate the boltzmann * temperature (kT) value
+    """
+    from scipy import constants
+
+    model = odijk(name)
+    model.defaults[f"{name}/Lc"].value = dna_length_kbp * um_per_kbp
+    model.defaults[f"{name}/Lp"].value = 50.0  # [3]
+    model.defaults[f"{name}/St"].value = 1200.0  # [4, 5]
+    model.defaults["kT"].value = (
+        1e21 * constants.k * constants.convert_temperature(temperature, "C", "K")
+    )
+    return model
+
+
 def inverted_odijk(name):
     """Odijk's Extensible Worm-Like Chain model with force as dependent variable (useful for 10 pN < F < 30 pN).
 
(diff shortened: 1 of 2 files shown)


lumicks/pylake/fitting/models.py
```diff
@@ -259,6 +259,43 @@ def odijk(name):
...
+    name : str
+        Name for the model. This name will be prefixed to the model parameter names.
+    kilobase_pairs : integer
+        The number of kilobase-pairs in the sample
+    kilobase_pairs_per_micron: float
+        The number of kilobase-pairs evaluating to 1 micron. This is used to convert the kilobase-pairs to microns for the fit function.
+    temperature: float
+        The temperature in celsius. This is used to calculate the boltzmann * temperature value
+    """
+    import scipy.constants as spy
```
JoepVanlier, line 283: I'd probably just import constants in this case and use it directly.
```python
from scipy import constants
```
And then go `constants.k` etc. It's a little funky to import constants and then pretend it's scipy.

lumicks/pylake/fitting/models.py
```diff
@@ -259,6 +259,43 @@ def odijk(name):
...
+    model.defaults[f"{name}/Lc"].value = kilobase_pairs * kilobase_pairs_per_micron
+    # Based on lambda dna C-trap experiments
+    model.defaults[f"{name}/Lp"].value = 56.0
+
+    # Convert temperature to kelvin in order to evaluate kT
+    temp = spy.convert_temperature(temperature, "C", "K")
+    # Conversion factor required to convert kT value from J/K to pN*nm/K
+    conversion_factor = 1e21
+    boltzmann = spy.k
+    model.defaults["kT"].value = conversion_factor * boltzmann * temp
```
JoepVanlier, line 295: Personally, I don't really see much added value in assigning these to variables first.
```python
# Conversion of 1e21 required to convert kT value from J/K to pN*nm/K
model.defaults["kT"].value = 1e21 * constants.k * temp
```

lumicks/pylake/fitting/models.py
```diff
@@ -259,6 +259,43 @@ def odijk(name):
...
+        The number of kilobase-pairs evaluating to 1 micron. This is used to convert the kilobase-pairs to microns for the fit function.
+    temperature: float
+        The temperature in celsius. This is used to calculate the boltzmann * temperature value
+    """
+    import scipy.constants as spy
+
+    model = odijk(name)
+    model.defaults[f"{name}/Lc"].value = kilobase_pairs * kilobase_pairs_per_micron
+    # Based on lambda dna C-trap experiments
+    model.defaults[f"{name}/Lp"].value = 56.0
```
JoepVanlier, line 288: I'd really prefer a citation over "# Based on lambda dna C-trap experiments" since our users can't really source that.

lumicks/pylake/fitting/models.py
```diff
@@ -259,6 +259,43 @@ def odijk(name):
...
+         28, 7016-7018 (1995).
+      2. M. D. Wang, H. Yin, R. Landick, J. Gelles, S. M. Block, Stretching
+         DNA with optical tweezers., Biophysical journal 72, 1335-46 (1997).
+
+    Parameters
+    ----------
+    name : str
+        Name for the model. This name will be prefixed to the model parameter names.
+    kilobase_pairs : integer
+        The number of kilobase-pairs in the sample
```
JoepVanlier, line 277: In the sample or in the tether / construct?

lumicks/pylake/fitting/models.py
```diff
@@ -259,6 +259,43 @@ def odijk(name):
...
+    kilobase_pairs_per_micron: float
+        The number of kilobase-pairs evaluating to 1 micron. This is used to convert the kilobase-pairs to microns for the fit function.
+    temperature: float
+        The temperature in celsius. This is used to calculate the boltzmann * temperature value
```
JoepVanlier, line 281: Not sure if `This is used to calculate the boltzmann * temperature value` means much to users.

JoepVanlier (changes requested):
I added some minor requests, but I am waiting to sign off on it until we have some citations for the parameter values.

I also think it's a good idea if @rpauszek has a look, since this model may be used / needed in DNA workflows.

lumicks/pylake/fitting/models.py
mcsmurf: Indeed. I have managed to obtain more citation data, will update as we go

lumicks/pylake/fitting/models.py
```diff
@@ -273,19 +273,21 @@ def dsdna(name, kilobase_pairs, kilobase_pairs_per_micron=0.34, temperature=24.5
...
+        The temperature in celsius. This is used to calculate the boltzmann * temperature (kT) value
     """
     import scipy.constants as spy
 
     model = odijk(name)
-    model.defaults[f"{name}/Lc"].value = kilobase_pairs * kilobase_pairs_per_micron
-    # Based on lambda dna C-trap experiments
-    model.defaults[f"{name}/Lp"].value = 56.0
+    model.defaults[f"{name}/Lc"].value = dna_length_kbp * um_per_bp
+    # Nucleic Acids: Structures, Properties, and Function. By Victor A. Bloomfield, Donald M. Crothers, and Ignacio Tinoco, Jr. University Science Books: Sausalito, CA; 2000. Clothbound, 800 pp, $88.00, ISBN 0-935702-49-0.
```
JoepVanlier, line 287: This citation should go in the docstring where it is publicly visible. Note that I can't check whether this value is indeed in this book since I don't have access to it.

lumicks/pylake/fitting/models.py
```diff
@@ -273,19 +273,21 @@ def dsdna(name, kilobase_pairs, kilobase_pairs_per_micron=0.34, temperature=24.5
...
+    model.defaults[f"{name}/Lp"].value = 50.0
+    # https://www.pnas.org/doi/pdf/10.1073/pnas.1407197111
```
JoepVanlier, line 289: Same as before. Should go in the docstring. We use APA style:
`Lipfert, J., Skinner, G. M., Keegstra, J. M., Hensgens, T., Jager, T., Dulin, D., ... & Dekker, N. H. (2014). Double-stranded RNA under force and torque: similarities to and striking differences from double-stranded DNA. Proceedings of the National Academy of Sciences, 111(43), 15408-15413.`

lumicks/pylake/fitting/models.py
```diff
@@ -259,7 +259,7 @@ def odijk(name):
     )
 
 
-def dsdna(name, kilobase_pairs, kilobase_pairs_per_micron=0.34, temperature=24.53608821):
+def dsDNA(name, dna_length_kbp, um_per_bp=0.34, temperature=24.53608821):
```
JoepVanlier, line 262: I think it should be in line with python conventions and be named `dsdna`.

JoepVanlier: I suspect it's either `um_per_kbp` or `nm_per_bp` , but not `um per bp` (I prefer the former since we work with microns for Lc).

lumicks/pylake/fitting/models.py
```diff
@@ -389,17 +391,17 @@ def ssdna(name, kilobases, kilobases_per_micron=0.56, temperature=24.53608821):
...
+    nm_per_b: float
+        The number of bases evaluating to 1 um. This is used to convert the length in kb to um as applied in the fit function.
     temperature: float
         The temperature in celsius. This is used to calculate the boltzmann * temperature value
     """
     import scipy.constants as spy
 
     model = freely_jointed_chain(name)
-    model.defaults[f"{name}/Lc"].value = kilobases * kilobases_per_micron
+    model.defaults[f"{name}/Lc"].value = dna_length_kb * nm_per_b
```
JoepVanlier, line 404: Remember that `Lc` is in microns. I'd opt for `um_per_kb`.

JoepVanlier (changes requested):

lumicks/pylake/fitting/models.py
mcsmurf: I'll see if I can understand this first indeed. These were based on the citation. :ok_hand:

lumicks/pylake/fitting/models.py
mcsmurf: Ditto on the citation. Will update

lumicks/pylake/fitting/models.py
JoepVanlier: Fixes look good.

lumicks/pylake/fitting/models.py
JoepVanlier: 👍

JoepVanlier (commented):
I think it looks good. Final thing is to verify those parameter values for ssdna.

@rpauszek do you think it's ok to have `ssdna` and `dsdna` in the global namespace, or did you have another plan for those?

JoepVanlier (approved):
👍

rpauszek (approved):
LGTM! :shipit:

PR #289, opened 2022-04-06
touchdown: make pylake responsible for downsampling

JoepVanlier:
**Why?**
It's more convenient to do this on the Pylake side than to make BL responsible for it.
As an additional benefit, this makes the downsampling more transparent to users wanting to reproduce their results.

### DIFF
--- a/lumicks/pylake/force_calibration/tests/test_touchdown.py
+++ b/lumicks/pylake/force_calibration/tests/test_touchdown.py
@@ -69,13 +69,34 @@ def test_sine_with_polynomial(amplitude, frequency, phase_shift, poly_coeffs):
     np.testing.assert_allclose(np.sum((sim - test_data) ** 2), 0, atol=1e-8)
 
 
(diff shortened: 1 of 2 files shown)


lumicks/pylake/force_calibration/tests/test_touchdown.py
```diff
@@ -70,12 +70,12 @@ def test_sine_with_polynomial(amplitude, frequency, phase_shift, poly_coeffs):
 
 
 def test_touchdown(mack_parameters):
-    stage_positions = np.arange(99.5, 103.5, 0.01)
+    stage_positions = np.arange(99.5, 103.5, 0.01 / 1502)
```
dean0x7d, line 73: The `/ 1502` factor seems a bit too magical for future readers. Can you maybe clarify it somehow? A named variable, comment, or a named function to generate the test data?

dean0x7d (approved):
Thanks for making this change!

PR #290, opened 2022-04-07
adjustments: adjust gamma with `ColorAdjustment`

JoepVanlier:
**Why this PR?**
This was a highly requested feature. Basically, it allows users to perform gamma adjustments on images.


lumicks/pylake/adjustments.py
```diff
@@ -47,19 +52,20 @@ def __init__(self, minimum, maximum, mode="absolute"):
...
+        minimum, maximum, gamma = (np.atleast_1d(x) for x in (minimum, maximum, gamma))
         self.mode = None if mode == "nothing" else mode
         if not self.mode:
             return
 
-        for bound in (minimum, maximum):
+        for bound in (minimum, maximum, gamma):
             if len(bound) not in (1, 3):
-                raise ValueError("Color value bounds should be of length 1 or 3")
+                raise ValueError("Color value bounds and gamma should be of length 1 or 3")
```
aafkevandenberg, line 62: Not sure whether it is necessary to have a different gamma for each color channel. But no harm to keep that option.

JoepVanlier: Code-wise it was actually simpler to offer this feature than not. I also didn't really see any harm in allowing it.

PR #291, opened 2022-04-11
docs: explain how to access channel data

aafkevandenberg:
i noticed that the docs do not explain that channel data can be accessed as file[group name][channel name].


JoepVanlier (approved):
Nice catch! 👍

PR #292, opened 2022-04-12
make KymoLine immutable

rpauszek:
**Why this PR?**
Because any modifications to `KymoLine` instances should be made via refinement methods or specifically allowed manual editing

This change is pulled out from #262, which is on hold pending validation of the algorithm


JoepVanlier (approved):
LGTM!

PR #299, opened 2022-04-25
Add smoothing spline baseline

aafkevandenberg:
I added functions to fit a baseline with a smoothing spline.
The smoothing factor is optimized using k-fold cross validation. 
Therefore I added a function to plot the mse on the test data vs the smoothing factors.

Let me know what you think.

### DIFF
--- a/lumicks/pylake/piezo_tracking/baseline.py
+++ b/lumicks/pylake/piezo_tracking/baseline.py
@@ -1,0 +1,80 @@
+import numpy as np
+import matplotlib.pyplot as plt
+from lumicks.pylake.channel import Slice, Continuous
+
+
+class ForceBaseLine:
+    def __init__(self, model, trap_data, force):
+        """Force baseline
+
+        Parameters
+        ----------
+        model : callable
+            Model which returns the baseline at specified points.
+        trap_data : lumicks.pylake.Slice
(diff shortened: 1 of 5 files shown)


lumicks/pylake/piezo_tracking/baseline.py
```diff
@@ -1,6 +1,90 @@
...
+
+    return x[m][ind], force.data[m][ind]
+
+
+def optimize_smoothing_factor(
+    trap_position,
+    force,
+    smoothing_factors,
+    n_repeats,
+    plot_smoothingfactor_mse,
```
JoepVanlier, line 32: Considering that the function has `smoothing_factor` in the name, I reckon `plot_mse` would be sufficient.

JoepVanlier (changes requested):
So, I've had a first look at this and come to the same conclusion as you, that the smoothing factor based optimization doesn't really work that well unless you also downsample. This suggests that we should provide a good default for this.

Adding a package involves adding it to `setup.py`. For conda, it might be a bit trickier since I don't see this package on conda-forge. I'd have to look into it.

To muddy the waters a bit more, another class of algorithms that's probably worth exploring for this purpose is LO(W)ESS regression.

PR #300, opened 2022-05-05
Add GenBank file parsing

rpauszek:
**Why this PR?**
We want to work toward automated calibration of kymographs from microns to genomic position. This PR adds a parsing function to extract feature annotations from a GenBank `.gb` file, which will store the locations of landmark fluorophores that can be used for calibration.


lumicks/pylake/kymotracker/detail/sequence.py
```diff
@@ -0,0 +1,62 @@
...
+    ----------
+    filename: str
+        Path to the file to be parsed.
+    """
+
+    with open(filename, "r") as gb_file:
+        record = gb_file.read()
+
+    # Extract the features table; located between the 'FEATURES' and 'ORIGIN' headers
+    feature_table = re.search(r"FEATURES.+\n([\s\S]*)ORIGIN", record).group(1)
```
JoepVanlier, line 49: Would be nice to have some error handling in case the file does not meet the specifications (i.e. if fields are missing or custom edits end up invalidating the format). I guess if this one comes up blank (no matches), you could flag that?

lumicks/pylake/kymotracker/detail/sequence.py
```diff
@@ -0,0 +1,62 @@
...
+
+    # Split the table into a list of repeating pattern: feature name, location, tags
+    # Discard the first item in the list; from the `re` docs:
+    #   If there are capturing groups in the separator and it matches at the start of the string,
+    #   the result will start with an empty string.
+    table_parts = re.split(r" {5}(\w+) +([\d\.]+)\n", feature_table)[1:]
+    features = [table_parts[j : j + 3] for j in range(0, len(table_parts), 3)]
+
+    # Parse tags string into individual items
+    regex_tag = re.compile(r" {21}/(\w+)=([\w\" \:]+)", re.MULTILINE)
```
JoepVanlier, line 59: I don't think this allows for special characters in the labels, which when I tried quickly with snapgene is allowed in the spec. For instance a label with `Test (yesyes)`  seems to truncate parsing it like this.

JoepVanlier (changes requested):
Some minor suggestions.

JoepVanlier (approved):
LGTM! 👍

PR #301, opened 2022-05-16
force_calib: make sure negative frequencies produce valid power spectral densities

JoepVanlier:
**Why this PR?**
While working with a deliberately aliased spectrum I noticed that for negative frequencies the hydrodynamically correct power spectral density fails.

The reason for this is that when implementing it, we always assumed the frequency would be positive (and thereby avoided using complex arithmetic). When simulating aliasing however, this is not the case anymore (since you have to evaluate negative frequencies as well as they wrap around).

To avoid future surprises like this, we should just fix this and deal with it properly.

Paper with equations here: https://arxiv.org/pdf/physics/0603037 (D4 - D6)


lumicks/pylake/force_calibration/detail/hydrodynamics.py
```diff
@@ -83,13 +83,14 @@ def calculate_complex_drag(f, gamma0, rho_sample, bead_radius, distance_to_surfa
     nu = gamma0 / (6.0 * np.pi * rho_sample * bead_radius)
     frequency_nu = nu / (np.pi * bead_radius**2)
 
-    freq_ratio = f / frequency_nu
+    freq_ratio = (np.asarray(f) / frequency_nu).astype(complex)
```
dean0x7d, line 86: Why do you need `np.asarray()`?

JoepVanlier: When calculating active calibration, this function is evaluated with a single value for `f`. This is typically just a regular `float` which doesn't have an `.astype` function. I could convert it to an array upstream if need be and demand this to be an `ndarray` on input, but I figured this would make this function slightly easier to use anyhow, since you don't have to think about whether you pass a `float` or a 1-element array now.

dean0x7d: Can't you simply do `complex(f / frequency_nu)` if you need the type to be `complex`?

JoepVanlier: That doesn't work for numpy arrays since it expects the output to be a scalar (`TypeError: only length-1 arrays can be converted to Python scalars`). This function needs to be able to handle both.

Could go with `freq_ratio = np.asarray(f / frequency_nu, dtype=complex)` or `scipy.complex_(f / frequency_nu)` instead. The former has my slight preference since it seems a bit funky to import scipy just for `complex_`.

dean0x7d: I see. I guess you just need to update the docstring then. Because it currently indicates `f : float` so I was assuming it was a single value.

dean0x7d (approved):
Cool! Seems like a natural way to extend the range. Just see my comment about about updating the docstring.

lumicks/pylake/force_calibration/detail/hydrodynamics.py
JoepVanlier: Ah yes, I see. The docstring was indeed incorrect. Fixing.

PR #302, opened 2022-05-16
force_calibration: prevent some precision loss when computing hydrodynamically correct power spectral density

JoepVanlier:
**Why this PR?**
In the hydro power spectral density we subtract two values with widely varying magnitudes while this is not necessary. You can start noticing the effects of this by an apparent sawtooth pattern at high frequencies. This PR fixes that.

Note that this likely did not affect any users yet, since for regular 78 kHz systems, the frequency at which this happens happens well above Nyquist.

image

### DIFF
--- a/lumicks/pylake/force_calibration/detail/hydrodynamics.py
+++ b/lumicks/pylake/force_calibration/detail/hydrodynamics.py
@@ -146,7 +146,8 @@ def passive_power_spectrum_model_hydro(
         f, gamma0, rho_sample, bead_radius, distance_to_surface
     )
     frequency_m = calculate_dissipation_frequency(gamma0, bead_radius, rho_bead)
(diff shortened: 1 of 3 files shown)


lumicks/pylake/force_calibration/detail/hydrodynamics.py
```diff
@@ -146,7 +146,8 @@ def passive_power_spectrum_model_hydro(
         f, gamma0, rho_sample, bead_radius, distance_to_surface
     )
     frequency_m = calculate_dissipation_frequency(gamma0, bead_radius, rho_bead)
-    denominator = (fc + f * im_drag - f**2 / frequency_m) ** 2 + (f * re_drag) ** 2
+    denominator = (fc + f * (im_drag - f / frequency_m)) ** 2 + (f * re_drag) ** 2
```
dean0x7d, line 149: Very subtle. Nice catch!

dean0x7d (approved):

PR #303, opened 2022-05-17
force_calibration: add private API for taking aliasing into account

JoepVanlier:
**Why this PR?**
Most systems have aliasing suppression in hardware, but it is still good to have the option to take it into account in the model for alternative detection methods.

image

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -11,6 +11,7 @@ from .detail.power_models import (
     passive_power_spectrum_model,
     sphere_friction_coefficient,
     theoretical_driving_power_lorentzian,
+    alias_spectrum,
 )
 from .detail.hydrodynamics import (
     passive_power_spectrum_model_hydro,
@@ -20,6 +21,7 @@ from .power_spectrum_calibration import CalibrationParameter
 
 from dataclasses import dataclass
 from typing import Callable
+from copy import copy
 
 
 def diode_params_from_voltage(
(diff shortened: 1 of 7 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -342,10 +343,27 @@ def __init__(
...
+        Parameters
+        ----------
+        sample_rate : int
+            Sample rate in Hz
+        num_aliases : int
+            Number of aliases to take into account.
+        """
+        self._calculate_power_spectral_density = alias_spectrum(
+            self._calculate_power_spectral_density, sample_rate, num_aliases
+        )
```
rpauszek, line 365: I'm a bit wary of reassigning a method like this, it seems a bit to me like introducing state to the object. I know this isn't public functionality, but I'm assuming it will be in the future? What is the intended usage?

I hesitate to suggest it, but it seems like this is an appropriate instance to use a factory.  In the `__init__()` you can just define
```
self._calculate_power_spectral_density = (
    self._passive_power_spectrum_model(f, fc, diffusion_constant) * self._filter(f, *filter_params)
)
```
and from this method here return a copy of the instance with the factory reset to the aliased version (as we do for other things).

Or if it makes more sense, have `alias=bool` as an optional argument and assign the proper function from the get go.

rpauszek (changes requested):
Nice feature, especially looking at the results you have in the intro description. My only concern is adding state to the class, let's see if you agree

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: I've added an explicit copy and return a new instance instead now. I don't think we need to move the construction of that function to the `__init__` for that and I'd rather keep this as a contained function if you're ok with it.

If this ends up being a full feature one day, then I would suggest moving it to the constructor, but not yet.

Note that some of the private ways in which this `CalibrationModel` are used are stateful. Note `_set_drag` for instance. We should probably move away from this, and it'd be relatively easy to do so, but this needs a little alignment with BL.

rpauszek: ah sure that seems fine

rpauszek (approved):
good to go

PR #304, opened 2022-05-18
stack: allow multidimensional indexing with `CorrelatedStack`

JoepVanlier:
**Why this PR?**
This PR adds some convenience when it comes to cropping and slicing stacks.

Notably, it allows one to directly use multidimensional indexing to select frames and crop the image. It also adds the ability to use negative indices to indicate "from end".

I also added a `.shape` property to quickly check the shape of a `CorrelatedStack`. Previously this required either collecting these pieces of information from various sources, or grabbing the entire stack's contents and calling `.shape` on it. This was very tedious and slow.

image

### DIFF
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -43,8 +43,28 @@ class CorrelatedStack:
         self.start_idx = 0
         self.stop_idx = self.src.num_frames
 
+    def _handle_cropping(self, item):
+        """Crop the stack based on tuple of slices"""
+
+        def interpret_crop(item):
+            if isinstance(item, slice):
+                if item.step is not None:
+                    raise IndexError("Slice steps are not supported when indexing")
+                return item.start, item.stop
+            else:
+                return item, item + 1
+
+        if len(item) > 3:
+            raise IndexError("Only three indices are accepted when slicing CorrelatedStacks.")
+
+        rows = interpret_crop(item[1])
+        columns = interpret_crop(item[2]) if len(item) >= 3 else [None, None]
+        return self.src.with_roi(np.array([columns, rows]).flatten()), item[0]
+
     def __getitem__(self, item):
         """All indexing is in frames"""
+        src, item = self._handle_cropping(item) if isinstance(item, tuple) else (self.src, item)
+
         if isinstance(item, slice):
             if item.step is not None:
                 raise IndexError("Slice steps are not supported")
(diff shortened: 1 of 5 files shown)


lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -220,6 +220,19 @@ def test_stack_roi():
...
 
+def test_roi_defaults():
+    first_page = np.arange(60).reshape((6, 10))
+    data = np.stack([first_page + (j * 60) for j in range(3)], axis=2)
+    stack_0 = TiffStack(MockTiffFile([data], times=make_frame_times(1)), align_requested=False)
+
+    np.testing.assert_equal(stack_0.with_roi([None, 7, 3, 6]).get_frame(0).data, data[3:6, :7])
+    np.testing.assert_equal(stack_0.with_roi([1, None, 3, 6]).get_frame(0).data, data[3:6, 1:])
+    np.testing.assert_equal(stack_0.with_roi([1, 7, None, 6]).get_frame(0).data, data[:6, 1:7])
+    np.testing.assert_equal(stack_0.with_roi([1, 7, 3, None]).get_frame(0).data, data[3:, 1:7, :])
```
mikhas, line 231: This line looks odd; why is it `data[3:, 1:7, :]` here?

mikhas (commented):
I ignored most of the tests, although they look very exhaustive to me.

lumicks/pylake/correlated_stack.py
```diff
@@ -45,6 +45,25 @@ def __init__(self, image_name, align=True):
...
+            else:
+                return item, item + 1
+
+        src = self.src
+        if isinstance(item, tuple):
+            if len(item) > 3:
+                raise IndexError("Only three indices are accepted when slicing CorrelatedStacks.")
+
+            first_axis = interpret_crop(item[1])
+            second_axis = interpret_crop(item[2]) if len(item) >= 3 else [None, None]
```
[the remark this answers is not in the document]
JoepVanlier: Good point. Changed.

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
JoepVanlier: Oversight actually. Should be `data[3:, 1:7]`, the last index does nothing here.

PR #306, opened 2022-05-23
docs: kernelspec instructions

JoepVanlier:
**Why this PR?**
In a previous version of the installation instructions, we recommended windows users adding the kernel to the `kernelspec` manually, but without a specific name (which defaulted it to `python3`). This was a mistake, because it overrides the autodiscovery of the current environment's kernel that normally happens. As a result, some people are having issue with kernel discovery when installing Pylake (and likely Python in general).

Because of this, these amendments to our installation instructions.

I expanded a bit on the instructions pertaining to the `kernelspec` list (in the FAQ).

In addition, I also explicitly reference the FAQ (where these instructions can be found) because people often don't end up reading the FAQ when they run into issues after installation.

Docs build here: https://lumicks-pylake.readthedocs.io/en/install_instructions_mk2/install.html

Long story short, there's two options for them to fix it. Either they register the kernel for their newly made environment (making it discoverable no matter what environment is active) or they delete the bad `python3` listing we added with the previous bad instruction (restoring the regular automatic discovery).

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -58,7 +58,6 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
     jupyter notebook
 
 Note that if you are used to starting Jupyter notebook from the Anaconda Navigator, you will have to set the environment to `pylake` for it to have access to the environment where you just installed Pylake.
-
 You can do this in the drop down menu that normally defaults to the environment `base`.
 
 If you are used to starting Jupyter from the Anaconda prompt, then remember to activate the correct environment (pylake) prior to starting the notebook.
@@ -67,6 +66,7 @@ You can activate this environment by calling::
     conda activate pylake
 
 That is all. Check out the :doc:`Tutorial </tutorial/index>` for some example code and Jupyter notebooks to get started.
+If importing the package fails please refer to the `Frequently asked questions`_ for troubleshooting options.
 
 
 .. rubric:: Linux
@@ -109,6 +109,7 @@ You can do this in the drop down menu that normally defaults to the environment 
 If you are used to starting Jupyter notebook from the terminal, then remember to activate the correct environment (pylake) prior to starting the notebook.
 
 That is all. Check out the :doc:`Tutorial </tutorial/index>` for some example code and Jupyter notebooks to get started.
+If importing the package fails please refer to the `Frequently asked questions`_ for troubleshooting options.
 
 
 .. rubric:: macOS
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -279,6 +281,69 @@ If you have already activated the correct environment, but you still do not see
...
+However, if an explicit `kernelspec` with that name has been created in the all-users or per-user kernel registry, Jupyter no longer performs auto-detection of the IPython kernel in the current conda environment.
+That means you are no longer able to start a Jupyter kernel from the currently active environment without explicitly installing it.
+
+*Can I just revert back to detecting which kernel to use based on the environment instead?*
+
+Yes. Alternatively, one can remove the kernelspec that is causing the issue, resulting in `Anaconda` reverting back to the default behaviour of using the kernel for the active environment::
+
+    jupyter kernelspec uninstall envname
+
+Where `envname` needs to be the name of the registered kernel that is being loaded. Usually this will be `python3`.
```
aafkevandenberg, line 324: Very clear! I finally start to understand what went wrong here.

docs/install.rst
```diff
@@ -279,6 +281,69 @@ If you have already activated the correct environment, but you still do not see
...
+
+**(Windows/Anaconda only) When I try to import pylake, I receive an ImportError: DLL load failed while importing win32api**
```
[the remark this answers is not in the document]
JoepVanlier: This pertains to a different error. This is an error with the ` win32api` stuff not being installed correctly, whereas the previous one is an issue with the kernelspec. Unfortunately, both can lead to a `DLL load` error (just a different reason for it).

docs/install.rst
JoepVanlier: :)

tiagobonetti:
I used the documentation to fix Lakeview not using the correct Jupyter kernel on Emma's Laptop.
10/10: Would read the documentation again.

docs/install.rst
```diff
@@ -279,6 +281,69 @@ If you have already activated the correct environment, but you still do not see
...
+
+If the output of this command lists any paths containing `AppData\\Roaming\\jupyter\\kernels` or `C:\\ProgramData\\jupyter\\kernels` (Windows); `.local/share/jupyter/kernels` or `/usr/share/jupyter/kernels` or `/usr/local/share/jupyter/kernels` (Mac/Linux); or `/Library/Jupyter/kernels` (Mac); then invoke::
+
+    python -m ipykernel install --user --name=envname
+
+Where `envname` should be replaced with the name of the environment you are using.
+
+Now restart the Jupyter Notebook, and make sure you open your Notebook using the `pylake` kernel that's now available in the list.
+
+*Why does this work?*
```
mikhas, line 306: This is a great approach, to also give the technical explanation in the FAQ, but to clearly mark it as such and to keep it separate from the answer.

mikhas (approved):
👍🏽

PR #309, opened 2022-05-31
kymotracker: better input validation

JoepVanlier:
**Why this PR?**
There are a few cases in which the kymotracking algorithms can fail with unclear error messages. Since this is user facing API, would be good to provide sensible error messages if an input is known to be invalid for that particular algorithm.

Especially the refinement one was a bit dodgy, since if you specified a line width of zero, centroid refinement would just silently do nothing.


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -79,10 +79,10 @@ def track_greedy(
     channel : str
         Kymograph channel.
     line_width : float
-        Expected line width in physical units.
+        Expected line width in physical units. Should be larger than zero.
```
rpauszek, line 82:
```suggestion
        Expected line width in physical units. Must be larger than zero.
```

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -79,10 +79,10 @@ def track_greedy(
...
     pixel_threshold : float
         Intensity threshold for the pixels. Local maxima above this intensity level will be
-        designated as a line origin.
+        designated as a line origin. Should be larger than zero.
```
rpauszek, line 85:
```suggestion
        designated as a line origin. Must be larger than zero.
```

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -99,7 +99,8 @@ def track_greedy(
         whether a peak in the next frame will be connected to this one. Increasing this value will
         make the algorithm allow more positional variation over time. If a particle disappears for
         a few frames and then reappears, points before and after the interval where the particle
-        was not visible are more likely to be connected with a higher diffusion setting.
+        was not visible are more likely to be connected with a higher diffusion setting. Should be
+        positive.
```
rpauszek, line 103:
```suggestion
        was not visible are more likely to be connected with a higher diffusion setting. Must be
        equal to or greater than zero.
```

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -202,7 +213,7 @@ def track_lines(
     channel : str
         Kymograph channel.
     line_width : float
-        Expected line width in physical units.
+        Expected line width in physical units. Should be larger than zero.
```
rpauszek, line 216:
```suggestion
        Expected line width in physical units. Must be larger than zero.
```

rpauszek (approved):
:shipit: just a couple small wording adjustments

PR #312, opened 2022-06-09
scan: include extra sample timestamp ranges

JoepVanlier:
**Why this PR?**
Currently `Scan.frame_timestamp_ranges(exclude=True)` returns the first and last DAQ sample of the scan. However, if you slice using these timestamps, you will miss one sample from the end. After this PR you will get the DAQ sample one beyond the end of the frame such that slicing behaves as expected.


lumicks/pylake/scan.py
```diff
@@ -87,12 +87,13 @@ def frame_timestamp_ranges(self, exclude=True):
         """
         ts_min = self._timestamps("timestamps", reduce=np.min)
         ts_max = self._timestamps("timestamps", reduce=np.max)
+        delta_ts = int(1e9 / self.infowave.sample_rate)  # We want the sample beyond the end
```
mikhas, line 90: Need to update the docstring. Right now it's misleading.

mikhas (approved):
👍🏽

lumicks/pylake/scan.py
mikhas: The example is excellent, ship it!

PR #313, opened 2022-06-16
msd: add uncertainty estimate and estimation parameters to `msd` return argument

JoepVanlier:
**Why this PR?**
This PR addresses two user requests:
1. It adds an uncertainty estimate for the diffusion constant.
2. It exposes the number of lags used to estimate the diffusion constant.

Note that I added new API since the function now returns a dataclass. I've decided to do this because grouping this information together is useful and more extensible without breaking the API again.  Note that another PR for a second method is incoming. Giving it a `__float__` dunder method allows us to still use it conveniently with `numpy`, i.e.:
```python
np.array([kymoline.estimate_diffusion() for kymoline in kymolines], dtype=float)
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
 * Added a few performance benchmarks.
 * Added offline piezo tracking functionality (documentation pending).
 * Allow cropping `CorrelatedStack` using multidimensional indexing, i.e. `stack[start_frame : end_frame, start_row : end_row, start_column : end_column]`. See [Correlated stacks](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/correlatedstacks.html#correlated-stacks) for more information.
+* Added `KymoLine.estimate_diffusion()` which provides additional information regarding the diffusion constant estimation. This dataclass can be converted to floating point to get just the estimate, but also contains the number of points used to compute the estimate, and the number of lags used in the analysis. In addition to that, it provides a `std_err` field which reports the standard error for the estimate.
 
 #### Bug fixes
 
@@ -19,6 +20,10 @@
 * Perform better input validation on `refine_lines_centroid`. Line width must now be at least one pixel. Previously, negative values produced a cryptic error message, while a line width smaller than one pixel would silently result in no refinement taking place.
 * Fixed bug in force calibration convenience function where setting `fixed_alpha` or `fixed_diode` to zero resulted in those parameters still being fitted.  After this change, setting `fixed_alpha` to zero will result in the diode model having a fixed `alpha` of zero, whereas setting `f_diode` to zero raises an exception.
 * Include one extra sample when requesting frame timestamp ranges from a scan (`Scan.frame_timestamp_ranges(exclude=True)`). Previously, when slicing using these timestamps, you would omit the last sample of the scan. Now this sample will be included.
+
+#### Deprecations
+
+* `KymoLine.estimate_diffusion_ols()` is now deprecated; use `KymoLine.estimate_diffusion(method="ols")` instead. This new method returns the same estimate of the diffusion coefficient as before but includes additional information about the fit.
 
 ## v0.12.0 | 2022-04-21
 
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -1,3 +1,4 @@
+from itertools import product
 import numpy as np
 import warnings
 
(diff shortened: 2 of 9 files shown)


lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -37,6 +63,90 @@ def calculate_msd(frame_idx, position, max_lag):
...
+    denominator = (n - i + 1.0) * (n - j + 1.0)
+    term2 = (min_ij**2 - min_ij**4) / denominator
+    heaviside = (i + j - n - 2) >= 0
+    term3 = heaviside * ((n + 1.0 - i - j) ** 4 - (n + 1.0 - i - j) ** 2) / denominator
+
+    # Covariance matrix if there was no localization uncertainty (Eq 8b)
+    base_covariance = (slope**2 / 3) * (term1 + term2 + term3)
+
+    # Intercept corresponds to a^2 in the paper, slope refers to sigma^2 in the paper
+    term4_nominator = intercept**2 * (1.0 + (i == j)) + 4 * intercept * slope * min_ij
```
rpauszek, line 102: numerator

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -37,6 +63,90 @@ def calculate_msd(frame_idx, position, max_lag):
...
+    inv_denominator = 1.0 / (num_lags * beta - alpha**2)
+    intercept = (beta * gamma - alpha * delta) * inv_denominator
+    slope = (num_lags * delta - alpha * gamma) * inv_denominator
+
+    # Determine variance on slope estimator (Eq. A1a from Appendix A of [2]).
+    covariance_matrix = _msd_diffusion_covariance(num_lags, num_points, intercept, slope)
+
+    i = np.tile(lag_idx, (num_lags, 1))
+    j = i.T
+    nominator = (i * num_lags - alpha) * (j * num_lags - alpha) * covariance_matrix
```
rpauszek, line 143: numerator

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -319,8 +320,8 @@ def plot_msd(self, max_lag=None, **kwargs):
         plt.xlabel("Lag time [s]")
         plt.ylabel(f"Mean Squared Displacement [{self._kymo._calibration.unit_label}$^2$]")
 
-    def estimate_diffusion_ols(self, max_lag=None):
-        """Perform an unweighted fit to the MSD estimates to obtain a diffusion constant.
+    def estimate_diffusion(self, max_lag=None):
```
rpauszek, line 323: as in the above comment, I think it'd be good to add the `method` argument from your next PR so that we don't immediately break this feature after release

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -345,19 +346,44 @@ def estimate_diffusion_ols(self, max_lag=None):
...
+        self-diffusion coefficients from molecular dynamics simulations. The Journal of Chemical
+        Physics, 153(2), 024116.
 
         Parameters
         ----------
         max_lag : int (optional)
-            Number of lags to include. When omitted, the method uses an optimal number of lags
-            as determined by [1].
+            Number of lags to include. When omitted, the method will choose an appropriate number
+            of lags to use.
```
rpauszek, line 357: kinda makes it sound like `pylake` is sentient 😆 maybe `the method determines an appropriate number of lags` or `the method calculates an appropriate number of lags`, I'm not sure

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -345,19 +346,44 @@ def estimate_diffusion_ols(self, max_lag=None):
...
         frame_idx, positions = np.array(self.time_idx, dtype=int), np.array(self.position)
         max_lag = max_lag if max_lag else determine_optimal_points(frame_idx, positions)[0]
         return estimate_diffusion_constant_simple(
             frame_idx, positions, self._line_time_seconds, max_lag
         )
 
+    @deprecated(
+        reason=(
+            "Diffusion constant estimation was moved to the function `estimate_diffusion` "
+            "which provides additional metadata."
```
rpauszek, line 368: Maybe just some slight re-wording
```suggestion
            "This method is replaced by `Kymoline.estimate_diffusion()`  to allow more flexibility"
            "in the choice of optimization algorithms and provides additional metadata."
```

rpauszek (changes requested):
Very nice, I like this API much better. The only thing, as mentioned in a couple comments, is I think we should add the required `method` argument straight away here since it will be implemented immediately after this anyway. Perhaps even a note in the changelog that other optimization algorithms will be added in the near future, so it doesn't seem like we're adding a random input argument that only has one choice

changelog.md
```diff
@@ -20,6 +21,10 @@
 * Fixed bug in force calibration convenience function where setting `fixed_alpha` or `fixed_diode` to zero resulted in those parameters still being fitted.  After this change, setting `fixed_alpha` to zero will result in the diode model having a fixed `alpha` of zero, whereas setting `f_diode` to zero raises an exception.
 * Include one extra sample when requesting frame timestamp ranges from a scan (`Scan.frame_timestamp_ranges(exclude=True)`). Previously, when slicing using these timestamps, you would omit the last sample of the scan. Now this sample will be included.
 
+#### Deprecations
+
+* `Kymoline.estimate_diffusion_ols()` is now deprecated.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -1,5 +1,31 @@
...
+    num_lags : int
+        Number of lags used to compute this estimate.
+    num_points : int
+        Number of points used to compute this estimate.
+    """
+
+    value: float
+    std_err: float
+    num_lags: int
+    num_points: int
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, good point. Didn't occur to me, but it indeed would immediately be a breaking change yeah. I'll add it now.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Oh man, that word has been a total blind spot for me. I've consistently been calling that thing nominator 😊

lumicks/pylake/kymotracker/kymoline.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -1,5 +1,34 @@
 import numpy as np
 import warnings
+from dataclasses import dataclass
+
+
+@dataclass
+class DiffusionEstimate:
```
rpauszek, line 7: something that just came to mind, but we might want to make this frozen from the get-go

rpauszek (approved):
LGTM, just a final thought about making the returned dataclass frozen if you agree it makes sense

PR #315, opened 2022-06-17
Gls estimator

JoepVanlier:
**Why this PR?**
The estimated mean squared differences which are used to estimate the diffusion constant are highly correlated. This comes from computing them from a single trace (every datapoint is used several times).

This is not a problem, even when using completely unweighted OLS as long as the number of lags included in the estimation is optimal [1]. As a default for the number of lags to include, we have a method to estimate this number based on [1]. This works, but there is a more robust alternative. This more robust alternative is called `generalized least squares` and it takes the full covariance matrix into account, leading to a more appropriate weighting of the data.

For example, look at the diffusion estimates below:
image
_Diffusion constants computed from 1000 simulated traces. The dashed line indicates the true parameter value._

What I did here was deliberately specify `max_lag` too high for the OLS estimator. As a result, we get a much broader distribution than necessary, which is clearly undesirable.

When we take into account the correct number of lags, we see that the GLS and OLS actually agree quite well.

image
_Diffusion constants computed from 1000 simulated traces. The dashed line indicates the true parameter value._

Docs build here: https://lumicks-pylake.readthedocs.io/en/gls_estimator/tutorial/kymotracking.html#studying-diffusion-processes


changelog.md
```diff
@@ -8,7 +8,8 @@
...
-* Added `KymoLine.estimate_diffusion()` which provides additional information regarding the diffusion constant estimation. This dataclass can be converted to floating point to get just the estimate, but also contains the number of points used to compute the estimate, and the number of lags used in the analysis. In addition to that, it provides a `std_err` field which reports the standard error for the estimate.
+* Added `Kymoline.estimate_diffusion()` which provides additional information regarding the diffusion constant estimation. This dataclass can be converted to floating point to get just the estimate, but also contains the number of points used to compute the estimate, and the number of lags used in the analysis. In addition to that, it provides a `std_err` field which reports the standard error for the estimate.
+* Added generalized least squares method as method option for `Kymoline.estimate_diffusion()`. Please refer to the [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more information.
```
rpauszek, line 12: nitpick, some subtle formatting I think looks better

```suggestion
* Added generalized least squares method as `method` option for `Kymoline.estimate_diffusion()`. Please refer to the [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more information.
```

docs/api.rst
```diff
@@ -72,6 +72,7 @@ Kymotracking
     :toctree: _api
 
     kymotracker.kymoline.KymoLine
+    kymotracker.kymoline.KymoLine.estimate_diffusion
```
rpauszek, line 75: I think this should already be generated from the line above

docs/zrefs.rst
```diff
@@ -28,6 +28,12 @@ The line algorithm was based on sections 1, 2 and 3 from :cite:`steger1998unbias
 geometric considerations. The greedy algorithm was based on two papers. It initially detects feature points based on
 :cite:`sbalzarini2005feature`, followed by line tracing inspired by :cite:`mangeol2016kymographclear`.
 
+Diffusion constant estimation
+-----------------------------
+
+We implemented the unweighted ordinary least squares estimation method (including the optimal number of lag computation) from :cite:`michalet2012optimal`.
+The generalized least squares method was based on :cite:`bullerjahn2020optimal`
```
rpauszek, line 35:
```suggestion
The unweighted ordinary least squares estimation method (including the optimal number of lag computation) is implemented from :cite:`michalet2012optimal`.
The generalized least squares method is based on :cite:`bullerjahn2020optimal`
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -192,22 +286,46 @@ def estimate_diffusion_constant_simple(frame_idx, coordinate, time_step, max_lag
         Positional coordinates.
     time_step : float
         Time step between each frame.
-    max_lag : int
-        Maximum delay to include in the estimate (must be larger than 1).
+    max_lag : int (optional)
+        Number of lags to include. When omitted, the method will choose an appropriate number
+        of lags to use.
+        When the method chosen is "OLS" an optimal number of lags is estimated as determined by
```
rpauszek, line 292: `"ols"`

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -192,22 +286,46 @@ def estimate_diffusion_constant_simple(frame_idx, coordinate, time_step, max_lag
...
+        [1]. When the method is set to "gls" all lags are included.
     method : str
-        Should be "ols".
+        Valid options are "ols" and "gls'.
+
+        - 'ols' : Ordinary least squares [1]. Determines optimal number of lags.
+        - 'gls' : Generalized least squares [2]. Takes into account covariance matrix (slower).
```
rpauszek, line 298: double quotes maybe?

lumicks/pylake/kymotracker/kymoline.py
```diff
@@ -338,32 +338,50 @@ def estimate_diffusion(self, method, max_lag=None):
...
         frame_idx, positions = np.array(self.time_idx, dtype=int), np.array(self.position)
-        max_lag = max_lag if max_lag else determine_optimal_points(frame_idx, positions)[0]
+        max_lag = (
+            max_lag
+            if max_lag
+            else determine_optimal_points(frame_idx, positions)[0]
+            if method == "ols"
+            else len(frame_idx)
+        )
+
```
rpauszek, line 384: oof, I love a good ternary but this may be too much. consider seeing if the usual `if-else` block would be more readable. or at the very least maybe an extra set of `()` for the first `else`

rpauszek (approved):
awesome, really great work with the docstrings. just some very superficial nitpicky stuff, but looks good to go

docs/api.rst
JoepVanlier: I thought that as well, but if I omit it, it doesn't appear. I think it's different from the other objects since in the other cases the objects are explicitly imported in the init.

If I omit it, I get this
```
checking consistency... C:\Dev\Pylake\pylake\docs\_api\lumicks.pylake.kymotracker.kymoline.KymoLine.estimate_diffusion.rst: WARNING: document isn't included in any toctree
```
and the link goes dead.

PR #317, opened 2022-06-23
force_calibration: change default driving frequency to 38 Hz

JoepVanlier:
Does this count as a single character PR?


dean0x7d (approved):
gr8

rpauszek (approved):
goed gedaan

PR #318, opened 2022-06-23
Add `Kymo._line_timestamp_ranges()`

rpauszek:
**Why this PR?**
For kymos generated from camera movies, per-pixel timestamps don't make much sense; however it would still be nice to know the start/stop timestamps for the frame integration. Additionally, a method to determine the kymo line ranges makes it simple to downsample channel data to the frequency of the lines; the functionality is similar to `Scan.frame_timestamp_ranges()`. I've kept it private for now, but it may be better as a public method.

I've implemented this as (yet another) factory because camera-based kymos will need a different implementation. Additionally, I've removed `Kymo._downsample_channel()` since that functionality is now achievable in 2 lines with this addition.

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -28,6 +28,22 @@ def _default_line_time_factory(self: "Kymo"):
         )
 
 
+def _default_line_timestamp_ranges_factory(self: "Kymo", exclude: bool):
+    """Get start and stop timestamp of each line in the kymo."""
+
+    ts_min = self._timestamps("timestamps", reduce=np.min)[0]
+
+    if exclude:
+        # Take the max value of each line to account for unfinished final line
+        # and add one sample to have proper slicing
+        delta_ts = int(1e9 / self.infowave.sample_rate)
+        ts_max = self._timestamps("timestamps", reduce=np.max).max(axis=0) + delta_ts
+    else:
+        line_time = ts_min[1] - ts_min[0]
+        ts_max = ts_min + line_time
+    return [(t1, t2) for t1, t2 in zip(ts_min, ts_max)]
+
+
 class Kymo(ConfocalImage):
     """A Kymograph exported from Bluelake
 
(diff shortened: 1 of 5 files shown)


lumicks/pylake/kymo.py
```diff
@@ -28,6 +28,22 @@ def _default_line_time_factory(self: "Kymo"):
         )
 
 
+def _default_line_timestamp_ranges_factory(self: "Kymo", exclude: "bool"):
```
JoepVanlier, line 31:
```suggestion
def _default_line_timestamp_ranges_factory(self: "Kymo", exclude: bool):
```
No need for the quotes on bool.

lumicks/pylake/kymo.py
```diff
@@ -267,7 +285,10 @@ def set_aspect_ratio(axis, ar):
             warnings.warn(
                 RuntimeWarning("Using downsampled force since high frequency force is unavailable.")
             )
-        force = self._downsample_channel(channel, reduce=reduce)
+
+        time_ranges = self._line_timestamp_ranges(exclude=True)
+        force = channel.downsampled_over(time_ranges, reduce=reduce, where="center")
```
JoepVanlier, line 290: Much nicer when it's composed like this.

lumicks/pylake/kymo.py
```diff
@@ -117,6 +134,19 @@ def pixel_time_seconds(self):
         """Pixel dwell time in seconds"""
         return (self.timestamps[1, 0] - self.timestamps[0, 0]) / 1e9
 
+    def _line_timestamp_ranges(self, exclude=True):
```
JoepVanlier, line 137: Any reason you'd prefer keeping this private? It seems like something that would benefit users as well?

JoepVanlier (changes requested):
Looks great!

JoepVanlier (approved):
LGTM!

PR #320, opened 2022-06-24
kymo: add `kymo.flip()`

JoepVanlier:
**Why this PR?**
This feature was a specific request from DNA WF.

The requirement is to flip the kymograph.

image
image


lumicks/pylake/kymo.py
```diff
@@ -483,6 +483,16 @@ def pixelcount_factory(_):
         result._pixelcount_factory = pixelcount_factory
         return result
 
+    def flip(self):
```
mikhas, line 486: What happens if you flip multiple times?
What happens if I use multiple but different image factories in sequence, say,

1. downsample,
2. flip,
3. downsample.

JoepVanlier: Double flip is explicitly tested. Downsample should also be fine (I can easily add a specific test for it, but it seems that it should be fine considering the double flip tests the factories in sequence already).

PR #322, opened 2022-06-27
kymolinegroup: allow estimating diffusion at the group level

JoepVanlier:
**Why this PR?**
For convenience, it's nice to not have to write this list comprehension every time.


mikhas (approved):
Works as advertised, thanks!

PR #323, opened 2022-06-28
Change PositionCalibration to take per pixel calibration value

rpauszek:
**Why this PR?**
Currently, the calibration is based relative to microns; it makes more sense to set the calibration per pixel


JoepVanlier (approved):
Seems reasonable to me.

PR #324, opened 2022-06-29
ci: clean up two warnings from the error log

JoepVanlier:
**Why this PR?**
We should probably not use ill-posed data or API we deprecated in the tests. This cleans up two warnings that pop up during the tests.


mwilkowski (approved):
finally a pylake PR I can review <3

PR #328, opened 2022-06-29
arithmetic: forward labels appropriately

JoepVanlier:
**Why this PR?**
I noticed during documenting the piezo tracking functionality that currently, whenever we do arithmetic on a slice, we lose all the labels. This is not necessary. Quite a few operations (subtraction, addition, negation) will preserve the units and thereby the y label. Additionally, we can construct titles for these slices by using the existing title strings and adding the operator.

`(f.force1x - f.force2x).plot()` results in:

Before:
image

After:
image

While I'm not a huge fan of the brackets, they are needed to ensure correctness if users were to repeatedly apply operations on a slice.

### DIFF
--- a/lumicks/pylake/tests/test_channels/test_arithmetic.py
+++ b/lumicks/pylake/tests/test_channels/test_arithmetic.py
@@ -12,18 +12,19 @@ slice_continuous_2 = Slice(Continuous([2, 2, 2, 2, 2], start=start, dt=1), calib
 slice_timeseries_1 = Slice(TimeSeries([1, 2, 3, 4, 5], time_series), calibration=calibration)
 slice_timeseries_2 = Slice(TimeSeries([2, 2, 2, 2, 2], time_series), calibration=calibration)
 
-# Operators and whether they preserve force calibration when operated with a scalar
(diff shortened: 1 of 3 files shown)


lumicks/pylake/tests/test_channels/test_arithmetic.py
```diff
@@ -52,7 +53,7 @@ def test_operator(first_slice, second_slice, operation):
         # With slices, force calibration should never be carried over (since it's invalid)
         assert not getattr(first_slice, operation)(second_slice).calibration
 
-    for operator, _ in operators:
+    for operator, _, _, _ in operators:
```
rpauszek, line 56: could also just do `*_`  at the end or did you want to keep the specific placeholders for each argument in the unpack?

rpauszek (approved):
🚢

lumicks/pylake/tests/test_channels/test_arithmetic.py
JoepVanlier: Good point. Applying that everywhere in that file.

PR #329, opened 2022-06-29
Piezo tracking documentation

JoepVanlier:
**Why this PR?**
This PR makes the piezo tracking public and adds a small tutorial for the piezo tracking (a larger example based on the notebook we have on Harbor will likely still be added as well).

Docs build here: https://lumicks-pylake.readthedocs.io/en/piezo_docs/

### DIFF
--- a/docs/tutorial/piezotracking.rst
+++ b/docs/tutorial/piezotracking.rst
@@ -1,0 +1,162 @@
+.. warning::
+    This is alpha functionality. While usable, this has not yet been tested in a large
+    number of different scenarios. The API may also still be subject to change.
+
+Piezo tracking
+==============
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+In this tutorial, we will determine the high frequency distance (piezo distance) between the beads from the piezo mirror position of trap 1 and the corresponding force data.
+We will also show how to use the same reference curve to apply baseline correction for force signals in post processing.
+
+Trap Positional Calibration
+---------------------------
+
+The first step is to calibrate the high frequency trap position from the piezo mirror to the low frequency bead-to-bead distance measured by the camera.
+
+For this calibration, we require a dataset (acquired in the absence of a tether) in which trap 1 is moved over the entire distance range intended for the experiment.
+The exported data file must contain the trap 1 position and camera-based distance channels.
+
+*Note that this dataset should be acquired such that the bead tracking templates do not overlap. In case of overlap, make sure to first slice the position and distance data such that this region is not included.*
+
+Let's load this dataset and perform the distance calibration by invoking::
+
+    no_tether_data = lk.File("Data/piezo_tracking_no_tether.h5")
+
+    distance_calibration = lk.DistanceCalibration(
+        no_tether_data["Trap position"]["1X"], no_tether_data.distance1, degree=2
+    )
+
+This class performs a polynomial regression between the trap position and bead tracking distance.
+Note that this bead-to-bead distance already has the bead radius subtracted, therefore it reflects the surface-to-surface distance.
+We can plot what this curve looks like by invoking `.plot()` on it::
+
+    distance_calibration.plot()
+
+.. image:: figures/piezotracking/mirror_calibration.png
+
+We can also inspect the residual, to determine how well the calibration model describes the data.
+We can see that there is some error (the discrepancy does not scatter randomly around zero), but for this experiment, 15 nm error is within an acceptable range::
+
+    distance_calibration.plot_residual()
+
+.. image:: figures/piezotracking/mirror_calibration_residual.png
+
+Baseline correction
+-------------------
+
+Let's have a look at the force data::
+
+    plt.subplot(2, 1, 1)
+    no_tether_data.force1x.plot()
+    no_tether_data.force2x.plot()
+
+    plt.subplot(2, 1, 2)
+    no_tether_data.distance1.plot()
+    plt.tight_layout()
+
+.. image:: figures/piezotracking/nonzero_force.png
+
+It seems that the force was not zeroed at the start of this experiment, but we can correct this in post-processing.
+We want to use the force when the beads are far apart (no interaction between them) which corresponds the beginning of the data. In this case, we'll use a quarter of a second::
+
+    f1_offset = np.mean(no_tether_data.force1x[:"0.25s"].data)
+    f2_offset = np.mean(no_tether_data.force2x[:"0.25s"].data)
+
+Let's correct our force data before we use it::
+
+    background_force1x = no_tether_data.force1x - f1_offset
+    background_force2x = no_tether_data.force2x - f2_offset
+
+Even without a tether, the measured force can be non-zero when the beads are close together due to trap/trap interactions.
+We can characterize this baseline force so that it can be subtracted from our experiment force data.
+In principle, this step is optional, but it can greatly improve the accuracy of your force-distance curves.
+We can quickly determine a polynomial baseline for both traps by invoking::
+
+    baseline_1x = lk.ForceBaseLine.polynomial_baseline(
+        no_tether_data['Trap position']['1X'], background_force1x, degree=7, downsampling_factor=100
+    )
+    baseline_2x = lk.ForceBaseLine.polynomial_baseline(
+        no_tether_data['Trap position']['1X'], background_force2x, degree=7, downsampling_factor=100
+    )
+
+Similarly as before, we can plot the fits to verify that they describe the data well::
+
+    baseline_1x.plot(label="baseline f1x")
+    baseline_2x.plot(label="baseline f2x")
+    plt.legend()
+
+.. image:: figures/piezotracking/baseline.png
+
+And the residuals::
+
+    baseline_1x.plot_residual(label="baseline f1x")
+    baseline_2x.plot_residual(label="baseline f2x")
+    plt.legend(loc='lower right')
+
+.. image:: figures/piezotracking/baseline_residual.png
+
+The residuals should ideally look like noise distributed around zero.
+
+Calculating the force-dependent bead displacements
+--------------------------------------------------
+
+When a tether is present, it exerts a force on the beads resulting in a displacement of the beads from the trap centers.
+If there are only small excursions from the trap center, this displacement is assumed linear with respect to force (proportional to the trap stiffness :math:`\kappa`).
+Therefore, we can compute the bead displacement :math:`\delta x` directly from the force signal.
+
+.. math::
+
+    \delta x_1 = \frac{F_{1x}}{\kappa_{1x}}
+
+Thus the surface-to-surface distance between the beads can be computed by correcting the trap-based distance with the correlated force data and their respective trap stiffnesses as follows.
+
+.. math::
+
+    d_{piezo} = d_{no\_tether} - \frac{F_{1x}}{\kappa_{1x}} + \frac{F_{2x}}{\kappa_{2x}}
+
+Here :math:`d_{piezo}` is the piezo distance and :math:`d_{no\_tether}` is the calibrated surface-to-surface distance without the tether.
+:math:`F_{1x}` and :math:`F_{2x}` are the forces measured on the beads and :math:`\kappa_{1x}` and :math:`\kappa_{2x}` are the trap stiffness for each trap.
+
+To do this in Pylake, we set up the piezo distance calibration as follows::
+
+    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
+
+We now have all the calibrations we need to do piezo tracking on our experimental data.
+
+Calculating the Fd Curve
+------------------------
+
+First, we load the data acquired in the presence of a tether::
+
+    pulling_curve = lk.File("Data/piezo_tracking_tether.h5")
+
+And determine the piezo distance and corrected force::
+
+    tether_length, corrected_force_1x, corrected_force_2x = piezo_calibration.force_distance(
+        pulling_curve['Trap position']['1X'], pulling_curve.force1x - f1_offset, pulling_curve.force2x - f2_offset, downsampling_factor=100
+    )
+    force_data = - corrected_force_2x
+
+Here the downsampling factor determines how much the data is downsampled prior to piezo-tracking and baseline correction.
+
+Which we can then plot::
+
+    plt.scatter(tether_length.data, force_data.data, s=1)
+    plt.xlabel('Distance [$\mu$m]')
+    plt.ylabel('Force [pN]')
+
+.. image:: figures/piezotracking/piezotracking_result.png
+
+We can compare this to the camera-based distance and raw force curve and see a clear difference::
+
+    plt.scatter(tether_length.data, force_data.data, s=1, label="corrected")
+    plt.scatter(pulling_curve.distance1.data, - (pulling_curve.downsampled_force2x.data - f2_offset), s=1, label="raw")
+    plt.xlabel('Distance [$\mu$m]')
+    plt.ylabel('Force [pN]')
+    plt.legend()
+
+.. image:: figures/piezotracking/comparison.png
--- a/lumicks/pylake/piezo_tracking/baseline.py
+++ b/lumicks/pylake/piezo_tracking/baseline.py
@@ -68,7 +68,7 @@ class ForceBaseLine:
         plt.scatter(self._trap_data.data, self._force.data, s=2, **kwargs)
         plt.plot(self._trap_data.data, self._model(self._trap_data.data), "k")
         plt.xlabel("Mirror position")
-        plt.ylabel(self._force.labels["y"])
+        plt.ylabel(self._force.labels.get("y", "y"))
         plt.title("Force baseline")
 
     def plot_residual(self, **kwargs):
@@ -86,7 +86,7 @@ class ForceBaseLine:
             **kwargs,
         )
         plt.xlabel("Mirror position")
-        plt.ylabel(f"Residual {self._force.labels['y']}")
+        plt.ylabel(f"Residual {self._force.labels.get('y', 'y')}")
         plt.title("Fit residual")
 
     @classmethod
--- a/lumicks/pylake/piezo_tracking/baseline.py
+++ b/lumicks/pylake/piezo_tracking/baseline.py
@@ -4,6 +4,8 @@ from lumicks.pylake.channel import Slice, Continuous
 
 
 class ForceBaseLine:
+    """A force baseline as a function of trap position"""
+
     def __init__(self, model, trap_data, force):
         """Force baseline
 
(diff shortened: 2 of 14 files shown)


docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+In this tutorial, we will determine a high frequency distance (piezo distance) between the beads from the piezo mirror position of trap 1 and the measured force data.
+We will also show how to use the same reference curve to apply baseline correction for force signals in post processing.
+
+Positional Calibration
+----------------------
+
+The first step is to perform a positional calibration to calibrate the trap position to the distance measured by the camera.
+
+To do this calibration, we require a dataset acquired in the absence of a tether.
+In this dataset, trap 1 is moved over range, while the trap 1 position and camera-based distance is recorded.
```
rpauszek, line 21:
```suggestion
For this calibration, we require a dataset (acquired in the absence of a tether) in which trap 1 is moved over the entire distance range intended for the experiment. The exported data file must contain the trap 1 position and camera-based distance channels.
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Note that this dataset should be acquired such that the bead tracking templates do not overlap.
+In case of overlap, make sure to slice the position and distance data such that this region is not included.
+Let's load this dataset and perform the distance calibration by invoking::
```
rpauszek, line 25: I think here it's nice to add some emphasis to this caveat and to separate it from the final line, since they are two different thoughts

```suggestion
*Note that this dataset should be measured such that the bead tracking templates do not overlap. In case of overlap, make sure to first slice the position and distance data such that this region is not included.*

Let's load this dataset and perform the distance calibration by invoking::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+    no_tether_data = lk.File("no_tether.h5")
+
+    distance_calibration = lk.DistanceCalibration(
+        calibration_data["Trap position"]["1X"], calibration_data.distance1, degree=2
+    )
+
+What this class does is regress a polynomial between the trap position and distance determined by the bead tracking.
+Note that this bead-to-bead distance typically already has the bead diameter subtracted, therefore it reflects the surface-to-surface distance.
+We can plot what this curve looks like by invoking `.plot()` on it::
```
rpauszek, line 35: Also `"typically"` seems weird here. Is there actually the possibility to have data without the radius subtracted? If so, do we have a way to detect this or any other kind of mitigation strategy?

```suggestion
This class performs a polynomial regression between the trap position and  bead tracking distance. Note that this bead-to-bead distance typically already has the bead radius subtracted, therefore it reflects the surface-to-surface distance.
We can plot the calibration curve by invoking `.plot()` on it::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+    distance_calibration.plot()
+
+.. image:: figures/piezotracking/mirror_calibration.png
+
+We can also inspect the residual, to determine how well the calibration model describes the data::
+
+    distance_calibration.plot_residual()
+
+.. image:: figures/piezotracking/mirror_calibration_residual.png
```
rpauszek, line 45: oof, this is a rough residual plot. Is it typical to see this kind of structure in the residuals? If so, we should probably mention something about it because it's very very apparent that this is not a particularly great fit

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Baseline correction
+-------------------
+
+Even without a tether, the measured force can be non-zero when the beads are close together.
+The next step is to characterize this baseline force so that we can subtract it from our force data.
+In principle, this step is optional, but it can improve the accuracy of your force-distance curves.
+We can quickly determine a polynomial baseline for both traps by invoking::
```
rpauszek, line 53:
```suggestion
Even without a tether, the measured force can be non-zero when the beads are close together due to trap/trap interactions. We can characterize this baseline force so it can be subtracted from our experiment force data.
In principle, this step is optional, but it can greatly improve the accuracy of your force-distance curves.

We can quickly determine a polynomial baseline for both traps by invoking::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+    baseline_1x = lk.ForceBaseLine.polynomial_baseline(
+        no_tether_data['Trap position']['1X'], no_tether_data.force1x, degree=7, downsampling_factor=100
+    )
+    baseline_2x = lk.ForceBaseLine.polynomial_baseline(
+        no_tether_data['Trap position']['1X'], no_tether_data.force2x, degree=7, downsampling_factor=100
+    )
+
+Similarly as before, we can plot these baselines to verify that they describe the data well::
```
rpauszek, line 62: is `"fit"` more appropriate here?
```suggestion
Similarly as before, we can plot the fits to verify that they describe the data well::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+And the residuals::
+
+    baseline_1x.plot_residual(label="baseline f1x")
+    baseline_2x.plot_residual(label="baseline f2x")
+    plt.legend(loc='lower right')
+
+.. image:: figures/piezotracking/baseline_residual.png
+
+The residuals should be distributed around zero and should ideally look like noise.
```
rpauszek, line 78: I'm also wondering if `"should look like noise"` is too generic, but for now should be fine
```suggestion
The residuals should ideally look like noise distributed around zero.
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+==============
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+In this tutorial, we will determine a high frequency distance (piezo distance) between the beads from the piezo mirror position of trap 1 and the measured force data.
+We will also show how to use the same reference curve to apply baseline correction for force signals in post processing.
+
+Positional Calibration
```
rpauszek, line 15:
```suggestion
Trap Position Calibration
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+    baseline_1x.plot_residual(label="baseline f1x")
+    baseline_2x.plot_residual(label="baseline f2x")
+    plt.legend(loc='lower right')
+
+.. image:: figures/piezotracking/baseline_residual.png
+
+The residuals should be distributed around zero and should ideally look like noise.
+
+Calculating the piezo distance and corrected force
```
rpauszek, line 80: I think we can make this a separate little section explaining the math and add a final section for generating the Fd curve
```suggestion
Correcting for force-dependent bead displacements
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+--------------------------------------------------
+
+When a tether is present, it exerts a force on the beads.
+This force results in the displacement of the beads from the trap centers.
```
rpauszek, line 84:
```suggestion
When a tether is present, it exerts a force on the beads resulting in a displacement of the beads from the trap centers.
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+Assuming only small excursions from the trap center, we can assume this relation to be linear (proportional to the trap stiffness :math:`\kappa`).
```
rpauszek, line 85:
```suggestion
If there are only small excursions from the trap center, this displacement is assumed to be linear (proportional to the trap stiffness :math:`\kappa`).
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+Therefore, we can compute the bead displacement :math:`\delta x` directly from the force signal.
+
+.. math::
+
+    \delta x_1 = \frac{F_{1x}}{\kappa_{1x}}
+
+This means that the surface-to-surface distance can be computed by correcting the calibrated distance that was based on the trap, with the force signals and their respective stiffnesses as follows.
```
rpauszek, line 92:
```suggestion
Thus the actual surface-to-surface distance between the beads can be computed by correcting the calibrated trap-to-trap distance with the correlated force data and their respective trap stiffnesses as follows.
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+.. math::
+
+    d_{piezo} = d_{no\_tether}(x_{trap1}) - \frac{F_{1x}}{\kappa_{1x}} + \frac{F_{2x}}{\kappa_{2x}}
```
rpauszek, line 96: Aren't all of these w.r.t. `x_trap{n}` (except for the kappas)? I know including the function arguments is technically correct, but I think for clarity we can omit it as it's assumed to be understood. Or conversely if you think it's better to be explicit, they should be everywhere
```suggestion
    d_{piezo} = d_{no\_tether} - \frac{F_{1x}}{\kappa_{1x}} + \frac{F_{2x}}{\kappa_{2x}}
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+Here :math:`d_{piezo}` reflects the piezo distance, :math:`d_{no tether}` reflects the calibrated surface-to-surface distance without the tether with the trap position given by :math:`x_{trap1}`.
```
rpauszek, line 98: Also remove the fragment referring to `x_trap1` if you decide to go with my suggestion above
```suggestion
Here :math:`d_{piezo}` reflects the piezo distance, :math:`d_{no_tether}` reflects the calibrated surface-to-surface distance without the tether with the trap position given by :math:`x_{trap1}`.
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+We now have all the calibrations we need to do piezo tracking on our experimental data.
+
+To do this in Pylake, we set up the piezo distance calibration as follows::
+
+    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
+
+Next, we load the data acquired in the presence of a tether::
+
+    pulling_curve = lk.File("pulling_experiment.h5")
```
rpauszek, line 109: This might be nice split up a bit to separate all of the calibration stuff from the finally let's look at an experiment stuff. What do you think?

```suggestion
To do this in Pylake, we set up the piezo distance calibration as follows::

    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
    
We now have all the calibrations we need to perform piezo tracking on our experimental data.

Calculating the Fd Curve
---------------------------    

First, we load the experiment data (acquired in the presence of a tether)::

    pulling_curve = lk.File("pulling_experiment.h5")
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+
+We can compare this to the camera-based distance and raw force curve and see a clear difference::
+
+    plt.scatter(tether_length.data, force_data.data, s=1, label="corrected")
+    plt.scatter(pulling_curve.distance1.data, - (pulling_curve.downsampled_force2x.data - f2_offset), s=1, color='r', label="raw")
+    plt.xlabel('Distance [$\mu$m]')
+    plt.ylabel('Force [pN]')
+    plt.legend()
+
+.. image:: figures/piezotracking/comparison.png
```
rpauszek, line 136: any reason the second curve here is red? seems a bit odd since all of the other plots use the standard blue, orange, ... mpl color cycle

rpauszek (approved):
Very nice stuff, I especially like the figures .... makes it very clear why you need to go through this whole process if you want good data.

There's a bunch of suggestions for some minor wording changes, feel free to pick and choose any that sound good to you

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,136 @@
...
+    number of different scenarios. The API may also still be subject to change.
+
+Piezo tracking
+==============
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+In this tutorial, we will determine a high frequency distance (piezo distance) between the beads from the piezo mirror position of trap 1 and the measured force data.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/piezotracking.rst
JoepVanlier: 👍

docs/tutorial/piezotracking.rst
JoepVanlier: Dropped typically indeed.

docs/tutorial/piezotracking.rst
JoepVanlier: Yeah, I agree. I think we should find a better dataset. I suspect something may have gone wrong with the template selection here. The template error for bead 1 also seems to depend on the mirror position.

docs/tutorial/piezotracking.rst
JoepVanlier: I'd prefer calculating, since we're not really correcting anything. I do agree with the rest of the suggestion though.

docs/tutorial/piezotracking.rst
JoepVanlier: Unfortunately, it's not trap-to-trap distance, since BL already subtracted the avg bead diameter. How about:
```Thus the surface-to-surface distance between the beads can be computed by correcting the trap-based distance with the correlated force data and their respective trap stiffnesses as follows.```

File: lumicks/pylake/piezo_tracking/piezo_tracking.py [hunk 18]
```diff
@@ -83,6 +85,8 @@ def from_file(cls, calibration_file, degree=1):
 
 
 class PiezoTrackingCalibration:
+    """Calibration to perform piezo tracking with"""
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, a bit. It doesn't give out the individual displacements though. How about:

```
    """Class to handle piezo tracking calibration.

    Allows calculating piezo distance from trap position and correlated force data"""
```

lumicks/pylake/piezo_tracking/baseline.py
```diff
@@ -4,6 +4,8 @@
 
 
 class ForceBaseLine:
```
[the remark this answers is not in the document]
JoepVanlier: Yes! Done! :)

docs/tutorial/piezotracking.rst
JoepVanlier: So looking at a few recent datasets, this residual magnitude is pretty typical. It's about 15 nm peak to peak; so that's about 45 base pairs. For pulling this is probably not super relevant (most constructs are much bigger and even the error in the um per kbp value is larger) and other sources of errors are dominant.

For folding/unfolding studies, the distance transitions are mostly gonna be dominated by the force part (since one doesn't move the mirror when looking at those).

I do think that is warrants some additional investigation, but for now, this is just what it is, and I still think that showing how to spin up the residual plot is useful.

I've added the following:
```
We can see that there is some error (the discrepancy does not scatter randomly around zero), but for this experiment, 15 nm error is within an acceptable range::
```

To explicitly note that this is a judgment call that needs to be made and not gloss over it. What do you think?

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,164 @@
...
+Trap Positional Calibration
+---------------------------
+
+The first step is to calibrate the high frequency trap position from the piezo mirror to the low frequency bead-to-bead distance measured by the camera.
+
+For this calibration, we require a dataset (acquired in the absence of a tether) in which trap 1 is moved over the entire distance range intended for the experiment.
+The exported data file must contain the trap 1 position and camera-based distance channels.
+
+*Note that this dataset should be acquired such that the bead tracking templates do not overlap. In case of overlap, make sure to first slice the position and distance data such that this region is not included.*
+Let's load this dataset and perform the distance calibration by invoking::
```
rpauszek, line 24:
```suggestion
*Note that this dataset should be acquired such that the bead tracking templates do not overlap. In case of overlap, make sure to first slice the position and distance data such that this region is not included.*

Let's load this dataset and perform the distance calibration by invoking::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,164 @@
...
+    no_tether_data.distance1.plot()
+    plt.tight_layout()
+
+.. image:: figures/piezotracking/nonzero_force.png
+
+It seems that the force was not reset to zero the force at the start of this experiment.
+We can still do this in post-processing however.
+To zero the force, we wish to use the force when the beads are far apart (no interaction between them).
+We can see in the curves that the distance decreases with time, so the best part to use would be the first section of data.
+In this case, we'll use a quarter of a second::
```
rpauszek, line 66:
```suggestion
It seems that the force was not zeroed at the start of this experiment, but we can correct this in post-processing.
We want to use the force when the beads are far apart (no interaction between them) which corresponds the beginning of the data. In this case, we'll use a quarter of a second::
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,164 @@
...
+
+    f1_offset = np.mean(no_tether_data.force1x[:".25s"].data)
+    f2_offset = np.mean(no_tether_data.force2x[:".25s"].data)
```
rpauszek, line 69:
```suggestion
    f1_offset = np.mean(no_tether_data.force1x[:"0.25s"].data)
    f2_offset = np.mean(no_tether_data.force2x[:"0.25s"].data)
```

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,164 @@
...
+
+The first step is to calibrate the high frequency trap position from the piezo mirror to the low frequency bead-to-bead distance measured by the camera.
+
+For this calibration, we require a dataset (acquired in the absence of a tether) in which trap 1 is moved over the entire distance range intended for the experiment.
+The exported data file must contain the trap 1 position and camera-based distance channels.
+
+*Note that this dataset should be acquired such that the bead tracking templates do not overlap. In case of overlap, make sure to first slice the position and distance data such that this region is not included.*
+Let's load this dataset and perform the distance calibration by invoking::
+
+    no_tether_data = lk.File("Data/20211210-161606 FD Curve 19 curve 7 return.h5")
```
rpauszek, line 26: do you think it would be good to rename these files to something a bit simpler/descriptive?

rpauszek (approved):
Looks good, just a few minor suggestions

docs/tutorial/piezotracking.rst
JoepVanlier: 👍

docs/tutorial/piezotracking.rst
JoepVanlier: 👍

docs/tutorial/piezotracking.rst
```diff
@@ -0,0 +1,164 @@
...
+    \delta x_1 = \frac{F_{1x}}{\kappa_{1x}}
+
+Thus the surface-to-surface distance between the beads can be computed by correcting the trap-based distance with the correlated force data and their respective trap stiffnesses as follows.
+
+.. math::
+
+    d_{piezo} = d_{no\_tether} - \frac{F_{1x}}{\kappa_{1x}} + \frac{F_{2x}}{\kappa_{2x}}
+
+Here :math:`d_{piezo}` reflects the piezo distance and :math:`d_{no_tether}` reflects the calibrated surface-to-surface distance without the tether.
+:math:`F_{1x}` and :math:`F_{2x}` are the forces measured on the bead and :math:`\kappa_{1x}` and :math:`\kappa_{2x}` represent the trap stiffness for each trap.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/piezotracking.rst
JoepVanlier: Will do!

Issue #327, opened 2022-06-29
FdFit objects cannot be pickled


b-dawes:
I tried to pickle and then load an FdFit object. When calling the loaded FdFit object, I get the error message copied below. 

I believe the issue comes from the call `self.datasets[id(model)]` as when the pickled file is loaded, the variable id numbers are not preserved.

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
File c:\Users\bdawes.DESKTOP-9ERP4NC\miniconda3\envs\PY571\lib\site-packages\IPython\core\formatters.py:343, in BaseFormatter.__call__(self, obj)
    341     method = get_real_method(obj, self.print_method)
    342     if method is not None:
--> 343         return method()
    344     return None
    345 else:

File c:\Users\bdawes.DESKTOP-9ERP4NC\miniconda3\envs\PY571\lib\site-packages\lumicks\pylake\fitting\fit.py:643, in Fit._repr_html_(self)
    640 out_string = "<h4>Fit</h4>\n"
    642 for model in self.models.values():
--> 643     datasets = "".join(f"{self.datasets[id(model)]._repr_html_()}<br>\n")
    644     out_string += f"<h5>Model: {model.name}</h5>\n"
    645     eqn = model.get_formatted_equation_string(tex=True)

KeyError: 2478746249168
lumicks.pylake.FdFit(models={DNA}, N=55)
```

b-dawes:
I looked into this a little bit more and wrote a tiny function that should reset the dictionary keys to be correct. This could be added as a function that would just need to be called when an object is loaded from a pickle. However, the root of the issue is using python id's as keys which seems like an odd choice to me.

```
def reset_ids(fit):
    old_ids = [*fit.models]
    for old_id in old_ids:
        new_id = id(fit.models[old_id])
        fit.datasets[new_id] = fit.datasets.pop(old_id)
        fit.models[new_id] = fit.models.pop(old_id)
```

JoepVanlier:
I have prepared a small PR that aims to fix this: https://github.com/lumicks/pylake/pull/330 . Let me know if it works for your use case with these changes.

JoepVanlier:
Closing since this feature was released in [v0.13.0](https://github.com/lumicks/pylake/releases/tag/v0.13.0)

PR #330, opened 2022-07-01
fitting: make `FdFit` 🥒-able

JoepVanlier:
**Why this PR?**
Using `id(model)` to identify models within an `FdFit` is not a good approach, since this value is only preserved over its lifetime. One consequence of this is that `FdFit` can't be pickled (see: https://github.com/lumicks/pylake/issues/327). This PR switches to associating a uuid with a model which is then used to uniquely identify it.


rpauszek (approved):
LGTM

PR #331, opened 2022-07-05
Add `min_length` parameter to tracking algorithms

rpauszek:
**Why this PR?**
It's common to filter tracked lines for a minimum length and this extra argument allows you to do that straight from the tracking algorithm rather than having to manually call `filter_lines()` after the tracking. The default `None` leaves the behavior unchanged compared to before. Also, this change allows us to de-duplicate some code in the Kymotracker widget.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 
 * Added function `Kymo.line_timestamp_ranges()` to obtain the start and stop timestamp of each scan line in a `Kymo`. Please refer to [Confocal images](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymographs.html) for more information.
 * Added `Kymo.flip()` to flip a Kymograph along its positional axis.
+* Added a `min_length` argument to `lk.track_greedy()` to automatically filter out tracked lines shorter than a minimum length.
 
 #### Breaking changes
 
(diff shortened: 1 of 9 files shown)


JoepVanlier (commented):
I have mixed feelings about this PR. I really like how is cleans up the widget, and I could definitely see use for the convenience of having this argument there. To me, it feels like it takes away from the composability of this API. The worry that I have is that `min_length` is a bit of a misnomer (my bad, sorry). It would better to call it `min_number_detections`. If you really want it to act as a minimum length, what should happen is `tracking -> refinement (w/interpolation) -> filtering`.

The reason I feel iffy about this is that I could easily see `filter_lines` changing, meaning that both these kymotracker functions then also get extra arguments and changed behaviour again.

Personally, I wonder if it would be better to go with a convenience function that does `tracking + refinement + filtering` in one go and leave the lower level API as granular as it is now, rather than tack filtering (and maybe refinement) onto the tracker.

I'm not sure about this however, so I'd be interested in hearing other thoughts on this.

rpauszek:
Yeah I see your point. Easier fix ended up just refactoring the behavior a bit in the widget itself in #335. So I'll close this one

PR #336, opened 2022-07-08
Download files directly from Zenodo.

JoepVanlier:
**Why this PR?**
This is the first step in making our docs runnable without storing the data on the repository itself.

This way, when users download a notebook from the docs, we can simply download the data from Zenodo in the notebook and use the data there.

Note that this PR adds two dependencies to Pylake. While it is possible to write it using the standard library `urllib`, it would be considerably less pretty.

Docs build here: https://lumicks-pylake.readthedocs.io/en/zenodo_dl/

image


lumicks/pylake/file_download.py
```diff
@@ -0,0 +1,139 @@
...
+    block_size : int
+        Block size to use when downloading
+    """
+    response = requests.get(url, stream=True)
+    size_bytes = int(response.headers.get("content-length", 0))
+
+    full_path = os.path.join(target_path, file_name)
+    dir_name = os.path.dirname(full_path)
+    if dir_name:
+        os.makedirs(dir_name, exist_ok=True)
```
mikhas, line 54: `dir_name` is just the same as `target_path`, no? Also, shouldn't it be `if not dir_name:`?

lumicks/pylake/file_download.py
```diff
@@ -0,0 +1,139 @@
...
+        while len(b) > 0:
+            m.update(b)
+            b = f.read(chunk_size)
+
+    return m.hexdigest() == reference_hash
+
+
+def download_from_doi(doi, target_path="", force_download=False, show_progress=True):
+    """Download files from a Zenodo DOI (i.e. 10.5281/zenodo.#######) and returns a file list in the
+    form of a List[str].
```
mikhas, line 83: Will users know how to get a fully qualified DOI for their data?

lumicks/pylake/file_download.py
```diff
@@ -0,0 +1,139 @@
...
+    show_progress : bool
+        Show a progress bar while downloading.
+    """
+    url = get_url_from_doi(doi)
+
+    if "zenodo" not in url:
+        raise RuntimeError("Only Zenodo DOIs are supported.")
+
+    record_number = url.split("/")[-1].strip()
+    rec = download_record_metadata(record_number)
```
mikhas, line 104: would prefer `record_meta` instead of `rec`

lumicks/pylake/tests/test_file_download.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Validate checksum
+    assert verify_hash(files[0], *record["files"][0]["checksum"].split(":"))
+
+    # Add a random character such that the checksum fails
+    with open(files[0], "ab") as f:
+        f.write(b"\x21")
+
+    # Validate that the hash is no longer correct
+    assert not verify_hash(files[0], *record["files"][0]["checksum"].split(":"))
```
mikhas, line 48: Does `verify_hash` show some info to the user in case of failure?

mikhas (commented):
Nice stuff! It downloads the kymos, and caches them accordingly. Not sure about the urllib dependency.

lumicks/pylake/file_download.py
JoepVanlier: Hmm, I think the confusion here comes from `file_name`  being a bit of a misnomer. `file_name` is actually a relative path provided by Zenodo, for the file currently being downloaded. Therefore, it's important to concatenate them, then re-extract the actual directory with `dirname` (otherwise you get a folder with the filename's name). I've renamed the `file_name` argument to `download_path` to clear up this confusion.

Regarding the if-statement, that is as intended. We only attempt to make folders if it needs to be done. Note that the `makedirs` call has `exist_ok=True`, which means that if the directory already exists, it just silently ignores it.

lumicks/pylake/file_download.py
JoepVanlier: Normally, they're in papers and the format is listed in the docstring. What else would you add?

lumicks/pylake/file_download.py
JoepVanlier: 👍

lumicks/pylake/tests/test_file_download.py
JoepVanlier: If it throws an exception, it will just bubble up. I'm not expecting a specific exception that we should catch and handle at this point.

lumicks/pylake/file_download.py
mikhas: Cool, just wanted to know, no change needed.

lumicks/pylake/file_download.py
mikhas: much better now

mikhas (approved):
👍🏽

lumicks/pylake/file_download.py
```diff
@@ -0,0 +1,139 @@
...
+    file_names = []
+    for file in record_metadata["files"]:
+        file_name, url = file["key"], file["links"]["self"]
+        full_path = os.path.join(target_path, file_name)
+
+        download = force_download  # By default, we skip files that already exist
+
+        # If the file doesn't exist, we can't skip it
+        if not os.path.exists(full_path):
+            download = True
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, not a bad idea, especially if we get additional stuff in between in the future. Less mutating state to think about :D

lumicks/pylake/file_download.py
```diff
@@ -0,0 +1,139 @@
...
+
+        # If the file is not the one we want, we can also not skip it (only check the hash if the
+        # file exists)
+        hash_algorithm, checksum = file["checksum"].split(":")
+        if not download and not verify_hash(full_path, hash_algorithm, checksum):
+            if not force_download:
```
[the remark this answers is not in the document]
JoepVanlier: Done.

mikhas (approved):
still works

PR #337, opened 2022-07-11
Manually redraw axes in kymotracker callbacks

rpauszek:
**Why this PR?**
After updating `jupyter/ipywidgets/mpl` I noticed that the interactive elements during horizontal panning and line connecting no longer worked. Manually calling `canvas.draw_idle()` seems to fix the problem without noticeable performance hit

Before:
tracker_bug

After:
tracker_fix


JoepVanlier (approved):
LGTM! 🚢

PR #341, opened 2022-07-15
Kymotracker line connection function

rpauszek:
**Why this PR?**
Pulled out the hit test for manual line connection so the functionality can be reused in other UI

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -152,10 +152,8 @@ class KymoWidget:
             if len(self.lines) == 0:
                 return
 
-            is_hit, line_info = _get_nearest(
-                self.lines, event.x, event.y, scale, cutoff_radius
-            )
-            if is_hit:
+            line_info = _get_nearest(self.lines, event.x, event.y, scale, cutoff_radius)
+            if line_info:
                 clicked_line_info = line_info
                 return True
 
@@ -165,7 +163,8 @@ class KymoWidget:
                 plotted_line.remove()
                 plotted_line = None
 
-            line, node_index, *_ = clicked_line_info
+            line = clicked_line_info["line"]
+            node_index = clicked_line_info["node_index"]
             plotted_line, *_ = self._axes.plot(
                 [line.seconds[int(node_index)], event.x],
                 [line.position[int(node_index)], event.y],
@@ -179,10 +178,8 @@ class KymoWidget:
                 plotted_line.remove()
                 plotted_line = None
 
-            is_hit, line_info = _get_nearest(
-                self.lines, event.x, event.y, scale, cutoff_radius
-            )
-            if is_hit:
+            line_info = _get_nearest(self.lines, event.x, event.y, scale, cutoff_radius)
+            if line_info:
                 released_line_info = line_info
 
                 # Explicit copy to make modifications. Current state pushed to undo stack on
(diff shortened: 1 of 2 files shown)


lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -482,6 +459,76 @@ def show(self, use_widgets, **kwargs):
         self._select_state({"value": "mode", "old": "", "new": "Track lines"})
 
 
+def _get_node_info(lines):
```
JoepVanlier, line 462: Would be good to add a test for `_get_nearest` now that it's just a free func.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -482,6 +459,76 @@ def show(self, use_widgets, **kwargs):
...
+                track.seconds,
+                track.position,
+            )
+        )
+        for j, track in enumerate(lines)
+    ]
+    return np.hstack(nodes).T
+
+
+def _get_nearest(lines, nodes, x, y, scale, cutoff_radius=0.05):
```
JoepVanlier, line 492: Same here. Would be nice to have a test for this. There's also ample room for optimization here, and it'd be nice to have a test in place if we wanna have a go at optimizing this.

JoepVanlier (commented):
Yeah, this is nice. Also nice to get rid of that ugly `nonlocal`. Would like to see a few tests added for the new free funcs, but otherwise it looks nice. :)

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -140,49 +140,25 @@ def set_xlim(drag_event):
 
         MouseDragCallback(self._axes, 1, set_xlim)
 
-    def _get_scale(self):
-        """Get scaling of the image axes"""
-        return tuple(lims[1] - lims[0] for lims in (self._axes.get_xlim(), self._axes.get_ylim()))
-
     def _connect_line_callback(self):
         canvas = self._axes.figure.canvas
+        scale = tuple(lims[1] - lims[0] for lims in (self._axes.get_xlim(), self._axes.get_ylim()))
```
mikhas, line 145: Perhaps rename `scale` to `visible_range`?

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -140,49 +140,25 @@ def set_xlim(drag_event):
...
             if len(self.lines) == 0:
                 return
 
-            nodes = get_node_info()
-            distance, idx = get_nearest(event.x, event.y)
-
-            if distance < cutoff_radius:
-                line_index, *line_info = nodes[idx]
-                clicked_line_info = [self.lines[int(line_index)], *line_info]
+            nodes = _get_node_info(self.lines)
```
mikhas, line 156: It's not a problem that I now have to call `_get_node_info()` myself, but it does leak implementation details that we don't really need, thus binding you a bit to this admittedly still private API.

All _I_ need, for the `_merge_lines()` to work, is the position of the datapoint inside a `KymoLine`. Maybe we can rethink this part a bit.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
mikhas, line 492: Tested it, works great!

mikhas (commented):
Agree with Joep that the tests need to change as well, even though one could argue that the existing integration test with the fake mouse events still covers this. Not too thrilled about the node info API, however.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
rpauszek: I agree, the API was a bit weird. basically I left it as it already was, but just pulled out into a separate function. I've updated it now to be a bit more straightforward I think. `_get_node_info()` is back to being truly internal, and `_get_nearest()` returns a `dict` with the relevant info if a node was clicked, else empty.

See last 3 commits for proposed changes, these'll be squashed on merge

rpauszek:
not needed anymore

PR #342, opened 2022-07-15
Removing sample rate defaults

JoepVanlier:
**Why this PR?**
Considering that not all systems have the same acquisition rate, providing a default `sample_rate` is risky. For this reason, we refrain from providing a default now.


lumicks/pylake/force_calibration/convenience.py
```diff
@@ -89,7 +90,7 @@ def calibrate_force(
         measurements performed deep in bulk.
     fast_sensor : bool, optional
          Fast sensor? Fast sensors do not have the diode effect included in the model.
-    sample_rate : float, optional
+    sample_rate : float
```
tobiasjj, line 93: Would it make sense to reorder the parameters in the docstring according to the order in the function header?

tobiasjj (approved):
With my limited opinion: This looks good and makes sense. I guess, the doc corrections should just go along with this PR?

PR #343, opened 2022-07-22
Slicing steps

tobiasjj:
**Why this PR?**
As a user I want to be able to define steps when indexing frames of `CorrelatedStacks` (i.e. decimate the frames).

### DIFF
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -49,6 +49,7 @@ class CorrelatedStack:
         )
         self.start_idx = 0
         self.stop_idx = self.src.num_frames
+        self._step = 1
 
     def _handle_cropping(self, item):
         """Crop the stack based on tuple of slices"""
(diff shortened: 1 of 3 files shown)


lumicks/pylake/correlated_stack.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, *image_names, align=True):
         )
         self.start_idx = 0
         self.stop_idx = self.src.num_frames
+        self.step = None
```
JoepVanlier, line 52: Why not `1` as default?

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -94,24 +135,33 @@ def compare_frames(original_frames, new_stack):
     compare_frames([0, 1, 2, 3, 4, 5], stack0[:-4])  # until negative index
     compare_frames([5, 6, 7], stack0[5:-2])  # mixed sign indices
     compare_frames([6, 7], stack0[-4:-2])  # negative indices slice
+    compare_frames([0, 2, 4, 6, 8], stack0[::2])  # all frames with steps
+    compare_frames([0, 3, 6, 9], stack0[::3])  # all frames with steps
+    compare_frames([3, 5], stack0[3:6:2])  # normal slice with steps
+    compare_frames([0, 2], stack0[:3:2])  # from beginning with steps
```
JoepVanlier, line 142: Might be nice to add one that does recursive slicing w/ steps.

```suggestion
    compare_frames([0, 2], stack0[:3:2])  # from beginning with steps
    compare_frames([1, 5, 9], stack0[1::2][::2])  # recursive slice with steps
```

lumicks/pylake/correlated_stack.py
```diff
@@ -425,10 +434,23 @@ def parse_tags(frame):
...
+        step = 1 if step is None else step
+        if step <= 0:
+            raise ValueError("slice step must be greater than 0")
+        self._step = step
+
     @property
     def num_frames(self):
         """Number of frames in the stack."""
-        return self.stop_idx - self.start_idx
+        return max(-1, (self.stop_idx - self.start_idx - 1)) // self.step + 1
```
JoepVanlier, line 453: I had to do a double take here, but it is indeed correct :)

JoepVanlier (changes requested):
Great work! The only thing I would like to see changed is to not make `CorrelatedStack` mutable by adding a setter for `step`. I also personally don't see the utility of having `None` as the default for the private property if it's always converted to `1` anyway, but maybe I'm overlooking something here?

lumicks/pylake/correlated_stack.py
```diff
@@ -64,33 +65,38 @@ def interpret_crop(item):
         if len(item) > 3:
             raise IndexError("Only three indices are accepted when slicing CorrelatedStacks.")
 
-        rows = interpret_crop(item[1])
+        rows = interpret_crop(item[1]) if len(item) >= 2 else [None, None]
         columns = interpret_crop(item[2]) if len(item) >= 3 else [None, None]
```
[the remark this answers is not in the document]
tobiasjj: Done

lumicks/pylake/correlated_stack.py
```diff
@@ -425,10 +434,23 @@ def parse_tags(frame):
...
 
+    @property
+    def step(self):
+        """Get the step value for slicing frames"""
+        return self._step
+
+    @step.setter
+    def step(self, step):
+        """Set the step value for slicing frames"""
+        step = 1 if step is None else step
```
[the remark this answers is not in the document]
tobiasjj: I set default values to `None` out of the habit not setting default values directly in the function header, but indeed, this is unnecessary overhead. I agree that making the class mutable creates more headaches than simply creating new correlated stacks with different steps -> Changed.

lumicks/pylake/correlated_stack.py
```diff
@@ -49,6 +49,7 @@ def __init__(self, *image_names, align=True):
         )
         self.start_idx = 0
         self.stop_idx = self.src.num_frames
+        self.step = 1
```
JoepVanlier, line 52: Please don't forget to make it private as well (preface with underscore) :)

tobiasjj: Done. Shouldn't we also make start_idx and stop_idx private, then?

JoepVanlier: Probably yes. Should probably make it private and add an entry in the changelog about it (since it is technically a breaking change).

tobiasjj: Done

JoepVanlier (approved):
Looks good to me.

One last minor nitpick; it's usually nice to keep commit messages short enough such that they don't exceed the display limit on git and git clients. While this isn't always possible, I think the second commit message could just be shortened to:
`corrstack: remove public attributes stop_idx and start_idx` 

Since that's the only thing that really matters to users. If more information is needed, the changelog already contains it.

tobiasjj:
Yihaa, thanks for taking the time for reviewing!

PR #347, opened 2022-08-02
ci: exclude zenodo downloads from ci

JoepVanlier:
**Why this PR?**
The ci tests for the download of files from Zenodo are brittle and fail when there is a timeout. This is annoying. I propose disabling these tests on ci, but adding a step to the release procedure to run them on demand.


rpauszek (approved):
🚀

PR #349, opened 2022-08-04
Same access of sample rate for `Continuous` and `TimeSeries` data

tobiasjj:
**Why this PR?**

As a user I want to easily access the sample rate of `Continuous` and `TimeSeries` data.

`TimeSeries` data usually has a constant sample rate. The property `sample_rate` now returns the `sample_rate` if it is constant, or if not, returns `None`. If the sample rate is variable, the user needs to do something else.

Additionally, this PR fixes wrong sample rates returned for downsampled `Continuous` data (see tests).

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -1,10 +1,10 @@
 import numpy as np
 import numpy.typing as npt
 import numbers
+from typing import Union
 
 from .detail.utilities import downsample
 from .detail.timeindex import to_timestamp
-from .calibration import ForceCalibration
 from .nb_widgets.range_selector import SliceRangeSelectorWidget
 
 
@@ -204,16 +204,23 @@ class Slice:
             return []
 
     @property
-    def sample_rate(self) -> int:
-        """The data frequency for continuous data sources or `None` if it's variable"""
+    def sample_rate(self) -> Union[float, None]:
+        """The data frequency for `Continuous` and `TimeSeries` data sources or `None` if it is not
+        available or variable"""
         try:
             return self._src.sample_rate
         except AttributeError:
             return None
 
     @property
-    def _timesteps(self):
-        return self._src._timesteps
+    def _timesteps(self) -> npt.ArrayLike:
+        """The unique timesteps for `Continuous` and `TimeSeries` data sources"""
+        try:
+            return self._src._timesteps
+        except AttributeError:
+            raise NotImplementedError(
+                f"`_timesteps` are not available for {self._src.__class__.__name__} data"
+            )
 
     def downsampled_over(self, range_list, reduce=np.mean, where="center"):
         """Downsample channel data based on timestamp ranges. The downsampling function (e.g. np.mean) is evaluated for
@@ -472,7 +479,7 @@ class Continuous:
         self._cached_data = None
         self.start = start
         self.stop = start + len(data) * dt
-        self.dt = dt
+        self.dt = dt  # ns
 
     def _with_data(self, data):
         return self.__class__(data, self.start, self.dt)
@@ -488,7 +495,7 @@ class Continuous:
     @staticmethod
     def from_dataset(dset, y_label="y", calibration=None):
         start = dset.attrs["Start time (ns)"]
-        dt = int(1e9 / dset.attrs["Sample rate (Hz)"])
+        dt = int(1e9 / dset.attrs["Sample rate (Hz)"])  # ns
         return Slice(
             Continuous(dset, start, dt),
             labels={"title": dset.name.strip("/"), "y": y_label},
@@ -506,12 +513,12 @@ class Continuous:
         return np.arange(self.start, self.stop, self.dt)
 
     @property
-    def sample_rate(self):
-        return int(1e9 / self.dt)
+    def sample_rate(self) -> float:
+        return 1e9 / self.dt  # Hz
 
     @property
-    def _timesteps(self):
-        return np.asarray([self.dt], dtype=int)
+    def _timesteps(self) -> npt.ArrayLike:
+        return np.asarray([self.dt], dtype=np.int64)  # ns
 
     def slice(self, start, stop):
         def to_index(t):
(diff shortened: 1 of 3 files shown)


lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -398,28 +438,31 @@ def test_channel(channel_h5_file):
...
+    sample_rate = 1e9 / dt  # Hz
+
+    s = channel.Slice(channel.Continuous([14, 15, 16, 17], start=with_offset(0), dt=dt))
     assert s.sample_rate == 1e8
 
     s2 = s.downsampled_by(2)
     np.testing.assert_allclose(s2.data, 14.5, 16.5)
     np.testing.assert_equal(s2.timestamps, with_offset([5, 25]))
-    assert s2.sample_rate == 0.5e8
+    assert s2.sample_rate == sample_rate / 2
```
JoepVanlier, line 450: 👍

JoepVanlier (changes requested):
Nice work! AFAIK `ValueError` is more intended for when the user provides an incorrect value to a function rather than an exception of this type.

Regarding the change from `None` to an exception, I'm still on the fence. I think `None` also communicates that it isn't defined well while being less cumbersome to handle upstream.

lumicks/pylake/channel.py
```diff
@@ -583,8 +591,16 @@ def stop(self):
             raise IndexError("End of empty time series is undefined")
 
     @property
-    def _timesteps(self):
-        return np.unique(np.diff(self.timestamps)).astype(int)
+    def sample_rate(self) -> float:
+        timestep = self._timesteps
+        if len(timestep) > 1:
+            raise ValueError("The sample rate is variable")
```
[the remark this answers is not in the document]
tobiasjj: Corrected!

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -116,23 +116,63 @@ def test_calibration_continuous_channels():
...
+    np.testing.assert_equal(s._timesteps, timesteps)
+
+    # `TimeTags`
     s = channel.Slice(channel.TimeTags(np.arange(0, size, dtype=np.int64)))
     assert len(s) == size
-    assert s.sample_rate is None
+    with pytest.raises(
+        NotImplementedError,
+        match="`sample_rate` is not available for TimeTags data",
+    ):
```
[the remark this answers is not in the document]
tobiasjj: Indeed handling of the `AttributError` is kinda cumbersome. It also is not optimal introducing another breaking change. I therefore reverted the functionality back to how it was and return a `None` in case no `sample_rate` is defined.

However, for `TimeSeries` data it still raises an `AttributeError` if the `sample_rate` is variable. I think otherwise users might be confused, when their code accessing `sample_rate` was usually working and then suddenly they are confronted with an empty `None` sample rate without an explanation. In this case maybe changing the `AttributeError` into a warning would make sense?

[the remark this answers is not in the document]
JoepVanlier: I think this could also be explained in the docstring though. When they get `None`, the first thing they will (/should) probably look at is the docstring.

While I'm not entirely opposed to a warning, this would get annoying if you wrote your code to actually gracefully deal with this case. Since you'd still emit the warning even if all they're doing is checking if it's `None`. Sure you can then filter that warning, but that's also not so nice.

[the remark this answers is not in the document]
tobiasjj: OK, convinced 😄

lumicks/pylake/channel.py
```diff
@@ -583,8 +591,17 @@ def stop(self):
...
     @property
-    def _timesteps(self):
-        return np.unique(np.diff(self.timestamps)).astype(int)
+    def sample_rate(self) -> Union[float, None]:
+        """The data frequency or `None` if it is variable"""
+        timestep = self._timesteps
+        if len(timestep) > 1:
+            return None
+        else:
+            return 1e9 / timestep[0]  # Hz
```
JoepVanlier, line 600: You could omit the `None` case, since that's the default return. No need to explicitly return it.
```suggestion
        if len(timestep) == 1:
            return 1e9 / timestep[0]  # Hz
```

lumicks/pylake/channel.py
```diff
@@ -583,8 +591,17 @@ def stop(self):
...
+
+    @property
+    def _timesteps(self) -> npt.ArrayLike:
+        return np.unique(np.diff(self.timestamps)).astype(int)  # ns
```
JoepVanlier, line 604: Why do you cast this to an `int` explicitly?

There's a risk associated with this for `TimeSeries`, because the precision may be insufficient to capture the timestep. If you downsample to 0.1 Hz for instance, you'll have a timestamp difference of `1e10` which when cast like this will overflow and therefore wrap around. Also note that timestamps are typically not regular `dtype=int` in Pylake, but `np.int64` because of the precision issues.

I'd omit the `astype(...)` if there's no specific reason to have it. If it's needed for downstream stuff, `np.int64` would be a better choice.

JoepVanlier (changes requested):
Nice! :)

Two remaining small comments, but I think it's mostly good to go.

lumicks/pylake/channel.py
tobiasjj: I removed the type conversion, which was a left over from old code. Also I ensured that property `_timesteps` in `Contiguous` returns an array of dtype `np.int64`.

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -120,12 +120,20 @@ def test_slice_properties():
...
     assert s.sample_rate == 1e9
-    timesteps = np.asarray([1], dtype=int)
+    timesteps = np.asarray([1])
+    np.testing.assert_equal(s._timesteps, timesteps)
+
+    # `Continuous` with sample rate < (1e9 / (2**31 - 1)), i.e. timesteps > max(int32)
+    s = channel.Slice(channel.Continuous(rng.random(size), start=0, dt=2**31))
+    assert len(s) == size
+    assert s.sample_rate == 1e9 / 2**31
+    timesteps = np.asarray([2**31])
```
JoepVanlier, line 136: Nice to add these 👍

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -142,6 +150,13 @@ def test_slice_properties():
     timesteps = np.unique(np.diff(timestamps_var))
     np.testing.assert_equal(s._timesteps, timesteps)
 
+    # `TimeSeries` with sample rate < (1e9 / (2**31 - 1)), i.e. timesteps > max(int32)
+    s = channel.Slice(channel.TimeSeries(rng.random(size), timestamps_far))
+    assert len(s) == len(timestamps_far)
+    assert s.sample_rate == 1e9 / 2**31
+    timesteps = np.unique(np.diff(timestamps_far))
+    np.testing.assert_equal(s._timesteps, timesteps)
```
JoepVanlier, line 158: 👍

JoepVanlier (approved):
I think it's good to go.

PR #350, opened 2022-08-04
Lazily load `TimeSeries` `data` and `timestamps`

tobiasjj:
**Why this PR?**
Found a comment in the source code stating that `data` and `timestamps` should be lazily loaded and took this as an opportunity to implement it.


JoepVanlier (approved):
LGTM! :)

PR #351, opened 2022-08-04
correlated_stack: change `frame_timestamp_ranges` to method

JoepVanlier:
**Why this PR?**
API uniformity basically. This way, it provides the exact same API as we have on `Scan` leading to fewer surprises for the user.

### DIFF
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -121,6 +121,6 @@ using timestamps obtained from the `CorrelatedStack`::
     file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges())
 
 By default, this averages only over the exposure time of the images in the stack.
-If you wish to average over the full time range from the start of the scan to the next scan, pass the extra parameter `exclude=False`::
+If you wish to average over the full time range from the start of the scan to the next scan, pass the extra parameter `include_dead_time=True`::
 
-    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges(exclude=False))
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges(include_dead_time=True))
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -118,4 +118,9 @@ downsampled over the video frames. This can be done using the function `Slice.do
 using timestamps obtained from the `CorrelatedStack`::
 
     # Determine the force trace averaged over frame 2...9.
-    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges)
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges())
+
+By default, this averages only over the exposure time of the images in the stack.
+If you wish to average over the full time range from the start of the scan to the next scan, pass the extra parameter `exclude=False`::
+
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges(exclude=False))
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -35,7 +35,7 @@ class CorrelatedStack(VideoExport):
         stack.plot_correlated(file.force1x)
 
         # Determine the force trace averaged over frame 2...9.
-        file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges)
+        file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges())
 
         # Loading multiple TIFFs into a stack.
         stack = pylake.CorrelatedStack("example.tiff", "example2.tiff", "example3.tiff")
@@ -364,7 +364,7 @@ class CorrelatedStack(VideoExport):
         from lumicks.pylake.nb_widgets.correlated_plot import plot_correlated
 
         title_factory = lambda frame: make_image_title(self, frame, show_name=False)
-        frame_timestamps = self.frame_timestamp_ranges
+        frame_timestamps = self.frame_timestamp_ranges()
 
         def frame_grabber(frame_idx):
             return self._get_frame(frame_idx)._get_plot_data(channel, adjustment=adjustment)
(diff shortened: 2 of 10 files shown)


lumicks/pylake/correlated_stack.py
```diff
@@ -487,15 +487,26 @@ def stop(self):
...
+        ----------
+        exclude : bool
+            Exclude dead time at the end of each frame.
+        """
+        ts_ranges = [frame.frame_timestamp_range for frame in self]
+        if exclude:
+            return ts_ranges
+        else:
+            frame_ts = [(start[0], stop[0]) for start, stop in zip(ts_ranges, ts_ranges[1:])]
+            frame_ts.append((ts_ranges[-1][0], ts_ranges[-1][1]))
```
tobiasjj, line 511: Can the frame rate of the `CorrelatedStacks` be assumed constant? If so one could calculate dt and add the last timestamp range like:
```suggestion
try:
    if len(ts_ranges) > 2:
        dt = ts_ranges[-1][0] - ts_ranges[-2][0]
        stop = ts_ranges[-1][0] + dt
    else:  # len <= 1
        stop = ts_ranges[-1][1]
    frame_ts.append((ts_ranges[-1][0], stop))
except IndexError:
    pass
```
This would also allow for empty timestamp ranges.

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -927,3 +927,18 @@ def test_alignment_multistack(rgb_alignment_image_data, gray_alignment_image_dat
...
+    stack = CorrelatedStack.from_dataset(
+        TiffStack(
+            [MockTiffFile(data=[np.ones((5, 5))] * 4, times=make_frame_times(4, step=6))],
+            align_requested=False,
+        )
+    )
+
+    for exclude, ranges in zip(
+            [False, True],
+            [[(10, 20), (20, 30), (30, 40), (40, 46)], [(10, 16), (20, 26), (30, 36), (40, 46)]],
```
tobiasjj, line 942: If you calculate the last timestamp range

```suggestion
            [[(10, 20), (20, 30), (30, 40), (40, 50)], [(10, 16), (20, 26), (30, 36), (40, 46)]],
```

tobiasjj (commented):
I like that this PR unifies`frame_timestamp_ranges` of `CorrelatedStack`s and `Scan`s! I found one error and have two minor suggestions:
* The documentation confuses the default value of `exclude`.
* Probably rename `exclude` to `include_dead_time` or similar
* Probably try to calculate the last timestamp range based on the frame rate

lumicks/pylake/correlated_stack.py
tobiasjj:
```suggestion
try:
    if len(ts_ranges) >= 2:
        dt = ts_ranges[-1][0] - ts_ranges[-2][0]
        stop = ts_ranges[-1][0] + dt
    else:  # len <= 1
        stop = ts_ranges[-1][1]
    frame_ts.append((ts_ranges[-1][0], stop))
except IndexError:
    pass
```

JoepVanlier: It can't be assumed constant. You can change the camera framerate while acquiring. One can not even assume that there is more than one frame since the stack may be a snapshot.

lumicks/pylake/correlated_stack.py
```diff
@@ -487,15 +487,26 @@ def stop(self):
...
+        Parameters
+        ----------
+        exclude : bool
+            Exclude dead time at the end of each frame.
+        """
+        ts_ranges = [frame.frame_timestamp_range for frame in self]
+        if exclude:
+            return ts_ranges
+        else:
+            frame_ts = [(start[0], stop[0]) for start, stop in zip(ts_ranges, ts_ranges[1:])]
```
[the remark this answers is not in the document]
JoepVanlier: Good point.

docs/tutorial/correlatedstacks.rst
```diff
@@ -118,4 +118,9 @@ downsampled over the video frames. This can be done using the function `Slice.do
 using timestamps obtained from the `CorrelatedStack`::
 
     # Determine the force trace averaged over frame 2...9.
-    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges)
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges())
+
+By default, this averages only over the exposure time of the images in the stack.
+If you wish to average over the full time range from the start of the scan to the next scan, pass the extra parameter `exclude=True`::
+
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges(exclude=True))
```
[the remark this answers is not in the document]
JoepVanlier: Oh, good catch! Thanks! I must have rushed through the docs a bit too fast 😅 

Personally, I'm not too bothered by `exclude`, but I'm used to it. I wouldn't mind changing the parameter name if we all agree it'd be worth breaking it for the extra descriptiveness.

On the `Scan` side, we could add a forced keyworded argument for `exclude_dead_time` and a deprecation notice for the old parameter name. That would give us a deprecation path, but some ugly code until fully deprecated.

What do you think @rpauszek ?

lumicks/pylake/correlated_stack.py
tobiasjj: I thought it might be useful to also allow for empty stacks and forgot that we do not allow for empty `CorrelatedStacks` when we slice.

tobiasjj (approved):
✈️

docs/tutorial/correlatedstacks.rst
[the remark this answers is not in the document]
JoepVanlier: @tobiasjj @rpauszek I've tacked on a last commit as a proposal for switching to `include_dead_time` while still keeping changes to a minimum. Let me know what you think, whether we should do it, or drop that commit and stick to `exclude`.

docs/tutorial/correlatedstacks.rst
```diff
@@ -118,4 +118,9 @@ downsampled over the video frames. This can be done using the function `Slice.do
 using timestamps obtained from the `CorrelatedStack`::
 
     # Determine the force trace averaged over frame 2...9.
-    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges)
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges())
+
+By default, this averages only over the exposure time of the images in the stack.
+If you wish to average over the full time range from the start of the scan to the next scan, pass the extra parameter `include_dead_time=True`::
+
+    file.force1x.downsampled_over(stack[2:10].frame_timestamp_ranges(include_dead_time=True))
```
tobiasjj, line 126: To me this flows much more intuitively!

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -927,3 +927,30 @@ def test_alignment_multistack(rgb_alignment_image_data, gray_alignment_image_dat
             full_stack_idx += 1
 
     assert full_stack.num_frames == len(ref_ts)
+
+
+def test_frame_timestamp_ranges_exclude_false():
```
tobiasjj, line 932: Probably rename the test function?
```suggestion
def test_frame_timestamp_ranges():
```
or
```suggestion
def test_frame_timestamp_ranges_include_true():
```

tobiasjj (approved):
I like the new name `include_dead_time`! Only one thing: Name of the function `test_frame_timestamp_ranges_exclude_false()` might be a left over from before renaming?!
✈️

docs/tutorial/correlatedstacks.rst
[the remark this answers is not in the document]
tobiasjj: I prefer to switch to `include_dead_time`.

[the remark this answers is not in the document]
JoepVanlier: Me too.

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
JoepVanlier: Good point! I'll fixup the name tomorrow.

docs/tutorial/correlatedstacks.rst
JoepVanlier: Yeah, I agree. Usually it's nicer when the default for something is `False`. Somehow it's more expected because you're asking for an action to be taken.

PR #353, opened 2022-08-04
force calibration: remove default value driving_frequency_guess

JoepVanlier:
**Why this PR?**
The new typical frequency is still under some debate. Since we didn't release yet with the 38 Hz, best to remove it for now until things are more certain.


rpauszek (approved):
LGTM

PR #355, opened 2022-08-08
msd: fix warnings in test

JoepVanlier:
**Why this PR?**
It's very minor, but we have some warnings in our own tests and they pollute the pytest results. This PR fixes them.


dean0x7d (approved):
I found them: https://github.com/lumicks/pylake/runs/7730132855?check_suite_focus=true#step:5:91. In general, it's good to mention which warnings are being fixed as part of the commit message/PR description.

Would it be a good idea to turn on warnings as errors on CI?

JoepVanlier:
> I found them: https://github.com/lumicks/pylake/runs/7730132855?check_suite_focus=true#step:5:91. In general, it's good to mention which warnings are being fixed as part of the commit message/PR description.
> 
> Would it be a good idea to turn on warnings as errors on CI?

Yes, it would. I'll do that in a future PR. We still plan to clean some of the mpl warnings. I'll turn on warnings as errors when doing that.

PR #357, opened 2022-08-17
Update installation instructions

aafkevandenberg:
When installing Pylake, users often type the first thing they see on the installation page. Right now, this is the instruction to use pip. However, most users are not really experienced with Python and should follow the instructions under 'Anaconda'.

docs build [here](https://lumicks-pylake.readthedocs.io/en/aafkevandenberg-patch-1/install.html)

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,26 +1,10 @@
 Installation
 ============
 
-Installation instructions
--------------------------
+Pylake can be installed on Windows, Linux and Mac and requires Python 3.7 or newer.
 
-.. _Python: https://www.python.org/
-.. _SciPy: http://www.scipy.org/
-
-Pylake can be installed on Windows, Linux or Mac, with the following prerequisites:
-
-* `Python`_ 3.7 or newer (Python 2.x is not supported)
-* The `SciPy`_ stack of scientific packages
-
-If you're already familiar with Python and have the above prerequisites, installing Pylake is just a simple case of using `pip`, Python's usual package manager::
-
-    pip install lumicks.pylake
-
-Alternatively, if you are using Anaconda, please see the more detailed instructions below.
-
-
-Anaconda
---------
+Installation using Anaconda
+---------------------------
 
 .. _Anaconda: https://www.anaconda.com/download/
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -145,6 +127,21 @@ If you are used to starting Jupyter from `Terminal`, then remember to activate t
...
+Installation using pip
+----------------------
+
+.. _Python: https://www.python.org/
+.. _SciPy: http://www.scipy.org/
+
+Pylake can also be installed using pip with the following prerequisites:
+
+* `Python`_ 3.7 or newer (Python 2.x is not supported)
+* The `SciPy`_ stack of scientific packages
```
rpauszek, line 139: having these lines here seems a bit weird since technically it's valid no matter how you install. Would probably add them back up at the top before the conda instructions

aafkevandenberg: When following the instructions under Anaconda, you don't need these prerequisites. I could also reformulate this sentence and leave it here?

dean0x7d: You can drop the SciPy prerequisite either way. No one needs to install it manually. I'd keep the Python version prerequisite since even with Anaconda you need to get a new-enough baseline installation.

rpauszek: yeah maybe rewording is good then, because even with `pip` you don't need the scipy stack installed, as it's a requirement for `pylake`.

aafkevandenberg: I just noticed that under the current Anaconda instructions, Python 3.8 or newer is advised, while for pip 3.7 or newer is advised. Which of the two is correct?

rpauszek: hmm interesting, maybe a typo. `pylake` supports `>=3.7` so we should stick with that

rpauszek (changes requested):
Changes you've made look great, but looking at the built docs I noticed one more thing. Under the Anaconda instructions for Linux and MacOS it still refers to `Python 3.8` and should be `3.7`

After this change, all of the commits can be squashed down to one and will be good to go (I can help with this if needed)

rpauszek (approved):
good to go

PR #359, opened 2022-08-18
Kymotracker: warn on low threshold value

rpauszek:
**Why this PR?**
If the signal threshold value is set too low, the tracker will start to track noise which can results in upwards of thousands of lines. In the widget, this can appear as the widget hanging while the algorithm runs. This PR adds another label in which warnings can be displayed.

tracker_warning

Things to note
- The cutoff value for "low threshold value" is not well known at the moment, but some tests on normal sized kymographs suggests this slowdown could appear at values corresponding to the 95th percentile of the data. However this value is likely variable based on SNR and kymo length/size. Other suggestions for cutoffs are welcome
- I used an `HTML` widget rather than the simple `Label` in order to be able to differentiate this warning label (with red font) from the existing status label.
- The ticket mentions fixups to the value used in the Harbor notebook and docs; however, with that example dataset the value stated is actually what you would get if you didn't manually specify the default slider starting value. So I think this fixup is best left for when we change the dataset out for newer better quality data
- I also piggybacked removing some extraneous code dealing with panning. I can put this in a separate PR if better

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -62,7 +62,7 @@ class KymoWidget:
         self.plotted_lines = []
         self._kymo = kymo
         self._channel = channel
-        self._label = None
+        self._labels = {}
         self._fig = None
         self._axes = None
         self.adding = True
@@ -242,9 +242,9 @@ class KymoWidget:
                 self.output_filename,
                 sampling_width=int(np.ceil(0.5 * self._line_width_pixels)),
             )
-            self._set_label(f"Saved {self.output_filename}")
+            self._set_label("status", f"Saved {self.output_filename}")
         except (RuntimeError, IOError) as exception:
-            self._set_label(str(exception))
+            self._set_label("status", str(exception))
 
     def save_lines(self, filename, delimiter=";", sampling_width=None):
         """Export KymoLineGroup to a csv file.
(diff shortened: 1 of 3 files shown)


tiagobonetti (approved):
Looks good to me.
Now I understand why you dislike `ipywidgets`. The API is questionable 😅 .

`honk`

JoepVanlier (changes requested):
I like the addition of having a warning label, but I noticed that if you open the widget with the threshold set below the limit, it doesn't show the warning. I think it'd be good if on first load, you check the conditions once as well.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -603,14 +597,31 @@ def wrapped_track_greedy(kymo, channel, min_length, **kwargs):
...
+        image = self._kymo.get_image(self._channel)
+        min_threshold = np.percentile(image, 80)
+
+        threshold_slider_index = list(self._algorithm_parameters.keys()).index("pixel_threshold")
+        slider_box.children[threshold_slider_index].children[0].observe(
+            lambda change: self._set_label(
+                "warning",
+                ""
+                if change["new"] > min_threshold
+                else f"Tracking with threshold of {change['new']} maybe be slow.",
```
JoepVanlier, line 617: `Tracking with threshold of 1 maybe be slow` sounds a bit odd to me because of the double `be`. Maybe something like `Tracking with threshold of 1 may be slow`

rpauszek: oops, typo. what you suggest is what I meant to put

rpauszek:
> I like the addition of having a warning label, but I noticed that if you open the widget with the threshold set below the limit, it doesn't show the warning. I think it'd be good if on first load, you check the conditions once as well.

Good point, although this leads to some pretty ugly code to get it to work. I think it's worth it in this case too. Also, added a changelog entry.

JoepVanlier (approved):
Cool! Looks good to me. :shipit:

PR #360, opened 2022-08-19
range_selector: force redraw on update

JoepVanlier:
**Why this PR?**
Aafke found that the range selector widget wasn't updating the plot anymore when the dataset was switched. This PR fixes that behaviour by adding a call to `draw_idle` analogously to https://github.com/lumicks/pylake/pull/337 .


rpauszek (approved):
🚀

PR #361, opened 2022-08-19
scans/stack: add slicing by timestamp and string

JoepVanlier:
**Why this PR?**
It's very convenient to slice by time and timestamp, hence we want to extend this functionality to `Scan` and `Stack`.

### DIFF
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -11,6 +11,10 @@ These videos can be opened and sliced using `CorrelatedStack`::
     stack = lk.CorrelatedStack("wf.tiff")  # Loading a stack.
     stack_slice = stack[2:10]  # Grab frame 2 to 9
 
+You can also slice by time::
+
+    stack_slice = stack["2s":"10s"]  # Slice from second second up to tenth second
+
 You can easily load multiple TIFF files by simply listing them consecutively::
 
     stack = lk.CorrelatedStack("wf.tiff", "wf2.tiff")  # Loading two tiff files in a single stack.
--- a/docs/tutorial/correlatedstacks.rst
+++ b/docs/tutorial/correlatedstacks.rst
@@ -13,7 +13,7 @@ These videos can be opened and sliced using `CorrelatedStack`::
 
 You can also slice by time::
 
-    stack_slice = stack["2s":"10s"]  # Slice from second second up to tenth second
+    stack_slice = stack["2s":"10s"]  # Slice from 2 to 10 seconds from beginning of the stack
 
 You can easily load multiple TIFF files by simply listing them consecutively::
 
--- a/docs/tutorial/images.rst
+++ b/docs/tutorial/images.rst
@@ -55,6 +55,16 @@ This will return a new `Scan` containing data equivalent to::
 
     multiframe_scan.get_image("rgb")[5:10, :, :, :]
 
+We can also slice the frames by time::
+
+    # get frames corresponding to the time range 5 through 10 seconds
+    sliced_scan = multiframe_scan["5s":"10s"]
+
+Or directly using timestamps::
+
+    # get frames that fall between the start and stop of a force channel
+    multiframe_scan[f.force1x.start:f.force1x.stop]
+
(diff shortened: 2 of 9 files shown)


lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -72,3 +77,30 @@ def plot(num):
         )
         line_ani.save(file_name, writer=writer)
         plt.close(fig)
+
+
+class FrameIndex:
+    def to_frame_index(self, time, use_start=False) -> Union[int, None]:
+        """Convert timestamps, timestamps or frame indices to frame indices
```
rpauszek, line 84: rebasing mishap?
```suggestion
        """Convert timestamps to frame indices
```

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -72,3 +77,30 @@ def plot(num):
...
+
+        Parameters
+        ----------
+        time : str or int or None
+            Time expressed in either string format (e.g. "5s"), timestamp in nanoseconds or frame.
+        use_start : bool
+            Are we dealing with the start or end of a frame here?
```
rpauszek, line 91: The term "use" here seems odd, I wasn't quite sure what the parameter meant. I think "is" might be better and more in line with convention for naming boolean args.

The description seems a bit off also; until I looked at the usage in the code I was thinking this arg meant something completely different. Not sure if I quite got the intended usage, but how about:
```suggestion
        is_frame_start : bool
            Whether the input time refers to the start or stop timestamp of the searched frame
```

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -72,3 +77,30 @@ def plot(num):
         )
         line_ani.save(file_name, writer=writer)
         plt.close(fig)
+
+
+class FrameIndex:
+    def to_frame_index(self, time, use_start=False) -> Union[int, None]:
```
rpauszek, line 83: should be private API?

also perhaps `_time_to_frame_index` might be more descriptive?

rpauszek (changes requested):
Looks pretty good to me. Just some concerns with the descriptions/docstrings I found confusing and future me will likely be completely lost, but let's see if you agree.

The big thing is I think the new method introduced through the mixin should be private API

docs/tutorial/correlatedstacks.rst
```diff
@@ -11,6 +11,10 @@ These videos can be opened and sliced using `CorrelatedStack`::
     stack = lk.CorrelatedStack("wf.tiff")  # Loading a stack.
     stack_slice = stack[2:10]  # Grab frame 2 to 9
 
+You can also slice by time::
+
+    stack_slice = stack["2s":"10s"]  # Slice from second second up to tenth second
```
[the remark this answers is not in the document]
JoepVanlier: second second 😅

docs/tutorial/images.rst
```diff
@@ -55,6 +55,14 @@ This will return a new `Scan` containing data equivalent to::
 
     multiframe_scan.get_image("rgb")[5:10, :, :, :]
 
+We can also slice the frames by time::
+
+    sliced_scan = multiframe_scan["5s":"10s"]
```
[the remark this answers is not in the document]
JoepVanlier: How would you feel about `# get frames corresponding to the time range 5 through 10 seconds`?

lumicks/pylake/detail/imaging_mixins.py
JoepVanlier: I agree with your suggestion. I think your names and docstring description are better here.

lumicks/pylake/detail/imaging_mixins.py
JoepVanlier: Yes, should definitely be private API. Good catch!

docs/tutorial/images.rst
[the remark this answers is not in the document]
rpauszek: yup that sounds good

rpauszek (approved):
cool, I think it's good to go

PR #363, opened 2022-08-25
ci: clear warnings caused by `mpl` argument deprecation

JoepVanlier:
**Why this PR?**
We've been getting a warning from `mpl` for a while. Since it only involves a simple parameter rename, I figured it's fine to support both for a while. I added a comment so that we don't forget to remove it once we upgrade to `>= 3.5.0`.

I also took the liberty of adding `Werror` to our `pytest` script, so that warnings are treated as errors from now on. Hopefully this'll make us a bit more proactive about clearing these out.


rpauszek (approved):
LGTM

PR #364, opened 2022-08-26
Fixup API docs

rpauszek:
**Why this PR?**
There are a number of problems with how our API docs stubs are currently built. 
- Regardless of the `autodoc` settings to hide private methods, these still show up. This seems to be because we use `autosummary` to build the TOC
- The methods section for each page is just a table with the method name and first line of the docstring; the actual docstrings including parameters aren't shown anywhere. 
- Linking to specific methods of a documented class requires adding a separate entry in the TOC so that a specific stub will be built for that method (you can't link to a row in the methods table).

The only solution I could find was to write custom templates which directly use `autodoc` directives; the settings from `conf.py` then get applied correctly. I also had to switch from `numpydoc` to the `napolean` extension to get formatting with the `deprecated` directive to display properly (although, see note below). This extension is also the preferred way to convert from numpy docstrings to rst according to the Sphinx documentation; it comes bundled with Sphinx already.

With the proposed change, each page for the class lists the full docstring for each method (including examples) and private API is not displayed (except for certain special methods which we turn on in the config). Linking to specific methods of a class also works properly now. A before and after example is below and full [docs are here](https://lumicks-pylake.readthedocs.io/en/docs_api/api.html).

**Before**
image

**After**
image

There are still a couple things that are less than optimal
- The formatting for the the parameter list for deprecated methods is slightly off (see [Kymo.plot_rgb()](https://lumicks-pylake.readthedocs.io/en/docs_api/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_rgb) for example, compare with the method below it). This also leads to warnings being emitted; it seems to be that with the inserted `.. deprecated` directive, the `Parameters` section title isn't recognized. I tried to figure out exactly what is going on but with no luck. For me, the benefits here far outweigh this minor formatting error and a couple warnings being emitted; especially for things that will be removed after a short time.
- The constructor signature is displayed even though most classes we don't want users constructing instances manually. This is actually true for what we have now. My idea is to use the `.. Attention:: ` directive in these docstrings to add a note explicitly mentioning that objects shouldn't be constructed by users but obtained in other ways (eg. `get a Kymo from File.kymos`). I didn't piggyback this here though. 
- With the proper display of all docstrings, there's a lot of apparent formatting mistakes on our part. I also didn't piggy back these needed changes here.
- Again, now that docstrings are displayed properly, there can be some less than ideal formatting for type annotations. For example see [Slice.data](https://lumicks-pylake.readthedocs.io/en/docs_api/_api/lumicks.pylake.channel.Slice.html#lumicks.pylake.channel.Slice.data) which is annotated with `npt.ArrayLike`

Again, even with all of these downsides, I still think the benefits of proper API documentation are worthwhile

### DIFF
--- a/docs/_templates/class.rst
+++ b/docs/_templates/class.rst
@@ -1,0 +1,10 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autoclass:: {{ objname }}
+   :members:
+   :special-members:
+   :inherited-members:
+
+
--- a/docs/_templates/function.rst
+++ b/docs/_templates/function.rst
@@ -1,0 +1,5 @@
+{{ fullname | escape | underline}}
+
+.. currentmodule:: {{ module }}
+
+.. autofunction:: {{ objname }}
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -7,6 +7,7 @@ If you are just looking to get started, read the :doc:`/tutorial/index` first.
 .. currentmodule:: lumicks.pylake
 
 .. autosummary::
+    :template: class.rst
     :toctree: _api
 
     File
@@ -24,12 +25,16 @@ Force calibration
 
 .. autosummary::
     :toctree: _api
+    :template: class.rst
+
+    PassiveCalibrationModel
+    ActiveCalibrationModel
+
+    :template: function.rst
 
     calibrate_force
     calculate_power_spectrum
     fit_power_spectrum
-    PassiveCalibrationModel
-    ActiveCalibrationModel
     force_calibration.power_spectrum.PowerSpectrum
     force_calibration.power_spectrum_calibration.CalibrationResults
     viscosity_of_water
@@ -40,9 +45,12 @@ FD Fitting
 
 .. autosummary::
     :toctree: _api
+    :template: class.rst
 
     fitting.model.Model
     FdFit
+
+    :template: function.rst
     parameter_trace
 
 .. _fd_models:
@@ -50,6 +58,7 @@ FD Fitting
 
 .. autosummary::
     :toctree: _api
+    :template: function.rst
 
     force_offset
     distance_offset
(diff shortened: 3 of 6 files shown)


aafkevandenberg (commented):
Thanks Ray, the updated formatting makes the docs much more readable and clear! 

And indeed, this does reveal some formatting errors, would be nice to fix those before the release of 1.0, eg:
- change formatting `Slice.data` and `Slice.timestamps`
- Would be nice to fix the links to internal classes or methods
- There is an example code for the `Kymo` Class that is not formatted as code.
- ...

JoepVanlier (commented):
I've been working through the warnings this Friday, to see if I can clear some of the formatting issues up in main, so that we can get a clean warning list in this PR.

There are indeed also a few things that we should ideally fix now that these docstrings become more visible to end users.
image

tobiasjj:
Great! I really appreciate the changes of the API docs and despite the remaining issues I think this greatly improves the readability and usability.

Regarding the weird formatting of `numpy.typing` return types, like in [Slice.data](https://lumicks-pylake.readthedocs.io/en/docs_api/_api/lumicks.pylake.channel.Slice.html#lumicks.pylake.channel.Slice.data), I might have found a viable solution:
Autodoc is automatically looking up the types defined by `npt.ArrayLike` and rendering it into an unreadable list of all types defined by it. One can prevent autodoc from doing so, by specifically defining an alias for `npt.ArrayLike` (and others) with the parameter [`autodoc_type_aliases`](https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#confval-autodoc_type_aliases). This would require the following steps:
* Add `from __future__ import annotations` to the top of the module that defines the return type (i.e. `channel.py`) to enable [Postponed Evaluation of Annotations (PEP 563)](https://peps.python.org/pep-0563/)
* Add `autodoc_type_aliases = {
    'npt.ArrayLike': 'npt.ArrayLike'
}` to `docs/conf.py`

This would require us to maintain an extra list of types we want to not be evaluated in `conf.py`, but probably this is worth the readability gain? What do you think?

JoepVanlier (approved):
Great work!

With the fix Tobias suggests and after playing with this for a bit, I'm pretty confident we can clear out the remaining issues (see also https://github.com/lumicks/pylake/pull/367 for an initial fixup of a bunch of issues).

It'll be really nice to actually have API docs :)

PR #365, opened 2022-08-29
"line" to "track" - part 2 / 3

rpauszek:
**Why this PR?**
Continuing with the terminology refactor. This PR tackles functions in `kymotracker.py` and some things in the widget to ensure the tests pass. There are some additional breaking changes that I've included since we're on one of those releases anyway and it cut down on some ugliness wrt argument order in the function signatures
- added defaults to the pixel threshold and track width args for `track_greedy`; this allowed for an aliasing/deprecation pathway for the former `line_width` argument
- `KymoWidget` and `KymoWidgetGreedy` now enforce keyworded arguments for everything beyond the first 2 (`kymo` and `channel`). Without this change, I would have had to put the new `track_width` parameter at the end of the constructor signature and would end up separated from the rest of the algorithm parameters
- the new `refine_tracks_centroid()` accepts `track_width` in physical units whereas the deprecated `refine_lines_centroid()` expects `line_width` to be in pixel units. This change was made so that `track_greedy`, `refine_tracks_centroid` and the widget all use physical units

I tried to keep the commits organized and fairly well separated. Again, most can be squashed before merging

Next and final PR will tackle the remaining terminology changes necessary in the widget

### DIFF
--- a/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
+++ b/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
@@ -102,7 +102,7 @@ view of our data.
 
 We will be using the greedy algorithm. The `threshold` should typically be chosen somewhere between the expected
 baseline photon count and the photon count of a true track (note that you can see the local photon count between square
-brackets while hovering over the kymograph). The `line width` should roughly be set to the expected line width (in the spatial dimension) of a
+brackets while hovering over the kymograph). The `track width` should roughly be set to the expected spot size (in the spatial dimension) of a
 track. The `window` should be chosen such that small gaps in a track can be overcome, but not so large that spurious
 points may be strung together as a track. `Sigma` controls how much the location can fluctuate from one time point to the
 next, while the `min length` determines how many peak points should be in a track for it to be considered a valid track.
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -49,7 +49,7 @@ use the region from 7 to 22 micrometer. Let's have a look at what that looks lik
     >>> plt.ylim([22, 7])
 
 If we zoom in a bit, we can see that our tracks in this image are about 0.3 microns wide, so let's set this as our
-`line_width`. Note that we invoked `plt.colorbar()` to add a little color legend here.
+`track_width`. Note that we invoked `plt.colorbar()` to add a little color legend here.
 
 .. image:: kymo_zoom.png
 
@@ -60,7 +60,7 @@ Considering that we only want to run the kymotracker on part of the image, we al
 a rectangle over which to track peaks. In this case, we track particles between 0 and 350 seconds, and 7 and 22 micron.
 Running the algorithm is easy using the function :func:`~lumicks.pylake.track_greedy`::
 
-    tracks = lk.track_greedy(kymo, "green", line_width=0.3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
+    tracks = lk.track_greedy(kymo, "green", track_width=0.3, pixel_threshold=3, window=6, rect=[[0, 7], [350, 22]])
 
 The result of tracking is a list of kymograph tracks::
 
@@ -68,10 +68,10 @@ The result of tracking is a list of kymograph tracks::
     7417
 
 Sometimes, we can have very short spurious tracks. To remove these from the list of detected tracks we can use
-:func:`~lumicks.pylake.filter_lines`. To omit all tracks with fewer than 4 detected points, we
+:func:`~lumicks.pylake.filter_tracks`. To omit all tracks with fewer than 4 detected points, we
 can invoke::
 
-    >>> tracks = lk.filter_lines(tracks, 4)
+    >>> tracks = lk.filter_tracks(tracks, 4)
 
 We can get the determined time and positions of these tracks from the `seconds` and `position`
 attributes. Let's plot the first track::
@@ -94,14 +94,14 @@ Centroid
 
 Once we are happy with the tracks found by the algorithm, we may still want to refine them. Since the algorithm finds
 tracks by determining local peaks and stringing these together, it is possible that some scan lines in the kymograph
-don't have an explicit point on the track associated with them. Using :func:`~lumicks.pylake.refine_lines_centroid` we
+don't have an explicit point on the track associated with them. Using :func:`~lumicks.pylake.refine_tracks_centroid` we
 can refine the tracks found by the algorithm. This function interpolates the tracks such that each time point gets its
 own point on the track. Subsequently, these points are then refined using a brightness weighted centroid. Let's perform
-line refinement and plot the longest track::
+track refinement and plot the longest track::
 
     longest_track_idx = np.argmax([len(track) for track in tracks])  # Get the longest track
 
-    refined = lk.refine_lines_centroid(tracks, line_width=0.3)
+    refined = lk.refine_tracks_centroid(tracks, track_width=0.3)
 
     plt.plot(refined[longest_track_idx].seconds, refined[longest_track_idx].position, '.')
     plt.plot(tracks[longest_track_idx].seconds, tracks[longest_track_idx].position, '.')
@@ -112,7 +112,7 @@ line refinement and plot the longest track::
 .. image:: kymo_refine.png
 
 We can see now that a few points were added post refinement (shown in blue). The others remain unchanged, since we used
-the same `line_width`.
+the same `track_width`.
 
 Fortunately, the signal to noise level in this kymograph is quite good. In practice, when the signal to noise is lower,
 one will have to resort to some fine tuning of the algorithm parameters over different regions of the kymograph to get
@@ -121,7 +121,7 @@ an acceptable result.
 Maximum Likelihood Estimation
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-The function `refine_lines_gaussian()` instead uses an MLE optimization of a Poisson likelihood with a Gaussian expectation
+The function `refine_tracks_gaussian()` instead uses an MLE optimization of a Poisson likelihood with a Gaussian expectation
 to characterize both the expected peak shape and photonic noise of the observed signal, adapted from :cite:`mortensen2010gauloc`.
 For each frame in the kymograph, we fit a small region around the tracked peak to the data by maximizing the following likelihood function:
 
@@ -142,7 +142,7 @@ photons/pixel.
 
 This function is called in a similar manner as the centroid refinement::
 
-    refined = lk.refine_lines_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip")
+    refined = lk.refine_tracks_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip")
 
 The number of pixels to be included in the fit is determined by the `window` argument, with a total size of `2*window+1` pixels.
 The exact value of this parameter is dependent on the quality of the data and should be balanced between including enough pixels to fully
@@ -179,9 +179,9 @@ Computing the appropriate photons/pixel background considering a Poissonian nois
 
     offset = np.mean(kymo_cropped.green_image)
 
-The independently determined offset (in photons per pixel) can then be provided directly to `lk.refine_lines_gaussian`::
+The independently determined offset (in photons per pixel) can then be provided directly to `lk.refine_tracks_gaussian`::
 
-    refined = lk.refine_lines_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip", fixed_background=offset)
+    refined = lk.refine_tracks_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip", fixed_background=offset)
 
 In this case the parameter will not be fitted, but fixed to the user specified value.
 This can help reduce the variance of the parameter estimates.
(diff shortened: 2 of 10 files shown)


mikhas (approved):
Looks good to me. I like the methodological approach to renaming & phasing out the old names, doing it commit by ommit.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -64,20 +65,20 @@ def track_greedy(
...
     ----------
     kymograph : lumicks.pylake.Kymo
-        Kymograph.
-    channel : str
-        Kymograph channel.
-    line_width : float or None
+        The kymograph to track.
+    channel : {'red', 'green', 'blue'}
+        Color channel to track.
+    track_width : float
```
JoepVanlier, line 88: or `None` ?

You explicitly state in the string describing it that it's an option.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -106,13 +107,15 @@ def track_greedy(
...
-        belongs to this trace. Lower values result in traces being more stringent in terms of
+        belongs to this track. Lower values result in tracks being more stringent in terms of
         continuing (default: 2.0).
     rect : tuple of two coordinates
         Only perform tracking over a subset of the image. When this argument is supplied, the peak
         detection and refinement is performed over the full image, but the results are then filtered
         to omit the peaks that fall outside of the rect. Coordinates should be given as:
         ((min_time, min_coord), (max_time, max_coord)).
+    line_width : float
+        **Deprecated** Forwarded to track_width.
```
JoepVanlier: `track_width`

JoepVanlier: ah... already done I see later

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
```diff
@@ -77,3 +77,15 @@ def test_greedy_algorithm_input_validation(kymo_integration_test_data):
...
+def test_default_parameters(kymo_integration_test_data):
+    ref_tracks = track_greedy(kymo_integration_test_data, "red", 0.35, 1)
+
+    tracks = track_greedy(kymo_integration_test_data, "red", line_width=None, pixel_threshold=1)
+    for ref, track in zip(ref_tracks, tracks):
+        np.testing.assert_allclose(ref.position, track.position)
+
+    tracks = track_greedy(kymo_integration_test_data, "red", line_width=0.35, pixel_threshold=None)
+    for ref, track in zip(ref_tracks, tracks):
+        np.testing.assert_allclose(ref.position, track.position)
```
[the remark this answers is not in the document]
rpauszek: good catch, added

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -579,6 +594,17 @@ def wrapped_track_greedy(kymo, channel, min_length, **kwargs):
         # update defaults to user-supplied parameters
         arg_names, _, _, values = inspect.getargvalues(inspect.currentframe())
         parameters = {name: values[name] for name in arg_names if name != "self"}
+
+        # TODO: remove line_width argument deprecation path
+        if line_width is not None:
```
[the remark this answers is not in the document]
JoepVanlier: I think otherwise you wouldn't see it until you actually track, which might be annoying.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -64,20 +65,20 @@ def track_greedy(
     together using a greedy forward search analogous to [2]. This in contrast with the linking
     algorithm in [1] which uses a graph based optimization approach.
 
-    The linking step traverses the kymograph, tracing lines starting from each frame. It starts with
-    the highest line and proceeds to lines with lower signal intensity. For every point along the
-    line, the algorithm makes a prediction for where the particle will be in the next frame. Points
-    are considered candidates for line membership when they fall within a cone parameterized by a
+    The linking step traverses the kymograph, tracing tracks starting from each frame. It starts with
+    the highest intensity track and proceeds to tracjs with lower signal intensity. For every point along the
```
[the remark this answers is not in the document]
rpauszek: oops, type

lumicks/pylake/nb_widgets/tests/test_kymotracker_widget.py
```diff
@@ -241,7 +241,9 @@ def test_stitch_anywhere(start, stop, same_line, kymograph, mockevent):
 
 
 def test_refine_line_width_units(kymograph, region_select):
-    kymo_widget = KymoWidgetGreedy(kymograph, "red", 1, line_width=2, use_widgets=False)
+    kymo_widget = KymoWidgetGreedy(
+        kymograph, "red", axis_aspect_ratio=1, line_width=2, use_widgets=False
```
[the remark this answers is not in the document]
rpauszek: this is fixed in a later commit

lumicks/pylake/kymotracker/tests/test_refinement.py
```diff
@@ -77,16 +80,20 @@ def test_refinement_line(loc, inv_sigma=0.3):
...
-def test_refinement_line(kymo_integration_test_data):
-    with pytest.raises(ValueError, match="line_width may not be smaller than 1"):
-        refine_lines_centroid([KymoTrack([0], [25], kymo_integration_test_data, "red")], 0)[0]
+def test_refinement_error(kymo_integration_test_data):
+    with pytest.raises(ValueError, match="track_width should be larger than zero"):
+        refine_tracks_centroid([KymoTrack([0], [25], kymo_integration_test_data, "red")], 0)[0]
+
+    with pytest.warns(DeprecationWarning):
+        with pytest.raises(ValueError, match="line_width may not be smaller than 1"):
+            refine_lines_centroid([KymoTrack([0], [25], kymo_integration_test_data, "red")], 0)[0]
```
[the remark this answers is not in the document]
JoepVanlier: Wanna make sure they still work until they're fully gone? Can remove it when we actually remove the function.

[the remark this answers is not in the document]
rpauszek: the first one handles the possible `"line_width"` key in the `slider_ranges` dict arg, the second handles the actual `line_width` arg

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -296,7 +296,9 @@ def set_value(value):
 
     def refine(self):
         if self.lines:
-            self.lines = refine_lines_centroid(self.lines, self._line_width_pixels)
+            self.lines = refine_tracks_centroid(
+                self.lines, self._algorithm_parameters["track_width"].value
+            )
             self.update_lines()
```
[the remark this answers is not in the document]
rpauszek: this and everything else dealing with the widget is taken care of in the next PR

lumicks/pylake/nb_widgets/kymotracker_widgets.py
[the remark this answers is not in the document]
rpauszek: the first handles the possible `"line_width"` key in the `slider_ranges` dict arg, the second handles the actual `line_width` arg

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -339,26 +348,60 @@ def refine_lines_centroid(lines, line_width):
...
+
+    Parameters
+    ----------
+    tracks : List[pylake.KymoTrack]
+        Detected tracks on a kymograph
+    track_width : float
+        Expected spatial track width in physical units. Must be larger than zero.
+        If `None`, the default is 0.35 (half the wavelength of the red limit of the visible spectrum)
+        for kymographs calibrated in microns and 1 for kymographs calibrated in kilobase pairs (based
+        on 0.34 nm/bp for duplex DNA)
```
JoepVanlier, line 372: I know we discussed this two days ago, but the double default did get me thinking, it does seem a bit odd that we use a different default depending on how the `Kymograph` is calibrated. For pixels, I totally get it, because you have no physical bearing, but for `kbp`, maybe it would make sense to have the default for kbp be `0.34/0.35`? Since that's what you actually expect for the line width, no?

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -280,6 +282,13 @@ def track_lines(
     )
 
 
+@deprecated(
```
JoepVanlier, line 292: The parameter list will throw a warning on the docs build with `napoleon`. Should probably axe it.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -321,6 +325,11 @@ def filter_tracks(tracks, minimum_length):
     return KymoTrackGroup([track for track in tracks if len(track) >= minimum_length])
 
 
+@deprecated(
```
JoepVanlier, line 329: Same as above.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -434,12 +440,52 @@ def refine_lines_gaussian(
...
+    fixed_background=None,
+):
+    """Refine the tracks by gaussian peak MLE.
+
+    Parameters
+    ----------
+    tracks : List[pylake.KymoTrack] or pylake.KymoTrackGroup
+        Detected tracks on a kymograph.
+    window : int
+        Number of pixels on either side of the estimated track to include in the optimization data.
```
JoepVanlier, line 468: To me, it seems strange that now we have one refinement method in pixels and one in physical units. I would make these consistent.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: I think the documentation should also clearly mention the rounding strategy used when determining how many pixels to use (`ceil`).

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -102,7 +102,7 @@ view of our data.
 
 We will be using the greedy algorithm. The `threshold` should typically be chosen somewhere between the expected
 baseline photon count and the photon count of a true track (note that you can see the local photon count between square
-brackets while hovering over the kymograph). The `line width` should roughly be set to the expected line width (in the spatial dimension) of a
+brackets while hovering over the kymograph). The `track width` should roughly be set to the expected spot size (in the spatial dimension) of a
```
JoepVanlier, line 105: Spot size is so much better :)

docs/tutorial/kymotracking.rst
```diff
@@ -94,14 +94,14 @@ Centroid
...
+don't have an explicit point on the track associated with them. Using :func:`~lumicks.pylake.refine_tracks_centroid` we
 can refine the tracks found by the algorithm. This function interpolates the tracks such that each time point gets its
 own point on the track. Subsequently, these points are then refined using a brightness weighted centroid. Let's perform
-line refinement and plot the longest track::
+track refinement and plot the longest track::
 
     longest_track_idx = np.argmax([len(track) for track in tracks])  # Get the longest track
 
-    refined = lk.refine_lines_centroid(tracks, line_width=0.3)
+    refined = lk.refine_tracks_centroid(tracks, track_width=0.3)
```
JoepVanlier, line 104: Shouldn't the width now be different?

docs/tutorial/kymotracking.rst
```diff
@@ -179,9 +179,9 @@ Computing the appropriate photons/pixel background considering a Poissonian nois
 
     offset = np.mean(kymo_cropped.green_image)
 
-The independently determined offset (in photons per pixel) can then be provided directly to `lk.refine_lines_gaussian`::
+The independently determined offset (in photons per pixel) can then be provided directly to `lk.refine_tracks_gaussian`::
 
-    refined = lk.refine_lines_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip", fixed_background=offset)
+    refined = lk.refine_tracks_gaussian(tracks, window=3, refine_missing_frames=True, overlap_strategy="skip", fixed_background=offset)
```
JoepVanlier, line 184: The window here is the same as with `centroid`; yet in the current implementation, one is defined in physical units and one is not. Seems something is wrong?

JoepVanlier (changes requested):
Quite the rename bonanza :)

Thanks for picking this up. It'll be great to have consistent naming across the board.

I still have a few concerns though, but you'll see those in the comments.

docs/tutorial/kymotracking.rst
rpauszek: It was wrong before because recently I changed it not realizing that `refine_lines_centroid()` expected pixel value. I think having everything working in consistent units now is a lot less confusing

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
```diff
@@ -77,3 +77,15 @@ def test_greedy_algorithm_input_validation(kymo_integration_test_data):
     for pixel_threshold in (-1, 0):
         with pytest.raises(ValueError, match="should be larger than zero"):
             track_greedy(test_data, "red", line_width=10, pixel_threshold=pixel_threshold)
+
+
+def test_default_parameters(kymo_integration_test_data):
```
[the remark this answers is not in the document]
rpauszek: good call. had to generate some new test data but adding these actually caught some bugs!

lumicks/pylake/kymotracker/kymotracker.py
rpauszek: for the first one, I hardcoded in the value because apparently I forgot that in python you can put math in your code 😒 

For the second comment, this sounds ok but I'll do it in a separate PR since this one is already getting a bit crazy

JoepVanlier: Right. Makes sense. 👍

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Right. In that case it's actually good that it has a different name and the user's attention is drawn to it a bit. Fully agree with your assessment.

JoepVanlier (approved):
image

PR #366, opened 2022-08-29
channel: fixed `TimeSeries.from_dataset()` not lazy loading data

tobiasjj:
The method `TimeSeries.from_dataset()` automatically loaded the data and timestamps from an HDF5 file upon creation of a `TimeSeries` instance. This is now avoided by creating the `TimeSeries` instance using the `h5py.Dataset.fields()` method of data and timestamps.

Just to add more information, how I realized this:

```
from lumicks import pylake as lk

filepath = "./data/Lakeview Showcase Data-20220616T152657Z-001/FD Curve_forw_DNA.h5"
file = lk.File(filepath)

a = file.h5["Distance"]["Distance 1"]["Value"]
b = file.h5["Distance"]["Distance 1"].fields("Value")

type(a), type(b)
```

The output of this is `(numpy.ndarray, h5py._hl.dataset.FieldsWrapper)`, so `a` is a numpy array directly loaded into memory, whereas `b` is a `FieldsWrapper`, enabling the lazy loading.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@
 * Added possibility to access property `sample_rate` for `TimeSeries` data with constant sample rate.
 * Added shape property to `Scan` and `Kymo`.
 * Added a warning to the Kymotracker widget if the threshold parameter is set too low, which may result in slow tracking and the widget hanging.
+* Pylake now depends on `h5py>=3.4, <4`. This change is required to still support `len()` with the lazy loading fix for `TimeSeries`.
 
 #### Bug fixes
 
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -30,6 +30,7 @@
 * Force distance models now have a non-zero lower bound for the contour length (`Lc`), persistence length (`Lp`), stretch modulus (`St`) and boltzmann constant times temperature (`kT`) instead of a lower bound of zero.
 * Force distance fits now raise a `RuntimeError` if any of the returned simulation values are NaN.
 * Fixed bug that resulted in profile likelihood automatically failing when an attempted step exceeded the bounds where the model could be simulated.
+* Fix method `TimeSeries.from_dataset()` reading all data from datasets upon creation of a `TimeSeries` instance instead of lazy loading
```
ab412, line 33: *Fixed

ab412 (approved):
The `pytest`  on my machine said it is working 🙂  Go for launch 🚀

JoepVanlier (changes requested):
If I try to create one of these, it fails when it tries to assert `len(data) == len(timestamps)` in the constructor with `TypeError: object of type 'FieldsWrapper' has no len()`.

Seems that this length property being defined depends on a newer version of `h5py` than we require. See: https://docs.h5py.org/en/stable/whatsnew/3.4.html

tobiasjj:
I bumped the required h5py version to `>=3.4, <4` and also changed the `TimeSeries.__len__()` method to use the uncached source data instead of using the cached `data` property.

JoepVanlier (approved):
Old:
```python
143.9609375
f = file.downsampled_force1x
146.203125
f.data
146.203125
```
New:
```python
143.6953125
f = file.downsampled_force1x
144.4921875
f.data
146.12109375
```

Works as advertised! :shipit:

PR #367, opened 2022-08-29
docs: fixing docs build warnings

JoepVanlier:
**Why this PR?**
With the API docs becoming visible, so does a bunch of our sins against the format. I noticed that when building the docs on the napoleon branch that many of the docs were throwing warnings, so I took the liberty of fixing a lot of the small issues that were there.

Whenever relevant, I also improved small parts of docstrings to hopefully make them a bit more useful to users.

A few not so nice things to note:
- It seems that our current doc building system does not play nice when we use a deprecation notice on a function that has a list of parameters. For now, I have removed these (in two places), since people should be using the new API anyhow. I am open to suggestions on how to make it play nice however.
- Deprecation notice also don't play nice with functions that don't have docstrings. In this case, I have just provided the function with a docstring.
- Docstring references are a bit of an issue. Having references with the same value end up on the same RST page results in warnings and broken references. I spent a while exploring options here, but didn't find anything particularly good ones.
a. I tried using footnotes, but they are also globally defined.
b. Manually make sure that no duplicates exist for any single class (you'll see this in `KymoTrack` for instance). The ugly downside is that if you ask for the docstring of a method here, the reference counting may start at an (from the user perspective) arbitrary number.
c. I tried hijacking the sphinx post-processor for this. Basically make the docstring just have regular values, but then in the sphinx post-processor rename them with a regular expression replacement. The problem here is that most references are multiple lines and the indent overhangs like so, so if you start replacing those counts with different numeric values, you need to adjust the spacing of the lines below it. The issue then comes with detecting the last line to do this for. It's certainly possible, but feels like it's going to be very brittle (kept finding corner cases and even though they were fixable, don't feel comfortable putting that in this PR).
```
[1] R. Pauszek, J. Vanlier, T. Jachowski, "what a lovely reference
    this is man", Journal of Questionable Sphinx Hacks, 2022.
```

At the time of this writing, the history is warning free on my local build :)

Docs build here: https://lumicks-pylake.readthedocs.io/en/docs_fixups/api.html#

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -20,7 +20,7 @@ class Slice:
     Parameters
     ----------
     data_source : Any
-        A slice data source. Can be `Continuous`, `TimeSeries`, 'TimeTags',
+        A slice data source. Can be `Continuous`, `TimeSeries`, `TimeTags`,
         or any other source which conforms to the same interface.
     labels : Dict[str, str]
         Plot labels: "x", "y", "title".
@@ -195,10 +195,15 @@ class Slice:
 
     @property
     def calibration(self) -> list:
-        """Calibration data slicing is deferred until calibration is requested to avoid
-        slicing values that may be needed."""
+        """List of force calibration items
+
+        The first element represents the calibration item that was active when this slice was
+        made.
+        """
         if self._calibration:
             try:
+                # Calibration data slicing is deferred until calibration is requested to avoid
+                # slicing values that may be needed.
                 return self._calibration.filter_calibration(self._src.start, self._src.stop)
             except IndexError:
                 return []
@@ -225,9 +230,9 @@ class Slice:
             )
 
     def downsampled_over(self, range_list, reduce=np.mean, where="center"):
-        """Downsample channel data based on timestamp ranges. The downsampling function (e.g. np.mean) is evaluated for
-        the time between a start and end time of each block. A list is returned that contains the data corresponding to
-        each block.
+        """Downsample channel data based on timestamp ranges. The downsampling function (e.g.
+        np.mean) is evaluated for the time between a start and end time of each block. A list is
+        returned that contains the data corresponding to each block.
 
         Parameters
         ----------
@@ -240,9 +245,11 @@ class Slice:
             cases, e.g. photon counts.
         where : str
             Where to put the final time point.
-            'center' time point is put at (timestamps_subset[0] + timestamps_subset[-1]) / 2, where timestamps_subset
-            are the timestamps corresponding to the samples being downsampled over.
-            'left' time point is put at start
+
+            - "center" : The new time points are set to
+              `(timestamps_subset[0] + timestamps_subset[-1]) / 2`, where `timestamps_subset` are
+              the timestamps corresponding to the samples being downsampled over.
+            - "left" : Time points are set to the starting timestamp of the downsampled data.
 
         Examples
         --------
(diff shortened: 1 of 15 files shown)


lumicks/pylake/correlated_stack.py
```diff
@@ -300,6 +299,7 @@ def crop_and_rotate(self, frame=0, channel="rgb", show_title=True, **kwargs):
         """Open a widget to interactively edit the image stack.
 
         Actions include:
+
```
rpauszek, line 302: I think it would be nice here to define our own `Actions` header to get some nice formatting. You can use `napoleon_custom_settings` in `conf.py`

https://sphinxcontrib-napoleon.readthedocs.io/en/latest/sphinxcontrib.napoleon.html#sphinxcontrib.napoleon.Config

JoepVanlier: Ah yes, great idea, that'd be cool. I can add that later this afternoon.

JoepVanlier: Added!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -260,11 +261,12 @@ def __len__(self):
         return len(self.coordinate_idx)
 
     def msd(self, max_lag=None):
-        """Estimate the Mean Square Displacement (MSD) for various time lags.
+        r"""Estimate the Mean Square Displacement (MSD) for various time lags.
 
-        The estimator for the MSD (rho) is defined as:
+        The estimator for the MSD (:math:`\rho`) is defined as:
```
rpauszek, line 266: 👍

rpauszek (approved):
Really nice stuff. Left a couple very superficial comments on some formatting, but otherwise it's good to go. Really happy we'll have pretty API docs now

lumicks/pylake/channel.py
```diff
@@ -240,9 +245,13 @@ def downsampled_over(self, range_list, reduce=np.mean, where="center"):
...
-            'center' time point is put at (timestamps_subset[0] + timestamps_subset[-1]) / 2, where timestamps_subset
-            are the timestamps corresponding to the samples being downsampled over.
-            'left' time point is put at start
+
+            - 'center'
+                The new time points are set to `(timestamps_subset[0] + timestamps_subset[-1]) / 2`,
+                where `timestamps_subset` are the timestamps corresponding to the samples being
+                downsampled over.
+            - 'left'
+                Time points are set to the starting timestamp of the downsampled data.
```
rpauszek, line 254: I think if you fix up the indentation so everything matches it'll get rid of the weird formatting here and be more like the list for the color adjustment


image

image

lumicks/pylake/channel.py
```diff
@@ -298,18 +307,27 @@ def downsampled_to(self, frequency, reduce=np.mean, where="center", method="safe
...
```
rpauszek, line 316: same with the indentation

lumicks/pylake/channel.py
```diff
@@ -298,18 +307,27 @@ def downsampled_to(self, frequency, reduce=np.mean, where="center", method="safe
...
+
+            - 'safe'
+                New sample time must be an exact multiple of the current sample time, else an
+                exception is raised.
+            - 'ceil'
+                Rounds the sample rate up to the nearest frequency which fulfills this condition.
+            - 'force'
+                Downsample data with the target input frequency; this will result in variable
+                sample times and a variable number of sampling contributing to each target sample
+                but must be used for variable-frequency data.
```
rpauszek, line 330: same with the indentation

rpauszek (approved):
one more thing I caught

lumicks/pylake/detail/confocal.py
```diff
@@ -459,9 +459,10 @@ def fast_axis(self):
 
     @property
     def timestamps(self) -> np.ndarray:
-        """Timestamps for image pixels, not for samples
+        """Timestamps for each image pixel.
 
-        The returned array has the same shape as the `*_image` arrays.
+        The returned array has the same shape as the `*_image` arrays. Timestamps are defined
```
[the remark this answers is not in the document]
JoepVanlier: fixed

lumicks/pylake/fitting/fit.py
```diff
@@ -453,7 +465,7 @@ def plot(
             # Have a quick look at what a stiffness of 5 would do to the fit.
             fit.plot("Control", overrides={"DNA/St": 5})
 
-            # When dealing with multiple models in one fit, one has to select the model first when we want to plot.
+            # When dealing with multiple models in one fit, one has to select the model first when  we want to plot.
```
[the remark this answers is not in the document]
JoepVanlier: yeah that should've been a linebreak

lumicks/pylake/fitting/fit.py
```diff
@@ -605,33 +621,38 @@ def aicc(self):
     def bic(self):
         """Calculates the Bayesian Information Criterion:
 
+        .. math::
             BIC = k ln(n) - 2 ln(L)
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/fitting/fit.py
```diff
@@ -562,17 +574,19 @@ def log_likelihood(self, params=[], sigma=None):
     def aic(self):
         """Calculates the Akaike Information Criterion:
 
+        .. math::
             AIC = 2 k - 2 ln(L)
```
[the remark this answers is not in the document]
rpauszek: yeah, I wonder if there's a way to add latex definitions somewhere in the configuration. Would be nice if we could do `\ln{L}`. Or maybe we could do it ourselves with the autodoc processing hood. Obviously not for right now though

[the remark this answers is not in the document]
JoepVanlier: Yeah. We can probably abuse that autodoc hook for this with a regex.

PR #369, opened 2022-08-31
Enforce odd number of pixels in centroid refinement

rpauszek:
**Why this PR?**

Brought up in the comments during review for #365


rpauszek:
Actually, discovered that this is already enforced in the internal function call

PR #370, opened 2022-09-01
ci: test docs build

JoepVanlier:
**Why this PR?**
After the work that went into cleaning up the docs and making sure that everything works, it'd be nice to maintain them in good working order. Therefore I propose this CI test to make sure we don't end up with weird oddities on our API docs again. If it turns out to be too brittle we can remove it again, but I think we should give it a shot.

Example of what a CI run will look like if it fails (here I had deliberately broken a docstring):
image

Note that this PR depends on all the PRs that clear out the warnings we have now before it can be merged and turned into an active test.

### DIFF
--- a/.github/workflows/pylake_docs_test.yml
+++ b/.github/workflows/pylake_docs_test.yml
@@ -1,0 +1,23 @@
+name: docs build
+
+on: [push]
+
+jobs:
+    build:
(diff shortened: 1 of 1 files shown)


rpauszek (approved):
Nice! I agree it's nice to have this check, hopefully we won't run into too many headaches with the docs

PR #372, opened 2022-09-05
packaging: make notebook dependencies optional for `pip`

JoepVanlier:
**Why this PR?**
Notebook used to be installed implicitly with `ipywidgets`. They dropped this requirement and as a result, installing `pylake` won't install `notebook` anymore.

Since installing all the notebook dependencies is still quite convenient, it'd be nice to still have this as an option. People who want to use Pylake with notebooks can install from pip using `pip install lumicks.pylake[notebook]`. This way, people who don't need the notebook dependencies aren't forced to install them. On `conda`, we will continue to ship both `notebook` and `ipywidgets` by default.

Note that this also adds a `ci` test to specifically test `pylake` without the dependency installed.


changelog.md
```diff
@@ -70,6 +70,8 @@
     * `export_kymolinegroup_to_csv()` was renamed to `export_kymotrackgroup_to_csv()`
     * `import_kymolinegroup_from_csv()` was renamed to `import_kymotrackgroup_from_csv()`
 * Updated the `KymoWidgetGreedy` UI to reflect changes in terminology.
+* Make `ipywidgets>=7.0.0` and `notebook>=4.4.1` an optional dependency for `pip`.
+* Make `notebook>=4.4.1` a mandatory dependency for `conda` release.
```
rpauszek, line 74: Should be `Made ...` to keep with the past tense used throughout the changelog

rpauszek (approved):
LGTM, will be nice to have this in the upcoming release. Just one super nit-picky note for the changelog

PR #373, opened 2022-09-06
Unify `export_tiff` for `Scan`, `Kymo` and `CorrelatedStack` and export metadata in all cases

tobiasjj:
**Why this PR?**
As a user, I want to have timestamps and metadata in TIFFs exported from Pylake (Confocal like WIT). As a developer I want to have common code in all different classes that handles exporting tiffs.

**Implementation details**
The new mixin class `TiffExport` implements the method `export_tiff()` to export tiff stacks. It uses the method `_tiff_frames()` to get frames to be written to a TIFF file. The method `_tiff_writer_kwargs()` generates the arguments used for the `TiffWriter.write()` method. The method  `_tiff_image_metadata()` generates the metadata to be written to the ImageDescription field of the exported tiff file pages and the method `_tiff_timestamp_ranges()` is used to generate a list of (start, stop) ranges, to set the (non standard) DateTime field for all pages. The `_tiff_*()` methods are meant to be overwritten by child classes to adjust the metadata to be exported. `CorrelatedStack`, `ConfocalImage`, `Scan` and `Kymo` now implement the new `TiffExport` mixin. I did not change the public facing `export_tiff()` method. I also added/updated the tests to separately test the mixin class `TiffExport` and the functionality of the child classes.

As a consequence, I removed the now unused `save_tiff()` function.

**What is still missing**
Checking of the `Camera` field in `CorrelatedStack` to avoid trying to access alignment matrices and a roundtrip test of a `Kymo`/`Scan` stored as a tiff and reloaded as a `CorrelatedStack`. A future PR might add these to ensure that we can guarantee tiffs exported from `Scan` and `Kymo` to be compatible with `CorrelatedStack`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 
 * Fixed and reintroduced lazy loading for `TimeSeries` data.
 * You can now add two `KymoTrackGroups` tracked on the same kymo together with the `+` operator.
+* TIFFs exported from `Scan` and `Kymo` now contain metadata. The `DateTime` tag indicated the start/stop timestamp of each frame. The `ImageDescription` tag contains additional information about the confocal acquisition parameters. 
 
 #### Other changes
 
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -1,14 +1,17 @@
-import numpy as np
 import os
-import tifffile
+from typing import Iterator, Union
+
+import numpy as np
+import numpy.typing as npt
 from deprecated.sphinx import deprecated
-from .detail.imaging_mixins import VideoExport, FrameIndex
+
 from .adjustments import ColorAdjustment
-from .detail.widefield import TiffStack
 from .detail.image import make_image_title
+from .detail.imaging_mixins import FrameIndex, TiffExport, VideoExport
+from .detail.widefield import TiffStack
 
 
-class CorrelatedStack(VideoExport, FrameIndex):
+class CorrelatedStack(FrameIndex, TiffExport, VideoExport):
     """CorrelatedStack acquired with Bluelake. Bluelake can export stacks of images to various
     formats. These can be opened and correlated to timeline data using CorrelatedStack.
 
@@ -230,7 +233,7 @@ class CorrelatedStack(VideoExport, FrameIndex):
             The color channel of the requested data.
             For single-color data, this argument is ignored.
         """
-        if self._src._description.is_rgb:
+        if self._src.is_rgb:
             channel_indices = {"red": 0, "green": 1, "blue": 2, "rgb": slice(None)}
             slc = (slice(None), slice(None), channel_indices[channel])
         else:
@@ -439,51 +442,39 @@ class CorrelatedStack(VideoExport, FrameIndex):
         Parameters
         ----------
         file_name : str
-            File name to export to.
+            The name of the TIFF file where the image will be saved.
         """
-        from . import __version__ as version
+        super().export_tiff(file_name, dtype=None, clip=False)
 
-        def parse_tags(frame):
-            # Parse original file tags into list of tuples
-            #   [(code, dtype, count, value, writeonce)]
-            #   code is defined by the TIFF specification
-            #   dtype is defined in tifffile
-            # Only tags that are not resolved automatically by `TiffWriter.save()` are needed
+    def _tiff_frames(self, iterator=False) -> Union[npt.ArrayLike, Iterator]:
+        """Create images for frames of TIFFs used by `export_tiff().`"""
+        return (frame.data for frame in self) if iterator else self.get_image()
 
-            # Orientation, uint16, len, ORIENTATION.TOPLEFT
-            orientation = (274, "H", 1, 1)
-
-            # SampleFormat, uint16, len, number of channels
-            n_channels = 3 if frame.is_rgb else 1
-            sample_format = (339, "H", n_channels, (1,) * n_channels)
-
-            # DateTime, str, len, start:stop
-            datetime = frame._page.tags["DateTime"].value
-            datetime = (306, "s", len(datetime), datetime)
-
-            return (orientation, sample_format, datetime)
-
+    def _tiff_image_metadata(self) -> dict:
+        """Create metadata stored in the ImageDescription field of TIFFs used by `export_tiff()`"""
         # re-name alignment matrices fields in image description
         # to reflect the fact that the image has already been processed
-        description = self._src._description.for_export
+        return self._src._description.for_export
+
+    def _tiff_timestamp_ranges(self) -> list:
+        """Create Timestamp ranges for DateTime field of TIFFs used by `export_tiff().`"""
+        return self.frame_timestamp_ranges()
+
+    def _tiff_writer_kwargs(self) -> dict:
+        """Create keyword arguments used for `TiffWriter.write()` in `self.export_tiff()`."""
+        from . import __version__ as version
 
         # add pylake to Software tag
         software = self._src._description.software
-        if "pylake" not in software:
-            software += f", pylake v{version}"
+        if "pylake" not in software.lower():
+            software += (", " if len(software) else "") + f"Pylake v{version}"
 
-        # write frames sequentially
-        with tifffile.TiffWriter(file_name) as tif:
-            for frame in self:
-                tif.write(
-                    frame.data,
-                    description=description,
-                    software=software,
-                    metadata=None,  # suppress tifffile default ImageDescription tag
-                    contiguous=False,  # needed to write tags on each page
-                    extratags=parse_tags(frame),
-                    photometric="rgb" if frame.is_rgb else "minisblack",
-                )
+        write_kwargs = {
+            "software": software,
+            "photometric": "rgb" if self._src.is_rgb else "minisblack",
+        }
+
+        return write_kwargs
 
     @property
     def num_frames(self):
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -1,14 +1,15 @@
 import json
 import numpy as np
 import cachetools
+from numpy import typing as npt
 from deprecated.sphinx import deprecated
 from dataclasses import dataclass
 from typing import List
 import warnings
 
-from .mixin import PhotonCounts
-from .mixin import ExcitationLaserPower
-from .image import reconstruct_image_sum, reconstruct_image, save_tiff
+from .mixin import ExcitationLaserPower, PhotonCounts
+from .image import reconstruct_image_sum, reconstruct_image
+from .imaging_mixins import TiffExport
 from .utilities import could_sum_overflow
 from ..adjustments import ColorAdjustment
 from matplotlib.colors import LinearSegmentedColormap
@@ -344,7 +345,7 @@ class BaseScan(PhotonCounts, ExcitationLaserPower):
         return self._metadata.center_point_um
 
 
-class ConfocalImage(BaseScan):
+class ConfocalImage(BaseScan, TiffExport):
     def _to_spatial(self, data):
         """Implements any necessary post-processing actions after image reconstruction from infowave"""
         raise NotImplementedError
@@ -390,6 +391,16 @@ class ConfocalImage(BaseScan):
             If enabled, the photon count data will be clipped to fit into the desired `dtype`.
             This option is disabled by default: an error will be raise if the data does not fit.
         """
+        super().export_tiff(filename, dtype=dtype, clip=clip)
+
+    def _tiff_frames(self, iterator=False) -> npt.ArrayLike:
+        """Create frames of TIFFs used by `export_tiff().`"""
+        image = self.get_image()
+        return image if image.ndim >= 4 else np.expand_dims(image, axis=0)
+
+    def _tiff_image_metadata(self) -> dict:
+        """Create metadata stored in the ImageDescription field of TIFFs used by `export_tiff()`"""
+        # Try to get the pixel time
         try:
             pixel_time_seconds = self.pixel_time_seconds
         except NotImplementedError:
@@ -398,17 +409,45 @@ class ConfocalImage(BaseScan):
                 "The corresponding metadata in the output file is set to `None`."
             )
             pixel_time_seconds = None
-        if self.get_image("rgb").size > 0:
-            save_tiff(
-                self.get_image("rgb"),
-                filename,
-                dtype,
-                clip,
-                pixel_sizes_um=self.pixelsize_um,
-                pixel_time_seconds=pixel_time_seconds,
-            )
-        else:
-            raise RuntimeError("Can't export TIFF if there are no pixels")
+
+        # Build metadata dict
+        metadata = {
+            "Camera": f"Confocal{self.__class__.__name__}",
+            "Scan axes": [
+                {
+                    "Axis": sa.axis,
+                    "Label": sa.axis_label.lower(),
+                    "Number of pixels": sa.num_pixels,
+                    "Pixel size (um)": sa.pixel_size_um,
+                }
+                for sa in self._metadata.scan_axes
+            ],
+            "Fast axis": self.fast_axis.lower(),
+            "Center point (um)": self.center_point_um,
+            "Pixel time (s)": pixel_time_seconds,
+        }
+
+        return metadata
+
+    def _tiff_writer_kwargs(self) -> dict:
+        """Create keyword arguments used for `TiffWriter.write()` in `self.export_tiff()`."""
+        from .. import __version__ as version
+
+        write_kwargs = {
+            "software": f"Pylake v{version}",
+            "photometric": "rgb",
+        }
+
+        # Add resolution if pixelsize is available
+        pixel_sizes_um = self.pixelsize_um
+        pixel_size_x, pixel_size_y = (
+            pixel_sizes_um[0],
+            pixel_sizes_um[1] if len(pixel_sizes_um) == 2 else pixel_sizes_um[0],
+        )
+        if pixel_size_x:
+            write_kwargs["resolution"] = (1e4 / pixel_size_x, 1e4 / pixel_size_y, "CENTIMETER")
+
+        return write_kwargs
 
     @deprecated(
         reason=(
--- a/lumicks/pylake/detail/imaging_mixins.py
+++ b/lumicks/pylake/detail/imaging_mixins.py
@@ -1,9 +1,101 @@
+import json
+import numpy as np
+import numpy.typing as npt
+import tifffile
+from typing import Iterator, Union
+
 from ..adjustments import ColorAdjustment
 from .timeindex import to_timestamp
-import numpy as np
-from typing import Union
 
 _FIRST_TIMESTAMP = 1388534400
+
+
+class TiffExport:
+    def export_tiff(self, filename, *, dtype=None, clip=False):
+        """Export frames to a TIFF image
+
+        Parameters
+        ----------
+        filename : str
+            The name of the TIFF file where the image will be saved.
+        dtype : np.dtype
+            The data type of a single color channel in the resulting image.
+        clip : bool
+            If enabled, the image data will be clipped to fit into the desired `dtype`. This option
+            is disabled by default: an error will be raise if the data does not fit.
+        """
+        # If the exported tiff should be cast to `dtype`, get the full image stack to later safely
+        # cast it. Otherwise, try to get an iterator, to save memory.
+        frames = self._tiff_frames(iterator=dtype is None)
+        timestamp_ranges = self._tiff_timestamp_ranges()
+
+        # Check length of timestamp ranges, as this is easier than checking the number of frames,
+        # which could be an iterator or an np.ndarray.
+        if len(timestamp_ranges) == 0:
+            raise RuntimeError("Can't export TIFF if there are no images.")
+
+        # Cast frames if a specific dtype is requested.
+        def cast_image(image, dtype=np.float32, clip=False):
+            # Check if requested dtype can fit image values without an overflow
+            info = np.finfo(dtype) if np.dtype(dtype).kind == "f" else np.iinfo(dtype)
+            if not clip and (np.min(image) < info.min or np.max(image) > info.max):
+                raise RuntimeError(
+                    f"Can't safely export image with `dtype={dtype.__name__}` channels."
+                    f" Switch to a larger `dtype` in order to safely store everything"
+                    f" or pass `force=True` to clip the data."
+                )
+
+            return image.astype(dtype)
+
+        frames = cast_image(frames, dtype=dtype, clip=clip) if dtype else frames
+
+        def extratags(timestamp_range):
+            """Create the extratags tuple used for the `TiffWriter.write()` method
+
+            Notes
+            -----
+            See TIFF specification for the definition of tags (or fields)
+            """
+            # DateTime, str, len, start:stop
+            datetime = f"{timestamp_range[0]}:{timestamp_range[1]}"
+            datetime = (306, "s", len(datetime), datetime)
+            return (datetime,)
+
+        # Save the tiff file page by page
+        with tifffile.TiffWriter(filename) as tif:
+            for frame, timestamp_range in zip(frames, timestamp_ranges):
+                tif.write(
+                    frame,
+                    contiguous=False,  # write tags on each page
+                    extratags=extratags(timestamp_range),
+                    metadata=None,  # suppress tifffile default ImageDescription tag
+                    description=json.dumps(self._tiff_image_metadata(), indent=4),
+                    **self._tiff_writer_kwargs(),
+                )
+
+    def _tiff_frames(self, iterator=False) -> Union[npt.ArrayLike, Iterator]:
+        """Create frames of TIFFs used by `export_tiff()`."""
+        raise NotImplementedError(
+            f"`{self.__module__}.{self.__class__.__name__}` does not implement `_tiff_frames()`."
+        )
+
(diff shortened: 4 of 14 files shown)


JoepVanlier (commented):
This is already looking pretty good!

I left a few comments already, but a few additional things to consider:

1. Not a must, but did you have a look at `CorrelatedStack`'s `export_tiff`? Do you think we could deduplicate between `CorrelatedStack` and `Confocal`. We do already have some imaging mixins going. This may be a good contender? You can still keep the actual tag generation in the specialized class.

2. I really like that you include the `DateTime`. Would be nice to include a test for a round trip (loading a `Scan` with `Stack`). One thing that irks me a little is that you ge a warning that the file doesn't contain alignment matrices when opening one of these as a `CorrelatedStack`. Could probably detect from the `Camera` type that alignment is not desired.

lumicks/pylake/detail/confocal.py
```diff
@@ -386,17 +459,32 @@ def export_tiff(self, filename, *, dtype=np.float32, clip=False):
...
-                pixel_sizes_um=self.pixelsize_um,
-                pixel_time_seconds=pixel_time_seconds,
-            )
-        else:
-            raise RuntimeError("Can't export TIFF if there are no pixels")
+
+        # Build metadata dict
+        metadata = {
+            "Camera": f"{self.__module__}.{self.__class__.__name__} exported as a tiff",
+            "ScanAxes": [
```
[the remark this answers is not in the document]
tobiasjj: I also combined the name and the unit in one key-value pair, as BL does it.

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,131 @@
...
+            else:
+                # `frames` is an np.ndarray. Check the size.
+                frames = frames if frames.size > 0 else None
+
+            if frames is None:
+                raise RuntimeError("Can't export TIFF if there are no pixels")
+
+            return frames
+
+        frames = _check_empty(frames)
```
JoepVanlier, line 53: I wish there was an easier way to do this.

I was hoping we could leave this statement out entirely and just iterate; and while this works (as long as you make sure that there is a check before `_cast_image` if there are no images. I was hoping that `TiffWriter` would not write a file until a frame comes in, but unfortunately it does.

However, since we are using the `_tiff_timestamp_ranges` anyhow, it might be worth considering just checking whether that is empty. Note though that if you go this route, you should store it, because they are somewhat expensive to grab.

```python
        ts_ranges = self._tiff_timestamp_ranges()
        if not ts_ranges:
            raise RuntimeError("Can't export TIFF if there are no pixels")
```

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,131 @@
...
+                if not clip and (np.min(image) < info.min or np.max(image) > info.max):
+                    raise RuntimeError(
+                        f"Can't safely export image with `dtype={dtype.__name__}` channels."
+                        f" Switch to a larger `dtype` in order to safely store everything"
+                        f" or pass `force=True` to clip the data."
+                    )
+
+                return image.astype(dtype)
+
+            frames = _cast_image(frames, dtype=dtype, clip=clip)
```
JoepVanlier, line 70: Personally, I don't like putting the entire function indented here and doing something like this (but that's more a personal preference):

```python
        # Cast frames if a specific dtype is requested.
        def _cast_image(image, dtype=np.float32, clip=False):
            # Check if requested dtype can fit image values without an overflow
            info = np.finfo(dtype) if np.dtype(dtype).kind == "f" else np.iinfo(dtype)
            if not clip and (np.min(image) < info.min or np.max(image) > info.max):
                raise RuntimeError(
                    f"Can't safely export image with `dtype={dtype.__name__}` channels."
                    f" Switch to a larger `dtype` in order to safely store everything"
                    f" or pass `force=True` to clip the data."
                )

            return image.astype(dtype)

        if dtype:
            frames = _cast_image(frames, dtype=dtype, clip=clip)
```

Or even:
```python
   frames = _cast_image(frames, dtype=dtype, clip=clip) if dtype else frames
```

lumicks/pylake/tests/test_imaging_mixins.py
```diff
@@ -0,0 +1,209 @@
...
+            def frames_iterator():
+                for frame in frames:
+                    yield frame
+
+            return frames_iterator()
+
+        return tiff_frames
+
+    def tiff_image_metadata():
+        return {"Writing": "tests is awesome!"}
```
JoepVanlier, line 31: are :D

lumicks/pylake/tests/test_imaging_mixins.py
```diff
@@ -0,0 +1,209 @@
...
+    export_empty = tiffexport_factory(np.array([]))
+    filename = tmp_path / "empty.tiff"
+
+    # Test empty `export_tiff` array
+    with pytest.raises(RuntimeError, match="Can't export TIFF if there are no pixels"):
+        export_empty.export_tiff(filename)
+
+    # Test empty `export_tiff` iterator
+    with pytest.raises(RuntimeError, match="Can't export TIFF if there are no pixels"):
+        export_empty.export_tiff(filename, dtype=np.float32)
```
JoepVanlier, line 106: 👍

JoepVanlier (changes requested):
Looking good! Here's my first 2c. I'm not quite done yet with the review, but I'll pick up where I left off tomorrow.

lumicks/pylake/detail/confocal.py
```diff
@@ -1,14 +1,16 @@
...
+import tifffile
 from deprecated.sphinx import deprecated
 from dataclasses import dataclass
+from itertools import zip_longest
 from typing import List
 import warnings
 
 from .mixin import PhotonCounts
 from .mixin import ExcitationLaserPower
-from .image import reconstruct_image_sum, reconstruct_image, save_tiff
```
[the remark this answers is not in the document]
tobiasjj: Done!

lumicks/pylake/detail/imaging_mixins.py
tobiasjj: I originally intended to also be able to export tiffs without timestamp ranges, which is why I did not think of checking the length of `_tiff_timestamp_ranges()`. As we ensure to provide timestamp ranges this is indeed a much cleaner solution!

Implemented!

lumicks/pylake/detail/imaging_mixins.py
tobiasjj: OK, I agree it is nicer to reduce indentations.

Changed.

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -372,18 +372,58 @@ def get_rectangle_data():
...
+        np.testing.assert_allclose(
+            tags["XResolution"][0] / tags["XResolution"][1],
+            kymo._tiff_writer_kwargs()["resolution"][0],
+            rtol=1e-1,
+        )
+        np.testing.assert_allclose(
+            tags["YResolution"][0] / tags["YResolution"][1],
+            kymo._tiff_writer_kwargs()["resolution"][1],
+            rtol=1e-1
+        )
```
rpauszek, line 416: why does the acceptance tolerance have to be so high here? that's a bit concerning

lumicks/pylake/tests/test_imaging_confocal/test_scan.py
```diff
@@ -247,23 +247,65 @@ def test_deprecated_plotting(test_scans):
...
+            for page in tif.pages:
+                page_tags = {}
+                for tag in page.tags.values():
+                    name, value = tag.name, tag.value
+                    try:
+                        page_tags[name] = literal_eval(value)
+                    except (ValueError, SyntaxError):
+                        page_tags[name] = value
+                tiff_tags.append(page_tags)
+        return tiff_tags
```
rpauszek, line 275: considering this is exactly duplicated from the kymo tests, is there a way to move this into `conftest.py`?

I feel like we've run across this kind of thing before but can't remember what we did. I think there's a problem where it's weird syntax to return functions from a fixture, and you want to avoid `from .conftest import ...` @JoepVanlier does this seem familiar to you?

lumicks/pylake/tests/test_imaging_mixins.py
```diff
@@ -0,0 +1,209 @@
...
+            if not iterator:
+                return frames
+
+            def frames_iterator():
+                for frame in frames:
+                    yield frame
+
+            return frames_iterator()
+
+        return tiff_frames
```
rpauszek, line 28: whoa, holy indentation batman! maybe:
```suggestion
    def tiff_frames_factory(frames):
        return lambda iterator: (frame for frame in frames) if iterator else frames
```

I gave it a quick try and all tests still pass

lumicks/pylake/tests/test_imaging_mixins.py
```diff
@@ -0,0 +1,209 @@
...
+    def tiff_timestamp_ranges_factory(number_of_frames):
+        def tiff_timestamp_ranges():
+            return [
+                np.array([i, i + exposure_time])
+                for i in range(
+                    start_time, start_time + frame_time * number_of_frames, start_time + frame_time
+                )
+            ]
+
+        return tiff_timestamp_ranges
```
rpauszek, line 42: normally I might argue against a multiline `lambda` but to save on the nesting might be nice here
```suggestion
    def tiff_timestamp_ranges_factory(number_of_frames):
        return lambda: [
            np.array([i, i + exposure_time])
            for i in range(
                start_time, start_time + frame_time * number_of_frames, start_time + frame_time
            )
        ]
```

rpauszek (changes requested):
wow, this thing is impressive, lots of good stuff. I've left some nitpicky comments, but there are one or two that I think should be addressed (breaking change to positional argument name, and some places where the code could be made more concise).

There's also a couple things that I've pointed out that we probably shouldn't address here (this PR is a beast as it is) but we should think about and make a decision on in the very near future. Great work overall though!

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,131 @@
...
+        filename : str
+            The name of the TIFF file where the image will be saved.
+        dtype : np.dtype
+            The data type of a single color channel in the resulting image.
+        clip : bool
+            If enabled, the image data will be clipped to fit into the desired `dtype`. This option
+            is disabled by default: an error will be raise if the data does not fit.
+        """
+        # If the the exported tiff should be casted to `dtype`, get the full image stack to later
+        # safely cast it. Otherwise, try to get an iterator, to save memory.
```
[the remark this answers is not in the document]
tobiasjj: 👍

lumicks/pylake/correlated_stack.py
```diff
@@ -420,57 +421,55 @@ def post_update(image_handle, image):
...
+        from . import __version__ as version
 
         # add pylake to Software tag
         software = self.src._description.software
-        if "pylake" not in software:
-            software += f", pylake v{version}"
+        if "pylake" not in software.lower():
+            if len(software) > 0:
+                software += ", "
+            software += f"Pylake v{version}"
```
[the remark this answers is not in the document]
tobiasjj: Yes, it is, and I like it :-)

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,112 @@
+import itertools
```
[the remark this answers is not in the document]
tobiasjj: removed ...

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,112 @@
...
+        frames = self._tiff_frames(iterator=dtype is None)
+        timestamp_ranges = self._tiff_timestamp_ranges()
+
+        # Check length of timestamp ranges, as this is easier than checking the number of frames,
+        # which could be an iterator or an np.ndarray.
+        if len(timestamp_ranges) == 0:
+            raise RuntimeError("Can't export TIFF if there are no images.")
+
+        # Cast frames if a specific dtype is requested.
+        def _cast_image(image, dtype=np.float32, clip=False):
```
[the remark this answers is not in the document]
tobiasjj: Yeah, removed the underscore...

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,112 @@
...
+
+        def extratags(timestamp_range):
+            """Create the extratags tuple used for the `TiffWriter.write()` method
+
+            Notes
+            -----
+            See TIFF specification for the definition of tags (or fields)
+            """
+            if timestamp_range is None:
+                return None
```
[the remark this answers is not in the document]
tobiasjj: Whoops, that is a left overt from when I initially wanted to support exporting tiffs without timestamp ranges. Removed it ...

lumicks/pylake/kymo.py
```diff
@@ -184,6 +185,23 @@ def line_timestamp_ranges(self, exclude=None, *, include_dead_time=None):
...
         return self._line_timestamp_ranges_factory(self, not include)
 
+    def _tiff_frames(self, iterator=False) -> npt.ArrayLike:
+        """Create frames of TIFFs used by `export_tiff().`"""
+        return np.expand_dims(self.get_image(), axis=0)
+
+    def _tiff_image_metadata(self) -> dict:
+        """Create metadata for the ImageDescription field of TIFFs used by `export_tiff()`."""
+        metadata = super()._tiff_image_metadata()
+        # Add `Kymo` specific metadata
```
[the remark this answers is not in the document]
tobiasjj: removed

lumicks/pylake/kymo.py
```diff
@@ -184,6 +185,23 @@ def line_timestamp_ranges(self, exclude=None, *, include_dead_time=None):
...
+        metadata["Line time (s)"] = self.line_time_seconds
+        return metadata
+
+    def _tiff_timestamp_ranges(self) -> list:
+        """Create Timestamp ranges for the DateTime field of TIFFs used by `export_tiff`."""
+        # As `Kymo` has only one frame, return a list with one timestamp range
```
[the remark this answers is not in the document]
tobiasjj: Actually, including extra metadata is fairly easy. I included 'Line timestamp ranges' in the method `_tiff_image_metadata()` for the kymo.

lumicks/pylake/kymo.py
```diff
@@ -184,6 +185,23 @@ def line_timestamp_ranges(self, exclude=None, *, include_dead_time=None):
 
         return self._line_timestamp_ranges_factory(self, not include)
 
+    def _tiff_frames(self, iterator=False) -> npt.ArrayLike:
+        """Create frames of TIFFs used by `export_tiff().`"""
+        return np.expand_dims(self.get_image(), axis=0)
```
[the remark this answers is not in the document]
tobiasjj: You are right, I forgot that we squeeze the output arrays and indeed a confocal `Scan` with just one frame will only have 3 dimensions.
Good idea! I generalized and moved this function into the parent class.

lumicks/pylake/correlated_stack.py
```diff
@@ -420,57 +421,55 @@ def post_update(image_handle, image):
...
-            datetime = (306, "s", len(datetime), datetime)
+        def image_iterator():
+            for frame in self:
+                yield frame.data
 
-            return (orientation, sample_format, datetime)
+        if iterator:
+            return image_iterator()
+        else:
+            return self.get_image()
```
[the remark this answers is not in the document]
tobiasjj: Yep, that is much cleaner!

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -372,18 +372,58 @@ def get_rectangle_data():
         kymo.plot_with_time_histogram(color_channel="red", pixels_per_bin=6)
 
 
-def test_save_tiff(tmpdir_factory, test_kymos):
+def test_export_tiff(tmp_path, test_kymos):
     from os import stat
 
-    tmpdir = tmpdir_factory.mktemp("pylake")
```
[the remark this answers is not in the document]
tobiasjj: I like to use the `/` operator to join paths, which is why I chose to use `tmp_path` over `tmp_dir`. `tmp_path*` fixtures are a ''newer'' version of the `tmp_dir*` fixtures: [pytest documentation](https://docs.pytest.org/en/stable/how-to/tmp_path.html).

I thought that temporarily storing a "one time use" tiff file in a directory does not require the session level fixture `tmp_path_factory`, but the function level fixture `tmp_path` is sufficient. Am I missing something here?

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
tobiasjj: The "XResolution" and "YResolution" are stored as a rational datatype with two 32 bit integers. This means these tags only give approximate resolution in many cases and the sole purpose (I can see) of these is solely for qualitative scaling and displaying purposes. Therefore, I also included the `pixel_size_um` in the ImageDescription tag (i.e. the `_tiff_image_metadata()` method). Still, I also included the resolution tags, as these were also set by the old `save_tiff()` function.
Note: The old tests did not take into account that the resolution tags were stored as rational, and only tested for the numerator.

lumicks/pylake/tests/test_imaging_mixins.py
tobiasjj: Thank you :-) Indeed, too many indentations!
As I really do not like lambdas, would you be satisified with:
```
def tiff_frames_factory(frames):
    def tiff_frames(iterator=False):
        return (frame for frame in frames) if iterator else frames
    return tiff_frames
```

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,112 @@
...
+        with tifffile.TiffWriter(filename) as tif:
+            for frame, timestamp_range in zip(frames, timestamp_ranges):
+                tif.write(
+                    frame,
+                    contiguous=False,  # write tags on each page
+                    extratags=extratags(timestamp_range),
+                    metadata=None,  # suppress tifffile default ImageDescription tag
+                    description=description,
+                    **tiff_write_kwargs,
+                )
```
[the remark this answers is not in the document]
tobiasjj: I wanted to avoid recalling the methods `_tiff_image_metadata()` and `_tiff_writer_kwargs()` for every single frame. But the performance impact is small, so OK -> changed.

lumicks/pylake/correlated_stack.py
```diff
@@ -1,14 +1,15 @@
 import numpy as np
+import numpy.typing as npt
 import os
-import tifffile
 from deprecated.sphinx import deprecated
-from .detail.imaging_mixins import VideoExport, FrameIndex
+from typing import Iterator, Union
```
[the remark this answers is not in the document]
tobiasjj: I reordered the imports with the pip package isort now (sorts imports according to PEP8). In general, we probably could rely on what isort is doing? What do you think?

lumicks/pylake/detail/confocal.py
```diff
@@ -386,17 +391,45 @@ def export_tiff(self, filename, *, dtype=np.float32, clip=False):
...
-                clip,
-                pixel_sizes_um=self.pixelsize_um,
-                pixel_time_seconds=pixel_time_seconds,
-            )
-        else:
-            raise RuntimeError("Can't export TIFF if there are no pixels")
+
+        # Build metadata dict
+        metadata = {
+            "Camera": f"{self.__module__}.{self.__class__.__name__} exported as a tiff",
```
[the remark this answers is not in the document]
tobiasjj: I renamed the cameras to `ConfocalKymo` and ``ConfocalScan`.

lumicks/pylake/tests/test_imaging_confocal/test_scan.py
tobiasjj: I created a fixture in `conftest.py`, and used it in the tests for `kymo` and `scan`. It works without the import.

lumicks/pylake/tests/test_imaging_mixins.py
tobiasjj: I do not like lambdas and am not so sure if the lambda improves the nesting situation. I like the verbose definition of the function `tiff_timestamp_ranges` that is finally returned. Should we vote? @JoepVanlier

JoepVanlier: My vote for this one would go to no lambda.

lumicks/pylake/tests/test_imaging_mixins.py
JoepVanlier: My vote for this one would go to lambda.

tobiasjj: OK, overruled. Will change it.

lumicks/pylake/detail/imaging_mixins.py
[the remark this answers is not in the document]
rpauszek: ah good point, didn't think of that. but if the performance impact is negligible then I guess it's ok

lumicks/pylake/kymo.py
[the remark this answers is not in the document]
rpauszek: yeah the problem isn't including extra information, I think we should spend some time planning it out though to avoid having a million different versions when we forget something etc. maybe let's set up a time for the 3 of us to look over everything and discuss?

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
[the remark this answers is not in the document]
rpauszek: ah I didn't realize it was a built-in `pytest` fixture. good call

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
rpauszek: yeah, I kinda suspected it had something to do with the stored datatype. that's fine then, we don't really rely on this tag anyway and with the `pixel_size_um` there that's good

lumicks/pylake/tests/test_imaging_mixins.py
rpauszek: yeah, in general I get your point about sometimes it's nice to have a named function to indicate what it's actually doing. In this particular case though I think having an extra nested named function is actually more confusing. Maybe using the lambda and just adding a comment line explaining why we need either a generator or just the list/array as a compromise?

lumicks/pylake/tests/test_imaging_mixins.py
rpauszek: ok yeah here I'm fine keeping the nested function

lumicks/pylake/detail/confocal.py
```diff
@@ -390,6 +391,16 @@ def export_tiff(self, filename, *, dtype=np.float32, clip=False):
             If enabled, the photon count data will be clipped to fit into the desired `dtype`.
             This option is disabled by default: an error will be raise if the data does not fit.
         """
+        super().export_tiff(filename, dtype=dtype, clip=clip)
+
+    def _tiff_frames(self, iterator=False) -> npt.ArrayLike:
+        """Create frames of TIFFs used by `export_tiff().`"""
+        image = self.get_image()
+        return image if image.ndim >= 4 else np.expand_dims(image, axis=0)
```
rpauszek, line 399: do we actually need the `expand_dims` still? did you test without?

rpauszek (approved):
LGTM!

JoepVanlier (approved):
Excellent work! :)

Really like how you set up the tests.

lumicks/pylake/detail/confocal.py
tobiasjj: Yes, we need `expand_dims`, as `export_tiff` iterates over all frames of the imagestack. Without `expand_dims`, `export_tiff` of one frame scans (and kymos) would try to iterate over the rows of the image zipped together with the timestamp range(s) and consequently would story only one image row.

changelog.md
```diff
@@ -5,6 +5,7 @@
 ### New features
 
 * Fixed and reintroduced lazy loading for `TimeSeries` data. 
+* Tiffs exported from `Scan` and `Kymo` now contain metadata. The `DateTime` tag indicated the start/stop timestamp of each frame. The `ImageDescription` tag contains additional information about the confocal acquisition parameters.
```
mikhas, line 8: Nit: Correct spelling would be TIFFs

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -1,11 +1,103 @@
...
 from ..adjustments import ColorAdjustment
 from .timeindex import to_timestamp
-import numpy as np
-from typing import Union
 
 _FIRST_TIMESTAMP = 1388534400
 
 
+class TiffExport:
+    def export_tiff(self, filename, *, dtype=None, clip=False):
```
mikhas, line 14: I would expect this function to be called `export`, or `export_tiff_stack` since you are not just exporting a single frame.

mikhas (commented):
I enjoyed reading the code.

mikhas:
Why is this PR not merged yet?

tobiasjj:
> Why is this PR not merged yet?

We had to postpone a final decision about the metadata. Will be merged today.

lumicks/pylake/kymo.py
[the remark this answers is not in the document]
tobiasjj: Change of plans: Instead of all `line timestamp ranges` there are now only two entries `Start pixel timestamp (ns)` and `Stop pixel timestamp (ns)`. With these and the number of lines the `Line timestamp ranges` can be reconstructed.

PR #376, opened 2022-09-08
channel: revert TimeSeries.from_dataset

rpauszek:
Using `.fields()` to try to get lazy loading leads to unreasonably long time to access data (on the order of a minute). Seems to have to do with the way this data is structured as `[(value, value) ...]` arrays. Dropping lazy loading for now; can revisit in the future if needed.


mikhas:
No regression test wrt. performance degradation?

JoepVanlier (changes requested):
I would also revert https://github.com/lumicks/pylake/commit/726cbc0bd83b4f53bc8029ffe18ec51ccd0a38aa since this lazy loading essentially does nothing since that indexing already returns numpy arrays.

JoepVanlier:
> No regression test wrt. performance degradation?
I pushed a commit which tests the performance of `TimeSeries` so that it won't happen again.

JoepVanlier (approved):
Looks good to me now, but considering I pushed one commit to this, it should probably not be reviewed only by me :D

PR #378, opened 2022-09-09
Lazy fix hotfix

tobiasjj:
**Why this PR?**
This PR reintroduces the lazy loading of `TimeSeries` and fixes the performance issue when loading data.

Long loading times were caused by calling `np.asarray(self._src_data)`, when accessing big datasets via `TimeSeries.data`. To support lazy loading of compound datasets, the method `fields()` was used during the initialization of `TimeSeries` objects. It returns a wrapper object around the original dataset. However, it does not implement the method `__array__()`. This caused `np.asarray(self._src_data)` to iterate over all elements via the `__getitem__()` method and thus causing long loading times. Also see https://github.com/h5py/h5py/issues/2144.

The quick fix  —until https://github.com/h5py/h5py/issues/2144 will be resolved— was to wrap the dataset in a small class that implements the `__array__()` method, instead of using the object returned by the `fields()` method.

I also added additional tests to check the functionality of `TimeSeries` to test the properties and functions that use the wrapped dataset. For completeness I added similar tests to `Continuous` and `TimeTags`.


mikhas:
I wouldn't commit to a hotfix unless it becomes clear we cannot fix it upstream in a reasonable time frame.

lumicks/pylake/channel.py
```diff
@@ -583,13 +584,51 @@ def _apply_mask(self, mask):
...
 
     @staticmethod
-    def from_dataset(dset, y_label="y", calibration=None):
+    def from_dataset(dset, y_label="y", calibration=None) -> Slice:
+        class LazyLoadedCompoundField:
+            """Wrapper to enable lazy loading of HDF5 compound datasets
+
+            Notes
+            -----
+            We only need to suppot the methods `__array__()` and `__len__()`, as we only access
```
JoepVanlier, line 593: I appreciate the little note for our future selves.

Minor typo though: `suppot` -> `support`

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.13.2 | tbd
+
+### New features
+* Fixed and reintroduced lazy loading for `TimeSeries` data.
```
JoepVanlier, line 6: There needs to be a linebreak between `New features` and the first item here.

JoepVanlier (approved):
Looks good to me. Two tiny requests, but otherwise :shipit:

lumicks/pylake/channel.py
tobiasjj: fixed

changelog.md
tobiasjj: fixed

PR #379, opened 2022-09-16
ci: don't use `cleanup` from the mpl testing decorators

JoepVanlier:
**Why this PR?**
We've been using the `cleanup` decorator from `matplotlib`, but it's [deprecated](https://github.com/matplotlib/matplotlib/pull/22697) now, so here is our own.

I also took the liberty of getting rid of a few deprecated API calls in the benchmark and adding a few `plt.close('all')` calls in a few of the tests, since auto-removal of overlapping axes is deprecated.

Longer term it might be nice to look into something such as [pytest-mpl](https://github.com/matplotlib/pytest-mpl). Maybe a fun friday afternoon project, but for now this at least gets us back up and running without using deprecated functionality.

### DIFF
--- a/lumicks/pylake/benchmark.py
+++ b/lumicks/pylake/benchmark.py
@@ -103,7 +103,7 @@ class _Tracking(_Benchmark):
 
     def context(self):
         kymo = _generate_kymo_for_tracking(12, 11)
-        yield lambda: lk.track_greedy(kymo, "red", line_width=1, pixel_threshold=20)
+        yield lambda: lk.track_greedy(kymo, "red", track_width=1, pixel_threshold=20)
 
 
 class _Refinement(_Benchmark):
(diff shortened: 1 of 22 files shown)


tobiasjj (changes requested):
Looks good to me beside one minor question regarding the context managers.

lumicks/pylake/benchmark.py
```diff
@@ -112,8 +112,8 @@ class _Refinement(_Benchmark):
 
     def context(self):
         kymo_tracking = _generate_kymo_for_tracking(1, 4)
-        lines = lk.track_greedy(kymo_tracking, "red", line_width=1, pixel_threshold=20)
-        yield lambda: lk.refine_lines_gaussian(
+        lines = lk.track_greedy(kymo_tracking, "red", track_width=1, pixel_threshold=20)
+        yield lambda: lk.refine_tracks_gaussian(
```
dean0x7d, line 116: I'd suggest mentioning this in `changelog.md` since it's pretty noticeable on the user side when running `lk.benchmark()`.

dean0x7d (approved):

lumicks/pylake/tests/test_decorators.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+def mpl_test_cleanup(func):
+    """Runs tests in a context manager and closes figures after each test"""
+
+    @functools.wraps(func)
+    def wrapped_callable(*args, **kwargs):
+        try:
+            orig_backend = plt.get_backend()
+            with plt.rc_context(), mpl.style.context(["classic", "_classic_test_patch"]):
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, I think you're right.. Fixing.

lumicks/pylake/benchmark.py
JoepVanlier: Good point. I'll add a line.

tobiasjj (approved):
LGTM! 🚴‍♂️

PR #380, opened 2022-09-16
diffusion: implement the covariance based estimator for diffusion

JoepVanlier:
**Why this PR?**
This is the first part of some of the new diffusion estimation work. This implements the CVE method.

Benefits:
- Provides a really fast unbiased estimator that provides a better alternative to OLS/GLS for cases where the `SNR > 1`.
- Future extension is possible towards methods where we take into account the bias caused by substrate fluctuations (whether this is relevant for us is still T.B.D).
- Future extension to incorporating the static localization uncertainty obtained from Gaussian refinement is possible, this reduces the problem by one parameter and reduces spread of the diffusion estimates.

Below you see a figure where this method was using on a 50 point trace with 20 random points missing. The solid lines with the symbols represent the mean, while the lines indicate `mean +- standard deviation`. You can see that all three methods are unbiased for `SNR > 1`.

image

Note that if you provide the localization uncertainty, the variance goes down (as expected).

image

One thing to keep in mind when dealing with these estimates is that the distribution of the estimates is not very normal. It is skewed towards positive values. While the `mean` is asymptotically correct and unbiased, the `median` will not be at the true value, and the `mean` can be outlier sensitive.

I have also compared the variance estimators that were provided (and implemented in this PR). In the results below we can see the difference between the theoretical variance estimate (variance estimator with the true parameters filled in) vs the variance actually obtained from running many traces (N=1500). Used here are traces of 50 points with 10 points randomly omitted (meaning 40 points effectively):

Left shows unknown localization uncertainty, right shows with localization uncertainty known (currently not exposed through the API):

image

The relevant papers are: [here](https://backend.orbit.dtu.dk/ws/files/90238574/PhysRevE.89.022726.pdf) and [here](https://arxiv.org/pdf/1504.05462.pdf).

Docs build can be found [here](https://lumicks-pylake.readthedocs.io/en/cve/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrack.html#lumicks.pylake.kymotracker.kymotrack.KymoTrack.estimate_diffusion) and [here](https://lumicks-pylake.readthedocs.io/en/cve/tutorial/kymotracking.html#studying-diffusion-processes). They are usable, but still a bit spartan. Reason for this is that I intend to restructure them when the ensemble MSD PR rolls around. Since that one requires some reworking in how the story is told, I didn't want to add too many docs changes to this PR (to avoid double work).

### DIFF
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -492,12 +492,13 @@ def _var_cve_unknown_var(
            diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
            022726.
     """
-    epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
-    numerator = 6.0 * avg_frame_steps**2 + 4.0 * epsilon * avg_frame_steps + 2.0 * epsilon**2
+    epsilon = variance_loc / dt - 2.0 * blur_constant * diffusion_constant
+    avg_diff = avg_frame_steps * diffusion_constant
+    numerator = 6.0 * avg_diff**2 + 4.0 * epsilon * avg_diff + 2.0 * epsilon**2
     denominator = num_points * avg_frame_steps**2
     term1 = numerator / denominator
-    term2 = (4.0 * (avg_frame_steps + epsilon) ** 2) / (num_points**2 * avg_frame_steps**2)
-    return (term1 + term2) * diffusion_constant**2
+    term2 = 4.0 * (avg_diff + epsilon) ** 2 / (num_points**2 * avg_frame_steps**2)
+    return term1 + term2
 
 
 def _var_cve_known_var(
@@ -542,11 +543,10 @@ def _var_cve_known_var(
     .. [2] Vestergaard, C. L. (2016). Optimizing experimental parameters for tracking of diffusing
            particles. Physical Review E, 94(2), 022401.
     """
-    epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
+    epsilon = variance_loc / dt - 2.0 * blur_constant * diffusion_constant
     blur_term = (avg_frame_steps - 2.0 * blur_constant) ** 2
-    numerator = diffusion_constant**2 * (
-        2.0 * avg_frame_steps**2 + 4.0 * epsilon * avg_frame_steps + 3.0 * epsilon**2
-    )
+    avg_diff = avg_frame_steps * diffusion_constant
+    numerator = 2.0 * avg_diff**2 + 4.0 * epsilon * avg_diff + 3.0 * epsilon**2
     denominator = num_points * blur_term
     term1 = numerator / denominator
     term2 = variance_variance_loc / (blur_term * dt**2)
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -1,4 +1,6 @@
 import numpy as np
+import numpy.typing as npt
+from typing import Optional
 import warnings
 from dataclasses import dataclass, field
 
@@ -12,7 +14,10 @@ class DiffusionEstimate:
     value : float
         Estimate for the diffusion constant.
     std_err : float
-        Standard error,
+        Standard error.
+
+        Note: while this provides some measure of uncertainty, these estimates are expected to be
+        biased and should not be used to weight multiple diffusion estimates!
     num_lags : int
         Number of lags used to compute this estimate.
     num_points : int
@@ -447,3 +452,211 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
         RuntimeWarning("Warning, maximum number of iterations exceeded. Returning best solution.")
     )
     return num_slope, num_intercept
+
+
+def _var_cve_unknown_var(
+    diffusion_constant: float,
+    variance_loc: float,
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+) -> float:
+    """Expected variance of the diffusion estimate obtained with CVE.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This function is based on eqn 17 from Vestergaard et al [1].
+
+    Parameters
+    ----------
+    diffusion_constant : float
+        Estimate of the diffusion constant
+    variance_loc : float
+        Estimate of the localization variance
+    dt : float
+        Time step
+    num_points : int
+        Number of points in the trace
+    blur_constant : float
+        Motion blur constant. See :func:`lumicks.pylake.kymotracker.detail.msd_estimation._cve` for
+        more information.
+
+    References
+    ----------
+    .. [1] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+           diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+           022726.
+    """
+    epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
+    term1 = (6.0 + 4.0 * epsilon + 2.0 * epsilon**2) / num_points
+    term2 = (4.0 * (1.0 + epsilon) ** 2) / (num_points**2)
+    return (term1 + term2) * diffusion_constant**2
+
+
+def _var_cve_known_var(
+    diffusion_constant: float,
+    variance_loc: float,
+    variance_variance_loc: float,
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+) -> float:
+    """Expected variance of the diffusion estimate obtained with CVE.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This function is based on eqn 18 from Vestergaard et al [1].
+
+    Parameters
+    ----------
+    diffusion_constant : float
+        Estimate of the diffusion constant
+    variance_loc : float
+        Estimate of the localization variance
+    variance_variance_loc : float
+        Variance of the localization variance estimate.
+    dt : float
+        Time step
+    num_points : int
+        Number of points in the trace
+    blur_constant : float
+        Motion blur constant. See :func:`lumicks.pylake.kymotracker.detail.msd_estimation._cve` for
+        more information.
+
+    References
+    ----------
+    .. [1] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+           diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+           022726.
+    """
+    epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
+    blur_term = (1.0 - 2.0 * blur_constant) ** 2
+    term1 = (
+        diffusion_constant**2
+        * (2.0 + 4.0 * epsilon + 3.0 * epsilon**2)
+        / (num_points * blur_term)
+    )
+    term2 = variance_variance_loc / (blur_term * dt**2)
+    return term1 + term2
+
+
+def _cve(
+    x: npt.ArrayLike,
+    dt: float,
+    blur_constant: float = 0,
+    variance_loc: Optional[float] = None,
+    variance_variance_loc: Optional[float] = None,
+) -> tuple:
+    """Covariance based estimator.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This estimator was introduced in the work of Vestergaard et al [1].
+
+    Note that this estimator is only valid in the case of pure diffusion without drift.
+
+    References
+    ----------
+    .. [1] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+           diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+           022726.
+
+    Parameters
+    ----------
+    x : array_like
+        Positions over time
+    dt : float
+        Time step
+    blur_constant : float
+        Motion blur coefficient. Should be between 0 and 1/4.
+
+        The normalized shutter function is defined as c(t), where c(t) represents whether the
+        shutter is open or closed. c(t) is normalized w.r.t. area. For no motion blur,
+        c(t) = delta(t_exposure), whereas for a constantly open shutter it is defined as
+        c(t) = 1 / dt.
+
+        With c(t) defined as before, the motion blur constant is defined as:
+
+        .. math::
+            R = \frac{1}{\Delta t} \int_{0}^{\Delta t}S(t) \left(1 - S(t)\right)dt
+
+        with
+
+        .. math::
+            S(t) = \int_{0}^{t} c(t') dt'
+
+        For no motion blur, we obtain: R = 0, while a continuously open shutter over the exposure
+        time results in R = 1/6. Note that when estimating both localization uncertainty and
+        the diffusion constant, the motion blur factor has no effect on the estimate of the
+        diffusion constant itself, but it does affect the calculated uncertainties. In the case of
+        a provided localization uncertainty, it does impact the estimate of the diffusion constant.
+    variance_loc : float
+        Variance of the localization error if it has been determined independently.
+    variance_variance_loc : float
+        The variance of the localization variance estimate (needed if variance_loc is provided).
+
+    Note
+    ----
+    With the SNR defined as:
+
+    .. math::
+        \sqrt{D \Delta t} / \sigma
+
+    With D the diffusion constant and :math:`\sigma`
+    This method requires SNR > 1. Below SNR = 1 it degrades rapidly.
+    """
+    if not 0 <= blur_constant < 0.25:
+        raise ValueError("Motion blur constant should be between 0 and 1/4")
+
+    dx = np.diff(x)
+    mean_dx_squared = np.mean(dx**2)
+
+    if not variance_loc:
+        # Estimate the variance based on this frame
+        mean_dx_consecutive = np.mean(dx[1:] * dx[:-1])
+        diffusion_constant = mean_dx_squared / (2 * dt) + mean_dx_consecutive / dt
+        variance_loc = (
+            blur_constant * mean_dx_squared + (2 * blur_constant - 1) * mean_dx_consecutive
+        )
+        var_diffusion = _var_cve_unknown_var(
+            diffusion_constant, variance_loc, dt, len(x), blur_constant
+        )
+    else:
+        # We know the variance in advance.
+        diffusion_constant = (mean_dx_squared - 2.0 * variance_loc) / (
+            2.0 * (1.0 - 2.0 * blur_constant) * dt
+        )
+        var_diffusion = _var_cve_known_var(
+            diffusion_constant,
+            variance_loc,
+            variance_variance_loc,
+            dt,
+            len(x),
+            blur_constant,
+        )
+
+    return diffusion_constant, var_diffusion, variance_loc
+
+
+def estimate_diffusion_cve(
+    frame_idx: npt.ArrayLike, coordinate: npt.ArrayLike, dt: float, unit: str, unit_label: str
+) -> DiffusionEstimate:
+    """Estimate diffusion constant based on covariance estimator
+
+    This function estimates the diffusion constant based on the covariance estimator and packs
+    the result in a `DiffusionEstimate`. Note that it is only suitable for confocal-based tracking
+    at the moment.
+    """
+    if np.unique(np.diff(frame_idx)).size > 1:
+        raise RuntimeError(
+            "This method can only be used for equidistantly sampled lines at the moment"
+        )
+
+    # We hardcode the blur constant for confocal for now (no motion blur)
+    diffusion, diffusion_variance, _ = _cve(coordinate, dt, blur_constant=0)
+    return DiffusionEstimate(
+        diffusion,
+        np.sqrt(diffusion_variance),
+        None,
+        len(coordinate),
+        "CVE",
+        unit,
+        unit_label,
+    )
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -16,8 +16,11 @@ class DiffusionEstimate:
     std_err : float
         Standard error.
 
-        Note: while this provides some measure of uncertainty, these estimates are expected to be
-        biased and should not be used to weight multiple diffusion estimates!
+        Note: While this provides some measure of uncertainty, the estimate should not be used for
+        calculating the weighted mean of multiple tracks. This is to prevent complications such as
+        bias due to correlations between the estimated parameters and estimated variances. Instead,
+        when calculating the weighted mean of estimates from time series of different lengths, the
+        length :math:`N` of a time series should be used as weight, since it is known exactly.
     num_lags : int
         Number of lags used to compute this estimate.
     num_points : int
@@ -43,14 +46,17 @@ class DiffusionEstimate:
 
 
 def calculate_msd(frame_idx, position, max_lag):
-    """Estimate the Mean Square Displacement (MSD) for various time lags.
+    r"""Estimate the Mean Squared Displacement (MSD) for various time lags.
 
-    The estimator for the MSD (rho) is defined as:
+    The estimator for the MSD (:math:`\rho`) is defined as:
 
-      rho_n = (1 / (N-n)) sum_{i=1}^{N-n}(r_{i+n} - r_{i})^2
+    .. math::
 
-    here N refers to the total frames, n to the lag time and r_i the spatial position at lag i.
-    This function produces a list of lag times and mean squared displacements for those lag times.
+      \rho_n = \frac{1}{N-n} \sum_{i=1}^{N-n}\left(r_{i+n} - r_{i}\right)^2
+
+    here :math:`N` refers to the total frames, :math:`n` to the lag time and :math:`r_i` to the
+    spatial position at lag :math:`i`. This function produces a list of lag times and mean squared
+    displacements for those lag times.
 
     Parameters
     ----------
@@ -80,7 +86,7 @@ def calculate_msd(frame_idx, position, max_lag):
 def _msd_diffusion_covariance(max_lags, n, intercept, slope):
     """Covariance matrix for the mean squared displacements.
 
-    Equation 8a from [2].
+    Equation 8a from Bullerjahn et al [1]_.
 
     Parameters
     ----------
@@ -95,7 +101,7 @@ def _msd_diffusion_covariance(max_lags, n, intercept, slope):
 
     References
     ----------
-    .. [2] Bullerjahn, J. T., von Bülow, S., & Hummer, G. (2020). Optimal estimates of
+    .. [1] Bullerjahn, J. T., von Bülow, S., & Hummer, G. (2020). Optimal estimates of
            self-diffusion coefficients from molecular dynamics simulations. The Journal of Chemical
            Physics, 153(2), 024116.
     """
@@ -126,7 +132,7 @@ def _msd_diffusion_covariance(max_lags, n, intercept, slope):
 
 
 def _diffusion_ols(mean_squared_displacements, num_points):
-    """Estimate the intercept, slope and standard deviation of the slope based on the msd
+    """Estimate the intercept, slope and standard deviation of the slope based on the msd [2]_
 
     Parameters
     ----------
@@ -205,7 +211,7 @@ def _diffusion_gls(mean_squared_displacements, num_points, tolerance=1e-4, max_i
     """Estimate the intercept, slope and standard deviation of the slope based on the msd
 
     This method takes into account the covariance matrix and thereby does not suffer from including
-    more lags than the optimal number of lags.
+    more lags than the optimal number of lags [3]_.
 
     Parameters
     ----------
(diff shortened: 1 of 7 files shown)


lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +452,257 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+    avg_frame_steps: float = 1,
+) -> float:
+    """Expected variance of the diffusion estimate obtained with CVE.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This function is based on eqn 22 from Vestergaard et al [2] adapted for 1D. See also eqn 17 from
+    Vestergaard et al [1].
```
ab412, line 469: Does this still align with the error message in `DiffusionEstimate`?  Do we still expect some bias when using cve as well? I'm referring to this line:
```HTML
        Note: while this provides some measure of uncertainty, these estimates are expected to be
        biased and should not be used to weight multiple diffusion estimates!
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +452,257 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    .. [1] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+           diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+           022726.
+    """
+    epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
+    numerator = 6.0 * avg_frame_steps**2 + 4.0 * epsilon * avg_frame_steps + 2.0 * epsilon**2
+    denominator = num_points * avg_frame_steps**2
+    term1 = numerator / denominator
+    term2 = (4.0 * (avg_frame_steps + epsilon) ** 2) / (num_points**2 * avg_frame_steps**2)
+    return (term1 + term2) * diffusion_constant**2
```
ab412, line 500: I would use built-in functions where applicable (`math.pow()`) - unless there's a reason not to. Perhaps also handle divide by zero scenarios.
```suggestion
    try
        epsilon = variance_loc / (diffusion_constant * dt) - 2.0 * blur_constant
        numerator = 6.0 * pow(avg_frame_steps,2) + 4.0 * epsilon * avg_frame_steps + 2.0 * pow(epsilon, 2)
        denominator = num_points * pow(avg_frame_steps, 2)
        term1 = numerator / denominator
        term2 = (4.0 * pow((avg_frame_steps + epsilon) ,2)) / pow(num_points,2) * pow(avg_frame_steps,2))
        return (term1 + term2) * pow(diffusion_constant,2)
    except ZeroDivisionError:
         raise ValueError("Diffusion constant, blur constant and time step parameters should not be 0")
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +452,257 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    variance_variance_loc: float,
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+    avg_frame_steps: float = 1,
+) -> float:
+    """Expected variance of the diffusion estimate obtained with CVE.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This function is based on eqn 24 from Vestergaard et al [2], but adapted for 1D.
```
ab412, line 515:
```suggestion
    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
    This function is based on equation 24 from Vestergaard et al [2], but adapted for 1D.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +452,257 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+        .. math::
+            R = \frac{1}{\Delta t} \int_{0}^{\Delta t}S(t) \left(1 - S(t)\right)dt
+
+        with
+
+        .. math::
+            S(t) = \int_{0}^{t} c(t') dt'
+
+        For no motion blur, we obtain: R = 0, while a continuously open shutter over the exposure
```
ab412, line 606:
```suggestion
        When there is no motion blur, we obtain: R = 0, whereas a continuously open shutter over the exposure
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Yes. The std error reported is biased. It's the best estimate we got at this point, but it's biased when inserting the estimates we obtained for `D` and `sigma` (I have also confirmed that there is a bias when simulating this). The only time we see no bias is when I compute it with the true values for `D` and `sigma`, but these will not be known when doing a real experiment. The paper also explicitly mentions that they should not be used to weight data. Look for `For importance weighting of estimates...` on page 7.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Why would I prefer `math.pow` over `**`? Personally I find the operator `**` much more readable.

That's a good point about division by zero. Though, the error message in the division by zero exception you suggest is not correct. The blur constant can definitely be zero (and in our case always is). The diffusion constant can actually also be zero numerically, and this indeed suggests that I should rewrite the equation a little bit.

Time step should not be zero; but I don't think it can be by construction.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
ab412: > Why would I prefer math.pow over **? Personally I find the operator ** much more readable and this code is not exactly performance critical.

Readability was the main thing, and also considering operator `**` has differing meanings based on context one might prefer the distinction. Comes down to preference and the intended audience I suppose. 

> Edit: I've also added a test which specifically includes a test with zero diffusion constant.

Even better!

lumicks/pylake/kymotracker/detail/msd_estimation.py
ab412: I see. Then perhaps the wording can be altered to indicate the reasoning behind the note? The idea being providing enough information from the API without needing to bounce back and forth with the source. What do you think of the something like following?

`While this provides some measure of uncertainty, the estimate should be avoided when weighting multiple diffusion estimates.  This is to prevent complications such as bias due to correlations between the estimated parameters and the estimated variances`

aafkevandenberg: And maybe add the  suggestion from the author in the same paragraph 'when cal-
culating the weighted mean of estimates from time series
of different lengths—the length N of a time series should
be used as weight when possible, since it is known exactly'

JoepVanlier: In the next PR, we'll be adding some functionality for actually weighting estimates coming from different tracks, so that the user never has to worry about this.

But yeah, I agree with both of these. How about for now we go with
```python
        Note: While this provides some measure of uncertainty, the estimate should not be used for
        weighting multiple diffusion estimates. This is to prevent complications such as bias due
        to correlations between the estimated parameters and estimated variances. Instead, when
        calculating the weighted mean of estimates from time series of different lengths—the length
        N of a time series should be used as weight when possible, since it is known exactly.
```

And then once we have the functionality on the group, we refer to the group functions there (e.g. note that you can compute diffusion constants for a group of tracks in the following way etc etc....).

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+def _var_cve_unknown_var(
+    diffusion_constant: float,
+    variance_loc: float,
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+    avg_frame_steps: float = 1,
+) -> float:
+    """Expected variance of the diffusion estimate obtained with CVE.
+
```
rpauszek, line 469:
```suggestion
    """Expected variance of the diffusion estimate obtained with CVE
         when the localization variance is not known a priori.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        Time step
+    num_points : int
+        Number of points in the trace
+    blur_constant : float
+        Motion blur constant. See :func:`lumicks.pylake.kymotracker.detail.msd_estimation._cve` for
+        more information.
+    avg_frame_steps : float
+        Average frame steps. This number is the average of the time between two localizations in
+        frames mean(delta time between frame / frame timestep). If all frames had successful
+        localization, this constant will be 1 [2].
```
rpauszek, line 540:
```suggestion
        localization, this constant will be 1 [3]_.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        Time step
+    num_points : int
+        Number of points in the trace
+    blur_constant : float
+        Motion blur constant. See :func:`lumicks.pylake.kymotracker.detail.msd_estimation._cve` for
+        more information.
+    avg_frame_steps : float
+        Average frame steps. This number is the average of the time between two localizations in
+        frames mean(delta time between frame / frame timestep). If all frames had successful
+        localization, this constant will be 1.
```
rpauszek, line 490:
```suggestion
    avg_frame_steps : float
        Average frame steps. This number is the average of the time between two localizations in
        frames mean(delta time between frame / frame timestep). If all frames had successful
        localization, this constant will be 1. [1]_
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    Note that this estimator is only valid in the case of pure diffusion without drift.
+
+    References
+    ----------
+    .. [1] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+           diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+           022726.
+    .. [2] Vestergaard, C. L. (2016). Optimizing experimental parameters for tracking of diffusing
+           particles. Physical Review E, 94(2), 022401.
```
rpauszek, line 583: this should move to the bottom to be consistent with other docstrings. also watch out for the numbering, needs to be 4 & 5 so the docs would be formatted properly

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    dt: float,
+    blur_constant: float = 0,
+    variance_loc: Optional[float] = None,
+    variance_variance_loc: Optional[float] = None,
+) -> tuple:
+    """Covariance based estimator.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This estimator was introduced in the work of Vestergaard et al [1]. The correction for
+    missing data was introduced in [2].
```
rpauszek, line 573:
```suggestion
    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
    This estimator was introduced in the work of Vestergaard et al [1]_. The correction for
    missing data was introduced in [2]_.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        The variance of the localization variance estimate (needed if variance_loc is provided).
+
+    Note
+    ----
+    With the SNR defined as:
+
+    .. math::
+        \sqrt{D \Delta t} / \sigma
+
+    With D the diffusion constant and :math:`\sigma`
```
rpauszek, line 628:
```suggestion
    With :math:`D` the diffusion constant and :math:`\sigma`
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    # Determine the average step size
+    average_frame_step = np.mean(np.diff(frame_indices))
+    avg_dt = average_frame_step * dt
+
+    dx = np.diff(x)
+    mean_dx_squared = np.mean(dx**2)
+
+    if len(x) < 3:
+        raise RuntimeError("Insufficient intervals for using the CVE")
```
rpauszek, line 642: maybe move towards the top and just throw right away?

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    if not variance_loc:
+        # Estimate the variance based on this frame
+        mean_dx_consecutive = np.mean(dx[1:] * dx[:-1])
+        # Equation 21 from [2] adapted for 1D.
+        diffusion_constant = mean_dx_squared / (2 * avg_dt) + mean_dx_consecutive / avg_dt
+        # Equation 15 from [1]. Note that static loc uncertainty is not affected by frame skipping.
```
rpauszek, line 649: the reference  numbers will have to be changed

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        variance_loc = (
+            blur_constant * mean_dx_squared + (2 * blur_constant - 1) * mean_dx_consecutive
+        )
+        var_diffusion = _var_cve_unknown_var(
+            diffusion_constant, variance_loc, dt, len(x), blur_constant, average_frame_step
+        )
+    else:
+        # We know the variance in advance. Equation 23 from [2] adapted for 1D.
```
rpauszek, line 657: same with the reference number

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        Time step
+    blur_constant : float
+        Motion blur coefficient. Should be between 0 and 1/4.
+    unit : str
+        Unit that the diffusion constant is specified in.
+    unit_label : str
+        Unit in TeX format used for plotting labels.
+    """
+
+    # We hardcode the blur constant for confocal for now (no motion blur)
```
rpauszek, line 703: 👍

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -336,6 +336,22 @@ def plot_msd(self, max_lag=None, **kwargs):
...
+
+        1. Covariance based estimator (CVE)
+
+        The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+        This estimator was introduced in the work of Vestergaard et al [5]_. The correction for
+        missing data was introduced in [6]_. The CVE is unbiased and practically optimal when the
+        signal-to-noise ratio (SNR) is bigger than 1. In this context, the SNR is defined
+        as: :math:`\sqrt{D \Delta t} / \sigma`.
+
+        2. MSD-based estimators
```
rpauszek, line 353: might be good to specifically state which ones these are (even though it's probably obvious)
```suggestion
        2. MSD-based estimators (OLS, GLS)
```

lumicks/pylake/kymotracker/tests/test_derived_quantities/test_msd.py
```diff
@@ -2,8 +2,11 @@
 import contextlib
 from lumicks.pylake.kymotracker.detail.msd_estimation import *
 from lumicks.pylake.kymotracker.detail.msd_estimation import (
+    _var_cve_known_var,
+    _var_cve_unknown_var,
     _msd_diffusion_covariance,
     _diffusion_ols,
+    _cve,
 )
```
rpauszek, line 10: hmm, why do we have import all and then import specific functions?

rpauszek (approved):
Nice, good to have an estimator that's not based on MSD. Just left some suggestions for minor formatting changes and a couple where the references got mixed up

lumicks/pylake/kymotracker/tests/test_derived_quantities/test_msd.py
JoepVanlier: Private functions are not imported when importing `*` unless explicitly listed in `__all__`.

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -12,7 +14,13 @@ class DiffusionEstimate:
...
         Estimate for the diffusion constant.
     std_err : float
-        Standard error,
+        Standard error.
+
+        Note: While this provides some measure of uncertainty, the estimate should not be used for
+        weighting multiple diffusion estimates. This is to prevent complications such as bias due
+        to correlations between the estimated parameters and estimated variances. Instead, when
+        calculating the weighted mean of estimates from time series of different lengths—the length
+        N of a time series should be used as weight when possible, since it is known exactly.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +455,259 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    With the SNR defined as:
+
+    .. math::
+        \sqrt{D \Delta t} / \sigma
+
+    With D the diffusion constant and :math:`\sigma`
+    This method requires SNR > 1. Below SNR = 1 it degrades rapidly.
+    """
+    if not 0 <= blur_constant < 0.25:
+        raise ValueError("Motion blur constant should be between 0 and 1/4")
```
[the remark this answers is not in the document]
JoepVanlier: Good catch! Fixing.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+           022726.
+    """
+    # Note that it uses a different definition of epsilon to circumvent a division by zero for D=0.
+    epsilon = variance_loc / dt - 2.0 * blur_constant * diffusion_constant
+    avg_diff = avg_frame_steps * diffusion_constant
+    numerator = 6.0 * avg_diff**2 + 4.0 * epsilon * avg_diff + 2.0 * epsilon**2
+    denominator = num_points * avg_frame_steps**2
+    term1 = numerator / denominator
+    term2 = 4.0 * (avg_diff + epsilon) ** 2 / (num_points**2 * avg_frame_steps**2)
+    return term1 + term2
```
tobiasjj, line 528: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    dt: float,
+    blur_constant: float = 0,
+    variance_loc: Optional[float] = None,
+    variance_variance_loc: Optional[float] = None,
+) -> tuple:
+    r"""Covariance based estimator.
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This estimator was introduced in the work of Vestergaard [11]_. The correction for missing
+    data was introduced in [12]_.
```
tobiasjj, line 599:
```suggestion
    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
    This estimator was introduced in the work of Vestergaard [12]_. The correction for missing
    data was introduced in [13]_.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    With :math:`D` the diffusion constant and :math:`\sigma`
+    This method requires SNR > 1. Below SNR = 1 it degrades rapidly [11]_.
+
+    References
+    ----------
+    .. [11] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+            diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+            022726.
+    .. [12] Vestergaard, C. L. (2016). Optimizing experimental parameters for tracking of diffusing
```
tobiasjj, line 657:
```suggestion
    .. [12] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
            diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
            022726.
    .. [13] Vestergaard, C. L. (2016). Optimizing experimental parameters for tracking of diffusing
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+    average_frame_step = np.mean(np.diff(frame_indices))
+    avg_dt = average_frame_step * dt
+
+    dx = np.diff(x)
+    mean_dx_squared = np.mean(dx**2)
+
+    if not variance_loc:
+        # Estimate the variance based on this frame
+        mean_dx_consecutive = np.mean(dx[1:] * dx[:-1])
+        # Equation 21 from [12] adapted for 1D.
```
tobiasjj, line 676:
```suggestion
        # Equation 14 from [12] and 21 from [13] adapted for 1D.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        diffusion_constant = mean_dx_squared / (2 * avg_dt) + mean_dx_consecutive / avg_dt
```
tobiasjj, line 677: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        # Equation 15 from [11]. Note that static loc uncertainty is not affected by frame skipping.
```
tobiasjj, line 678:
```suggestion
        # Equation 15 from [12]. Note that static loc uncertainty is not affected by frame skipping.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        variance_loc = (
+            blur_constant * mean_dx_squared + (2 * blur_constant - 1) * mean_dx_consecutive
+        )
+        var_diffusion = _var_cve_unknown_var(
+            diffusion_constant, variance_loc, dt, len(x), blur_constant, average_frame_step
+        )
+    else:
+        # We know the variance in advance. Equation 23 from [12] adapted for 1D.
```
tobiasjj, line 686:
```suggestion
        # We know the variance in advance. Equation 16 from [12] and 23 from [13] adapted for 1D.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        diffusion_constant = (mean_dx_squared - 2.0 * variance_loc) / (
+            2.0 * (avg_dt - 2.0 * blur_constant * dt)
+        )
```
tobiasjj, line 689: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    # Determine the average step size
+    average_frame_step = np.mean(np.diff(frame_indices))
+    avg_dt = average_frame_step * dt
+
+    dx = np.diff(x)
+    mean_dx_squared = np.mean(dx**2)
+
+    if not variance_loc:
+        # Estimate the variance based on this frame
```
tobiasjj, line 674: Is it just one frame? I thought x (and therefore dx) are determined from many consecutive frames?

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+        Time step
+    blur_constant : float
+        Motion blur coefficient. Should be between 0 and 1/4.
+    unit : str
+        Unit that the diffusion constant is specified in.
+    unit_label : str
+        Unit in TeX format used for plotting labels.
+    """
+
+    # We hardcode the blur constant for confocal for now (no motion blur)
```
tobiasjj, line 732: Hm, I would probably move this comment into the calling function of `kymotrack` (see comment below).

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -397,6 +421,16 @@ def estimate_diffusion(self, method, max_lag=None):
             )
 
         frame_idx, positions = np.array(self.time_idx, dtype=int), np.array(self.position)
+        unit_labels = {
+            "unit": f"{self._kymo._calibration.unit}^2 / s",
+            "unit_label": f"{self._kymo._calibration.unit_label}$^2$/s",
+        }
+
+        if method == "cve":
+            return estimate_diffusion_cve(
```
tobiasjj, line 430:
```suggestion
            # We hardcode the blur constant for confocal for now (no motion blur)
            return estimate_diffusion_cve(
```

tobiasjj (approved):
Nice work! Only some minor nit picks and suggestions for adding references to the initial publication of Vestergaard from the year 2014 (for symmetry reasons). LGTM!

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -38,14 +46,17 @@ def __float__(self):
 
 
 def calculate_msd(frame_idx, position, max_lag):
-    """Estimate the Mean Square Displacement (MSD) for various time lags.
+    r"""Estimate the Mean Square Displacement (MSD) for various time lags.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+def _var_cve_known_var(
+    diffusion_constant: float,
+    variance_loc: float,
+    variance_variance_loc: float,
+    dt: float,
+    num_points: int,
+    blur_constant: float = 0,
+    avg_frame_steps: float = 1,
+) -> float:
+    r"""Expected variance of the diffusion estimate obtained with CVE.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+    The covariance-based diffusion estimator provides a simple unbiased estimator of diffusion.
+    This function is based on equation 24 from Vestergaard [10]_, but adapted for 1D.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -447,3 +469,274 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
...
+
+            \frac{1}{N} \sum_{i=1}^{N_{frames - 1}} \frac{t_{i+1} - t_{i}}{dt}
+
+        where :math:`dt` is the time step. If all frames had successful localization, this constant
+        will be 1 [10]_.
+
+    References
+    ----------
+    .. [10] Vestergaard, C. L. (2016). Optimizing experimental parameters for tracking of diffusing
+            particles. Physical Review E, 94(2), 022401.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Good catch! Clearly that should've been `track`. Fixing.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Yeah... hmm... I wonder why I put it here. My guess is that I moved the =0 upstream at some point during writing this and forgot to move the comment with. Anyhow, good catch! 👍

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/msd_estimation.py
tobiasjj: 👍

PR #381, opened 2022-09-20
Rename fdfit functions

tobiasjj:
**Why this PR**
As a user I want to have consistent naming of fitting models.

I renamed the public facing models as follows:

`inverted_marko_siggia_simplified` -> `wlc_marko_siggia_distance`
`marko_siggia_simplified` -> `wlc_marko_siggia_force`
`marko_siggia_ewlc_distance` -> `ewlc_marko_siggia_distance`
`marko_siggia_ewlc_force` -> `ewlc_marko_siggia_force`
`odijk` -> `ewlc_odijk_distance`
`inverted_odijk` -> `ewlc_odijk_force`
`freely_jointed_chain` -> `efjc_distance`
`inverted_freely_jointed_chain` -> `efjc_force`
`twistable_wlc` -> `twlc_distance`
`inverted_twistable_wlc` -> `twlc_force`

I refrained from simplifying `wlc_marko_siggia_*` to `wlc_*`, as I am unsure, if we should simplify the names already at this level and not only when we provide convenience functions to assemble a composition of models.

I also renamed the functions of the model implementations accordingly. ~However, there an `i` indicates that the model was inverted to calculate the `*_idistance` and `*_iforce`.~

~Additionally, I took the opportunity to reorder the functions in `models` and `model_implementation`, bundled into a separate commit, so it can be easily reverted, if you do not agree.~ Postponed for now, probably I will open another PR.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,17 @@
 ### Other changes
 
 * Updated benchmark to not use deprecated functions and arguments. Prior to this change, running the benchmark would produce deprecation warnings.
+* Renamed force distance fitting functions. They are deprecated now and will be removed in the future:
+    *  `inverted_marko_siggia_simplified` -> `wlc_marko_siggia_distance`
+    *  `marko_siggia_simplified` -> `wlc_marko_siggia_force`
+    *  `marko_siggia_ewlc_distance` -> `ewlc_marko_siggia_distance`
+    *  `marko_siggia_ewlc_force` -> `ewlc_marko_siggia_force`
+    *  `odijk` -> `ewlc_odijk_distance`
+    *  `inverted_odijk` -> `ewlc_odijk_force`
+    *  `freely_jointed_chain` -> `efjc_distance`
+    *  `inverted_freely_jointed_chain` -> `efjc_force`
+    *  `twistable_wlc` -> `twlc_distance`
+    *  `inverted_twistable_wlc` -> `twlc_force`
 
 ## v0.13.1 | 2022-09-08
 
--- a/lumicks/pylake/fitting/tests/test_fd_models.py
+++ b/lumicks/pylake/fitting/tests/test_fd_models.py
@@ -5,74 +5,97 @@ import numpy as np
 import pytest
 
 
+def test_deprecated_models():
+    """Check that functions are deprecated and point to the corresponding new ones"""
+    old_new = [
+        (marko_siggia_simplified, wlc_marko_siggia_force),
+        (inverted_marko_siggia_simplified, wlc_marko_siggia_distance),
+        (marko_siggia_ewlc_distance, ewlc_marko_siggia_distance),
+        (marko_siggia_ewlc_force, ewlc_marko_siggia_force),
+        (odijk, ewlc_odijk_distance),
+        (inverted_odijk, ewlc_odijk_force),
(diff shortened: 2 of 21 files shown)


aafkevandenberg (commented):
Very nice! The FD fitting code is now much more readable.

When testing the functions, I realized that `fjc_distance` and `fjc_force` have a stretch modulus. So this is actually not the freely jointed chain model... For consistency, it should be called _extensible_ freely jointed chain, or 
```
efjc_distance
efjc_force
```
What do you think?

lumicks/pylake/fitting/tests/test_fd_models.py
```diff
@@ -129,15 +160,15 @@ def test_model_reprs():
...
-        (model_impl.marko_siggia_ewlc_solve_distance, ["Lp", "Lc", "St", "kT"]),
-        (model_impl.tWLC, ["Lp", "Lc", "St", "kT"]),
-        (model_impl.invtWLC, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.ewlc_odijk_iforce, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.ewlc_odijk_distance, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.efjc_distance, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.efjc_solve_iforce, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.wlc_marko_siggia_force, ["Lp", "Lc", "kT"]),
+        (model_impl.ewlc_marko_siggia_force, ["Lp", "Lc", "St", "kT"]),
+        (model_impl.ewlc_marko_siggia_distance, ["Lp", "Lc", "St", "kT"]),
```
aafkevandenberg, line 169: Why is it that the `ewlc_marko_siggia` has no 'inverted' version?

aafkevandenberg (approved):
LGTM!

changelog.md
```diff
@@ -9,6 +9,17 @@
 ### Other changes
 
 * Updated benchmark to not use deprecated functions and arguments. Prior to this change, running the benchmark would produce deprecation warnings.
+* Renamed force distance fitting functions. Old names will be deprecated in the future:
```
JoepVanlier, line 12: They are deprecated now and will be removed in the future.

JoepVanlier (approved):
Yeah much nicer. This way we don't bother users with implementational details. I have one minor request for the changelog.

Also, while doing extensive changes/renaming, I think it'd be nice to add an `__all__` variable to `fitting/models.py`. That way you both have a nice manifest of what's inside at the top, and an easy way to import all the models elsewhere. What do you think?

lumicks/pylake/fitting/tests/test_fd_models.py
tobiasjj: Original reason is that independent and dependent variable are on both sides of the equation. After a discussion with @JoepVanlier we came to the conclusion to remove the `i` everywhere in favor of an even cleaner naming convention, as the `i` indicates an unimportant implementation detail to the user.

aafkevandenberg: Maybe such details could be discussed in a theory section.

PR #382, opened 2022-09-20
ci: mpl setup/teardown again

JoepVanlier:
**Why this PR?**
Turns out the way I solved this in https://github.com/lumicks/pylake/pull/379 can lead to rare [test failures](https://github.com/lumicks/pylake/actions/runs/3089114566/jobs/4996352870) if you forget the decorator. For that reason, I propose just setting this up as an autouse fixture instead, basically because fewer things can go wrong when writing tests.

Note that it required a bit more rework than I would've liked, since it turned out that we really needed the test configuration defined at a level higher (since we keep some tests with their submodules rather than all in `/tests/`).

I was a little concerned that this may have a performance impact, but it seems that it's not measurable:

Always using the decorator:
```
======================= 884 passed, 4 skipped in 41.29s ========================
```

Without always using context:
```
======================= 884 passed, 4 skipped in 41.92s ========================
```

Note: The first commit is there just to show what was forgotten in the last PR, but it would be dropped if I merge this.


lumicks/pylake/conftest.py
```diff
@@ -0,0 +1,92 @@
...
+    try:
+        orig_backend = plt.get_backend()
+        with plt.style.context(["classic", "_classic_test_patch"]):
+            # Use a headless backend for testing, note that passing it as a context parameter
+            # did not work.
+            plt.switch_backend("agg")
+            yield
+    finally:
+        plt.close("all")
+        plt.switch_backend(orig_backend)
```
dean0x7d, line 92: I think that `plt.close("all")` makes sense here, but with `autouse=True` does it still make sense to keep switching backends as part of the fixture? The backend is switched between every test, but effectively, every test sees the `agg` backend. So why not just do it once on startup in `conftest.py`?

JoepVanlier: Yeah, that's a good point. It's excessive. I'll switch the `backend` once.

dean0x7d (approved):
`@lk_pr_approved`

PR #383, opened 2022-09-20
Ensure same source data for all tracks in group

rpauszek:
**Why this PR?**
Some downstream analyses on kymo tracks require information from the source `Kymo`  (`refine_tracks_centroid` and `refine_tracks_gaussian`). Until now we've assumed that all tracks have the same source instance and just used the first track in the group to get this information. Here we specifically enforce this so our assumptions are always valid.

In addition, I added properties to look up the source `Kymo` and color channel, and made some minor changes to downstream functions that ensure we're working on a `KymoTrackGroup` rather than a raw `list` or `tuple` of `KymoTrack`s

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -429,6 +429,17 @@ class KymoTrackGroup:
 
     def __init__(self, kymo_tracks):
         self._src = kymo_tracks
+        if self._src:
+            self._validate_single_source(kymo_tracks)
+
+    def _validate_single_source(self, kymo_tracks):
+        kymos = set([track._kymo._id for track in kymo_tracks])
+        channels = set([track._channel for track in kymo_tracks])
+
+        assert len(kymos) == 1, "All tracks must have the same source kymograph."
+        assert len(channels) == 1, "All tracks must be from the same color channel."
+
+        return next(iter(kymos)), next(iter(channels))
 
     def __iter__(self):
         return self._src.__iter__()
@@ -441,6 +452,9 @@ class KymoTrackGroup:
 
     def __setitem__(self, item, value):
         raise NotImplementedError("Cannot overwrite KymoTracks.")
+
+    def __bool__(self):
+        return bool(self._src)
 
     def __copy__(self):
         return KymoTrackGroup(copy(self._src))
@@ -487,10 +501,17 @@ class KymoTrackGroup:
         return len(self._src)
 
     def extend(self, other):
+        other = [other] if isinstance(other, KymoTrack) else other
+
+        other_kymo, other_channel = self._validate_single_source(other)
+        if self:
+            assert self[0]._kymo._id == other_kymo, "All tracks must have the same source kymograph."
+            assert self._channel == other_channel, "All tracks must be from the same color channel."
+
         if isinstance(other, self.__class__):
             self._src.extend(other._src)
-        elif isinstance(other, KymoTrack):
-            self._src.extend([other])
+        elif isinstance(other, list):
+            self._src.extend(other)
         else:
             raise TypeError(
                 f"You can only extend a {self.__class__} with a {self.__class__} or " f"{KymoTrack}"
@@ -524,7 +545,7 @@ class KymoTrackGroup:
         self._src = [track for track in self._src if not track.in_rect(rect)]
 
     def __repr__(self):
-        return f"{self.__class__.__name__}(N={len(self._src)})"
+        return f"{self.__class__.__name__}(N={len(self)})"
 
     def save(self, filename, delimiter=";", sampling_width=None):
         """Export kymograph tracks to a csv file.
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -27,7 +27,7 @@ def export_kymotrackgroup_to_csv(filename, kymotrack_group, delimiter, sampling_
         raise RuntimeError("No kymograph tracks to export")
 
     time_units = "seconds"
-    position_units = kymotrack_group[0]._kymo._calibration.unit
+    position_units = kymotrack_group._kymo._calibration.unit
 
     idx = np.hstack([np.full(len(track), idx) for idx, track in enumerate(kymotrack_group)])
     coords_idx = np.hstack([track.coordinate_idx for track in kymotrack_group])
@@ -429,7 +429,7 @@ class KymoTrackGroup:
 
     def __init__(self, kymo_tracks):
         self._src = kymo_tracks
-        if self._src:
+        if self:
             self._validate_single_source(kymo_tracks)
 
     def _validate_single_source(self, kymo_tracks):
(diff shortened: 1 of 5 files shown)


JoepVanlier (commented):
It's nice to enforce this assumption we've been making so far and have an easy to use accessor for us. I'm not sure about using `id` to identify Kymo's though.

Note that there is still a reference to `lines[0]._kymo` on line `346`: `track_width = line_width * lines[0]._kymo.pixelsize[0]`. Would be good to fix that one up as well.

Also note the request on the id's.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -486,10 +499,22 @@ def __len__(self):
         return len(self._src)
 
     def extend(self, other):
+        other = [other] if isinstance(other, KymoTrack) else other
+
+        if len(self):
+            new_kymos = set([id(track._kymo) for track in other])
```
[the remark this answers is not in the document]
rpauszek: good idea. I agree it's too much to piggyback on this PR, but I added the placeholder property and just use `id()` for now

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -444,6 +449,14 @@ def __setitem__(self, item, value):
     def __copy__(self):
         return KymoTrackGroup(copy(self._src))
 
+    @property
+    def _kymo(self):
+        return self[0]._kymo
+
+    @property
+    def _channel(self):
+        return self[0]._channel
```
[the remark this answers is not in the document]
tobiasjj: I think the same!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -496,11 +504,11 @@ def extend(self, other):
...
             new_kymos = set([track._kymo._id for track in other])
             assert (
-                len(new_kymos) == 1 and self[0]._kymo._id == list(new_kymos)[0]
+                len(new_kymos) == 1 and self._kymo._id == list(new_kymos)[0]
             ), "All tracks must have the same source kymograph."
             new_channels = set([track._channel for track in other])
             assert (
-                len(new_channels) == 1 and self[0]._channel == list(new_channels)[0]
+                len(new_channels) == 1 and self._channel == list(new_channels)[0]
             ), "All tracks must be from the same color channel."
```
tobiasjj, line 512: Would it make sense to have one central function like `check_kymo_compatiblity()`, which can be called from here and the `KymoTrackGroup.__init__()`? This would deduplicate the code to check if kymos are compatible.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -428,6 +428,11 @@ class KymoTrackGroup:
 
     def __init__(self, kymo_tracks):
         self._src = kymo_tracks
+        if self._src:
+            n_kymos = len(set([track._kymo._id for track in kymo_tracks]))
+            assert n_kymos == 1, "All tracks must have the same source kymograph."
+            n_channels = len(set([track._channel for track in kymo_tracks]))
+            assert n_channels == 1, "All tracks must be from the same color channel."
```
tobiasjj, line 435: Would it make sense to have one central function, which could be called from here and further down (line 504-512)? This would deduplicate some of the code to check if kymos are compatible.
```
def validate_compatibility(self, kymo_tracks, compare_to_self=False):
    new_kymos = set([track._kymo._id for track in kymo_tracks])
    assert (
        len(new_kymos) == 1 and (not compare_to_self or self._kymo._id == list(new_kymos)[0])
    ), "All tracks must have the same source kymograph."
    new_channels = set([track._channel for track in other])
    assert (
        len(new_channels) == 1 and (not compare_to_self or self._channel == list(new_channels)[0])
    ), "All tracks must be from the same color channel."
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -486,10 +499,22 @@ def __len__(self):
...
+
+        if len(self):
+            new_kymos = set([track._kymo._id for track in other])
+            assert (
+                len(new_kymos) == 1 and self._kymo._id == list(new_kymos)[0]
+            ), "All tracks must have the same source kymograph."
+            new_channels = set([track._channel for track in other])
+            assert (
+                len(new_channels) == 1 and self._channel == list(new_channels)[0]
+            ), "All tracks must be from the same color channel."
```
tobiasjj, line 512: See comment above.

lumicks/pylake/kymotracker/kymotrack.py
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: good idea, I switched it up a bit from your suggestion because I find that negating things tends to be more confusing than using the ternary

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -428,6 +428,19 @@ class KymoTrackGroup:
...
         self._src = kymo_tracks
+        if self._src:
+            self._validate_compatibility(kymo_tracks, compare_to_self=False)
+
+    def _validate_compatibility(self, kymo_tracks, compare_to_self=False):
+        new_kymos = set([track._kymo._id for track in kymo_tracks])
+        new_channels = set([track._channel for track in kymo_tracks])
+
+        assert len(new_kymos) == 1 and (
+            self._kymo._id == list(new_kymos)[0] if compare_to_self else True
```
JoepVanlier, line 439: You don't have to convert the set to a `list` to extract one element. Just do `next(iter(new_kymos))`. Same for channels.

Personally, I find the `compare_to_self` argument a bit smelly. You can just return the unique kymo and unique channel and throw if it's not unique. Then compare upstream to the kymo already associated with the group if need be. There's only one place where you actually need to do that.

JoepVanlier (changes requested):

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -486,10 +507,15 @@ def __len__(self):
         return len(self._src)
 
     def extend(self, other):
+        other = [other] if isinstance(other, KymoTrack) else other
+
+        if len(self):
```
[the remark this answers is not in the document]
rpauszek: It's possible to end up with an empty group (if you have really bad tracking parameters for instance). Without testing for length this will throw an `IndexError` because the empty group tries to access `self[0]`. However, I think we should still allow an empty group to be extended

If you unintentionally try to access `group._kymo` or `group._channel` you do get the original `IndexError` thrown from `__getitem__()`. I thought about adding a more descriptive error message in these properties themselves but I think for internal API it's a bit overkill?

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: good points, I like it better now

lumicks/pylake/kymotracker/kymotrack.py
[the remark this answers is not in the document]
JoepVanlier: Ah right. I meant more why are you checking the length specifically. All we want to check is empty, so can't we just check `if self._src:` since an empty list is falsy?

JoepVanlier (approved):
Good work. Only one minor thing regarding the `len()`, but otherwise looks solid :)

PR #384, opened 2022-09-20
Pixel threshold for centroid refinement

rpauszek:
**Why this PR?**
Sometimes on refinement you don't want to include interpolated points where the signal level is too low for reliable localization (eg, can be detrimental for MSD analysis). This PR adds an optional keyword argument to do this filtering.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -3,6 +3,7 @@ from .detail.scoring_functions import kymo_score
 from .kymotrack import KymoTrack, KymoTrackGroup, _sample_from_image
 from .detail.gaussian_mle import gaussian_mle_1d, overlapping_pixels
 from .detail.peakfinding import (
+    blur_dilate_image,
     peak_estimate,
     refine_peak_based_on_moment,
     merge_close_peaks,
@@ -10,7 +11,6 @@ from .detail.peakfinding import (
 )
 from .detail.localization_models import GaussianLocalizationModel
 import numpy as np
-from scipy.ndimage import gaussian_filter, grey_dilation
 import warnings
 from deprecated.sphinx import deprecated
 
@@ -379,6 +379,7 @@ def refine_tracks_centroid(tracks, track_width=None, *, pixel_threshold=None):
         # Must be positive otherwise refinement fails
         raise ValueError(f"track_width should be larger than zero")
 
+    print(tracks)
     track_width_pixels = np.ceil(track_width / tracks._kymo.pixelsize[0])
     half_width = np.ceil(0.5 * track_width_pixels).astype(int)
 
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -377,6 +383,38 @@ def refine_tracks_centroid(tracks, track_width=None):
     track_width_pixels = np.ceil(track_width / tracks._kymo.pixelsize[0])
 
     interpolated_tracks = [track.interpolate() for track in tracks]
+    image = tracks._kymo.get_image(tracks._channel)
+
+    if pixel_threshold is not None:
+        if pixel_threshold <= 0:
+            raise ValueError(f"pixel_threshold should be larger than zero")
+
+        blurred = gaussian_filter(image, [0.5, 0])
```
JoepVanlier, line 392: I would really prefer it if we don't duplicate this filter. I fear that if we add this here, we will forget about the fact that there is another one like these in the actual peak detection and they risk going out of sync. I would pull this out into a small subfunction or leave a note on both sides. I'd probably prefer the subfunction way.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -377,6 +383,38 @@ def refine_tracks_centroid(tracks, track_width=None):
...
+        blurred_tracks = [
+            KymoTrack(track.time_idx, track.coordinate_idx, blurred_kymo, tracks._channel)
+            for track in interpolated_tracks
+        ]
+
+        sample_window = int(track_width_pixels / 2)
+        keep_nodes = [
+            np.array(track.sample_from_image(sample_window, np.max)) >= pixel_threshold
+            for track in blurred_tracks
+        ]
```
JoepVanlier, line 411: To me, this seems a bit smelly and it feels like there should be a low level API that just does this low level.

mikhas: The feature itself seems to work correctly (tested it).

JoepVanlier: I tried it as well, and it changes some tracks even when I did no stitching operation.

```python
kymo_ds = kymo.downsampled_by(time_factor=2)
lines = lk.track_greedy(kymo_ds, "green", **settings)
refined_lines = lk.refine_tracks_centroid(
  lines,
  track_width=settings["track_width"],
  pixel_threshold=settings["pixel_threshold"]
)  # AFAIK, this line should've been a nullop but it's not.
```

Yellow is after refinement.
image

JoepVanlier (changes requested):
Would be good to check why the results are changing.

lumicks/pylake/kymotracker/kymotracker.py
rpauszek: I was worried about that too, I was just wondering if it was too much for basically a single line function. But I think it's worth it to avoid bugs here. I also thought we should be comparing against the blurred and dilated image, but I'm not 100% sure on this one. I pulled this stuff out into a separate function, what do you think?

lumicks/pylake/kymotracker/kymotracker.py
rpauszek: > To me, this seems a bit smelly and it feels like there should be a low level API that just does this low level.

yeah you're right, pulled the image sampling stuff out to a separate function that returns the sub-array basically and left the `reduce` in the method. thoughts?

rpauszek:
The unexpected changes in results mentioned above result from tracks being too close together. Because these are not taken into account during this refinement, it leads to bias towards the middle of the two tracks. Closing this for now; it seems that gaussian refinement will be the only proper way to handle situations like this

PR #385, opened 2022-09-22
Docs] curate / update files and channels

aafkevandenberg:
As a first step in updating the tutorials, I updated file.rst file in the docs:

A summary of the changes:
- Make sure that all the code runs at once
- Updated the text on some places, to make it more clear and concise. 
- I used the file 'WTCas9_tether=0011_replicate=0000_date=2021_12_03__13_28_26.h5', which is in the folder 'Pylake Docs Datasets' on Google drive.

### DIFF
--- a/docs/tutorial/file.rst
+++ b/docs/tutorial/file.rst
@@ -5,10 +5,10 @@ Files and channels
 
     :nbexport:`Download this page as a Jupyter notebook <self>`
 
-Opening a Bluelake HDF5 file is very simple::
+Open a Bluelake HDF5 file as follows::
 
     import lumicks.pylake as lk
-
+    
     file = lk.File("example.h5")
 
 Contents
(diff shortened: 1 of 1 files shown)


docs/tutorial/file.rst
```diff
@@ -102,32 +121,25 @@ All channel data can be accessed using the above method. High frequency force da
...
 The `timestamps` attribute returns the measurement time in nanoseconds since epoch (January 1st 1970, midnight UTC/GMT).
-Note that since these values are typically very large, they cannot be converted to floating point without losing precision::
 
-    >>> t = f1x_timestamps[0]
-    >>> roundtrip_t = np.int64(np.float64(t))
-    >>> print(t - roundtrip_t)
-    24
-
-The reason for this is that timestamps exceed the maximum integer value representable by the mantissa.
 The relative time values in seconds can also be accessed directly::
```
JoepVanlier, line 138: Did you remove the warning about the lossiness of converting timestamps to/from floats to try and not distract from the main line of the story?

Would it be possible to add it back somewhere else? Or devote a sentence that care must be taken that these timestamps cannot be converted to floating point without losing some precision (maybe without the explicit example so that it doesn't distract too much)? Reason being that we have seen this go wrong a few times.

docs/tutorial/file.rst
```diff
@@ -305,4 +322,4 @@ particular, high frequency channels tend to take up a lot of space, and aren't a
 
     file.save_as("no_hf.h5", omit_data={"Force HF/*"})  # Omit high frequency force data from export
 
-We use `fnmatch` patterns for specifying which fields to omit from the saved `h5` file.
+We use `fnmatch` patterns for specifying which fields to omit from the saved `h5` file; the ``*`` in the line of code above means that any channel starting with ``""Force HF/""`` should be omitted from the file.
```
JoepVanlier, line 325: 👌

JoepVanlier: Might be nice to link the `fnmatch` to here: https://docs.python.org/3/library/fnmatch.html in case people want to look up the exact syntax.

```suggestion
We use `fnmatch <https://docs.python.org/3/library/fnmatch.html>`_ patterns for specifying which fields to omit from the saved `h5` file; the ``*`` in the line of code above means that any channel starting with ``""Force HF/""`` should be omitted from the file.
```

docs/tutorial/file.rst
```diff
@@ -5,10 +5,10 @@ Files and channels
 
     :nbexport:`Download this page as a Jupyter notebook <self>`
 
-Opening a Bluelake HDF5 file is very simple::
+Open a Bluelake HDF5 file as follows::
 
     import lumicks.pylake as lk
-
+    
     file = lk.File("example.h5")
```
JoepVanlier, line 12: Is the file still going to be called `example.h5`?

JoepVanlier (changes requested):
Great! Thanks for updating this! :)

A few small suggestions, but overall looks good!

docs/tutorial/file.rst
aafkevandenberg: in the future there should be a link to Zenodo on this page, I guess we can just keep it like this for now?

docs/tutorial/file.rst
aafkevandenberg: didn't realize that this was actually an error that pops up regularly. Put it back, with the example.

docs/tutorial/file.rst
JoepVanlier: It would likely still have a filename. But yes, it's fine to rename it later. 👍

docs/tutorial/file.rst
JoepVanlier: 👍

JoepVanlier (approved):
LGTM!

PR #386, opened 2022-09-22
Fixup unescaped backslash in docstrings

rpauszek:
**Why this PR?**
Unescaped `\` in latex markdown in some docstrings was causing warning -> error. Also took the opportunity to use the `\ln{}` latex operator instead of the `mathrm{ln}`


JoepVanlier (approved):
Looks good Ray. Passes on my machine now too.

PR #387, opened 2022-09-26
Make function mentions in the documentation link to pages with the docstring

tobiasjj:
**Why this PR?**
As a user, I would like function mentions in the documentation to actually link to pages with the docstrings for those functions

To link python objects -mentioned in the documentation- to their corresponding documentation in the api section, I (tried to) replace(d) all default code blocks `PYTHON_OBJECT` with corresponding python roles (https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#python-roles). Additionally, I ensured to use the correct path and name of the python objects. The result is working links.

Additionally, I used a custom css file, to change the default color of the links from "black" to the same color of the sample code blocks (compare `KymoTrack` in the following images), as I find the colored links in the text easier to identify as clickable links:
Black
Colored

However, this has the side effect, that also the links on the API overview page are "orange". What do you think?

Please, find the compiled version of the docs there: https://lumicks-pylake.readthedocs.io/en/sphinx_links/

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -329,7 +329,7 @@ How to do active calibration in Pylake
 
 Using `pylake`, the procedure to use active calibration is not very different from passive calibration.
 However, it does require some additional data channels as inputs.
-Instead of using the :class:`~.PassiveCalibrationModel` presented in the previous section, we now use the :class:`ActiveCalibrationModel`.
+Instead of using the :class:`~.PassiveCalibrationModel` presented in the previous section, we now use the :class:`~lumicks.pylake.ActiveCalibrationModel`.
 
 In this tutorial, we're going to assume the nanostage was used as driving input::
 
--- a/lumicks/pylake/fdcurve.py
+++ b/lumicks/pylake/fdcurve.py
@@ -15,7 +15,7 @@ class FdCurve(DownsampledFD):
     """An FD curve exported from Bluelake
 
     By default, the primary force and distance channels are `downsampled_force2`
-    and `distance1`. Alternatives can be selected using `FdCurve.with_channels()`.
+    and `distance1`. Alternatives can be selected using :func:`FdCurve.with_channels()`.
     Note that it does not modify the FD curve in place but returns a copy.
 
     Attributes
@@ -218,7 +218,7 @@ class FdCurve(DownsampledFD):
         Parameters
         ----------
         **kwargs
-            Forwarded to `~matplotlib.pyplot.scatter`.
+            Forwarded to :func:`matplotlib.pyplot.scatter`.
         """
         import matplotlib.pyplot as plt
 
--- a/lumicks/pylake/file.py
+++ b/lumicks/pylake/file.py
@@ -66,7 +66,7 @@ class File(Group, Force, DownsampledFD, BaselineCorrectedForce, PhotonCounts, Ph
 
     @classmethod
     def from_h5py(cls, h5py_file):
-        """Directly load an existing `h5py.File`"""
+        """Directly load an existing `h5py.File <https://docs.h5py.org/en/latest/high/file.html>`_"""
         new_file = cls.__new__(cls)
         new_file.h5 = h5py_file
         new_file._check_file_format()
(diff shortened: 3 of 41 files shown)


aafkevandenberg (commented):
Nice, a very useful addition!

The change of color of the links on the API page is fine with me.

I don't find it entirely intuitive that the clickable and non-clickable code have the same color (see the example you gave above). Even though the clickable code is bold. Could you maybe make the clickable code dark blue? That's what they did in the numpy documentation: https://numpy.org/doc/stable/user/misc.html (I also like the shade of pink they use for the code :P )

lumicks/pylake/file.py
```diff
@@ -66,7 +66,7 @@ def _check_file_format(self):
 
     @classmethod
     def from_h5py(cls, h5py_file):
-        """Directly load an existing `h5py.File`"""
+        """Directly load an existing `h5py.File <https://docs.h5py.org/en/latest/high/file.html>`_"""
```
JoepVanlier, line 69: Love also adding some linkouts 🏆

lumicks/pylake/force_calibration/detail/power_models.py
```diff
@@ -24,7 +24,7 @@ def fit_analytical_lorentzian(ps):
         - `fc` : corner frequency [Hz]
         - `D`: diffusion constant [V^2/s]
         - `sigma_fc`, `sigma_D`: 1-sigma confidence intervals for `fc` and `D`
-        - `ps_fit`: `PowerSpectrum` object with model fit
+        - `ps_fit`: :func:`~lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum` object with model fit
```
JoepVanlier, line 27: Isn't this a class?

docs/tutorial/force_calibration.rst
```diff
@@ -329,7 +329,7 @@ How to do active calibration in Pylake
 
 Using `pylake`, the procedure to use active calibration is not very different from passive calibration.
 However, it does require some additional data channels as inputs.
-Instead of using the :class:`~.PassiveCalibrationModel` presented in the previous section, we now use the :class:`ActiveCalibrationModel`.
+Instead of using the :class:`~.PassiveCalibrationModel` presented in the previous section, we now use the :class::class:`~lumicks.pylake.ActiveCalibrationModel`.
```
JoepVanlier, line 333:
```suggestion
Instead of using the :class:`~.PassiveCalibrationModel` presented in the previous section, we now use the :class:`~lumicks.pylake.ActiveCalibrationModel`.
```
find/replace mishap?

lumicks/pylake/fdcurve.py
```diff
@@ -15,7 +15,7 @@ class FdCurve(DownsampledFD):
     """An FD curve exported from Bluelake
 
     By default, the primary force and distance channels are `downsampled_force2`
-    and `distance1`. Alternatives can be selected using `FdCurve.with_channels()`.
+    and `distance1`. Alternatives can be selected using :func:`FdCurve.with_channels`.
```
JoepVanlier, line 18:
```suggestion
    and `distance1`. Alternatives can be selected using :func:`FdCurve.with_channels()`.
```

JoepVanlier (approved):
Nice work. A few small comments. I think the color change is nice too.

lumicks/pylake/force_calibration/detail/power_models.py
tobiasjj: Yep, it's a class. Left over from before the commit where I corrected the API falsly referencing `PowerSpectrum` as a function. Fixed.

docs/tutorial/force_calibration.rst
tobiasjj: 👍 : Yep, found also another additional one. Fixed!

lumicks/pylake/fdcurve.py
tobiasjj: Not necessary, as sphinx automatically adds the braces, if it is a function. But indeed nicer especially in the raw docstrings. -> fixed

PR #390, opened 2022-09-27
kymotrackgroup: ensure that api functions when passed iostrings

JoepVanlier:
#### Why this PR?
We want to ensure that if the API calls for saving/loading tracked lines are passed a `IOString`, the whole thing functions.

I have also added an inspection check whether the whole export function is parametrized (so that if we add arguments in the future, we don't forget to expand the tests).


mwilkowski (approved):
Looks good to me! Thanks for this

Issue #358, opened 2022-08-18
`KymoLineGroup._merge_lines()` silently assumes that the parameters are ordered relative to time


mikhas:
This is the signature of `KymoLineGroup`'s `_merge_lines()`:

~~~python
def _merge_lines(self, starting_line: KymoLine, starting_node: int, ending_line: KymoLine, ending_node: int)
~~~

For this method to produce the _expected_ result, it is assumed that

~~~python
starting_line.seconds[starting_node] < ending_line.seconds[ending_node]
~~~

Failing this, the method will still "work" as the assumption isn't checked, turning it into a rather versatile footgun.

PR #392, opened 2022-09-30
kymotrack: ensure proper order on merge

rpauszek:
**Why this PR?**
Addresses issue #358

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -499,15 +499,42 @@ class KymoTrackGroup:
         self._src.remove(ending_track)
 
     def _merge_tracks(self, starting_track, starting_node, ending_track, ending_node):
+        """Connect two tracks from any given nodes, removing the points in between.
+
+        Note: Any specialized refinement (eg. gaussian refinement) will be lost when
+        tracks are merged.
+
+        Parameters
+        ----------
+        starting_track: KymoTrack
+            First track to connect
+        starting_node : int
+            Index of the node in the track to connect from
+        ending_track: KymoTrack
+            Second track to connect
+        ending_node: int
+            Index of the node in the track to connect to
+        """
         if starting_track not in self._src or ending_track not in self._src:
             raise RuntimeError("Both tracks need to be part of this group to be merged")
 
-        starting_node = int(starting_node) + 1
+        starting_node = int(starting_node)
         ending_node = int(ending_node)
 
+        start_time_idx = starting_track.time_idx[starting_node]
+        end_time_idx = ending_track.time_idx[ending_node]
+        assert start_time_idx != end_time_idx, "Cannot connect two points with the same time index."
+
+        # ensure that tracks are properly ordered so resulting merge track
+        # has coordinates sorted in ascending time indices
+        if start_time_idx > end_time_idx:
+            starting_track, ending_track = ending_track, starting_track
+            starting_node, ending_node = ending_node, starting_node
+
+        # up to and *including* starting_node
         first_half = starting_track._with_coordinates(
-            starting_track.time_idx[:starting_node],
-            starting_track.coordinate_idx[:starting_node],
+            starting_track.time_idx[: starting_node + 1],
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -491,12 +491,17 @@ def _merge_tracks(self, starting_track, starting_node, ending_track, ending_node
         if starting_track not in self._src or ending_track not in self._src:
             raise RuntimeError("Both tracks need to be part of this group to be merged")
 
-        starting_node = int(starting_node) + 1
+        starting_node = int(starting_node)
         ending_node = int(ending_node)
 
+        assert (
+            starting_track.time_idx[starting_node] < ending_track.time_idx[ending_node]
+        ), "The starting node must occur before the ending node."
```
mikhas, line 499: It's good that the assumption was made explicit, but that means I still have to manually swap the starting and ending node. Why can't you just swap it here, given that you already check for the condition?

mikhas (commented):
I get a nice exception now, but looks like I will have to maintain the workaround.

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: yeah, not sure why I didn't do it that way. probably because the original work was for the widget so my head jumped straight to handling it there for some reason. I'll move the functionality to here

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -497,12 +497,21 @@ def _merge_tracks(self, starting_track, starting_node, ending_track, ending_node
         if starting_track not in self._src or ending_track not in self._src:
             raise RuntimeError("Both tracks need to be part of this group to be merged")
 
-        starting_node = int(starting_node) + 1
+        starting_node = int(starting_node)
         ending_node = int(ending_node)
 
+        start_time_idx = starting_track.time_idx[starting_node]
+        end_time_idx = ending_track.time_idx[ending_node]
+        assert start_time_idx != end_time_idx, "Cannot connect two points with the same time index."
```
mikhas, line 505: Good catch!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -497,12 +497,21 @@ def _merge_tracks(self, starting_track, starting_node, ending_track, ending_node
...
+
+        if start_time_idx > end_time_idx:
```
mikhas, line 507: I would add a code comment here explaining very briefly why ordering is important.

mikhas (approved):
Good stuff. Tested it and can confirm that it works. Thanks!

rpauszek:
> Do you need to mention the bugfix in the changelog as well?

I don't think so because it's private API and we don't usually add that to changelog. Also, I guess in a way it's not technically a bug because the only place it's publicly used is in the widget and the sorting was performed there previously

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -497,12 +497,23 @@ def _merge_tracks(self, starting_track, starting_node, ending_track, ending_node
...
+        end_time_idx = ending_track.time_idx[ending_node]
+        assert start_time_idx != end_time_idx, "Cannot connect two points with the same time index."
+
+        # ensure that tracks are properly ordered so resulting merge track
+        # has coordinates sorted in ascending time indices
+        if start_time_idx > end_time_idx:
+            starting_track, ending_track = ending_track, starting_track
+            starting_node, ending_node = ending_node, starting_node
+
+        # up to and *including* starting_node
```
JoepVanlier, line 513: 👍

JoepVanlier (commented):
One thing that occurred to me is that this makes one lose any specialized localization for this `KymoTrack` since you end up overwriting it with just flat coordinates when you use `_with_coordinates`. This should probably be mentioned in the docstring so we don't cut our fingers on this.

For all our current use cases that is fine, but maybe good to add something on the backlog to think about whether we can abstract those localization models such that they can be sliced as though they were just arrays so that we can do stitching and whatnot without losing the extra properties as well.

PR #393, opened 2022-10-04
kymotracker: bugfixes

JoepVanlier:
**Why this PR?**
There were bugs. They are fixed now.

The bugs in question were:
- If the ROI selection resulted in no tracks remaining, `track_greedy` would crash. Now it returns an empty group.
- If no points are above the threshold `track_greedy` would return an empty list. Now it returns a group (more consistent with its regular output).
- `refine_tracks_*` used to raise with a numpy error if there were no tracks in the group. Now it just returns an empty group.
- `fit_binding_times(1)` and `plot_binding_histogram()` used to raise with a numpy error if there were no tracks, now you get a clean exception.


lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
```diff
@@ -61,6 +62,13 @@ def test_kymotracker_greedy_algorithm_integration_tests(kymo_integration_test_da
     np.testing.assert_allclose(tracks[0].time_idx, np.arange(15, 25))
 
 
+def test_greedy_algorithm_empty_result(kymo_integration_test_data):
+    test_data = kymo_integration_test_data
+    tracks = track_greedy(test_data, "red", track_width=300, pixel_threshold=400)
```
mikhas, line 76: It is not immediately clear that you will get an empty result due to absurd parameters: they aren't absurd enough.
Perhaps add a code comment?

mikhas (commented):
I am not sure I trust the 3rd commit but the other commits in this patch set look fine to me.

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
JoepVanlier: I've added a small comment :)

JoepVanlier:
> I am not sure I trust the 3rd commit but the other commits in this patch set look fine to me.

I changed the third commit slightly to make it less of a change.

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
```diff
@@ -39,6 +39,15 @@ def test_kymotracker_subset_test_greedy(kymo_integration_test_data):
     np.testing.assert_allclose(tracks[0].sample_from_image(1), [40] * np.ones(10))
 
 
+def test_kymotracker_regression_test_subset_comes_up_empty(kymo_integration_test_data):
+    """Test whether we gracefully handle a kymotracker where the rect results in no lines. Before
+    this threw a cryptic error message."""
```
tobiasjj, line 44: Do we need the comment about what happened before?

tobiasjj (approved):
LGTM.

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
JoepVanlier: I added it because it's a regression and if it happens again, it might provide some information on what happened last time this came up. That said, I don't feel particularly attached to this part of the comment, so I can remove it if you think it's not helpful.

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
tobiasjj: I'm fine with it, doesn't hurt to leave it in. Was not sure if you included it purposefully.

lumicks/pylake/kymotracker/tests/test_greedy_algorithm.py
JoepVanlier: I changed it to:
```python
"""Test whether we gracefully handle the case where the ROI results in no lines."""
```

mikhas (approved):
Tested, works!

JoepVanlier:
> Tested, works!

Thank you.

PR #394, opened 2022-10-06
Update docs confocal

aafkevandenberg:
I updated image.rst. Some of the changes are:

- Used the single-frame scan and multiframe scan from the Pylake data wishlist. I tuned all the parameters to fit these two data files.
- Make sure everything runs
- Restructured the text and change headings for a better structure
- When exporting video, I changed` end_frame` -> `stop_frame`
- `scan.red_photons` -> `scan.red_photon_count`

### DIFF
--- a/docs/tutorial/images.rst
+++ b/docs/tutorial/images.rst
@@ -12,24 +12,105 @@ inside of it, run::
 
     import lumicks.pylake as lk
 
-    file = lk.File("example.h5")
+    file = lk.File("Scan_WTCas9_2Markers_unspecific.h5")
     list(file.scans)  # e.g. shows: "['reference', 'bleach', 'imaging']"
 
-Once again, :attr:`.scans <lumicks.pylake.File.scans>` is a regular Python dictionary so we can
-easily iterate over it::
+:attr:`.scans <lumicks.pylake.File.scans>` is a regular Python dictionary so we can iterate over it::
 
     # Plot all scans in a file
     for name, scan in file.scans.items():
         scan.plot(channel="rgb")
         plt.savefig(name)
 
-Or just pick a single one::
+Or just pick a single one by providing the name of the scan as ``scan=file.scans["name"]``::
 
-    scan = file.scans["name"]
+    scan = file.scans["41"]
     scan.plot("red")
 
-Scan data and details
----------------------
+Plotting and Exporting
+----------------------
+
+As shown above, there are convenience functions for plotting either the full RGB image or a single
+color channel::
+
+    scan.plot(channel="rgb")
+
+The `channel` argument accepts the strings `“red”`, `“green”`, `“blue”`, or `“rgb”`. Multi-frame scans are also supported::
+
+    multiframe_file = lk.File("Scan_WTCas9_2Markers_binding on and off_stack.h5")
+    multiframe_scan = multiframe_file.scans["46"]
+
+    print(multiframe_scan.num_frames)
+    print(multiframe_scan.get_image("blue").shape)  # (self.num_frames, h, w) -> single color channel
+    print(multiframe_scan.get_image("rgb").shape)  # (self.num_frames, h, w, 3) -> three color channels
+
+    # plot frame at index 3 (first frame is index 0)
+    # defaults to the first frame if no argument is given
+    multiframe_scan.plot("green", frame=3)
+
+Sometimes a few bright pixels can dominate the colormap of a scan.
+When this is the case, it may be beneficial to manually set the color limits for each of the channels.
+This can be accomplished by providing a :class:`~lumicks.pylake.ColorAdjustment` to plotting or export functions::
+
+    scan.plot(channel="rgb", adjustment=lk.ColorAdjustment([0, 0, 0], [4, 4, 4]))
+
+Here the first array gives the minimal values for the color scale of red, green and blue respectively (here `[0, 0, 0]`) and the second array gives the maximum values for the color scales.
+The color scale is linear by default, but `Gamma correction <https://en.wikipedia.org/wiki/Gamma_correction>`_ can be applied in addition to the bounds by supplying an extra argument named `gamma`.
+For example, a gamma adjustment of `0.1` to the green channel can be applied as follows::
+
+    scan.plot(channel="rgb", adjustment=lk.ColorAdjustment([0, 0, 0], [4, 4, 4], gamma=[1, 0.1, 1]))
+
+The limits can also be specified in percentiles::
+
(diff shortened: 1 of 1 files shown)


docs/tutorial/images.rst
```diff
@@ -12,73 +12,88 @@ inside of it, run::
...
+As shown above, there are convenience functions for plotting either the full RGB image or a single
+color channel::
 
-    rgb = scan.get_image("rgb")  # matrix with `shape == (h, w, 3)`
-    blue = scan.get_image("blue")  # single color so `shape == (h, w)`
+    scan.plot(channel="rgb")
 
-    # Plot manually
-    plt.imshow(rgb)
+The `channel` argument accepts the strings “red”, “green”, “blue”, or “rgb”. Multi-frame scans are also supported::
```
JoepVanlier, line 38:
```suggestion
The `channel` argument accepts the strings `“red”`, `“green”`, `“blue”`, or `“rgb”`. Multi-frame scans are also supported::
```
Very minor, but might be nice to emphasize that they're literal code strings by making them a code block.

docs/tutorial/images.rst
```diff
@@ -12,73 +12,88 @@ inside of it, run::
...
+Sometimes a few bright pixels can dominate the colormap of a scan.
+When this is the case, it may be beneficial to manually set the color limits for each of the channels.
+This can be accomplished by providing a :class:`~lumicks.pylake.ColorAdjustment` to plotting or export functions::
 
-    sliced_scan = multiframe_scan[5:10]
+    scan.plot(channel="rgb", adjustment=lk.ColorAdjustment([0, 0, 0], [4, 4, 4]))
 
-This will return a new :class:`~lumicks.pylake.scan.Scan` containing data equivalent to::
+Gamma adjustments can be applied in addition to the bounds by supplying an extra argument named `gamma`.
+For example, a gamma adjustment of `0.1` to the green channel can be applied as follows::
```
JoepVanlier, line 58: Could be nice to add a few sample curves of what curve such a gamma adjustment has? Or alternatively, maybe even just make the `Gamma adjustments` text a link to wikipedia or another source so people can read about gamma adjustments there.

docs/tutorial/images.rst
```diff
@@ -12,73 +12,88 @@ inside of it, run::
...
 
-    multiframe_scan.get_image("rgb")[5:10, :, :, :]
+    scan.plot(channel="rgb", adjustment=lk.ColorAdjustment([0, 0, 0], [4, 4, 4], gamma=[1, 0.1, 1]))
```
JoepVanlier, line 60: Does this adjustment make the plot look better?

docs/tutorial/images.rst
```diff
@@ -12,73 +12,88 @@ inside of it, run::
...
+
+For other video formats such as `.mp4` or `.avi`, ffmpeg must be installed. See
+:ref:`installation instructions <ffmpeg_installation>` for more information on this.
 
 The images contain pixel data where each pixel represents summed photon counts.
-For an even lower-level look at data, the raw photon count samples can be accessed::
+The photon count per pixel can be accessed as follows::
 
-    photons = scan.red_photons
+    photons = scan.red_photon_count
```
JoepVanlier, line 86: Nice catch! 👍

docs/tutorial/images.rst
```diff
@@ -93,69 +108,58 @@ There are also several properties available for convenient access to the scan me
...
+    multiframe_scan[multiframe_file.force1x.start:multiframe_file.force1x.stop]
 
-We can downsample channel data according to the frames in a scan. We can use :func:`~lumicks.pylake.scan.Scan.frame_timestamp_ranges()` for this::
+Correlating a multiframe scan with data channels
+-------------------------------------------------
+The frames of a multiframe scan can be correlated to the force or other data channels.
+Downsample channel data according to the frames in a scan using :func:`~lumicks.pylake.scan.Scan.frame_timestamp_ranges()`::
 
-    frame_timestamp_ranges = scan.frame_timestamp_ranges()
+    frame_timestamp_ranges = multiframe_scan.frame_timestamp_ranges()
```
JoepVanlier, line 155: It might be nice to mention the options here (`include_dead_time` being `True` or `False`).

docs/tutorial/images.rst
```diff
@@ -93,69 +108,58 @@ There are also several properties available for convenient access to the scan me
...
 
 This returns a list of start and stop timestamps that can be passed directly to :func:`~lumicks.pylake.channel.Slice.downsampled_to`, which will then return a :class:`~lumicks.pylake.channel.Slice` with a datapoint per frame::
```
JoepVanlier, line 157:
```suggestion
This returns a list of start and stop timestamps that can be passed directly to :func:`~lumicks.pylake.channel.Slice.downsampled_over`, which will then return a :class:`~lumicks.pylake.channel.Slice` with a datapoint per frame::
```
I assume this should be `downsampled_over`?

JoepVanlier (changes requested):
Nice! I think it indeed flows better! :)

A few suggestions and questions from my side but looks good!

docs/tutorial/images.rst
aafkevandenberg: The min and max values do make the plot look better. All the colors are clearly visible for this range.
The value of gamma does not improve the plot necessarily, I just picked a value that is visibly different from setting gamma=1

docs/tutorial/images.rst
aafkevandenberg: Good catch, I overlooked this error in the current docs.

docs/tutorial/images.rst
aafkevandenberg: In the original example, gamma was > 1, but I believe gamma < 1 is a more common correction in image analysis. This way both low and high intensity features can both be visible.

docs/tutorial/images.rst
aafkevandenberg: How about https://en.wikipedia.org/wiki/Gamma_correction ?

docs/tutorial/images.rst
aafkevandenberg: Good suggestion, included it!

docs/tutorial/images.rst
JoepVanlier: Yeah, that'd work :)

docs/tutorial/images.rst
JoepVanlier: Makes sense, yes.

docs/tutorial/images.rst
aafkevandenberg: I added the link, and also a bit more explanation on the color scale (that it is linear by default, but a gamma correction can be applied)

JoepVanlier (approved):
Looks good to me!

:shipit:

PR #396, opened 2022-10-07
KymoTrackGroup: Add `+` operator

JoepVanlier:
**Why this PR?**
I did this PR for a few reasons (found mostly while doing testing on the new msd methods):

- It's very convenient to be able to add `KymoTrackGroups` and to be able to add lines to one and getting a shallow copy of the group. Similarly, being able to add groups together is quite convenient when creating test data (which will come in handy when writing the theory section).
- There were some weird cornercases in `_extend`. Extending with an empty list produced an error because it doesn't have a Kymo and `_validate_single_source` fails.
- Analogously, extending with an incorrect datatype didn't manage to make it to the correct `TypeError` for the same reason.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -459,6 +459,11 @@ class KymoTrackGroup:
     def __copy__(self):
         return KymoTrackGroup(copy(self._src))
 
+    def __add__(self, other):
+        new_group = copy(self)
+        new_group.extend(other)
+        return new_group
+
     @property
     def _kymo(self):
         try:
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -519,21 +524,25 @@ def __len__(self):
...
+        if not other:
+            return self
+
+        if not (isinstance(other, KymoTrack) or isinstance(other, self.__class__)):
+            raise TypeError(
+                f"You can only extend a {self.__class__.__name__} with a {self.__class__.__name__} "
+                f"or {KymoTrack.__name__}"
+            )
 
+        other = [other] if isinstance(other, KymoTrack) else other
```
rpauszek, line 536: looking at this again, I wonder if it's just better to cast to `KymoTrackGroup` rather than `list` up front? Then below after the assertions we can completely get rid of the whole `if` block.
```suggestion
        other = KymoTrackGroup([other]) if isinstance(other, KymoTrack) else other
```

rpauszek (approved):
Good stuff. And good catch with the extending by empty group. I caught the other way around but missed this corner case, thanks!

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Good point. Fixing.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -459,6 +459,11 @@ def __bool__(self):
     def __copy__(self):
         return KymoTrackGroup(copy(self._src))
 
+    def __add__(self, other):
+        new_group = copy(self)
+        new_group.extend(other)
```
tobiasjj, line 464: Probably, I'm missing something, but why do you make a shallow copy of the `KymoTrackGroup`? As I understand, the method `extend()` appends kymos to the list `self._src`. So after adding there are two different `KymoTrackGroup`s pointing to the same list of kymos, with the original `KymoTrackGroup` "silently" beeing changed. I would expect when performing something like `ktg3 = ktg1 + ktg2` that `ktg1` would stay unchanged.

tobiasjj: Yes, mystery solved. I originally stumbled because I missed the method `KymoTrackGroup.__copy__()`, which returns a copy of the list.

PR #397, opened 2022-10-07
diffusion: disallow estimating diffusion after downsampling

JoepVanlier:
**Why this PR?**
We currently offer downsampling kymographs, but one can still compute diffusion constants from them. However, the assumptions that go into the method are then not met.

This PRs adds a  boolean property to a `Kymo` which indicates whether the Kymo is the result from downsampling. This boolean is checked before doing diffusion analysis.

*What to look out for?*
1. I chose not to disallow computing MSDs (since technically, they are just defined as they are; the assumptions come in once people start inferring things on them). I am on the fence whether we should issue a warning though.

2. I did not disallow dwelltime analyses because I think these are still valid. The downstream methods there just make use of the actually achieved minimum times, and therefore should be fine.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -389,6 +389,13 @@ class KymoTrack:
         if method not in ("gls", "ols"):
             raise ValueError('Invalid method selected. Method must be "gls" or "ols"')
 
+        if not self._kymo.contiguous:
+            raise NotImplementedError(
+                "Estimating diffusion constants from data which has been integrated over disjoint "
+                "sections of time is not supported. To estimate diffusion constants, do not "
+                "downsample the kymograph temporally prior to tracking."
+            )
+
         frame_idx, positions = np.array(self.time_idx, dtype=int), np.array(self.position)
         max_lag = (
             max_lag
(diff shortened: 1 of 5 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -389,6 +389,13 @@ def estimate_diffusion(self, method, max_lag=None):
         if method not in ("gls", "ols"):
             raise ValueError('Invalid method selected. Method must be "gls" or "ols"')
 
+        if not self._kymo.contiguous:
+            raise NotImplementedError(
+                "Estimating diffusion constants from data with discontiguous pixels not supported. "
+                "To estimate diffusion constants, do not downsample the kymograph temporally prior "
+                "to tracking."
+            )
```
rpauszek, line 397: I wonder if we should include the little bit from the `contiguous` docstring? or somehow mention that the actual problem is integrating over disjoint time. Maybe ...
```suggestion
        if not self._kymo.contiguous:
            raise NotImplementedError(
                "Estimating diffusion constants from data which has been integrated over disjoint sections "
                "of time is not supported. To estimate diffusion constants, do not downsample the kymograph "
                " temporally prior to tracking."
            )
```

rpauszek (approved):
Nice. Good point that it's only the diffusion analysis that's affected by this. This way we can keep the majority of downstream analysis for users that really need to downsample but might not need diffusion

Just left a small suggestion for the `NotImplementedError` message (if you go with it, it'll have to be changed in the test too)

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Makes sense. It's a bit more clear. I'll go with your suggestion.

PR #398, opened 2022-10-10
kymotracker: ensure threshold default is never 0

JoepVanlier:
**Why this PR?**
With an empty image, the threshold could be set to zero. It should at a minimum be `1`.

I noticed that the properties that depend on the `Kymo` were untested, so I added tests for these as well.


mikhas (approved):
Good testcase!

PR #399, opened 2022-10-10
kymotracker widget: extended descriptions

JoepVanlier:
**Why this PR?**
We want to provide some more insightful tooltips and for this we need some description fields.

### DIFF
--- a/lumicks/pylake/nb_widgets/kymotracker_widgets.py
+++ b/lumicks/pylake/nb_widgets/kymotracker_widgets.py
@@ -713,6 +713,7 @@ class KymotrackerParameter:
     lower_bound: Number
     upper_bound: Number
     ui_visible: bool
+    extended_description: str
 
     def __post_init__(self):
         if self.ui_visible and (self.lower_bound is None or self.upper_bound is None):
@@ -732,57 +733,102 @@ def _get_default_parameters(kymo, channel):
 
     return {
         "pixel_threshold": KymotrackerParameter(
-            "Threshold",
-            "Set the pixel threshold.",
+            "Minimum intensity",
+            "Set the pixel intensity threshold.",
             "int",
             max(1, int(np.percentile(data.flatten(), 98))),
             *(1, max(2, np.max(data))),
             True,
-        ),
-        "track_width": KymotrackerParameter(
-            "Track width",
-            "Estimated spot width.",
-            "float",
-            4 * position_scale,
-            *(0.0, 15.0 * position_scale),
-            True,
-        ),
-        "window": KymotrackerParameter(
-            "Window", "How many frames can a track disappear.", "int", 4, *(1, 15), True
+            r"Minimum intensity defines the minimum signal required in a single pixel for it to be "
+            r"considered part of a track. This parameter should be chosen slightly above the "
+            r"background photon count. Higher values reject more noise, but parts of the track may "
+            r"be missed.",
         ),
         "sigma": KymotrackerParameter(
-            "Sigma",
+            "Positional search range",
             "How much does the track fluctuate?",
             "float",
             2 * position_scale,
             *(1.0 * position_scale, 5.0 * position_scale),
             True,
+            r"Search range defines how much the position of a particle is allowed to fluctuate "
+            r"from one time point to the next while still being considered part of the same "
+            r"track. Larger values will result in a wider range in which points are added to a "
+            r"track.",
         ),
-        "vel": KymotrackerParameter(
-            "Velocity",
-            "How fast does the particle move?",
-            "float",
-            0.0,
-            *(-5.0 * vel_calibration, 5.0 * vel_calibration),
+        "window": KymotrackerParameter(
+            "Maximum gap",
+            "How many frames can a track disappear.",
+            "int",
+            4,
+            *(1, 15),
             True,
-        ),
-        "diffusion": KymotrackerParameter(
-            "Diffusion", "Expected diffusion constant.", "float", 0.0, *(None, None), False
-        ),
-        "sigma_cutoff": KymotrackerParameter(
-            "Sigma cutoff",
-            "Number of standard deviations from the expected trajectory a particle no longer belongs to a trace.",
-            "float",
-            2.0,
-            *(None, None),
-            False,
+            r"Maximum gap defines how many scan lines without detected points can occur between "
+            r"two points and still be connected within a single track. This can occur, for "
+            r"instance, due to fluorophore blinking or the signal falling slightly below the "
+            r"intensity threshold. This value should be chosen such that small gaps in a track "
+            r"can be overcome and tracked as one, but not so large that separate tracks are strung "
+            r"together.",
         ),
         "min_length": KymotrackerParameter(
-            "Min length",
+            "Minimum length",
             "Minimum number of frames a spot has to be detected in to be considered.",
             "int",
             3,
             *(1, 10),
             True,
+            r"Minimum length defines the minimum number of points a tracked line must contain for "
+            r"it to be considered valid. Reducing this parameter can be effective in reducing "
+            r"tracking noise. Note that this length refers to the number of detected points, not "
+            r"length in time!",
+        ),
+        "track_width": KymotrackerParameter(
+            "Expected spot size",
+            "How big a particle needs to appear to be tracked as a single molecule.",
+            "float",
+            4 * position_scale,
+            *(0.0, 15.0 * position_scale),
+            True,
+            r"Expected spot size defines how big a bound particle needs to appear as on the "
+            r"kymograph for it to be tracked as a single molecule. This parameter should be set to "
+            r"roughly the width of the point spread function. Setting it larger rejects more "
+            r"noise, but at the cost of potentially merging tracks that are close together.",
+        ),
+        "vel": KymotrackerParameter(
+            "Expected velocity",
+            "How fast does the particle move?",
+            "float",
+            0.0,
+            *(-5.0 * vel_calibration, 5.0 * vel_calibration),
+            True,
+            r"Expected velocity defines how fast and in which direction particles are expected "
+            r"to move along the DNA on average. When tracking, the algorithm searches for points "
+            r"in future scan lines to connect. Points within a certain distance from the expected "
+            r"future position are connected.",
+        ),
+        "diffusion": KymotrackerParameter(
+            "Diffusion",
+            "Expected diffusion constant.",
+            "float",
+            0.0,
+            *(None, None),
+            False,
+            r"When tracking, the algorithm searches for points in future frames to connect. Points "
+            r"within a certain distance from the expected future position are connected. The "
+            r"diffusion parameter determines how quickly this distance grows over time.",
+        ),
+        "sigma_cutoff": KymotrackerParameter(
+            "Sigma cutoff",
+            "Number of standard deviations from the expected trajectory a particle no longer "
+            "belongs to a trace.",
+            "float",
+            2.0,
+            *(None, None),
+            False,
+            r"Search range (sigma) controls how much the location can fluctuate from one time "
+            r"point to the next while still being considered part of the same track. Sigma cutoff "
+            r"controls the actual threshold applied to the expected positional variability. Larger "
+            r"values for sigma or sigma_cutoff will be more permissive when it comes to connecting "
+            r"detected peaks into a track.",
         ),
     }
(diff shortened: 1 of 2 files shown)


mikhas:
Can you rebase on `main` or at least change target branch? First commit does not belong here.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
 
     return {
         "pixel_threshold": KymotrackerParameter(
-            "Threshold",
-            "Set the pixel threshold.",
+            "Minimum intensity",
+            "Set the intensity threshold.",
```
rpauszek, line 737: I think here it's still ok to keep the reference to "pixel", it does make it clear what is actually being checked
```suggestion
            "Minimum intensity",
            "Set the pixel intensity threshold.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
             "int",
             max(1, int(np.percentile(data.flatten(), 98))),
             *(1, max(2, np.max(data))),
             True,
+            r"Minimum pixel intensity defines how many photons are required in a single pixel for "
+            r"something to be considered part of a track. This parameter should be chosen slightly "
+            r"above the baseline photon count. Higher values reject more noise, but actual peaks "
+            r"that should be part of the track may be missed.",
```
rpauszek, line 745: One thing that has come up in discussions with the HW team is that the signal we observe technically isn't the true number of photons. This is because the detector efficiency also plays a role, so for complete correctness we should refer to these as `"counts"`. I think `"photon counts"` is still accurate also. 

Also in general, I don't think we should expect for users to understand that underlying all this there is a peak finding algorithm, so I'd rather avoid that term in these descriptions.

So maybe just a minor change:
```suggestion
            r"Minimum pixel intensity defines the minimum signal required in a single pixel for "
            r"it to be considered part of a track. This parameter should be chosen slightly "
            r"above the baseline photon count. Higher values reject more noise, but actual points "
            r"that should be part of the track may be missed.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
             "float",
-            4 * position_scale,
-            *(0.0, 15.0 * position_scale),
+            2 * position_scale,
+            *(1.0 * position_scale, 5.0 * position_scale),
             True,
+            "Maximum search range defines how much the position of a particle is allowed to "
+            "fluctuate from one time point to the next while still being considered part of the "
+            "same track. Larger values will result in a wider range in which points are added to "
+            "a track.",
```
rpauszek, line 757: I think "maximum" is implied by the term "range"
```suggestion
            "Search range defines how much the position of a particle is allowed to "
            "fluctuate from one time point to the next while still being considered part of the "
            "same track. Larger values will result in a wider range in which points are added to "
            "a track.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
-            "Window", "How many frames can a track disappear.", "int", 4, *(1, 15), True
+            "Maximum gap duration",
+            "How many frames can a track disappear.",
+            "int",
+            4,
+            *(1, 15),
+            True,
+            r"Maximum gap duration defines how big the gap between two points can be for them to "
+            r"connect. It should be chosen such that small gaps in a track can be overcome and "
+            r"tracked as one, but not so large that separate tracks are strung together.",
```
rpauszek, line 768:
```suggestion
            r"Maximum gap duration defines how many scan lines without detected points can occur "
            r"between two points and still be connected within a single track.  This can occur, for instance, "
            r"due to fluorophore blinking or the signal falling slightly below the intensity threshold. "
            r"This value should be chosen such that small gaps in a track can be overcome and "
            r"tracked as one, but not so large that separate tracks are strung together.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
+        "min_length": KymotrackerParameter(
+            "Minimum length",
+            "Minimum number of frames a spot has to be detected in to be considered.",
+            "int",
+            3,
+            *(1, 10),
+            True,
+            "Minimum length defines the minimum number of points a tracked line must have for it "
+            "to show up. Reducing this parameter is very effective in reducing noise. Note that "
+            "this length refers to the number of detections, not length in time!",
```
rpauszek, line 779:
```suggestion
            "Minimum length defines the minimum number of points a tracked line must contain for it "
            "to be considered valid. Reducing this parameter can be effective in reducing tracking noise. " 
            "Note that this length refers to the number of detected points, not length in time!",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
             "float",
-            2 * position_scale,
-            *(1.0 * position_scale, 5.0 * position_scale),
+            4 * position_scale,
+            *(0.0, 15.0 * position_scale),
             True,
+            r"Minimum particle width defines how big a particle needs to appear as on the "
+            r"kymograph for it to be tracked as a single molecule. This parameter should be set to "
+            r"roughly the width of the point spread function. Making it larger rejects more noise, "
+            r"but at the cost of potentially merging tracks that are close together.",
```
rpauszek, line 791:
```suggestion
            r"Expected spot size defines how big a bound particle needs to appear on the "
            r"kymograph for it to be tracked as a single molecule. This parameter should be set to "
            r"roughly the width of the point spread function. Setting it larger rejects more noise, "
            r"but at the cost of potentially merging tracks that are close together.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
+            "Expected velocity",
             "How fast does the particle move?",
             "float",
             0.0,
             *(-5.0 * vel_calibration, 5.0 * vel_calibration),
             True,
+            "Minimum particle velocity defines how fast particles can move on along the DNA on "
+            "average. When tracking, the algorithm searches for points in future scan lines to "
+            "connect. Points within a certain distance from the expected future position are "
+            "connected.",
```
rpauszek, line 803:
```suggestion
            "Expected velocity defines how fast and in which direction particles are expected "
            "to move on along the DNA on average. When tracking, the algorithm searches for "
            "points in future scan lines to connect. Points within a certain distance from the "
            "expected future position are connected.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
-            "Diffusion", "Expected diffusion constant.", "float", 0.0, *(None, None), False
+            "Diffusion",
+            "Expected diffusion constant.",
+            "float",
+            0.0,
+            *(None, None),
+            False,
+            "When tracking, the algorithm searches for peaks in future frames to connect. Peaks "
+            "within a certain distance from the expected future position are connected. The "
+            "diffusion parameter determines how quickly this distance grows over time.",
```
rpauszek, line 814: Not super important since this one isn't displayed, but just for consistency:

```suggestion
            "When tracking, the algorithm searches for points in future frames to connect. Points "
            "within a certain distance from the expected future position are connected. The "
            "diffusion parameter determines how quickly this distance grows over time.",
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,98 @@ def _get_default_parameters(kymo, channel):
...
-            "Minimum number of frames a spot has to be detected in to be considered.",
-            "int",
-            3,
-            *(1, 10),
-            True,
+            "Search range controls how much the location can fluctuate from one time point to the "
+            "next while still being considered part of the same track. Sigma cutoff controls the "
+            "actual threshold applied to the expected positional variability. Larger values for "
+            "sigma or sigma_cutoff will be more permissive when it comes to connecting detected "
+            "peaks into a track.",
```
rpauszek, line 828: Again, not shown currently, but considering at the moment we're still referring to this one with "sigma" but not the other, I think the description should make the connection between the 2 parameters that are related

```suggestion
            "Search range (sigma) controls how much the location can fluctuate from one time point to the "
            "next while still being considered part of the same track. Sigma cutoff controls the "
            "actual threshold applied to the expected positional variability. Larger values for "
            "sigma or sigma_cutoff will be more permissive when it comes to connecting detected "
            "peaks into a track.",
```

rpauszek (commented):
some minor phrasing suggestions

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -713,6 +713,7 @@ class KymotrackerParameter:
     lower_bound: Number
     upper_bound: Number
     ui_visible: bool
+    info: str
```
mikhas, line 716: I would prefer `extended_description` so the relationship between the two fields is made clear.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: 👍

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: Fixed!

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -732,57 +733,102 @@ def _get_default_parameters(kymo, channel):
...
-            *(1.0 * position_scale, 5.0 * position_scale),
+            4 * position_scale,
+            *(0.0, 15.0 * position_scale),
             True,
+            r"Expected spot size defines how big a bound particle needs to appear as on the "
+            r"kymograph for it to be tracked as a single molecule. This parameter should be set to "
+            r"roughly the width of the point spread function. Setting it larger rejects more "
+            r"noise, but at the cost of potentially merging tracks that are close together.",
         ),
         "vel": KymotrackerParameter(
```
mikhas, line 797: rename `vel` to `velocity`? it stands out as the only parameter key/name that is abbreviated.

mikhas (approved):

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: That would be a breaking change. Not one I want to piggyback on this PR.

PR #401, opened 2022-10-11
Common plot method call signature for `CorrelatedStack`, `Kymo`, `Scan` and `PointScan`

tobiasjj:
**Why?**
As a user I want a sensible docstring for the `plot()` method of `CorrelatedStack`, `Kymo`, `Scan` and `PointScan`. Additionally, I would like to have a sensible and similar call signature for all `plot()` implementations.

This PR aligns the different plot functions and involves the following changes:
  * A new module `plotting`, which provides common utility functions for plotting
  * `Scan`, `Kymo` and `PointScan`: Made argument `channel` optional.
  * `Scan`: Added argument `show_title`
  * `Kymo`: Added arguments `image_handle` and `show_title`.
  * `CorrelatedStack`: Reordered the keyword arguments and enforced all parameters after `axes` to be keyword arguments.
  * The method call to `CorrelatedStack.plot()` has a decorater, which checks if the method was called with the deprecated order of arguments.
  * Tests for the new `plotting` functions and deprecation of the argument order of `CorrelatedStack.plot()`.

Before merging, I would squash all commits into one: `plotting: common plot method call signature for ...`

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,13 @@
 * TIFFs exported from `Scan` and `Kymo` now contain metadata. The `DateTime` tag indicated the start/stop timestamp of each frame. The `ImageDescription` tag contains additional information about the confocal acquisition parameters.
 * Added covariance-based estimator (cve) option to `KymoTrack.estimate_diffusion()`. See [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more details.
 * Added the optional `min_length` parameter to `KymoTrackGroup.estimate_diffusion()` to discard tracks shorter than a specified length from the analysis.
+* Harmonized method `plot()` call signature for `Scan`, `Kymo`, `PointScan` and `CorrelatedStack`:
+  * `Scan`, `Kymo` and `PointScan`: Made argument `channel` optional
+  * `Scan` and `PointScan`: Added argument `show_title`
+  * `Kymo`: Added arguments `image_handle` and `show_title`.
+  * `CorrelatedStack`: See deprecation changelog entry.
+* `Kymo.plot()` now returns a handle of the plotted image
+* `PointScan.plot()` now returns a list of handles of the plotted lines
 
 #### Other changes
 
@@ -18,6 +25,8 @@
 #### Deprecations
 
 * Deprecated property `CorrelatedStack.src`.
+* Reordered the keyword arguments of the method `CorrelatedStack.plot()` and enforced all parameters after `channel` to be keyword arguments. For details see the [docstring](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.correlated_stack.CorrelatedStack.html#lumicks.pylake.correlated_stack.CorrelatedStack.plot).
+* Enforced the argument `axes` of the method `plot()` for `Scan`, `Kymo` and `PointScan` to be a keyword argument.
 
 #### Bugfixes
 
@@ -32,6 +41,7 @@
 * Fixed a bug where the `pixel_threshold` could be set to zero for an empty image. Now the minimum `pixel_threshold` is one.
 * Fixed a bug where single pixel detections in a `KymoTrackGroup` would contribute values with a dwell time of zero. These are now dropped, the correct minimally observable time is set appropriately and a warning is issued.
 * Fixed slicing of a `Kymo` where slicing from a time point inside the last line to the end (e.g. `kymo["5s":]`) resulted in a `Kymo` which returned errors upon trying to access its contents. 
+
 ### Other changes
 
 * Updated benchmark to not use deprecated functions and arguments. Prior to this change, running the benchmark would produce deprecation warnings.
--- a/lumicks/pylake/correlated_stack.py
+++ b/lumicks/pylake/correlated_stack.py
@@ -8,7 +8,58 @@ from deprecated.sphinx import deprecated
 from .adjustments import ColorAdjustment
 from .detail.image import make_image_title
 from .detail.imaging_mixins import FrameIndex, TiffExport, VideoExport
+from .detail.plotting import get_axes, show_image
 from .detail.widefield import TiffStack
+
+
+def _deprecate_plot_arguments(plot):
+    """Decorator to deprecate old arguments of the method `CorrelatedStack.plot()`"""
+    import functools
+    import matplotlib
+    import warnings
+
+    def is_old_order(args):
+        # old arguments: frame, channel, show_title, axes, adjustment
+        old_arg_types = [int, str, bool, matplotlib.axes.Axes, ColorAdjustment]
+        nones_allowed = [False, True, False, True, False]
+        old_order = 1 <= len(args) <= 5
+        for arg, old_type, none_allowed in zip(args, old_arg_types, nones_allowed):
+            old_order = old_order and (isinstance(arg, old_type) or none_allowed and arg is None)
+        return old_order
+
+    @functools.wraps(plot)
+    def wrapper(self, *args, **kwargs):
+        # The plot function might be called with up to 5 positional arguments with the old order
+        # We can gracefully convert the old ordered positional arguments into keyword arguments
+        if is_old_order(args):
+            keys = ["frame", "channel", "show_title", "axes", "adjustment"]
+            defaults = [0, "rgb", True, None, ColorAdjustment.nothing()]
+            warn_on_keys = []
+            for arg, key, default in zip(args, keys, defaults):
+                if key in kwargs:
+                    raise TypeError(
+                        f"`{self.__class__.__name__}.plot()` got multiple values for argument `{key}`"
+                    )
+                kwargs[key] = default if arg is None else arg
+                if key != "channel":
+                    warn_on_keys.append(key)
+            warn_keys = (
+                f"`{warn_on_keys[0]}`"
+                if len(warn_on_keys) == 1
+                else f"`{'`, `'.join(warn_on_keys[:-1])}` and `{warn_on_keys[-1]}`"
+            )
+            warnings.warn(
+                DeprecationWarning(
+                    f"The call signature of `plot()` has changed: Please, provide {warn_keys} as "
+                    f"keyword argument{'s' if len(warn_on_keys) > 1 else ''}."
+                ),
+                stacklevel=2,
+            )
+            return plot(self, **kwargs)
+        else:
+            return plot(self, *args, **kwargs)
+
+    return wrapper
 
 
 class CorrelatedStack(FrameIndex, TiffExport, VideoExport):
@@ -241,56 +292,53 @@ class CorrelatedStack(FrameIndex, TiffExport, VideoExport):
 
         return np.stack([frame.data[slc] for frame in self], axis=0).squeeze()
 
+    @_deprecate_plot_arguments
     def plot(
         self,
-        frame=0,
         channel="rgb",
-        show_title=True,
-        axes=None,
-        adjustment=ColorAdjustment.nothing(),
         *,
+        frame=0,
+        adjustment=ColorAdjustment.nothing(),
+        axes=None,
         image_handle=None,
+        show_title=True,
         **kwargs,
     ):
-        """Plot image from image stack
+        """Plot a frame from the image stack for the requested color channel(s)
 
         Parameters
         ----------
+        channel : {"red", "green", "blue", "rgb"}, optional
+            Color channel to plot.
         frame : int, optional
             Index of the frame to plot.
-        channel : 'rgb', 'red', 'green', 'blue', None; optional
-            Channel to plot for RGB images (None defaults to 'rgb')
-            Not used for grayscale images
-        show_title : bool, optional
-            Controls display of auto-generated plot title
-        axes : mpl.axes.Axes or None
+        adjustment : lk.ColorAdjustment
+            Color adjustments to apply to the output image.
+        axes : matplotlib.axes.Axes, optional
             If supplied, the axes instance in which to plot.
         image_handle : matplotlib.image.AxesImage or None
             Optional image handle which is used to update plots with new data rather than
             reconstruct them (better for performance).
+        show_title : bool, optional
+            Controls display of auto-generated plot title
         **kwargs
-            Forwarded to :func:`matplotlib.pyplot.imshow`.
-        """
-        import matplotlib.pyplot as plt
+            Forwarded to :func:`matplotlib.pyplot.imshow`. These arguments are ignored if
+            `image_handle` is provided.
 
-        if axes is None:
-            axes = plt.gca() if image_handle is None else image_handle.axes
-        elif image_handle is not None and id(axes) != id(image_handle.axes):
-            raise ValueError("Supplied image_handle with a different axes than the provided axes")
+        Returns
+        -------
+        matplotlib.image.AxesImage
+            The image handle representing the plotted image.
+        """
+        axes = get_axes(axes=axes, image_handle=image_handle)
+        image = self._get_frame(frame)._get_plot_data(channel, adjustment=adjustment)
 
         default_kwargs = dict(cmap="gray")
         kwargs = {**default_kwargs, **kwargs}
 
-        image = self._get_frame(frame)._get_plot_data(channel, adjustment=adjustment)
-
-        if not image_handle:
-            image_handle = axes.imshow(image, **kwargs)
-        else:
-            # Updating the image data in an existing plot is a lot faster than re-plotting with
-            # `imshow`.
-            image_handle.set_data(image)
-
-        adjustment._update_limits(image_handle, image, channel)
+        image_handle = show_image(
+            image, adjustment, channel, image_handle=image_handle, axes=axes, **kwargs
+        )
 
         if show_title:
             axes.set_title(make_image_title(self, frame))
@@ -302,7 +350,7 @@ class CorrelatedStack(FrameIndex, TiffExport, VideoExport):
 
         Parameters
         ----------
-        axes : mpl.axes.Axes or None
+        axes : matplotlib.axes.Axes or None
             If supplied, the axes instance in which to plot.
         **kwargs
             Forwarded to :func:`matplotlib.pyplot.plot`.
--- a/lumicks/pylake/detail/confocal.py
+++ b/lumicks/pylake/detail/confocal.py
@@ -1,18 +1,19 @@
 import json
-import numpy as np
-import cachetools
-from numpy import typing as npt
-from deprecated.sphinx import deprecated
+import warnings
 from dataclasses import dataclass
 from typing import List
-import warnings
 
-from .mixin import ExcitationLaserPower, PhotonCounts
-from .image import reconstruct_image_sum, reconstruct_image
-from .imaging_mixins import TiffExport
-from .utilities import could_sum_overflow
-from ..adjustments import ColorAdjustment
+import cachetools
+import numpy as np
+from deprecated.sphinx import deprecated
 from matplotlib.colors import LinearSegmentedColormap
+from numpy import typing as npt
+
+from ..adjustments import ColorAdjustment
+from .image import reconstruct_image, reconstruct_image_sum
+from .imaging_mixins import TiffExport
+from .mixin import ExcitationLaserPower, PhotonCounts
+from .utilities import could_sum_overflow
 
 linear_colormaps = {
     "red": LinearSegmentedColormap.from_list("red", colors=[(0, 0, 0), (1, 0, 0)]),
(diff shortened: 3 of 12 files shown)


lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
```diff
@@ -399,6 +399,44 @@ def test_correlated_stack_plotting(rgb_alignment_image_data):
...
+        [0],
+        [1, None],
+        [0, "blue"],
+        [0, None, True],
+        [0, None, True, "SUPPLY_AXES"],  # "SUPPLY_AXES" will be replaced with an mpl.axes object
+        [0, None, True, None, ColorAdjustment.nothing()]
+    ]:
+        frame = args[0]
+        channel = "blue" if args[-1] == "blue" else "rgb"
+        args[-1] = plt.subplots()[1] if args[-1] == "SUPPLY_AXES" else args[-1]
```
JoepVanlier, line 416: Why not just put the axes object in immediately?

lumicks/pylake/detail/plotting.py
```diff
@@ -0,0 +1,23 @@
...
+
+def get_axes(axes=None, image_handle=None):
+    if axes is None:
+        axes = plt.gca() if image_handle is None else image_handle.axes
+    elif image_handle and axes is not image_handle.axes:
+        raise ValueError("Supplied image_handle with a different axes than the provided axes")
+    return axes
+
+
+def show_image(image, image_handle=None, axes=None, **kwargs):
```
JoepVanlier, line 12: Nice dedup :)

JoepVanlier (changes requested):
Personally, I feel `axes` should probably be a `kwarg`, but other than that, I think it looks nice!

I think in terms of order, I would sort them by importance if we're in the sorting business anyhow now. I see `image_handle`, `show_title` and `axes` almost like an implementational detail. I think `adjustment` should precede those.

If I were to suggest an order, it'd probably be:
`frame` (if applicable), `adjustment`, `axes`, `handle`, `show_title`, `**kwargs`

What do you think?

changelog.md
```diff
@@ -15,6 +20,7 @@
 #### Deprecations
 
 * Deprecated property `CorrelatedStack.src`.
+* Reordered the keyword arguments of the method `CorrelatedStack.plot()` and enforced all parameters after `axes` to be keyword arguments. For details see the docstring.
```
rpauszek, line 23: would be good to link to the actual page
```suggestion
* Reordered the keyword arguments of the method `CorrelatedStack.plot()` and enforced all parameters after `axes` to be keyword arguments. For details see the [docstring](https://lumicks-pylake.readthedocs.io/en/stable/_api/lumicks.pylake.correlated_stack.CorrelatedStack.html#lumicks.pylake.correlated_stack.CorrelatedStack.plot).
```

lumicks/pylake/correlated_stack.py
```diff
@@ -241,54 +284,46 @@ def get_image(self, channel="rgb"):
...
+        show_title=True,
         **kwargs,
     ):
-        """Plot image from image stack
+        """Plot a frame from the image stack for the requested color channel(s)
 
         Parameters
         ----------
+        channel : {"red", "green", "blue", "rgb"}, optional
+            Color channel to plot.
```
rpauszek, line 304: 👍 nice that we're updating to the `{option1, option2, ...}` format as we're finding them

lumicks/pylake/correlated_stack.py
```diff
@@ -241,54 +284,46 @@ def get_image(self, channel="rgb"):
...
             Optional image handle which is used to update plots with new data rather than
             reconstruct them (better for performance).
+        adjustment : lk.ColorAdjustment
+            Color adjustments to apply to the output image.
+        show_title : bool, optional
+            Controls display of auto-generated plot title
         **kwargs
-            Forwarded to :func:`matplotlib.pyplot.imshow`.
+            Forwarded to :func:`matplotlib.pyplot.imshow`. Only used, if `image_handle` is not
+            provided.
```
rpauszek, line 318: Generally I prefer to avoid speaking in the negative sense for stuff like this, I think it's clearer

```suggestion
            Forwarded to :func:`matplotlib.pyplot.imshow`. These arguments are ignored
            if `image_handle` is provided.
```

lumicks/pylake/detail/plotting.py
```diff
@@ -0,0 +1,23 @@
...
+    if image_handle and axes and axes is not image_handle.axes:
+        raise ValueError("Supplied image_handle with a different axes than the provided axes")
+
+    if image_handle:
+        # Increase plotting speed by updating the image data in an already existing plot
+        image_handle.set_data(image)
+    else:
+        # Fall back to slower re-plotting with `imshow`.
+        image_handle = axes.imshow(image, **kwargs)
+    return image_handle
```
rpauszek, line 23: can we add in the `adjustment._update_limits(image_handle, image, channel)` line here? since it's common to all uses as far as I see. obviously would require passing it in, but is probably worthwhile

lumicks/pylake/point_scan.py
```diff
@@ -22,19 +23,20 @@ def _get_plot_data(self, channel):
         """Get photon count :class:`~lumicks.pylake.channel.Slice` for requested channel."""
         return getattr(self, f"{channel}_photon_count")
 
-    def _plot(self, channel, axes, **kwargs):
+    def plot(self, channel="rgb", axes=None, **kwargs):
```
rpauszek, line 26: for consistency, would be nice to add `show_title` here too

rpauszek (changes requested):
I really like this, especially the deprecation path for changing `CorrelatedStack.plot()`'s signature rather than breaking. It will be very nice to have a unified signature for all fluorescence plotting.

I agree with Joep that enforcing `axes` to be keyword only is the way to go; however, this would be a breaking change. I think instead we should set up a deprecation path similar to what you did for `CorrelatedStack`. In this case it's much simpler: if there is more than 1 positional argument, the final one is `axes`.

I wonder too if it'd be worthwhile to add a test for the call signature between these 4 plotting methods to ensure that everything stays synced. I think this is preferable to the other option of trying to have some mixin class handle it. I played around with it a bit, and you can use `inspect.signature` to get an `OrderedDict` of the arguments. You can then assert on the order of the positional arguments and also on any keyword arguments that should be common among all or a subset of these classes. Could avoid this headache in the future.

Since we're enforcing almost all args to be keyworded, I think it's also nice to think about the order we put them in so that the docs are nicer. More common ones up front, ones that are meant more for internal details at the end. I would suggest:
```
plot(
    self, 
    channel="rgb", 
    *, 
    adjustment,  # not needed for PointScan
    frame,  # not needed for PointScan, Kymo
    axes, 
    image_handle,  # not needed for PointScan
    show_title, 
    **kwargs)
```

lumicks/pylake/correlated_stack.py
```diff
@@ -241,54 +284,46 @@ def get_image(self, channel="rgb"):
 
         return np.stack([frame.data[slc] for frame in self], axis=0).squeeze()
 
+    @_deprecate_plot_arguments
     def plot(
         self,
-        frame=0,
         channel="rgb",
-        show_title=True,
         axes=None,
```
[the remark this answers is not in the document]
tobiasjj: 👍

lumicks/pylake/correlated_stack.py
```diff
@@ -8,9 +8,52 @@
...
+            for arg, key, default in zip(args, keys, defaults):
+                if key not in kwargs:
+                    kwargs[key] = default if arg is None else arg
+                else:
+                    raise TypeError(
+                        f"{self.__class__.__name__}.plot() got multiple values for argument '{key}'"
+                    )
+            warnings.warn(
+                DeprecationWarning(
+                    "The call signature of `plot()` has changed. Please, use the new signature."
```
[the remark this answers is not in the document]
tobiasjj: 👍 
I now explicitly provide the user with the information which parameter he/she should provide as a keyword.

lumicks/pylake/scan.py
```diff
@@ -404,35 +405,41 @@ def _to_spatial(self, data):
         else:
             return data
 
-    def _plot(
+    def plot(
         self,
-        channel,
-        axes,
+        channel="rgb",
+        axes=None,
```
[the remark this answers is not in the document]
tobiasjj: 👍

lumicks/pylake/tests/test_imaging_camera/test_correlated_stack.py
tobiasjj: This is due to the internal state of matplotlib keeping the current axes (usually the one created recently). Calling `plt.close()` will per default close the figure belonging to the current axes. As `plt.close()` is called in the for loop several times before "SUPPLY_AXES" is evaluted, the axes created before the for loop would have been closed by then.
I added a comment explaining why this is done as it is.

lumicks/pylake/detail/plotting.py
tobiasjj: 😄

tobiasjj:
> I wonder too if it'd be worthwhile to add a test for the call signature between these 4 plotting methods to ensure that everything stays synced. I think this is preferable to the other option of trying to have some mixin class handle it. I played around with it a bit, and you can use `inspect.signature` to get an `OrderedDict` of the arguments. You can then assert on the order of the positional arguments and also on any keyword arguments that should be common among all or a subset of these classes. Could avoid this headache in the future.
I added a test for the signature of the plotting "interface' for all classes supporting it (i.e. `Kymo`, `Scan`, `CorrelatedStack` and `PointScan`). The test centrally lives in `test_plotting.py`.

changelog.md
tobiasjj: 👍

lumicks/pylake/correlated_stack.py
tobiasjj: 😃

lumicks/pylake/correlated_stack.py
tobiasjj: 👍

lumicks/pylake/point_scan.py
tobiasjj: 👍

lumicks/pylake/detail/plotting.py
```diff
@@ -0,0 +1,36 @@
...
+from ..adjustments import ColorAdjustment
+
+
+def get_axes(axes=None, image_handle=None):
+    """Return `axes` or the axes of the provided `image_handle` or ensure both axes are the same. If
+    neither `axes` nor `image_handle` are provided, fallback to the current `matplotlib` axes"""
+    if axes is None:
+        axes = plt.gca() if image_handle is None else image_handle.axes
+    elif image_handle and axes is not image_handle.axes:
+        raise ValueError("Supplied image_handle with a different axes than the provided axes")
```
rpauszek, line 12: I think it would be nicer to `assert` rather than raise a `ValueError` here. The logic could be a tad simpler also:
```suggestion
    if axes is None:
        axes = plt.gca() if image_handle is None else image_handle.axes
    if image_handle:
            assert (
                axes == image_handle.axes
            ), "Supplied image_handle with a different axes than the provided axes"
```

lumicks/pylake/detail/tests/test_plotting.py
```diff
@@ -0,0 +1,155 @@
...
+    """
+    supported_interfaces = ["general"]
+    if image:
+        supported_interfaces.append("image")
+    if stack:
+        supported_interfaces.append("stack")
+
+    POSITIONAL_OR_KEYWORD = Parameter.POSITIONAL_OR_KEYWORD
+    KEYWORD_ONLY = Parameter.KEYWORD_ONLY
+    VAR_KEYWORD = Parameter.VAR_KEYWORD
```
rpauszek, line 103: I would get rid of this part and just use `Parameter.VALUE` in the list below. It's a bit of repetition but I think it's fine

lumicks/pylake/detail/tests/test_plotting.py
```diff
@@ -0,0 +1,155 @@
...
+        [("channel", POSITIONAL_OR_KEYWORD), "general"],
+        [("frame", KEYWORD_ONLY), "stack"],
+        [("adjustment", KEYWORD_ONLY), "image"],
+        [("axes", KEYWORD_ONLY), "general"],
+        [("image_handle", KEYWORD_ONLY), "image"],
+        [("show_title", KEYWORD_ONLY), "general"],
+        [("kwargs", VAR_KEYWORD), "general"],
+    ]
+
+    return [arg[0] for arg in arguments if arg[1] in supported_interfaces]
```
rpauszek, line 116: you can get the same effect with less braces, I think it looks cleaner
```suggestion
    arguments = [
        # name, parameter kind, interface type
        ["channel", POSITIONAL_OR_KEYWORD, "general"],
        ["frame", KEYWORD_ONLY, "stack"],
        ["adjustment", KEYWORD_ONLY, "image"],
        ["axes", KEYWORD_ONLY, "general"],
        ["image_handle", KEYWORD_ONLY, "image"],
        ["show_title", KEYWORD_ONLY, "general"],
        ["kwargs", VAR_KEYWORD, "general"],
    ]

    return [arg[0] for arg in arguments if arg[-1] in supported_interfaces]
```

lumicks/pylake/kymo.py
```diff
@@ -267,25 +276,47 @@ def pixelsize(self):
...
-        axes : mpl.axes.Axes
-            The axes instance in which to plot.
         adjustment : lk.ColorAdjustment
             Color adjustments to apply to the output image.
-
+        axes : mpl.axes.Axes, optional
+            If supplied, the axes instance in which to plot.
+        image_handle : `matplotlib.image.AxesImage` or None
+            Optional image handle which is used to update plots with new data rather than
+            reconstruct them (better for performance).
```
rpauszek, line 302: should these have the same annotations? maybe `axes : mpl.axes.Axes or None` is nicer and/or I think that's what we use more often? would be good on some rainy day to go through and check for stuff like this

lumicks/pylake/kymo.py
```diff
@@ -299,11 +330,20 @@ def _plot(self, channel, axes, adjustment=ColorAdjustment.nothing(), **kwargs):
...
+            **{**default_kwargs, **kwargs},
+        )
         axes.set_xlabel("time (s)")
         axes.set_ylabel(f"position ({self._calibration.unit_label})")
-        axes.set_title(self.name)
-        adjustment._update_limits(image_handle, image, channel)
+        if show_title:
+            axes.set_title(self.name)
+
+        return image_handle
```
rpauszek, line 346: so technically adding this I guess could be considered a breaking change 😕 but this would only affect things if someone was checking that the returned value was `None` which would've always be `True` which is .... less than useful. I say we keep this without doing a breaking release, because it's nice to have consistency, but let's put a note in the changelog to just highlight the change

lumicks/pylake/point_scan.py
```diff
@@ -22,23 +23,36 @@ def _get_plot_data(self, channel):
...
             axes.plot(time, count.data, **{"color": channel, "label": channel, **kwargs})
-            axes.set_xlabel("time (s)")
-            axes.set_ylabel(r"photon count")
+
+        axes.set_xlabel("time (s)")
+        axes.set_ylabel(r"photon count")
+        if show_title:
             axes.set_title(self.name)
+
+        return axes.get_lines()
```
rpauszek, line 58: same thing about a note in the changelog

lumicks/pylake/scan.py
```diff
@@ -337,6 +339,7 @@ def plot_correlated(
             scan.plot_correlated(file.force1x, channel="red")
         """
         from lumicks.pylake.nb_widgets.correlated_plot import plot_correlated
+
```
rpauszek, line 342: does `black` add this?

rpauszek (approved):
Looks really good! Left some suggestions for minor adjustments to help with clarity, but overall it's really great, especially appreciate the non-trivial deprecation paths 🎉

changelog.md
```diff
@@ -16,6 +21,8 @@
 #### Deprecations
 
 * Deprecated property `CorrelatedStack.src`.
+* Reordered the keyword arguments of the method `CorrelatedStack.plot()` and enforced all parameters after `channel` to be keyword arguments. For details see the [docstring](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.correlated_stack.CorrelatedStack.html#lumicks.pylake.correlated_stack.CorrelatedStack.plot).
+* Enforced the argument `axes` of the method `plot()` for `Scan`, `Kymo` and `PointScan` to be a keyword argument.
```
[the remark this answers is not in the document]
tobiasjj: Yes, we enforce almost all of the arguments to be keywords. However, the original and now removed method `Confocal.plot()` in the superclass only exposed `channel` and `axes` as named arguments, all other arguments where enforced to be keywords by `**kwargs`. So the only breaking change is `axes` being keyword enforced.

lumicks/pylake/detail/plotting.py
tobiasjj: Yep, that is easier to grasp! Will change it.

lumicks/pylake/correlated_stack.py
```diff
@@ -8,9 +8,52 @@
...
+            for arg, key, default in zip(args, keys, defaults):
+                if key in kwargs:
+                    raise TypeError(
+                        f"`{self.__class__.__name__}.plot()` got multiple values for argument `{key}`"
+                    )
+                kwargs[key] = default if arg is None else arg
+            warnings.warn(
+                DeprecationWarning(
+                    f"The call signature of `plot()` has changed: Please, provide `{key}` as a "
+                    "keyword argument."
```
JoepVanlier, line 46: Nitpick: This can also trigger for non-keyword args and will then give you a suggestion that is more constrained than it needs to be:
```python
cs.plot(3, "red")
```
Will net you a warning to make `channel` keyworded; but it would actually be totally fine having that one not be keyworded. Not really something to worry about though. As `kwarg` will just work. It's just slightly more restrictive than it needs to be.

lumicks/pylake/correlated_stack.py
```diff
@@ -8,9 +8,52 @@
...
+                ),
+                stacklevel=2,
+            )
+            return plot(self, **kwargs)
+        else:
+            return plot(self, *args, **kwargs)
```
JoepVanlier, line 52: There's a second failure mode that is maybe a bit more notorious which is that:

```python
cs.plot("red", 3)
```
goes here and complains that there are too many arguments. Unavoidable I guess. Not a big deal though, since even in the deprecated order, this was the wrong argument order.

lumicks/pylake/scan.py
```diff
@@ -404,35 +407,47 @@ def _to_spatial(self, data):
...
 
-    def _plot(
+    @_deprecate_basescan_plot_args
+    def plot(
         self,
-        channel,
-        axes,
+        channel="rgb",
+        *,
         frame=0,
```
JoepVanlier, line 415: Unfortunately, I found a counter example for the `Scan.plot(...)` deprecation path I'm afraid. 
```python
scan.plot("red", plt.gca(), 5)
```
Valid in the old API, raises `TypeError: plot() takes from 1 to 2 positional arguments but 4 were given` in the new one.

lumicks/pylake/kymo.py
JoepVanlier: I'd be fine with it.

lumicks/pylake/point_scan.py
```diff
@@ -22,23 +23,36 @@ def _get_plot_data(self, channel):
...
+        axes = get_axes(axes=axes)
         for channel in channels:
             count = self._get_plot_data(channel)
             time = (count.timestamps - count.timestamps[0]) * 1e-9
             axes.plot(time, count.data, **{"color": channel, "label": channel, **kwargs})
-            axes.set_xlabel("time (s)")
-            axes.set_ylabel(r"photon count")
+
+        axes.set_xlabel("time (s)")
+        axes.set_ylabel(r"photon count")
```
JoepVanlier, line 54: also nice catch btw :)

lumicks/pylake/detail/confocal.py
```diff
@@ -148,6 +149,36 @@ def from_json(cls, json_string):
...
+    import functools
+    import warnings
+    import matplotlib
+
+    @functools.wraps(plot)
+    def wrapper(self, *args, **kwargs):
+        # The plot function might be called with up to 2 positional arguments. We can gracefully
+        # convert the second positional argument (axes) into a keyword argument.
+        # old arguments: channel, axes
+        if len(args) == 2 and (isinstance(args[1], matplotlib.axes.Axes) or args[1] is None):
```
JoepVanlier, line 163: I think this should handle the case `len(args) > 2` (see below). The annoying part is that the signature for `Scan` and `Kymo` are different. The positional args for `Scan` and `Kymo` are `channel, axes, [frame, handle], adjustments, other kwargs` between `[]` are `Scan` exclusive.

JoepVanlier (changes requested):
I found a cornercase for the `Scan.plot()` deprecation path that raises on code that would've run with the previous version.

JoepVanlier (approved):

lumicks/pylake/detail/confocal.py
```diff
@@ -316,28 +347,10 @@ def _get_plot_data(self, channel):
         """Get data for plotting requested channel."""
         raise NotImplementedError
 
-    def _plot(self, channel, axes, **kwargs):
+    def plot(self, channel="rgb", **kwargs):
```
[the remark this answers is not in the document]
tobiasjj: Added a comment in `plot()`

lumicks/pylake/detail/tests/test_plotting.py
tobiasjj: removed.

lumicks/pylake/detail/tests/test_plotting.py
tobiasjj: I would like to keep the braces, as it emphasizes that the parameter tuple consisting of the name and the parameter kind are one unit and are expected with the higher-ranking interface type. I have inverted the order of the interface type and the parameter tuple, to express the classifying character of the interface type over the parameter tuple. Would you be fine with that?

lumicks/pylake/kymo.py
tobiasjj: 👍 I changed it to `matplotlib.*` in the `plot()` functions.

lumicks/pylake/kymo.py
tobiasjj: Added a note in the changelog.

lumicks/pylake/point_scan.py
tobiasjj: Added a changelog entry.

lumicks/pylake/scan.py
tobiasjj: That was isort, which unfortunately does not restrict itself to the module wide imports. Reverted it.

lumicks/pylake/correlated_stack.py
tobiasjj: Fixed it: Now all supplied and deprecated only positional arguments are mentioned in the deprecation warning.

lumicks/pylake/correlated_stack.py
tobiasjj: Yeah, I noticed that, too, but accepted it with the same reasoning you mentioned.

PR #404, opened 2022-10-13
Validate the lengths of tracks analyzed in `KymoTrackGroup.estimate_diffusion()`

rpauszek:
**Why this PR?**
When using `KymoTrackGroup.estimate_diffusion()`, if any tracks are shorter than the required length (3 for `"cve"` or 5 for the others), an error would be thrown. Instead we want to gracefully handle this and discard tracks that do not meet the requirement and issue a warning to the user.

Here I add a `min_length` argument where the user can specify a length, or defaults to `None` in which the appropriate min is used based on the specified method. I also changed up the docstring to reference `KymoTrack.estimate_diffusion()` in order to properly document the updated call signature without duplicating (most of it).

Relevant docs changes are [here](https://lumicks-pylake.readthedocs.io/en/msd_checker/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.html#lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup)

I also checked that there is a simple way to catch the emitted warning. You can do this with the `warnings` package and the following context manager:

```
# collect emitted warnings in a list `w`
with warnings.catch_warnings(record=True) as w:
    # turn on all warnings
    warnings.simplefilter("always")

    # warning is emitted here
    tracks.estimate_diffusion("ols")]

    # access relevant warning information for validation
    warning_type = w[0].category, RuntimeWarning)
    warning_msg = str(w[0].message))
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,11 +7,13 @@
 * Fixed and reintroduced lazy loading for `TimeSeries` data.
 * You can now add two `KymoTrackGroups` tracked on the same kymo together with the `+` operator.
 * TIFFs exported from `Scan` and `Kymo` now contain metadata. The `DateTime` tag indicated the start/stop timestamp of each frame. The `ImageDescription` tag contains additional information about the confocal acquisition parameters.
-* Added covariance-based estimator (cve) option to `KymoTrack.estimate_diffusion`. See [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more details.
+* Added covariance-based estimator (cve) option to `KymoTrack.estimate_diffusion()`. See [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more details.
+* Added the optional `min_length` parameter to `KymoTrackGroup.estimate_diffusion()` to discard tracks shorter than a specified length from the analysis.
 
 #### Other changes
 
 * All `KymoTrack` intances must have the same source `Kymo` and color channel in order to be in the same `KymoTrackGroup` instance. While this behavior was required previously for some downstream analyses on the tracks, it is now explicitly enforced upon `KymoTrackGroup` construction.
+* When calling `KymoTrackGroup.estimate_diffusion()` without specifying the `min_length` parameter, tracks which are shorter than the required length for the specified method will be discarded from analysis and a warning emitted. Previously, if any tracks were shorter than required, an error would be raised.
 
 #### Deprecations
 
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1,4 +1,5 @@
 from copy import copy
+import warnings
 from deprecated.sphinx import deprecated
 from sklearn.neighbors import KernelDensity
 from ..detail.utilities import use_docstring_from
@@ -386,9 +387,7 @@ class KymoTrack:
 
         Parameters
         ----------
-        method : str
-            Valid options are "cve", "ols" and "gls".
-
+        method : {"cve", "ols", "gls"}
             - "cve" : Covariance based estimator [5]_. Optimal if SNR > 1. Can only be used when
               track is equidistantly sampled.
             - "ols" : Ordinary least squares [3]_. Determines optimal number of lags.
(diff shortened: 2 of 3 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -798,6 +797,40 @@ def _histogram_binding_profile(self, n_time_bins, bandwidth, n_position_points):
...
+    def estimate_diffusion(self, method, *args, min_length=None, **kwargs):
+        r"""Estimate diffusion constant for each track in the group.
+
+        Calls :meth:`KymoTrack.estimate_diffusion` for each track. See the documentation for that
+        method for more detailed information and references.
+
+        Parameters
+        ----------
+        method : {"cve", "ols", "gls"}
+            - "cve" : Covariance based estimator [5]_. Optimal if SNR > 1. Can only be used when
```
JoepVanlier, line 809: Should either add the references or remove the `[5]_` etc because this will lead to errors in the docs build.

changelog.md
```diff
@@ -8,10 +8,12 @@
 * You can now add two `KymoTrackGroups` tracked on the same kymo together with the `+` operator.
 * TIFFs exported from `Scan` and `Kymo` now contain metadata. The `DateTime` tag indicated the start/stop timestamp of each frame. The `ImageDescription` tag contains additional information about the confocal acquisition parameters.
 * Added covariance-based estimator (cve) option to `KymoTrack.estimate_diffusion`. See [kymotracker documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html#studying-diffusion-processes) for more details.
+* Added the optional `min_length` argument to `KymoTrackGroup.estimate_diffusion` to discard tracks shorter than a specified length from the analysis.
```
mikhas, line 11: What you added was the optional parameter to the function _signature_. The argument is how the function is called.

mikhas (commented):
:shipit: Catching the warning in case filtering took place seems to work as intended. Haven't tested though if the new `min_length` parameter works.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -801,9 +800,43 @@ def _histogram_binding_profile(self, n_time_bins, bandwidth, n_position_points):
...
+        min_length : None or int (optional)
+            Discards tracks shorter than a certain length from the analysis. If `None` (the default)
+            tracks shorter than 3 points if `method == "cve"` or 5 points if `method == "ols" or "gls"`
+            will be discarded.
+        """
+        required_length = (3 if method == "cve" else 5) if min_length is None else min_length
+        filtered_tracks = [track for track in self if len(track) >= required_length]
+        n_discarded = len(self) - len(filtered_tracks)
+
+        if n_discarded:
```
JoepVanlier, line 831: Weren't we going to omit the message if `min_length` is specified?

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -685,11 +685,96 @@ def test_diffusion_cve(blank_kymo):
...
+        assert len(d) == 2
+
+    with pytest.warns(RuntimeWarning, match=warning_string(4)):
+        d = tracks.estimate_diffusion("ols", min_length=6)
+        assert len(d) == 2
+
+    # test empty result
+    with pytest.warns(RuntimeWarning, match=warning_string(6)):
+        d = tracks.estimate_diffusion("ols", min_length=200)
+        assert d == []
```
JoepVanlier, line 768: I think these should not have the warning emitted.

JoepVanlier (changes requested):
I thought we agreed on not emitting a warning if the filter action was explicit?

changelog.md
rpauszek: How would you word this then? I know it's more technically correct, so far this is the way we've been describing these kinds of changes in the changelog but I'd be ok changing if it sounds ok. Maybe
```suggestion
* Added the optional `min_length` parameter to the `KymoTrackGroup.estimate_diffusion()` signature. Use this parameter to discard tracks shorter than a specified length from the analysis.
```

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: no, we still want to emit the warning. If `min_length` isn't specified the value is selected based on the chosen method (`3 if "cve" else 5`) so we still want to notify users if tracks are discarded

JoepVanlier: Doesn't that entirely defeat the purpose of `min_length`? Then you might as well force people to use `filter_tracks` instead and not offer the option at all.

rpauszek: sorry, yeah I was confused about this. you're right

JoepVanlier (approved):
Looks solid. :shipit:

PR #405, opened 2022-10-14
docs: update fdfitting.rst

aafkevandenberg:
Updated fdfitting.rst:

- make sure everything runs using the given data files
- Some explanations were duplicated in the tutorial, I removed those.
- Reword some of the text.

The structure is largely the same.
 
Some remarks/discussion points:
- The variable 'fit' is overwritten all the time, so it is important to run the Notebook from beginning to end. What do you think, should the variable renamed for each example?
- Since i only had one set of data for RecA, I used the AdK data from Harbor for the 'global fitting' and 'multiple models' sections.
- There are 2 images to illustrate a 'bad' Likelihood profile. I decided to not add images and also not add the corresponding code, to save space. So the images are only visible on the webpage.

Docs build here: https://lumicks-pylake.readthedocs.io/en/update_docs_fdfitting/tutorial/fdfitting.html


docs/tutorial/fdfitting.rst
```diff
@@ -290,45 +273,121 @@ Calculating per point contour length
...
+Sometimes, you may want to add multiple data sets for one condition, to perform a global fit.
+Consider two lists of distance and force vectors stored in `distances` and `forces`::
+
+    file_adk1 = lk.File("Data\\adk5_curve1.h5")
+    d_adk1 = file_adk1.fdcurves['adk5_curve1'].d['0s':'13s'].data
+    f_adk1 = file_adk1.fdcurves['adk5_curve1'].f['0s':'13s'].data
+
+    file_adk2 = lk.File("Data\\adk5_curve2.h5")
+    d_adk2 = file_adk2.fdcurves['adk5_curve2']['0s':'13s'].d.data
+    f_adk2 = file_adk2.fdcurves['adk5_curve2']['0s':'13s'].f.data
```
JoepVanlier, line 378:
```suggestion
    d_adk2 = file_adk2.fdcurves["adk5_curve2"]["0s":"13s"].d.data
    f_adk2 = file_adk2.fdcurves["adk5_curve2"]["0s":"13s"].f.data
```

docs/tutorial/fdfitting.rst
```diff
@@ -380,82 +439,59 @@ Note that this piece of code will lead to parameters `DNA/Lc_0`, `DNA/Lc_1` etc.
...
+Next, add data after unfolding to `fit[model2]`::
 
-Confidence intervals and standard errors
-****************************************
+    d_adk1 = file_adk1.fdcurves['adk5_curve1'].d['21s':'29s'].data
+    f_adk1 = file_adk1.fdcurves['adk5_curve1'].f['21s':'29s'].data
 
-Once parameters have been fitted, standard errors can easily be obtained as follows::
+    d_adk2 = file_adk2.fdcurves['adk5_curve2']['20s':'29s'].d.data
+    f_adk2 = file_adk2.fdcurves['adk5_curve2']['20s':'29s'].f.data
```
JoepVanlier, line 467:
```suggestion
    d_adk2 = file_adk2.fdcurves["adk5_curve2"]["20s":"29s"].d.data
    f_adk2 = file_adk2.fdcurves["adk5_curve2"]["20s":"29s"].f.data
```

docs/tutorial/fdfitting.rst
```diff
@@ -380,82 +439,59 @@ Note that this piece of code will lead to parameters `DNA/Lc_0`, `DNA/Lc_1` etc.
...
-    fit["DNA/Lc"].stderr
+    distances2 = [d_adk1, d_adk2]
+    forces2 = [f_adk1, f_adk2]
 
-Assuming that the parameters are not at the bounds, the sum of random variables with finite moments converges to a
-Gaussian distribution. This allows for the computation of confidence intervals using the Wald test
-:cite:`press1990numerical`. To get these asymptotic intervals, we can use the member function `.ci` with a desired
-confidence interval::
+    for i, (distance, force) in enumerate(zip(distances2, forces2)):
+        fit[model2].add_data(f"After unfolding {i}", force[distance>0], distance[distance>0])
```
JoepVanlier, line 473: There's data with negative distance? Is this due to noise, or does the average distance actually go negative?

docs/tutorial/fdfitting.rst
```diff
@@ -131,25 +113,23 @@ If we also expect an offset in the dependent parameter, we can simply add an off
...
-    kT               4.11  [pN*nm]   False               0              8
+    DNA/Lp          40     [nm]      True              0.001            100
+    DNA/Lc          16     [micron]  True            0.00034            inf
+    DNA/St        1500     [pN]      True                  1            inf
+    kT               4.11  [pN*nm]   False              3.77              8
     DNA/f_offset     0.01  [pN]      True               -0.1            0.1
 
-From the above example, you can see how easy it is to composite models. Sometimes, models become more complicated. For
-instance, we may have two worm like chain models that we wish to add, and then invert. For the Odijk model, this can be
+Sometimes, models become more complicated. For instance, we may have two worm like chain models that we wish to add, and then invert. For the Odijk model, this can be
```
JoepVanlier, line 122: I wonder if it would make sense to discuss when you'd add them by distance (construct) and when you'd add them by force (multiple tethers).

docs/tutorial/fdfitting.rst
```diff
@@ -131,25 +113,23 @@ If we also expect an offset in the dependent parameter, we can simply add an off
...
 done as follows::
 
     model = lk.ewlc_odijk_distance("DNA") + lk.ewlc_odijk_distance("protein") + lk.distance_offset("offset")
     model = model.invert()
 
 Note how we added three models and then inverted the composition of those models. Models inverted via `invert()` will
```
JoepVanlier, line 128: Now that we have API docs, it would be nice to link the docs page for `.invert()` here.
```suggestion
Note how we added three models and then inverted the composition of those models. Models inverted via :meth:`.invert() <lumicks.pylake.fitting.model.Model.invert()>` will
```

docs/tutorial/fdfitting.rst
```diff
@@ -62,36 +59,21 @@ We can simulate the model by passing a dictionary with parameters values::
...
-    DNA/Lp    40     [nm]      True                  0            inf
-    DNA/Lc    16     [micron]  True                  0            inf
-    DNA/St  1500     [pN]      True                  0            inf
-    kT         4.11  [pN*nm]   False                 0              8
-
-This seems to be what we want: force as a function of distance. Let's assume we have acquired data, but upon analysis
+Let's assume we have acquired data, but upon analysis
 we notice that the template matching wasn't completely optimal. In addition, we experienced some force drift, while
 forgetting to reset the force back to zero. In this case, we can incorporate offsets in our model. We can introduce an
 offset in the independent parameter, by calling `.subtract_independent_offset()` on our model::
```
JoepVanlier, line 79:
```suggestion
offset in the independent parameter, by calling :meth:`.subtract_independent_offset() <lumicks.pylake.fitting.model.Model.subtract_independent_offset()>` on our model::
```

docs/tutorial/fdfitting.rst
```diff
@@ -131,25 +113,23 @@ If we also expect an offset in the dependent parameter, we can simply add an off
...
 done as follows::
 
     model = lk.ewlc_odijk_distance("DNA") + lk.ewlc_odijk_distance("protein") + lk.distance_offset("offset")
     model = model.invert()
 
 Note how we added three models and then inverted the composition of those models. Models inverted via `invert()` will
-typically be slower than the pre-inverted counterparts. This is because the inversion is done numerically rather than
-analytically. For more complex examples on how this inversion may be used, please see the examples.
+typically fit slower than the pre-inverted counterparts. This is because the inversion is done numerically rather than
+analytically. For example, using `lk.ewlc_odijk_force` would be faster to use than `lk.ewlc_odijk_distance.invert()`. For more complex examples on how this inversion may be used, please see the examples.
```
JoepVanlier, line 130:
```suggestion
analytically. For example, using :func:`lk.ewlc_odijk_force() <lumicks.pylake.ewlc_odijk_force()>` would be faster to use than `lk.ewlc_odijk_distance.invert()`. For more complex examples on how this inversion may be used, please see the examples.
```

JoepVanlier (changes requested):
Great work. I think it reads a lot more fluently this way.

Thanks for picking this up Aafke! And my apologies that I'm so late with the review.

I just have some suggestions here and there. Most of it pretty minor.

docs/tutorial/fdfitting.rst
aafkevandenberg: It would indeed be good to elaborate on that somewhere. Maybe that would fit in a theory section, rather than a tutorial?

docs/tutorial/fdfitting.rst
```diff
@@ -159,95 +139,98 @@ the parameters involved in the models and data, and will allow you to interact w
...
+    model = lk.ewlc_odijk_force("DNA")
     fit = lk.FdFit(model)
 
 Adding data to the fit
 **********************
 
 To do a fit, we have to add data. Let's assume we have two data sets. One was acquired in the presence of a ligand, and
 another was measured without a ligand. We expect this ligand to only affect the contour length of our DNA. Let's add the
-first data set which we name `Control`. Adding it to the fit is simple::
+first data set which we name `Control`. Since the extensible worm-like chain works up to 30 pN, we select forces < 30pN::
```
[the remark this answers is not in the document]
aafkevandenberg: I agree, changed it.

docs/tutorial/fdfitting.rst
aafkevandenberg: The distance is actually always positive (the >0 part was copy-pasted from the AdK Notebook on Harbor). I removed  ' `[distance>0]` '

docs/tutorial/fdfitting.rst
JoepVanlier: Yeah, fair enough. That makes sense. Might still make sense to mention that you add by distances (length)? It is something a user might run into, if they try to add a force to a distance or vice versa.

docs/tutorial/fdfitting.rst
```diff
@@ -290,45 +273,121 @@ Calculating per point contour length
...
+
+    profile = fit.profile_likelihood("model/Lc", num_steps=1000)
+    
+The lower and upper bound of the 95% confidence interval of the given parameter (`Lc` in this example) can be obtained as::
+
+    [profile.lower_bound, profile.upper_bound] # [lower bound, upper bound]
+
+Note that these profiles require iterative computation and are therefore time consuming to produce.
+For a well parametrized model with sufficient data, a profile plot results in a (near) parabolic shape, where the line
+of the parabola intersects with the confidence interval lines (dashed). The confidence intervals are then determined to
```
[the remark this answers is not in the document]
aafkevandenberg: I think it is better not to emphasize that. We explain a few lines above that using the likelihood profile has advantages. Users don't know in advance whether they will get a well defined profile or not, so it is just better to stick to this method.

docs/tutorial/fdfitting.rst
aafkevandenberg: I added an explanation on why methods are added in terms of distance. I do feel that this particular topic deserves an extra theory section, but we can add that when there is time.

docs/tutorial/fdfitting.rst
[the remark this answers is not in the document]
aafkevandenberg: Btw, I asked for other unfolding data then AdK. We have hairpin data, but that unfolding event is rather small and not well visible on the low resolution fd curves. The demolab does not have any other protein unfolding data that can be shared. The nucleosome unwrapping data are too complicated and don't fit well as only Lc changes.

docs/tutorial/fdfitting.rst
aafkevandenberg: Update: The distance is always positive _or_ 0. So the >0 statement was there to remove the distance = 0 datapoints, resulting from bad camera tracking. I put the >0 back, but at a different location, right when `distances` and `forces` are defined.

JoepVanlier: Right, that makes sense. But it should probably be mentioned in the text somewhere, since it's something that may surprise someone.

docs/tutorial/fdfitting.rst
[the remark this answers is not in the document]
JoepVanlier: > We explain a few lines above that using the likelihood profile has advantages. Users don't know in advance whether they will get a well defined profile or not, so it is just better to stick to this method.

I agree that they are better. But one reason could be that profiles are a lot slower and may become prohibitively slow in fits with many datasets and complex inverted models. But anyway, I agree, it probably takes away from the tutorial nature of the section.

docs/tutorial/fdfitting.rst
```diff
@@ -380,82 +440,59 @@ Note that this piece of code will lead to parameters `DNA/Lc_0`, `DNA/Lc_1` etc.
...
-You can add as many data sets as you want to both models and fit it all at once.
-
-Plotting is straightforward in this setting. We can plot the data sets corresponding to model 1 and 2 as follows::
-
-    fit[model1].plot()
-    fit[model2].plot()
-
-Accessing the model parameters for a specific data set is a little more complicated in this setting. If we want to
-obtain the parameters for "data for model 1", we'd have to invoke::
+Rather then directly adding the data from after the unfolding event, and fitting everything together, we are going to fit in incrementally::
```
JoepVanlier, line 457:
```suggestion
Rather than directly adding the data from after the unfolding event, and fitting everything together, we are going to fit incrementally::
```

docs/tutorial/fdfitting.rst
```diff
@@ -159,95 +138,99 @@ the parameters involved in the models and data, and will allow you to interact w
...
 
     distance = np.arange(2.0, 5.0, 0.01)
-    simulation_result = model(distance, fit["Control"])
+    simulated_force = model(distance, fit["Control"])
+    plt.plot(distance, simulated_force)
 
-Basically what happens here is that `fit["Control"]` grabs those parameters needed to simulate the condition
-corresponding to the dataset with the name `control`. By providing specifically those parameters to the model, we can
-simulate that condition.
+Here `fit["Control"]` grabs the parameters needed to simulate the condition corresponding to the dataset with the name "Control".
```
JoepVanlier, line 228:
```suggestion
Here `fit["Control"]` grabs the parameters needed to simulate the condition corresponding to the dataset with the name `"Control"`.
```

docs/tutorial/fdfitting.rst
```diff
@@ -380,82 +440,59 @@ Note that this piece of code will lead to parameters `DNA/Lc_0`, `DNA/Lc_1` etc.
...
+    fit[model2].plot()
 
-    profile.plot_relations()
 
-These inferred relations can provide information on the coupling between different parameters. This can be quite
-informative when diagnosing fitting issues. For example, when fitting a contour length in the presence of an distance
-offset, we can observe that the two are related. To produce the following figure, we set a lower bound and upper bound
-of -0.1 and 0.1 for the distance respectively. We can see that the profile is perfectly flat until the distance reaches
-the bound. Only then does the profile suddenly jump.
+Accessing the model parameters for a specific data set is a little more complicated in this setting. If we want to obtain the parameters for “data for model 1”, we’d have to invoke::
```
JoepVanlier, line 495: I'm unsure about this one, but shouldn't this be:
```suggestion
Accessing the model parameters for a specific data set is a little more complicated in this setting. If we want to obtain the parameters for `“Before unfolding 1”`, we’d have to invoke::
```

JoepVanlier (changes requested):
I went over it again. I noticed a few small things, but looking good!

docs/tutorial/fdfitting.rst
```diff
@@ -9,22 +9,19 @@ Fd Fitting
...
 ------
-
-When fitting data, everything revolves around models. One of these models is the so-called
-extensible worm-like-chain model by Odijk et al. Let's have a look at it. We can construct this
-model using the supplied function :func:`lk.ewlc_odijk_distance()
-<lumicks.pylake.ewlc_odijk_distance()>`.
-
-Note that we also have to give it a name. This name will be prefixed to model specific parameters in this model::
+Pylake has several models implemented that can be used for fitting force extension data. For a full list of models invokie `help(lk.fitting.models)`
+or see :ref:`fd_models`. For example, a model that is frequently used for fitting force extension data of dsDNA is the so-called
```
ab412, line 17: You can ignore this if it has been mentioned already elsewhere
```suggestion
or see :ref:`fd_models`. For example, a model that is frequently used for fitting force extension data of double-stranded DNA (dsDNA) is the so-called
```

docs/tutorial/fdfitting.rst
```diff
@@ -36,24 +33,24 @@ Entering model prints the model equation and its default parameters::
...
 
-Note how most of the parameters listed have the model name prefixed. For example, the persistence length in this model
-is now named `DNA/Lp`. Here `DNA` refers to the name we gave to the model.
+As we can see from the model equation, the model is given as distance as a function of force.
+Note also how the model name is prefixed to the model specific parameters, forming a key-value pair like so: `"name/parameter": value`. For example, the persistence length in this model
+is now named `DNA/Lp`.
 
-One parameter stands out, which is kT as it doesn't have `DNA` prefixed. The reason for this is that this parameter is
-not model specific (in this case the parameter represents the Boltzmann constant times the temperature).
+kT does not have `DNA` prefixed, because this parameter is not model specific (the parameter represents the Boltzmann constant times the temperature).
```
ab412, line 45:
```suggestion
kT does not have `DNA` prefixed since it represents the product of the Boltzmann constant times the temperature, and is therefore not model specific.
```

docs/tutorial/fdfitting.rst
```diff
@@ -62,39 +59,24 @@ We can simulate the model by passing a dictionary with parameters values::
...
+
 
-Note how the model name is prefixed to the model specific parameters, forming a key-value pair like so: `"name/parameter": value`.
 
 Model composition and inversion
 -------------------------------
 
-In practice, we would typically want to fit force as a function of distance however. For this we have the inverted
-Odijk model::
+Above, we investigated a model with distance as a function of force. In practice, the force is typically fitted as a function of distance. This can be done using :func:`lk.ewlc_odijk_force() <lumicks.pylake.ewlc_odijk_force()>`::
```
ab412, line 72:
```suggestion
Above, we investigate a model with distance as a function of force. In practice, the force is typically fitted as a function of distance. This can be done using :func:`lk.ewlc_odijk_force() <lumicks.pylake.ewlc_odijk_force()>`::
```

docs/tutorial/fdfitting.rst
```diff
@@ -62,39 +59,24 @@ We can simulate the model by passing a dictionary with parameters values::
...
-
-    Name      Value  Unit      Fitted      Lower bound    Upper bound
-    ------  -------  --------  --------  -------------  -------------
-    DNA/Lp    40     [nm]      True                  0            inf
-    DNA/Lc    16     [micron]  True                  0            inf
-    DNA/St  1500     [pN]      True                  0            inf
-    kT         4.11  [pN*nm]   False                 0              8
-
-This seems to be what we want: force as a function of distance. Let's assume we have acquired data, but upon analysis
+Let's assume we have acquired data, but upon analysis
```
ab412, line 76:
```suggestion
Let's assume we have acquired data, and upon analysis
```

docs/tutorial/fdfitting.rst
```diff
@@ -131,25 +113,22 @@ If we also expect an offset in the dependent parameter, we can simply add an off
...
+    DNA/Lp          40     [nm]      True              0.001            100
+    DNA/Lc          16     [micron]  True            0.00034            inf
+    DNA/St        1500     [pN]      True                  1            inf
+    kT               4.11  [pN*nm]   False              3.77              8
     DNA/f_offset     0.01  [pN]      True               -0.1            0.1
 
-From the above example, you can see how easy it is to composite models. Sometimes, models become more complicated. For
-instance, we may have two worm like chain models that we wish to add, and then invert. For the Odijk model, this can be
-done as follows::
+Sometimes, models become more complicated. For instance, we may have two worm like chain models, one for a DNA tether and the other for an unfolded protein. The total length of the construct is then the sum of the length of the DNA and the protein and the total distance is given by::
```
ab412, line 122:
```suggestion
Sometimes models become more complicated. For instance, we may have two worm-like chain models, one for a DNA tether and the other for an unfolded protein. The total length of the construct is then the sum of the length of the DNA and the protein and the total distance is given by::
```

docs/tutorial/fdfitting.rst
```diff
@@ -131,25 +113,22 @@ If we also expect an offset in the dependent parameter, we can simply add an off
...
 
     model = lk.ewlc_odijk_distance("DNA") + lk.ewlc_odijk_distance("protein") + lk.distance_offset("offset")
     model = model.invert()
 
-Note how we added three models and then inverted the composition of those models. Models inverted via `invert()` will
-typically be slower than the pre-inverted counterparts. This is because the inversion is done numerically rather than
-analytically. For more complex examples on how this inversion may be used, please see the examples.
+Note how the three models are all defined as distance as a function of force. Since fitting is best done for force as a function of distance, we then inverted the composited model. Note that models inverted via :meth:`.invert() <lumicks.pylake.fitting.model.Model.invert()>` will
```
ab412, line 127:
```suggestion
Note how the three models all define distance as a function of force. Since fitting is best done for force as a function of distance, we then invert the composited model. Note that models inverted via :meth:`.invert() <lumicks.pylake.fitting.model.Model.invert()>` will
```

JoepVanlier (approved):
The new doc page looks great, thank you!

PR #411, opened 2022-10-27
Add `DwelltimeModel.rate_constants` property

rpauszek:
**Why this PR?**

Although trivial, quite often the rate constant is quantity that you want, not the lifetime. This stemmed from a request for additional documentation on the relationship between the two parameters and the underlying assumptions for this analysis, also added here.

Docs build [here](https://lumicks-pylake.readthedocs.io/en/rate_constants/tutorial/population_dynamics.html#dwell-time-analysis) and [here](https://lumicks-pylake.readthedocs.io/en/rate_constants/tutorial/population_dynamics.html#a-note-on-rate-constants)


docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,44 @@ the distribution values are directly available through the properties `Dwelltime
...
+
+    \mathrm{DNA} + \mathrm{protein} \rightarrow \left[ \mathrm{DNA} \cdot \mathrm{protein} \right]
+
+This reaction is described by the second order association rate constant :math:`k_\mathrm{on}` with units of :math:`\mathrm{M}^{-1}\mathrm{sec}^{-1}`.
+Under the first order approximation, this can be determined by fitting the appropriate dwell times to the exponential model and multiplying
+the resulting rate constant by the concentration of the protein in solution.
+
+.. note::
+
+    This procedure relies on having an accurate measurement of the bulk concentration which can be difficult to determine when working
```
tobiasjj, line 211:
```suggestion
    This procedure relies on having an accurate measurement of the bulk concentration, which can be difficult to determine when working
```

tobiasjj (changes requested):
I like the clear and brief explanation of how rate constants are modeled and what one needs to consider when measuring and interpreting them.
The PR looks good to me, but I would like to have one test checking the functionality of the `rate_constants` property.

aafkevandenberg (commented):
Thanks for adding this! The explanation is really nice.

I am a bit confused about the purpose of this section. The title just says ' note on rate constants'. It would be good to make a more descriptive title and an intro to make clear to the reader when this section would be helpful. And here you only elaborate on situations that can be described by a single exponential, was that on purpose?

rpauszek:
> ... I would like to have one test checking the functionality of the `rate_constants` property.

added

rpauszek:
> Thanks for adding this! The explanation is really nice.
> 
> I am a bit confused about the purpose of this section. The title just says ' note on rate constants'. It would be good to make a more descriptive title and an intro to make clear to the reader when this section would be helpful. And here you only elaborate on situations that can be described by a single exponential, was that on purpose? (Eg what about a paused state that results in two-step unbinding?)

Changed up the title and added a couple intro sentences, hopefully it's better. Always open to suggestions though, sometimes I struggle with this kind of stuff

Hadn't really thought about multi-exponential case as I was mainly concerned with the conversion/interpretation of the rate constants specifically but this is a good idea. Added an extra section. Not sure if I quite like it, so again suggestions are always welcome

tobiasjj (approved):
Nice! I also like the improved documentation.
🚴

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
 `DwelltimeModel.bootstrap.lifetime_distributions` which return the data as a `numpy` array with
 shape `[# components, # bootstrap samples]`.
 
+.. _rate_constants:
+
+Assumptions and limitations on determining rate constants
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
aafkevandenberg, line 180: Much better!

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
...
+
+A common example of this is the binding of a protein to a DNA strand:
+
+.. math::
+
+    \mathrm{DNA} + \mathrm{protein} \rightarrow \left[ \mathrm{DNA} \cdot \mathrm{protein} \right]
+
+This reaction is described by the second order association rate constant :math:`k_\mathrm{on}` with units of :math:`\mathrm{M}^{-1}\mathrm{sec}^{-1}`.
+Under the first order approximation, this can be determined by fitting the appropriate dwell times to the exponential model and multiplying
+the resulting rate constant by the concentration of the protein in solution.
```
aafkevandenberg, line 210: So, what is 'this' in the sentence, 'this can be determined'? 
The rate that you get out of the dwell time distribution is effectively kon * [Protein], right? So you would have to divide the fitted rate constant by concentration if you want to obtain kon? Do I think about it wrongly, or were you trying to say something else?

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
...
+
+.. note::
+
+    This procedure relies on having an accurate measurement of the bulk concentration, which can be difficult to determine when working
```
aafkevandenberg, line 214: I would be more specific on what 'this procedure' is.

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
...
+    in the nanomolar regime, as nonspecific adsorption can lower the
+    effective concentration at the experiment.
+
+A warning on reliability and interpretation of multi-exponential kinetics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Care must also be taken when interpreting a mixture of exponential distributions. Sometimes, a double exponential distribution is required to accurately
+describe the observed data. This occurs when a system consists of two states with different kinetics that emit the same observable signal. For instance,
+the dissociation rate of a bound protein might depend on the microscopic conformation of the molecule that does not affect the emission intensity of
+an attached fluorophore used for tracking.
```
aafkevandenberg, line 224: Why did you add the statement 'that does not affect the emission intensity'. Are you thinking about specific situations where the emission intensity is affected?

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
...
+
+However, in the setting of a limited number of observations, the optimization of the exponential mixture can
+become non-identifiable, meaning that there are multiple sets of parameters that result in near equal likelihoods. A good first check on the quality of
+the optimization is to run a bootstrap simulation (as described above) and check the shape of the resulting distributions. Very wide, flat, or skewed
+distributions can indicate that the model was not fitted to a sufficient amount of data.
```
aafkevandenberg, line 229: Maybe end the section using an advice on how to handle a bit fit, eg  'for processes that are best described by two exponentials, it may be necessary to acquire more data for a good fit'.

aafkevandenberg (commented):
Much better!

I think a whole page can be written on this topic, but this should already help the user a lot!

docs/tutorial/population_dynamics.rst
```diff
@@ -157,6 +174,60 @@ the distribution values are directly available through the properties `Dwelltime
...
+
+.. note::
+
+    This procedure relies on having an accurate measurement of the bulk concentration, which can be difficult to determine when working
+    in the nanomolar regime, as nonspecific adsorption can lower the
+    effective concentration at the experiment.
+
+A warning on reliability and interpretation of multi-exponential kinetics
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
```
aafkevandenberg, line 220: I would start this section more positive/neutral, just by saying, 'sometimes a process can best be described by a two or more exponential distributions'. 
Then give an example of such a process, and then explain that care has to be taken that the quality of the fit is good, and that you need enough data. Now users may feel that a mixture of exponential distributions is always a bad thing.

docs/tutorial/population_dynamics.rst
aafkevandenberg: And I see you mention '_appropriate_ dwell times'. I guess the word 'appropriate' is hiding a whole discussion on itself :P

JoepVanlier (approved):
It wasn't added in this PR, but reading the docs again, I was wondering, is there a reason you specifically show the `BIC` and not the `AIC` in the overfitting case? Did the `AIC` point to the wrong model?

Not suggesting to piggyback it now, but if we ever expand this theory section, it might be nice to at some point in the future show the bootstrap for some failure modes rather than tell about them.

I know from experience that asymptotics like the AIC and BIC, albeit better than just an LR ratio alone, tend to not always fare that well. Anyways, for now, this looks like a big step up from what we had, so I would say :shipit:

aafkevandenberg (approved):
The updated Dwell time analysis section is really nice! Very instructive and clearly written.

docs/tutorial/population_dynamics.rst
rpauszek: no, you're right. I was apparently just not thinking when I wrote this

docs/tutorial/population_dynamics.rst
rpauszek: I guess for dwell time analysis my mind immediately jumps to some more advanced experiments like PIFE. For those experiments the conformation of the protein/complex leads to brighter or dimmer fluorescence in which case you can then tell the difference of the state. As this class is written for general analysis, I thought it would be nice to mention it here.

It also holds true for the convenience method `KymoTrackGroup.fit_binding_times()` that we have as loss of emission can be considered emission intensity being affected I guess.

rpauszek:
> It wasn't added in this PR, but reading the docs again, I was wondering, is there a reason you specifically show the `BIC` and not the `AIC` in the overfitting case? Did the `AIC` point to the wrong model?
> 
> Not suggesting to piggyback it now, but if we ever expand this theory section, it might be nice to at some point in the future show the bootstrap for some failure modes rather than tell about them.
> 
> I know from experience that asymptotics like the AIC and BIC, albeit better than just an LR ratio alone, tend to not always fare that well. Anyways, for now, this looks like a big step up from what we had, so I would say :shipit:

No, I just picked one to display so as not to overwhelm users that might not be used to this stuff in the tutorial section. I agree, we should expand on the descriptions for these tests when we split off the separate theory section

rpauszek:
> The updated Dwell time analysis section is really nice! Very instructive and clearly written. Just one suggestion: a customer was asking what exact function is being fitted. It would be nice to spell out the single and double exponential function.

The likelihood function was already written out in the `The Exponential (Mixture) Model` subsection, but I can see how this all might be a bit unclear to many users as MLE is not as well known as the more typical least square fitting technique. With this PR there is a bit of repetition at the top now showing the exponential PDF so hopefully that'll make things a bit clearer. I'll merge this for now so we can get the feature in, but as we're going to be splitting off a separate `Theory` section in the docs so we can go into more technical data, I think we can expand on this point more there when we get to it.  

But do let us know if there's still customer questions after this update and we can definitely try to address it.

PR #412, opened 2022-10-27
calibration: fix analytic fit failing is `b < 0` and potentially returning `diff < 0`

JoepVanlier:
In rare cases, it is possible that during the analytic fit, the diffusion constant returned is negative. This makes no sense as an initial guess and leads to numerical problems down the road. For this reason, an alternative method to guesstimate it is used when this occurs.

In addition, `a > 0` is not a strong enough condition for the fc estimator, because the square root operation can still produce an exception if `b < 0` while `a > 0`.

### DIFF
--- a/lumicks/pylake/force_calibration/detail/power_models.py
+++ b/lumicks/pylake/force_calibration/detail/power_models.py
@@ -46,8 +46,9 @@ def fit_analytical_lorentzian(ps):
     )
 
     # Having a and b, calculating fc and D is trivial.
-    if a > 0:
-        fc = math.sqrt(a / b)  # corner frequency [Hz]
+    a_div_b = a / b
+    if a_div_b > 0:
+        fc = math.sqrt(a_div_b)  # corner frequency [Hz]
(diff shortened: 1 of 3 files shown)


rpauszek (approved):
🚢

lumicks/pylake/force_calibration/detail/power_models.py
```diff
@@ -46,16 +46,32 @@ def fit_analytical_lorentzian(ps):
...
+        # estimating this quantity. The power spectral density at frequency zero is given by:
+        #
+        #   P0 = D / (pi**2 * fc**2)
+        #
+        # Therefore, an alternative method to get a rough estimate for this quantity would be:
+        #
+        #   D_guess = pi**2 * fc**2 * power[0]
+        #
+        # Where power[0] is the lowest frequency in the spectrum.
+        D = math.pi**2 * fc**2 * ps.power[0]
```
tobiasjj, line 74: Nice fallback!

PR #413, opened 2022-10-28
Make tests in `test_imaging_confocal` less brittle: preparational work

tobiasjj:
**Why?**
As a developer, I do not want to spent two days updating all the tests whenever I want to change one of the source images so that I don't waste a lot of time.

**Explanation**
The tests depending on the mocked h5-file are very brittle. Many tests rely on hard coded magic numbers, which are only valid for the hard coded sample data. Additionally, the hard coded sample data has a wrong format that never would be produced as such by Bluelake. 

To mitigate the issue, I have extended the `mock_confocal` module to not only generate test data, but also create reference objects holding the expected values that can be used to assert against in the test functions, instead of the magic numbers.

This PR implements the basic infrastructure for further PRs that implement the actual changes in the test functions.

**Details**
* move the existing tests from `test_imaging_confocal` to `test_imaging_confocal_old`
* `generate_scan()` now ensures that the created `Scan` will reconstruct the image as it was supplied to `generate_scan()`, regardless of the order of the axes
* create a `ConfocalRef` reference object, holding expected data to assert against, created by functions `generate_kymo_with_ref()` and `generate_scan_with_ref()`
* create a function `timestamp_references()` that calculates the timestamps and timestamp ranges to be expected (the calculation is solely based on information known _before_ creation of a `Kymo` or a `Scan`
* Add a test function to test the generation of the timestamp references
* Add unit test functions for `ScanAxis` and `ScanMetadata`

**Additional changes**
I took the liberty to change further smaller things:
* move fixture `grab_tiff_tags` from global `contest` into `conftest` of test_imaging_confocal_old`
* add multicolor support for generation of kymos and scans from images 
* few variables were renamed

**Outlook**
Further PRs will follow, which will incrementally re-implement functional duplicates of the old tests using the new infrastructure. In the long run, the old tests should be removed, when the functionality and coverage of the new tests was properly tested.


tobiasjj:
As discussed, I will rework this PR and dependent PRs #418 #419 #420. WIP ...

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +211,268 @@ def make_slice(data):
...
+    if not with_ref:
+        return confocal
+
+    # Assume the x axis to be the slow and the y axis to be the fast indexed one: The x axis
+    # contains the `lines_per_frame` and the y axis the `pixels_per_line`:
+    number_of_frames, pixels_per_line, lines_per_frame = (
+        [1, *image.shape[:2]] if len(image.shape) == 2 + multi_color else image.shape[:3]
+    )
+
+    x_axis_fast = len(axes) == 2 and axes[0] < axes[1]
```
JoepVanlier, line 305: Could also be Y axis fast (if the second axis is Z).

JoepVanlier (changes requested):
Great start with these unit tests for the metadata. A few comments from my side.

lumicks/pylake/tests/data/test_generate_timestamps.py
```diff
@@ -0,0 +1,89 @@
...
+    [
+        (1, 2, 2, 0, 1, 1, 0, False, False),  # kymo, minimal, y_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, False, True),  # kymo, minimal, x_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, True, False),  # scan, minimal, y_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, True, True),  # scan, minimal, x_axis_fast
+        (1, 2, 3, 20e9, 16e9, 7, 15, False, False),  # kymo, y_axis_fast
+        (1, 2, 3, 20e9, 16e9, 7, 15, False, True),  # kymo, x_axis_fast
+        (1, 2, 3, 20e9, 16e9, 7, 15, True, False),  # scan, 1 frame, y_axis_fast
+        (1, 2, 3, 20e9, 16e9, 7, 15, True, True),  # scan, 1 frame, x_axis_fast
+        (1, 2, 3, 20e9, 16e9, 6, 14, False, False),  # kymo, even, y_axis_fast
```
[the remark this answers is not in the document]
tobiasjj: Yup, I missed that, as I was enforcing `start` and `dt` to be `np.int64` in `generate_confocal()`. Good you caught that one! Now `generate_timestamps()` ensures `start >= 0` and `dt > 0` and enforces them to be of type int. Additionally, `generate_timestamps()` checks for overflow errors and raises in that case. I also added tests to check for the errors.

lumicks/pylake/tests/data/test_generate_timestamps.py
```diff
@@ -0,0 +1,89 @@
...
+    (
+        "number_of_frames, lines_per_frame, pixels_per_line, start, dt, samples_per_pixel, "
+        "line_padding, scan, x_axis_fast"
+    ),
+    [
+        (1, 2, 2, 0, 1, 1, 0, False, False),  # kymo, minimal, y_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, False, True),  # kymo, minimal, x_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, True, False),  # scan, minimal, y_axis_fast
+        (1, 2, 2, 0, 1, 1, 0, True, True),  # scan, minimal, x_axis_fast
+        (1, 2, 3, 20e9, 16e9, 7, 15, False, False),  # kymo, y_axis_fast
```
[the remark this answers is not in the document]
tobiasjj: `start` and `dt` are now cast to int in `generate_timestamps()`.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +211,268 @@ def make_slice(data):
...
+        samples_per_pixel=samples_per_pixel,
+        line_padding=line_padding,
+        multi_color=multi_color,
+    )
+
+    confocal_class = Scan if scan else Kymo
+    confocal = confocal_class(name, confocal_file, start, stop, metadata)
+
+    if not with_ref:
+        return confocal
```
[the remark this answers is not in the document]
tobiasjj: I separeted the function `generate_confocal()` into `generate_confocal()` and `generate_confocal_with_ref()`, as the old tests still need the original functionality of `generate_confocal()` without returning a reference and I do not want to change a bunch of the old tests.

lumicks/pylake/tests/test_imaging_confocal/test_confocal.py
```diff
@@ -0,0 +1,98 @@
...
+cp1 = np.array([10.0, 1.0, 0.1])
+
+
+@pytest.mark.parametrize(
+    "scan_axes, scan_order, center_point, num_frames",
+    [
+        ([sa0], [0], cp0, 10),
+        ([sa0, sa1], [0, 1], cp1, 20),
+        ([sa1, sa0], [1, 0], cp0, 30),
+        ([sa2, sa0], [1, 0], cp1, 40),
```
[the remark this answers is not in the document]
tobiasjj: :-)

lumicks/pylake/tests/data/mock_confocal.py
tobiasjj: Right, what do you think about `first_physcial_axis_fast`? It's long but descriptive.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +211,344 @@ def make_slice(data):
...
+    -------
+    tuple(np.ndarray, np.ndarray)
+        The the pixel timestamps and the line (kymo) or frame (scan) timestamp ranges.
+    """
+    if start < 0:
+        raise ValueError("`start` needs to be a non negative int")
+    if dt <= 0:
+        raise ValueError("`dt` needs to be a positive int")
+    start = int(start)
+    dt = int(dt)
```
JoepVanlier, line 471: This doesn't protect against roundoff that happens on creation though? If you specify the input argument as a float, then the damage has already been done at this point. It's probably better to just assert on whether they're integer, no?

lumicks/pylake/tests/data/mock_confocal.py
JoepVanlier: sounds good

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +214,347 @@ def make_slice(data):
...
+    confocal_file,
+    metadata,
+    image,
+    multi_color,
+    axes,
+    start,
+    dt,
+    samples_per_pixel,
+    line_padding,
+):
```
rpauszek, line 364: I wonder, for separation of concerns, should we split from here down in a separate file like `mock_confocal_reference`? Just an idea as, personally, I find really long files difficult to work with, especially when you're constantly scrolling up and down to see the web of calls

lumicks/pylake/tests/data/test_generate_timestamps.py
```diff
@@ -0,0 +1,142 @@
...
+        (1, 2, 3, int(20e9), int(16e9), 6, 14, False, False),  # kymo, even, y_axis_fast
+        (1, 2, 3, int(20e9), int(16e9), 6, 14, False, True),  # kymo, even, x_axis_fast
+        (1, 2, 3, int(20e9), int(16e9), 6, 15, True, False),  # scan, even, 1 frame, y_axis_fast
+        (1, 2, 3, int(20e9), int(16e9), 6, 15, True, True),  # scan, even, 1 frame, x_axis_fast
+        (2, 2, 3, int(20e9), int(16e9), 7, 15, True, True),  # scan, 2 frames, x_axis_fast
+        (2, 2, 3, int(20e9), int(16e9), 7, 15, False, True),  # scan (auto), 2 frames, x_axis_fast
+        (3, 2, 3, int(20e9), int(16e9), 7, 15, True, True),  # scan, 3 frames, x_axis_fast
+        (3, 3, 2, int(20e9), int(16e9), 7, 15, False, True),  # scan (auto), 3 frames, x_axis_fast
+    ],
+)
```
rpauszek, line 37: this is some epic parameterization 🤣

lumicks/pylake/tests/data/test_generate_timestamps.py
```diff
@@ -0,0 +1,142 @@
+import numpy as np
+import pytest
+from lumicks.pylake.tests.data.mock_confocal import generate_timestamps
```
rpauszek, line 3: I might move this file into the actual tests folder. I think it's kinda nice to keep tests and mocking stuff separated. On the other hand, yes it is kinda weird to have tests of our data generation mixed in with actual tests ... thoughts?

Also, as a side note, I think our convention for tests (decided somewhere in the middle, so not entirely consistent) is to use absolute imports for `lumicks.pylake` and relative imports for testing stuff. So just `from .mock_confocal import generate_timestamps`

rpauszek (approved):
LGTM, I see most of the major points were taken care of after the first review. Just left some minor suggestions for clarity and organization; feel free to implement or not as you see fit

lumicks/pylake/tests/data/mock_confocal.py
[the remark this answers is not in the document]
JoepVanlier: Yeah sounds fine.

JoepVanlier (approved):
Love that parametrization :D

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +214,347 @@ def make_slice(data):
...
+
+def generate_kymo_with_ref(
+    name,
+    image,
+    pixel_size_nm=10.0,
+    start=20 * 10**9,
+    dt=625 * 10**8,
+    samples_per_pixel=5,
+    line_padding=3,
+):
```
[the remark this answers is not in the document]
tobiasjj: Unfortunately, we do use `generate_kymo()` and `generate_scan()` quite a lot in different tests. Otherwise, I would have liked to default to always return a reference with the function calls `generate_kymo()` and `generate_scan()`. 

This would also have had the beneficial side effect that the code for the generation of the test kymos/scans and the generation of the reference would be bundled together. The risk of having divergent parameters in how the test data would be created and what would be stored in the reference itself would be minimized. Also, the function `_scan_axes()` would not be needed to synchronize and ensure same parameters upon creation of the test data and reference.

However, splitting the test data generation and reference creation into separate functions and providing wrapper functions that return both the data and the reference, while still leaving the data generation functions marked as public, seemed a viable solution to avoid functions whose return values depend on the optional parameter `with_ref`.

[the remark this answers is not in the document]
tobiasjj: In the long run we probably could replace all calls of `generate_kymo()` and `generate_scan()`?!

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -78,29 +86,35 @@ def split_pixel(x, num_samples):
...
         """Generate a line of photon counts and pad on both sides with padding"""
         return np.hstack(
             (padding, np.hstack([split_pixel(pixel, samples_per_pixel) for pixel in line]), padding)
         )
 
-    if image_data.ndim == 3:
+    if image_data.ndim == 3 + multi_color:
         # Flattens lines of a multi-frame image. This concatenates the lines of different frames.
         image_data = np.hstack([img for img in image_data])
+    image_data = np.expand_dims(image_data, -1) if not multi_color else image_data
```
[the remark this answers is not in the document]
tobiasjj: 👍 I added some comments.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -78,29 +86,35 @@ def split_pixel(x, num_samples):
...
 
-    return np.hstack([generate_infowave_line(line) for line in image_data.T]), np.hstack(
-        [generate_photon_count_line(line) for line in image_data.T]
+    pixels_per_line, number_of_lines = image_data.shape[:2]
+    return (
+        np.tile(generate_infowave_line(pixels_per_line), number_of_lines),
+        tuple(
+            np.hstack([generate_photon_count_line(line) for line in channel.T])
+            for channel in np.moveaxis(image_data, -1, 0)
+        ),
```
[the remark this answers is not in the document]
tobiasjj: 👍

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -186,31 +214,347 @@ def make_slice(data):
...
+        pixel_sizes_nm=pixel_sizes_nm,
+        axes=_scan_axes(image, axes),
+        start=start,
+        dt=dt,
+        samples_per_pixel=samples_per_pixel,
+        line_padding=line_padding,
+    )
+
+
+def _generate_confocal(
```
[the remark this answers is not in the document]
tobiasjj: 👍 Will do that

lumicks/pylake/tests/data/test_generate_timestamps.py
tobiasjj: 👍

tobiasjj: You are right, it would be nice to keep mocking and testing stuff separate. However, this specific test is really testing the mocking itself. What do you think about renaming the file from `test_generate_timestamps.py` to `test_mock_confocal.py` to make clear, that you test mocking stuff?

lumicks/pylake/tests/data/mock_confocal.py
tobiasjj: I would like to keep the generation of scans/kymos and the corresponding reference data together, to emphasize that they should be treated as a "unit" and belong together. As stated above, in the long run, we probably could get rid of the separation of the functions `generate_scan()`/`generate_kymo()` and `generate_scan_with_ref()`/`generate_kymo_with_ref()` and thereby reduce number of lines.

lumicks/pylake/tests/data/test_generate_timestamps.py
tobiasjj: :-)

PR #414, opened 2022-10-28
ci: update workflows

JoepVanlier:
**Why this PR?**
There are newer version of our GH workflows available.

I also enabled pip caching.

Note that I also piggybacked a small fix for a warning we were getting in our tests.


tobiasjj (approved):
LGTM! 🚴

PR #416, opened 2022-11-03
Handle ROI for binding profile histogram

rpauszek:
**Why this PR?**
Currently if tracks are generated using the `rect` argument to `track_greedy` the histogram will be generated still for the entire kymograph. We want to focus just on the tracked ROI. To fix this, added the optional `roi` parameter, which is the same form as `rect` (we're discussing changing `rect` anyway so I went with the more informative `roi` here)

Also, I changed how the y-data is returned. Previously each line was returned as a full probability density (integral == 1), but this doesn't show properly the relative height of peaks between the individual plotted lines. So now we weight each density by the number of data points that went into the estimate. Finally, we now "normalize" the entire set of y-data to the global heighest peak. This way the required offset between lines to plot isn't as strongly dependent on the number of time bins or the data. Since we don't show the units and this is purely a qualitative visualization, it's ok

Before
image

After
image

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -814,15 +814,21 @@ class KymoTrackGroup:
         for bin_index in np.arange(n_time_bins) + 1:
             binned_positions = positions[bin_labels == bin_index][:, np.newaxis]
             try:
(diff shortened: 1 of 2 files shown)


mikhas (commented):
:shipit: works!

JoepVanlier (approved):
Looks good.

PR #417, opened 2022-11-04
Add plotting methods for tracks

rpauszek:
**Why this PR?**
It's nice to have plotting convenience methods 😏 

Since we want to be able to forward formatting arguments to `plt.plot()` we need a way to deal with the aliased ones (you can use `color` or `c` for instance), so I wrote a little utility for that.

Also once I started updating the documentation it became clear that it would actually be easier to re-do almost everything using the new example data we tend to release. So all of that is tacked on here too (except for the diffusion section, as that still needs significant work and a proper dataset). I left it as separate commits because there's quite a bit of changes, but those will all be squashed down before merging.

Docs build [here](https://lumicks-pylake.readthedocs.io/en/plot_tracks/tutorial/kymotracking.html)


mikhas (commented):
Docs look great!

JoepVanlier (changes requested):
Might be worth considering path effects for this.

Note: I did not review the docs yet. Might be nice to split that into its own PR?

docs/tutorial/kymotracking.rst
```diff
@@ -19,72 +19,133 @@ particles moving directed in a particular orientation appear as diagonal tracks
...
+    tracks = lk.track_greedy(kymo40, "green")
+
+The result is a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup`::
+
+    print(type(tracks))  # <class 'lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup'>
+    print(len(tracks))  # 123
 
-The result of tracking is a list of kymograph tracks::
+This is a custom list of :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` objects. We can access the individual tracks by indexing just
+like an other list::
```
tobiasjj, line 67:
```suggestion
like an ordinary list::
```

docs/tutorial/kymotracking.rst
```diff
@@ -19,72 +19,133 @@ particles moving directed in a particular orientation appear as diagonal tracks
...
+    plt.ylim(8.8, 9.8)
+    plt.colorbar(image)
+    plt.show()
+
+.. image:: figures/kymotracking/zoom_track_colorbar.png
+
+We can see the tracks are about 0.3 microns wide and the signal level is around 10-14 counts. We can use this information when tracking
+by setting the `track_width` and `pixel_threshold` parameters. We also see that sometimes a particle momentarily disappears or drops below
+the average signal, due to blinking for instance. To still connect these in a single track, we want to allow for some gaps in the connection step.
+We can set this with the `window` parameter; let's use a `window` size of 6 in this test::
```
tobiasjj, line 122:
```suggestion
We can set this with the `window` parameter; let's use a `window` size of 6 px in this test::
```
?

docs/tutorial/kymotracking.rst
```diff
@@ -557,32 +691,34 @@ Use of the CVE is indicated when the SNR is bigger than 1. For smaller values fo
...
 the bound dwelltime as the length of each track in seconds.
 
-Note: tracks which start in the first frame of the kymograph or end in the last frame are excluded from the analysis. This is because, such tracks have
-ambiguous binding times as the start or end of the track is not known definitively. If these tracks were included in the analysis, this could lead to minor
-biases in the results, especially if the number of tracks that meet this criteriion is large relative to the total number.
-This behavior can be overridden with the keyword argument `exclude_ambiguous_dwells=False`.
+.. note::
+    Tracks which start in the first frame of the kymograph or end in the last frame are excluded from the analysis. This is because, such tracks have
+    ambiguous binding times as the start or end of the track is not known definitively. If these tracks were included in the analysis, this could lead to minor
+    biases in the results, especially if the number of tracks that meet this criteriion is large relative to the total number.
```
tobiasjj, line 700:
```suggestion
    biases in the results, especially if the number of tracks that meet this criterion is large relative to the total number.
```

tobiasjj (commented):
My two cents: Figures look nice, just wondering if we should ensure all of them have proper axis labels (see docs/tutorial/figures/kymotracking/longest_track.png)? Also in the long run, we should probably unify the labels to be "Upper" or "lower" case. Also, I have spotted some typos.

docs/tutorial/kymotracking.rst
```diff
@@ -224,7 +337,7 @@ The second algorithm present is an algorithm that works purely on signal derivat
 the image, and then performing sub-pixel accurate line detection. It can be a bit more robust to low signal levels,
 but is generally less temporally and spatially accurate due to the blurring involved::
 
-    tracks = lk.track_lines(kymo, "green", line_width=0.3, max_lines=50)
+    lines_tracks = lk.track_lines(kymo40, "green", line_width=0.3, max_lines=50)
```
JoepVanlier, line 340: I really wish I hadn't called this algorihtm `track_lines` because `lines_tracks` just triggers me now. Especially with the whole lines / tracks renaming from before 😅 

That said, I don't know a better name in this context.

docs/tutorial/kymotracking.rst
```diff
@@ -310,7 +436,7 @@ computing an offset from a brightness-weighted centroid in a small neighborhood
 
 .. math::
 
-    offset = \frac{1}{m} \sum_{i^2 < w^2} i I(x + i)
+    \mathrm{offset} = \frac{1}{m} \sum_{i^2 < w^2} i I(x + i)
```
JoepVanlier, line 439: Nice :)

docs/tutorial/kymotracking.rst
```diff
@@ -557,32 +683,34 @@ Use of the CVE is indicated when the SNR is bigger than 1. For smaller values fo
...
 
-The lifetime of the bound state(s) can be determined using `KymoTrackGroup.fit_binding_times()`. This method defines
+The lifetime of the bound state(s) can be determined using :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.fit_binding_times()`. This method defines
 the bound dwelltime as the length of each track in seconds.
 
-Note: tracks which start in the first frame of the kymograph or end in the last frame are excluded from the analysis. This is because, such tracks have
-ambiguous binding times as the start or end of the track is not known definitively. If these tracks were included in the analysis, this could lead to minor
-biases in the results, especially if the number of tracks that meet this criteriion is large relative to the total number.
-This behavior can be overridden with the keyword argument `exclude_ambiguous_dwells=False`.
+.. note::
```
JoepVanlier, line 689: 👍

JoepVanlier (changes requested):
The plot function looks great!

I have a few small questions/suggestions about the docs.

Should we warn that if you are dealing with high backgrounds or tracks that are very close together, you _should_ use Gaussian refinement currently? Given that centroid would be biased.

docs/tutorial/kymotracking.rst
```diff
@@ -123,38 +211,55 @@ an acceptable result.
...
+This function is called in a similar manner as the centroid refinement. Since the MLE optimization is significantly slower than the centroid
+method, let's refine just a single long track::
+
+    # track a subsection of the kymo
+    cropped_kymo = kymo40.crop_by_distance(4.2, 5.8)
+    tracks = lk.track_greedy(cropped_kymo, "green", track_width=0.3, window=10, pixel_threshold=9)
+    tracks = lk.filter_tracks(tracks, minimum_length=10)
+
+    # perform the gaussian refinement
+    refined = lk.refine_tracks_gaussian(tracks, window=3, refine_missing_frames=False, overlap_strategy="skip")
```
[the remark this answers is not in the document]
rpauszek: for the first comment, I think the paragraph overall reads better in the order it is now. I also would hope that `"multiple"` isn't used in most cases; ideally the experiment should be performed such that you get well separated tracks (although I get that this can't always be the case). Maybe though we should mention that this is the goal since very congested kymos are just difficult to analyze properly anyway (increased chance of track crossing, merging, etc. which our algorithm can't handle anyway)?

docs/tutorial/kymotracking.rst
rpauszek: `tracked_lines`?

JoepVanlier: yeah sounds fine

docs/tutorial/kymotracking.rst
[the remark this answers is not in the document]
JoepVanlier: In principle true, but one can likely get a better separation between offset and Gaussian tail if you use a bigger window. Allowing overlap would allow you to increase the window size and include a bigger chunk to get a better estimate of the background (even though the peaks would be distinct).

But for now, I'm ok with leaving the order as is, since we haven't done much testing to assess how big an effect this would have on bias/variance.

rpauszek:
Added one final commit, moving the widget section of the tutorial to the `Notebook widgets` tutorial (with cross-referencing in both sections). I think this makes more sense, since all other interactive widgets live here and the widget section was just a small part of this tutorial

JoepVanlier (approved):
LGTM!

PR #421, opened 2022-11-09
ci: use `platformdirs` for jupyter

JoepVanlier:
Adds environment variable to acknowledge that we want to use `platformdirs` to determine which paths to use. Basically they changed how the path lookup is done and added a deprecation warning to notify us about this. See https://github.com/jupyter/jupyter_core/pull/292 for more information.


tobiasjj (approved):
LGTM! 🚴

PR #422, opened 2022-11-10
diffusion: add option to compute ensemble cve

JoepVanlier:
**Why this PR?**
This functionality is provided so that we have a tried and tested method of computing a correctly weighted average diffusion.

If we don't, I fear people might just average the constants they get from `estimate_diffusion()` which can lead to suboptimal results.

To test the method, I've done some bootstrapping. Below are the results. The diffusion constant was set to a range of values (according to `(snr * noise)**2 / dt)` to simulate various levels of SNR. The noise level was kept fixed to `0.1`. I then simulate `M` "datasets" of `R` tracks. On every set of tracks I apply the methods. At the end, the ensemble method gives me an estimate and an estimate for the standard error. I aggregate these standard errors for realisations to see if there's a bias in them. I do this by squaring them, adding them and then taking the square root.

The shaded area indicate the `mean` +- `stdev` of the bootstrap. So on the individual results obtained with the method. The lines indicate the `aggregated mean` +- `aggregated stdev` returned by the method. Ideally, the lines should extent exactly to the shaded area. If they don't, then the estimate is bad.

CVE_250reps_equal
*Figure 1: 250 replications with all tracks (`R=12`) of equal length (`N=40` points). Weighting doesn't do anything here, but it's just to show the reduced variance that can be obtained.*

CVE_1500reps
*Figure 2: `1500` replicates of a set of tracks where `3` tracks have full length (`N=40`) and `9` tracks are much shorter (`N=5`). Here we would expect weighting of the results to kick in, giving more weight to the long tracks.*

image
*Figure 3: `1500` replicates of a set of tracks where `2` tracks have full length (`N=40`) and `8 tracks are much shorter (`N=6`). Here we would expect weighting of the results to really kick in, giving more weight to the long tracks. Here we can see naive averaging compared to optimal averaging. You can see that naive averaging leads to barely any variance reduction since the estimates are dominated by the short tracks. What you can also see is that there is some bias in the variance estimate for cases where the variance is large. This seems to only occur when there is a large difference between the track lengths.*

image
*Figure 4: Close up of the most relevant area*

**Why are there no docs?**
Documentation is pending. I plan to rewrite the diffusion part of the documentation, and it will include some plots like the one above.

Small note:
I know that in the function `ensemble_diffusion` I have a function argument with literally one option, but I don't want to omit it in case we release before the `ols` variant makes it in. Users should make the decision of which method to use explicitly, and therefore I think it should not be defaulted (otherwise we'd immediately introduce a breaking change).

### DIFF
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -971,3 +971,48 @@ def estimate_diffusion_cve(
         unit,
         unit_label,
     )
+
+
+def ensemble_cve(kymotracks):
+    """Calculate ensemble-based CVE.
+
+    Determines the weighted average of the mean and standard error of the CVE estimates.
+
+    See docstring for _cve for more information on the estimator itself. The averaging is handled
+    through equation 57 and 58 from [14]_.
+
+    Parameters
+    ----------
+    kymotracks : lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup
+        Group of kymotracks
+
+    References
+    ----------
+    .. [14] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+            diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+            022726.
+    """
+    cve_based = kymotracks.estimate_diffusion(method="cve")
+
+    if len(cve_based) == 1:
+        return cve_based[0]
+
+    estimates = np.array([c.value for c in cve_based])
+    counts = np.array([c.num_points for c in cve_based])
+    counts_sum = np.sum(counts)
+
+    # Equation 58 and 57 from Vestergaard et al [14].
+    ensemble_mean = np.sum(estimates * counts) / counts_sum
+    ensemble_mean_var = np.sum(counts * (estimates - ensemble_mean) ** 2) / (
+        (counts.size - 1) * counts_sum
+    )
+
+    return DiffusionEstimate(
+        value=ensemble_mean,
+        std_err=np.sqrt(ensemble_mean_var),
+        num_lags=np.nan,
+        num_points=counts_sum,
+        method="ensemble cve",
+        unit=cve_based[0].unit,
+        _unit_label=cve_based[0].unit,
+    )
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1,8 +1,6 @@
 from copy import copy
-import warnings
 from deprecated.sphinx import deprecated
 from sklearn.neighbors import KernelDensity
-from ..detail.utilities import use_docstring_from
 from .detail.msd_estimation import *
 from .detail.localization_models import LocalizationModel
 from .. import __version__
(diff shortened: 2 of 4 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -868,6 +866,31 @@ def estimate_diffusion(self, method, *args, min_length=None, **kwargs):
...
+              :meth:`KymoTrack.estimate_diffusion` for more detailed information and references.
+
+        References
+        ----------
+        .. [7] Vestergaard, C. L., Blainey, P. C., & Flyvbjerg, H. (2014). Optimal estimation of
+               diffusion coefficients from single-particle trajectories. Physical Review E, 89(2),
+               022726.
+        """
+        if method == "cve":
+            return ensemble_cve(self)
```
rpauszek, line 892: maybe `else` throw? seems overkill maybe but I could imagine someone having a typo or using `"CVE"` in which case this will just silently do nothing right?

rpauszek (approved):
looks good, and the bootstrapping results are very convincing

just one minor suggestion about checking the `method` string

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -971,3 +971,48 @@ def estimate_diffusion_cve(
...
+    if len(cve_based) == 1:
+        return cve_based[0]
+
+    estimates = np.array([c.value for c in cve_based])
+    counts = np.array([c.num_points for c in cve_based])
+    counts_sum = np.sum(counts)
+
+    # Equation 58 and 57 from Vestergaard et al [14].
+    ensemble_mean = np.sum(estimates * counts) / counts_sum
+    ensemble_sd = np.sum(counts * (estimates - ensemble_mean) ** 2) / (
```
tobiasjj, line 1006: Hm, does `ensemble_sd` stand for ensemble _standard deviation_? Isn't it more like a variance of the means? I would probably name it `ensemble_mean_var`, or similar.

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -861,3 +861,38 @@ def test_ensemble_api(blank_kymo):
...
+    single_group_msd = KymoTrackGroup([kymotracks._src[0]]).ensemble_diffusion("cve")
+    single_track_msd = kymotracks._src[0].estimate_diffusion("cve")
+    np.testing.assert_allclose(single_group_msd.value, single_track_msd.value)
+    np.testing.assert_allclose(single_group_msd.std_err, single_track_msd.std_err)
+
+
+def test_invalid_ensemble_diffusion(blank_kymo):
+    """Tests whether we can call this function at the diffusion level"""
+    kymotracks = KymoTrackGroup([KymoTrack([], [], blank_kymo, "red")])
+    with pytest.raises(ValueError, match=re.escape("Invalid method (potato) selected")):
```
tobiasjj, line 897: Do you test us reviewing? I would like to see `egg` instead.

tobiasjj (approved):
This PR turned out to be much shorter than I anticipated! Nice work. I have one nitpick about the naming of `ensemble_sd`. Otherwise, LGTM 🚴 

And one addition: I really like the bootsrapping figures (and results)!

lumicks/pylake/kymotracker/tests/test_kymotrack.py
JoepVanlier: egg it is :D

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: Good catch! The name is indeed bad. Thanks!

PR #426, opened 2022-11-22
Dwell-times from Gaussian Mixture Models

rpauszek:
**Why this PR?**
We have two main functionalities currently in the `population` module: Gaussian mixture models and Dwell-time analysis. This feature allows for the results from the first to be used as input to the second. Specifically, channel data can be fit with the GMM and a labeled state-path extracted. This state-path is converted into lists of dwell-times for each state and these can then be fit with the exponential distribution.

While I was working on the tutorial update I noticed the need for this feature. So there are no docs updates here; a second PR with those changes will follow shortly

It might also be useful to have some classmethod on `DwelltimeModel` to read in state-paths directly

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 
 * Added `KymoTrack.plot()` and `KymoTrackGroup.plot()` methods to conveniently plot the coordinates of tracked lines. See the [kymotracking documentation](https://lumicks-pylake.readthedocs.io/en/latest/tutorial/kymotracking.html) for more details.
 * Added `KymoTrackGroup.ensemble_diffusion()` for estimating an average diffusion for a collection of tracks.
+* Added the `lk.GaussianMixtureModel.extract_dwell_times()` method to calculate dwell-times for all states from channel data.
 
 ## v0.13.2 | 2022-11-15
 
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -480,3 +480,55 @@ def _exponential_mle_optimize(
 
     # output parameters as [amplitudes, lifetimes], -log_likelihood
     return result.x, -result.fun
+
+
+def _dwellcounts_from_statepath(statepath, exclude_ambiguous_dwells=True):
+    """Calculate the dwell counts and slicing indices for all states in a state path trajectory.
+
+    Note: the counts are the number of frames or time points. To convert to proper
+    dwelltimes, multiply the counts by the time step between points.
+
+    Parameters
+    ----------
+    statepath : numpy.typing.ArrayLike
+        Time-ordered array of state labels
+    exclude_ambiguous_dwells : bool
+            Determines whether to exclude dwelltimes which are not exactly determined. If `True`, tracks which
+            start in the first frame or end in the last frame of the kymograph are not used in the analysis,
+            since the exact start/stop times of the binding event are not definitively known.
+
+    Returns
+    -------
+    dict:
+        Dictionary of all dwell counts for each state. Keys are state labels.
+    dict:
+        Dictionary of slicing indices for all dwells for each state. Keys are state labels.
+    """
+    unique_states = np.unique(statepath)
+    # pad with extra state to catch first and last dwells
+    # this also effectively shifts our indexing by one so we
+    # keep the exclusive last index
+    assert np.all(np.isfinite(statepath))
+    padded_statepath = np.hstack((np.nan, statepath, np.nan))
+
+    # store list slicing indices for each dwell in a state
+    state_ranges = {}
+    for state in unique_states:
+        mask = np.array(padded_statepath == state).astype(int)
+        diff_mask = np.diff(mask)
+
+        # find slicing indices as [start:stop) pairs
+        # as rows in array
+        idx = np.argwhere(diff_mask != 0).squeeze()
+        idx = idx.reshape((-1, 2))
+
+        if exclude_ambiguous_dwells:
+            start = 1 if idx[0, 0] == 0 else 0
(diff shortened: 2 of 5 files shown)


lumicks/pylake/population/dwelltime.py
```diff
@@ -480,3 +480,55 @@ def _exponential_mle_optimize(
...
+    dwelltimes, multiply the counts by the time step between points.
+
+    Parameters
+    ----------
+    statepath : numpy.typing.ArrayLike
+        Time-ordered array of state labels
+    exclude_ambiguous_dwells : bool
+            Determines whether to exclude dwelltimes which are not exactly determined. If `True`, tracks which
+            start in the first frame or end in the last frame of the kymograph are not used in the analysis,
+            since the exact start/stop times of the binding event are not definitively known.
```
JoepVanlier, line 498: Copy/paste? Kymograph doesn't seem relevant here? Also, the indent is not correct.

lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -62,3 +63,39 @@ def test_bootstrap(exponential_data):
     np.testing.assert_allclose(mean, 0.4642469883372174, rtol=1e-5)
     np.testing.assert_allclose(ci, (0.3647038711684928, 0.5979550940729152), rtol=1e-5)
     np.random.seed()
+
+
+def test_dwellcounts_from_statepath():
```
JoepVanlier, line 68: Include a test for an empty state path?

lumicks/pylake/population/mixture.py
```diff
@@ -121,6 +122,26 @@ def label(self, trace):
...
+        Parameters
+        ----------
+        statepath : numpy.typing.ArrayLike
+            Time-ordered array of state labels. This can be obtained using the `GaussianMixtureModel.label()` method.
+        dt_seconds :
+            Time between data points in seconds.
+        exclude_ambiguous_dwells : bool
+            Determines whether to exclude dwelltimes which are not exactly determined. If `True`, tracks which
+            start in the first frame or end in the last frame of the kymograph are not used in the analysis,
+            since the exact start/stop times of the binding event are not definitively known.
```
JoepVanlier, line 137: Copy/paste? Kymograph is mentioned again.

lumicks/pylake/population/mixture.py
```diff
@@ -121,6 +122,26 @@ def label(self, trace):
         output_states = np.argsort(self._map)  # output model state labels in wrapped model order
         return output_states[labels]  # output model labels
 
+    def dwell_times(self, statepath, dt_seconds, exclude_ambiguous_dwells):
```
JoepVanlier, line 125: Why is this a method? You don't use anything contained in `self`.

The fact that it doesn't, makes me wonder about the API though. Why not take a slice, then you can label it for the user, extract the time step etc?

Also, would it make sense to make `exclude_ambiguous_dwells` a keyword argument?

changelog.md
```diff
@@ -1,5 +1,10 @@
 # Changelog
 
+## v0.13.3 | tbd
+
+#### New features
+* Added the `lk.GaussianMixtureModel.dwell_times()` method to calculate dwell-times for all states in a time-ordered state trajectory.
```
JoepVanlier, line 6:
```suggestion
#### New features

* Added the `lk.GaussianMixtureModel.dwell_times()` method to calculate dwell-times for all states in a time-ordered state trajectory.
```

JoepVanlier (changes requested):
Nice PR! Especially like the little private method for calculating the dwell times.

Not so sure about the API on the MixtureModel though.

lumicks/pylake/population/dwelltime.py
```diff
@@ -480,3 +480,55 @@ def _exponential_mle_optimize(
...
+
+def _dwellcounts_from_statepath(statepath, exclude_ambiguous_dwells=True):
+    """Calculate the dwell counts and slicing indices for all states in a state path trajectory.
+
+    Note: the counts are the number of frames or time points. To convert to proper
+    dwelltimes, multiply the counts by the time step between points.
+
+    Parameters
+    ----------
+    statepath : numpy.typing.ArrayLike
```
JoepVanlier, line 493: I think here just using `array_like` is fine? The `numpy.typing` stuff was for the type annotations.

lumicks/pylake/population/mixture.py
```diff
@@ -121,6 +122,26 @@ def label(self, trace):
...
+        """Calculate lists of dwelltimes for each state in a time-ordered statepath array.
+
+        Parameters
+        ----------
+        statepath : numpy.typing.ArrayLike
```
JoepVanlier, line 130:
```suggestion
        statepath : array_like
```

lumicks/pylake/population/mixture.py
rpauszek: yeah you're right. I've changed it now and I'll add a section in the docs for how to use this with `DwelltimeModel` in the next PR

lumicks/pylake/population/mixture.py
```diff
@@ -121,6 +122,34 @@ def label(self, trace):
...
+
+        Returns
+        -------
+        dict:
+            Dictionary of all dwell times (in seconds) for each state. Keys are state labels.
+        dict:
+            Dictionary of slicing indices for all dwell ranges for each state. Keys are state labels.
+        """
+        statepath = self.label(trace)
+        dt_seconds = 1 / trace.sample_rate
```
JoepVanlier, line 145: nitpick: `1.0 / trace.sample_rate`

JoepVanlier (approved):
Few minor suggestions but other than that it looks good to me.

PR #427, opened 2022-11-24
docs: update fdcurves.rst

aafkevandenberg:
I added figures and updated the text for fdcurves.rst

The docs can be found here: https://lumicks-pylake.readthedocs.io/en/update_fdcurves2/tutorial/fdcurves.html

### DIFF
--- a/docs/tutorial/fdcurves.rst
+++ b/docs/tutorial/fdcurves.rst
@@ -5,37 +5,46 @@ FD curves
 
     :nbexport:`Download this page as a Jupyter notebook <self>`
 
-The following code loads an HDF5 file and lists all of the FD curves inside of it::
+Load an HDF5 file and list all FD curves inside the file::
 
-    import lumicks.pylake as lk
+    >>> file = lk.File("20200430-192424 FD Curve FD_5_control_forw.h5")
+    >>> list(file.fdcurves)
+    ['FD_5_control_forw']
 
-    file = lk.File("example.h5")
-    list(file.fdcurves)  # e.g. shows: "['baseline', '1', '2']"
 
 To visualizes an FD curve, you can use the built-in :meth:`.plot_scatter()
 <lumicks.pylake.fdcurve.FdCurve.plot_scatter()>` method::
 
-    # Pick a single FD curve
-    fd = file.fdcurves["baseline"]
+    fd = file.fdcurves["FD_5_control_forw"]
     fd.plot_scatter()
 
+.. image:: figures/fdcurves/fdcurves_scatter.png
+
 Here, :attr:`.fdcurves <lumicks.pylake.File.fdcurves>` is a standard Python dictionary, so we can
-do standard `dict` thing with it. For example, we can iterate over all the FD curve in a file and plot them::
+do all the things you can do with a regular dictionary. For example, we can iterate over all the FD curves in a file and plot them::
 
     for name, fd in file.fdcurves.items():
         fd.plot_scatter()
-        plt.savefig(name)
+
+.. image:: figures/fdcurves/fdcurves_scatter.png
 
 By default, the FD channel pair is `downsampled_force2` and `distance1`.
 This assumes that the force extension was done by moving trap 1, which is the most common.
 In that situation the force measured by trap 2 is more precise because that trap is static.
 The channels can be switched with the following code::
 
-    alt_fd = fd.with_channels(force='1x', distance='2')
+    alt_fd = fd.with_channels(force='1x', distance='1')
     alt_fd.plot_scatter()
 
-    # or as quick one-liner for plotting
-    fd.with_channels(force='2y', distance='2').plot_scatter()
+.. image:: figures/fdcurves/fdcurves_scatter_f1x.png
+
+or as quick one-liner for plotting::
+
+    fd.with_channels(force='1x', distance='1').plot_scatter()
+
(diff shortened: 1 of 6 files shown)


docs/tutorial/fdcurves.rst
```diff
@@ -9,33 +9,48 @@ The following code loads an HDF5 file and lists all of the FD curves inside of i
 
     import lumicks.pylake as lk
 
-    file = lk.File("example.h5")
-    list(file.fdcurves)  # e.g. shows: "['baseline', '1', '2']"
+    file = lk.File("20200430-192424 FD Curve FD_5_control_forw.h5")
+
+List all the FD curves inside the file::
+
+    >>>list(file.fdcurves)
```
JoepVanlier, line 16: Shouldn't there be a space between `>>>` and the command?

```suggestion
    >>> list(file.fdcurves)
```

docs/tutorial/fdcurves.rst
```diff
@@ -9,33 +9,48 @@ The following code loads an HDF5 file and lists all of the FD curves inside of i
...
 
-    # Pick a single FD curve
-    fd = file.fdcurves["baseline"]
+    fd = file.fdcurves["FD_5_control_forw"]
     fd.plot_scatter()
 
+.. image:: figures/fdcurves/fdcurves_scatter.png
+
 Here, :attr:`.fdcurves <lumicks.pylake.File.fdcurves>` is a standard Python dictionary, so we can
 do standard `dict` thing with it. For example, we can iterate over all the FD curve in a file and plot them::
```
JoepVanlier, line 29:
```suggestion
Here, :attr:`.fdcurves <lumicks.pylake.File.fdcurves>` is a standard Python dictionary, so we can
do all the things you do with a regular dictionary with it. For example, we can iterate over all the FD curve in a file and plot them::
```

docs/tutorial/fdcurves.rst
```diff
@@ -9,33 +9,48 @@ The following code loads an HDF5 file and lists all of the FD curves inside of i
...
 
     for name, fd in file.fdcurves.items():
         fd.plot_scatter()
-        plt.savefig(name)
+        plt.savefig(name + ".png")
```
JoepVanlier, line 33: Shouldn't we mention in the text that we are saving it to a file here?

Personally, I wonder if this saving even adds anything. Might be fine to just omit it.

docs/tutorial/fdcurves.rst
```diff
@@ -47,17 +62,37 @@ The raw data can be accessed as well::
...
+`fd_ensemble = lk.FdEnsemble(ensemble_file.fdcurves)`. If the fd curves are in different files, the ensemble can be defined as follows::
+
+    ensemble_file1 = lk.File("20210728-115145 FD Curve FD hairpin forward.h5")
+    ensemble_file1.fdcurves
 
-    fd_ensemble = lk.FdEnsemble(file.fdcurves)
+::
+
+    ensemble_file2 = lk.File('20210728-115145 FD Curve FD hairpin back.h5')
+    ensemble_file2.fdcurves
```
JoepVanlier, line 95: I don't see the part where you merge them?

docs/tutorial/fdcurves.rst
```diff
@@ -72,6 +107,12 @@ and distance from such an ensemble using::
...
     f = fd_ensemble.f
     d = fd_ensemble.d
+    plt.scatter(d, f, s=1)
+    plt.ylabel("Force (pN)")
+    plt.xlabel("Distance $\mu$m")
+    plt.title("Two aligned fd curves")
+
+.. image:: figures/fdcurves/fdcurves_aligned.png
 
 Baseline Correction
```
JoepVanlier, line 117: You didn't change this, but I noticed that the indent below is a bit strange. Normal convention is two spaces after the code, and one after the hash.

```python
    file = lk.File("example.h5")
    fd = file.fdcurves["baseline"]  # low frequency, uncorrected force magnitude
    fd_bl = fd.with_baseline_corrected_x()  # low frequency, baseline corrected force x-component
```

Also, it might be nice to move the note _after_ the baseline correction is explained? Otherwise you start reading it before even knowing about baseline correction.

JoepVanlier (changes requested):
Nice work!

I am missing the part where you merge fd curves from two files into one ensemble though? It's mentioned in the text, but I don't see it happening in the code block.

docs/tutorial/fdcurves.rst
aafkevandenberg: Indeed forgot to add that part, updated it

docs/tutorial/fdcurves.rst
aafkevandenberg: True, I removed the saving part

docs/tutorial/fdcurves.rst
aafkevandenberg: Good point, included the suggestions

JoepVanlier (approved):
Great work! Thank you!

Sorry for not noticing this last time, but I see `pylake` is imported in this page (line 10). Thing is, all these pages are already prefaced with a standard section which already imports `pylake`, `matplotlib.pyplot` and `numpy`, so that bit could be left out.

The standard section looks like this:
```python
import numpy as np

import lumicks.pylake as lk

%matplotlib inline
```

Other than that, looks great!

PR #430, opened 2022-11-30
diffusion: expose localization variance

JoepVanlier:
**Why this PR?**
The localization variance `sigma**2` is a good quantity to know as it can be used to determine the diffusive SNR (defined as `sqrt(D dt / sigma**2)`).
I noticed in writing the theory section for diffusive analysis that it was missing and I needed it.
Given that we basically do the work anyway, I figured I would add it as a public attribute of the estimate.

I decided to go with variance instead of the standard deviation since: 
a. This is what we actually calculate.
b. Estimate can go negative; and negatives should not be discarded or invalid to avoid biases when calculating population averages).
c. Variances are additive and there is less risk of user error if they get something they can safely average.

Here are some estimates for some real diffusive data (albeit a pretty crowded kymo):
image

For simulations it looks like it's calculating the correct thing.
image

Note that more detailed documentation is part of the docs rewrite (I decided to spin this out first since it was easy to compartmentalize).

### DIFF
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -26,18 +26,22 @@ class DiffusionEstimate:
         Number of lags used to compute this estimate.
     num_points : int
         Number of points used to compute this estimate.
+    localization_variance : float
+        Estimated localization variance.
+
+        Note: It is possible to obtain negative estimates for the localization variance when
+        estimating diffusion based on single tracks.
     method : str
         String identifying which method was used to estimate the parameters.
     unit : str
         Unit that the diffusion constant is specified in.
-    _unit_label : str
-        Unit in TeX format used for plotting labels.
     """
 
     value: float
     std_err: float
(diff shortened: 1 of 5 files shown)


lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -586,17 +590,18 @@ def estimate_diffusion_constant_simple(
...
-        unit,
-        unit_label,
+        value=slope * to_time,
+        std_err=np.sqrt(var_slope) * to_time,
+        num_lags=max_lag,
+        num_points=len(coordinate),
+        localization_variance=intercept / 2.0,
+        method=method,
+        unit=unit,
+        _unit_label=unit_label,
```
tobiasjj, line 604: 👍 appreciate the explicit naming!

tobiasjj (approved):
I think `localization_variance` might be useful. I appreciate the explicit naming of parameters when creating `DiffusionEstimate`s. Couldn't find any issues. LGTM! 🚴

PR #431, opened 2022-12-01
tests: compare functionality of kymo based on `MockConfocalFile` and H5 like `MockDataFile_v2`

tobiasjj:
**Why?**
While rewriting the `test_confocal_imaging` tests I realized that almost all of the old tests used test kymos with a `MockConfocalFile` as data source. There was no specific integration test with a kymo directly using an H5 like `MockDataFile_v2` as data source. An issue caused by the interaction of a kymo with an H5 like data source might stay unnoticed.

Therefore, I implemented a test that compares the behavior of two kymos, one with a `MockConfocalFile` and the other with a `MockDataFile_v2` as data source. Together with the (to be added) unit tests of kymos this should ensure the correct interaction of kymos and H5 like `MockDataFile_v2`s.

This PR is based and a follow up on #413.

### DIFF
--- a/lumicks/pylake/tests/data/mock_confocal.py
+++ b/lumicks/pylake/tests/data/mock_confocal.py
@@ -1,14 +1,17 @@
 from dataclasses import dataclass
 from numbers import Integral
 from typing import List
+from pathlib import Path
 
 import numpy as np
 from lumicks.pylake.channel import Continuous, Slice, empty_slice
 from lumicks.pylake.detail.confocal import ConfocalImage, ScanMetaData
 from lumicks.pylake.detail.image import InfowaveCode
+from lumicks.pylake.file import File
 from lumicks.pylake.kymo import Kymo
 from lumicks.pylake.scan import Scan
 
+from .mock_file import MockDataFile_v2
 from .mock_json import mock_json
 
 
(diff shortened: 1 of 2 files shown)


lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -0,0 +1,96 @@
...
+        start=ref.start,
+        dt=ref.timestamps.dt,
+        force_data=kymo.file.force1x.data,
+        force_start=ref.start,
+        force_channel="1x",
+    ) as file:
+
+        kymo_h5 = file.kymos[name]
+
+        # Test if kymos with different data sources work the same
```
JoepVanlier, line 87: Any reason you picked these attributes to test specifically?

Why not test only the source ones, but then all of them (for example, I'm missing the photon channels)?

For example `get_image()` and `.shape`  are derived from other attributes through code that would need its own dedicated tests for `Kymo`/`Scan`.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -119,6 +155,123 @@ def generate_photon_count_line(line):
     )
 
 
+def confocal_mock_file_h5(
```
JoepVanlier, line 158: I'm curious for which tests you intend to use `confocal_mock_file_h5` versus `generate_kymo_with_ref` vs just generating a simple `Kymo` from an image directly with `_kymo_from_array`.

I imagine the last is still preferred for many downstream tasks (since it is much more performant than actually churning out an `infowave`) unless you are specifically testing properties that explicitly test reconstruction, reshaping or interaction with other channels.

It might make sense to detail the use cases for each of them somewhere so we have some guidelines for when the code is less fresh in all our memories.

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -0,0 +1,96 @@
...
+        force = np.r_[
+            np.ones(first_pixels) * head, np.ones(number_of_samples - first_pixels) * tail
+        ]
+
+        monkeypatch.setattr(
+            kymo.file,
+            channel,
+            Slice(Continuous(force, ref.start, ref.timestamps.dt)),
+            raising=False,
+        )
```
JoepVanlier, line 43: This relies on that you access it over the attribute rather than through `__getitem__`. I guess that if you'd need more, you would sooner go through the full file creation with the `Mock` classes?

lumicks/pylake/tests/test_imaging_confocal/test_kymo.py
```diff
@@ -0,0 +1,96 @@
...
+# value of the property, ensure to clear the `_cache` attribute before and after testing. To achieve
+# both, you can monkeypatch the `_cache` attribute with an empty dict.
+@pytest.fixture(scope="module")
+def test_kymos():
+    return {"standard": generate_kymo_with_ref("standard", np.random.poisson(10, (5, 4, 3)))}
+
+
+@pytest.fixture
+def add_force_channel(monkeypatch):
+    def do_add_force_channel(kymo, ref, head=30, tail=10, head_lines=2, channel="force1x"):
```
JoepVanlier, line 23: Minor nitpick, but I wouldn't be able to guess from `head` and `tail` what they'd do. `early_force`, `late_force`? I also wonder if `attribute` is a better descriptor than `channel` here, since you're not really adding a channel, but specifically that attribute.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -119,6 +155,123 @@ def generate_photon_count_line(line):
...
+    for color, counts in zip(("Red", "Green", "Blue"), (red_counts, green_counts, blue_counts)):
+        mock_manager.make_continuous_channel("Photon count", color, start, dt, counts)
+
+    if force_data is not None:
+        force_start = start if force_start is None else np.int64(force_start)
+        mock_manager.make_continuous_channel(
+            "Force HF", f"Force {force_channel.lower()}", force_start, dt, force_data
+        )
+
+    class MockFileContext(File):
```
JoepVanlier, line 265: Always love me a context manager.

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -119,6 +155,123 @@ def generate_photon_count_line(line):
...
+    start = infowave.start if start is None else np.int64(start)
+    dt = infowave.dt if dt is None else np.int64(dt)
+    dset.attrs["Start time (ns)"] = start
+    dset.attrs["Stop time (ns)"] = start + len(infowave) * dt
+    mock_manager.make_continuous_channel("Info wave", "Info wave", start, dt, infowave)
+    for color, counts in zip(("Red", "Green", "Blue"), (red_counts, green_counts, blue_counts)):
+        mock_manager.make_continuous_channel("Photon count", color, start, dt, counts)
+
+    if force_data is not None:
+        force_start = start if force_start is None else np.int64(force_start)
```
JoepVanlier, line 260: Why cast?

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -119,6 +155,123 @@ def generate_photon_count_line(line):
...
+    """
+    mock_class = MockDataFile_v2
+    group = "Scan" if scan else "Kymograph"
+    mock_manager = mock_class(path / f"{group}_{mock_class.__name__}.h5")
+    mock_manager.write_metadata()
+
+    dset = mock_manager.make_json_data(group, name, json_confocal)
+
+    start = infowave.start if start is None else np.int64(start)
+    dt = infowave.dt if dt is None else np.int64(dt)
```
JoepVanlier, line 252: Why cast?

JoepVanlier (commented):
I don't have that many requests on the code specifically. It looks nice.

Mostly have questions about the intended use cases for the different parts :D

PR #434, opened 2022-12-08
diffusion: Add ensemble OLS method.

JoepVanlier:
**Why this PR?**
This method provides an appropriate solution for estimating a diffusion constant when we have a large number of relatively short tracks with low SNR.

image

image
Graph with the obtained standard error vs number of included points. Shown here is the standard deviation of `5000` runs of `30` tracks with `25` steps each. Dashed line indicates the recommended number of lags suggested by the OLS method.

image
image
image
Graph with the obtained standard error vs number of included points. Shown here is the standard deviation of `100` runs of `20` tracks with `15` steps each. Dashed line indicates the recommended number of lags suggested by the OLS method.

image
Very long tracks but with a very poor SNR (leading to many lags being required). Here `steps=150` and `num_tracks=20`.

What is seen in this images is that most of the time, the optimal number is very close to the actual optimal variance.

Docs are part of a third PR that's still coming up.
If you want to read them, they are here: https://lumicks-pylake.readthedocs.io/en/diffusion_docs/theory/diffusion/diffusion.html

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -461,7 +461,7 @@ class KymoTrack:
                diffusing particles. Physical Review E, 94(2), 022401.
         """
         if method not in ("cve", "gls", "ols"):
-            raise ValueError('Invalid method selected. Method must be "gls" or "ols"')
+            raise ValueError('Invalid method selected. Method must be "cve", "gls" or "ols"')
 
         if not self._kymo.contiguous:
             raise NotImplementedError(
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -947,7 +947,7 @@ class KymoTrackGroup:
 
         return [k.estimate_diffusion(method, *args, **kwargs) for k in filtered_tracks]
 
-    def ensemble_diffusion(self, method):
+    def ensemble_diffusion(self, method, *, max_lag=None):
         """Determine ensemble based diffusion estimates.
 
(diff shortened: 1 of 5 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -461,7 +461,7 @@ def estimate_diffusion(
                diffusing particles. Physical Review E, 94(2), 022401.
         """
         if method not in ("cve", "gls", "ols"):
-            raise ValueError('Invalid method selected. Method must be "gls" or "ols"')
+            raise ValueError('Invalid method selected. Method must be "cve", "gls" or "ols"')
```
rpauszek, line 464: nice catch

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -971,6 +976,8 @@ def ensemble_diffusion(self, method):
         """
         if method == "cve":
             return ensemble_cve(self)
+        elif method == "ols":
+            return ensemble_ols(self, max_lag)
         else:
             raise ValueError(f'Invalid method ({method}) selected. Method must be "cve".')
```
rpauszek, line 982:
```suggestion
            raise ValueError(f'Invalid method ({method}) selected. Method must be "cve" or "ols".')
```

rpauszek (approved):
looks good, just one little fixup

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Good catch! 👍

PR #435, opened 2022-12-09
docs: improve technical documentation for the diffusion quantification methods

JoepVanlier:
Sorry, it's a big one. A comparison of the statistical properties of the various methods we offer was in order. Previously we only had a few guidelines in the kymotracker chapter without any simulations to back them up.

I would also still like to have a go at the diffusion section in the kymotracker section itself. Right now I severely shortened it (which I think is not bad), but it would be nice to have a go at it with some real diffusive data and really tutorialize that section (focusing more on the what to look for and what the APIs are like).

Docs build here: https://lumicks-pylake.readthedocs.io/en/diffusion_docs/theory/diffusion/diffusion.html


docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+This section details some of the available methods and details some pitfalls with each of them.
+In this chapter, we will be making use of simulated data, so that we can study the quality of our estimates while knowing the ground truth.
+Note that the code to perform these simulation is included, so that you can play with them and produce similar graphs which are more tailored to your own experimental conditions.
+For more information on tracking the particles in a kymograph, refer to :doc:`Kymotracking </tutorial/kymotracking>` and specifically :ref:`the section on diffusion <kymotracking-diffusion>` for specifics regarding diffusion analysis on kymographs.
+
+Simulating 1D free diffusion
+----------------------------
+
+The displacements of a particle undergoing free diffusion are uncorrelated over time and distributed according to a normal distribution.
+The standard deviation of this distribution is given by Einstein's equation.
```
aafkevandenberg, line 19: The standard deviation of this distribution follows from the diffusion law, $\sigma = \sqrt{2 \Delta} t$.

The Einstein relation, which relates the diffusion constant to the Boltzman's constant and viscosity, does not give the standard deviation?

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+Pylake supports a number of methods for studying diffusive processes.
+These methods rely on quantifying the movement of the fluorophore.
+This section details some of the available methods and details some pitfalls with each of them.
+In this chapter, we will be making use of simulated data, so that we can study the quality of our estimates while knowing the ground truth.
+Note that the code to perform these simulation is included, so that you can play with them and produce similar graphs which are more tailored to your own experimental conditions.
```
aafkevandenberg, line 12: Awesome that you included the simulation code!

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+Considering that diffusion is a stochastic process, it is important to average over many such increments to obtain a result not subject to a lot of stochastic error.
+The exact details of how to obtain a diffusion constant depend on the method in question and will be discussed below.
+
+Mean Squared Displacements
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Simply averaging the increments will yield an estimate of zero (since the original increment distribution had a mean of zero).
+Instead, a measure of distance travelled is required.
+One easy way to do this is to square these displacements prior to averaging.
+When dealing with pure diffusive motion (a complete absence of drift) in an isotropic medium, the averaged squared displacements follow the following relation:
```
aafkevandenberg, line 56: Suggestion for this paragraph:

Diffusive particles have an average displacement of zero. Therefore distance travelled is not a meaningful quantity when studying diffusive processes. Instead, the mean squared displacement is often used. When dealing with pure diffusive motion (a complete absence of drift) in an isotropic medium, the mean squared displacements follow the following relation:

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
```
aafkevandenberg, line 57: Did you on purpose say 'averaged squared displacements' instead of mean squared displacements?

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+.. math::
+
+    \rho[n] = \langle\left(x_{i} - x_{i + n}\right)^2\rangle = 2 D n \Delta t
+
+Here the brackets :math:`\langle\ldots\rangle` indicate the expected value, :math:`\rho[n]` corresponds to the average squared displacement for lag :math:`n` and :math:`D` is the diffusion constant we seek.
```
aafkevandenberg, line 62: and $\Delta t$ is the time step of one lag.

I know you defined this quantity above, but I always find it annoying to scroll up to find definitions.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+If all tracks experience the same diffusion, the ensemble averaged motion of a collection of independent molecules results in the same diffusion estimates as the time-averaged motion.
+Considering that we may be interested in differences in diffusion constants between different tracks, we use the following time-averaged estimator:
+
+.. math::
+
+    \hat{\rho}[n] = \frac{1}{N - n} \sum_{i=1}^{N-n}\left(x_{i+n} - x_{i}\right)^2
+
+Here :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` is the number of time points in the track, and :math:`x_i` is the track position at time frame :math:`i`.
+This particular estimator is referred to as the Time Averaged MSD (TAMSD), owing its name to the fact that the estimates for each lag are averaged over all time increments.
+Consider the definition of the TAMSD, it can be seen that the displacements are not independent from another.
```
aafkevandenberg, line 75:
```suggestion
Consider the definition of the TAMSD, it can be seen that the displacements are not independent from one another.
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+Instead, a measure of distance travelled is required.
+One easy way to do this is to square these displacements prior to averaging.
+When dealing with pure diffusive motion (a complete absence of drift) in an isotropic medium, the averaged squared displacements follow the following relation:
+
+.. math::
+
+    \rho[n] = \langle\left(x_{i} - x_{i + n}\right)^2\rangle = 2 D n \Delta t
+
+Here the brackets :math:`\langle\ldots\rangle` indicate the expected value, :math:`\rho[n]` corresponds to the average squared displacement for lag :math:`n` and :math:`D` is the diffusion constant we seek.
+This is the reason diffusive motion is often quantified using Mean Squared Displacements (MSD).
```
aafkevandenberg, line 63:
```suggestion
As you can see, the Mean Squared Displacement (MSD) gives a simple relation between the diffusion constant and the observed displacements, which is why MSD is often used for diffusion analysis.
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+There are multiple ways to estimate this quantity in practice.
+We can average over different tracks (ensemble averaging) or over time (time averaging).
+If all tracks experience the same diffusion, the ensemble averaged motion of a collection of independent molecules results in the same diffusion estimates as the time-averaged motion.
```
aafkevandenberg, line 66:
```suggestion
If all tracks experience the same diffusion, the ensemble averaged motion of a collection of independent molecules results in the same diffusion estimates as the time-averaged motion; the diffusive process is 'ergodic'.
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+Considering that we may be interested in differences in diffusion constants between different tracks, we use the following time-averaged estimator:
```
aafkevandenberg, line 67: I would start a new (sub) section here, to separate the general introduction from the TAMSD

aafkevandenberg, line 67:
```suggestion
If the diffusive process is not ergodic, and the goal of the analysis is to find potential differences in diffusion constants between different tracks, the following time-averaged estimator can be used:
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
+.. math::
+
+    \hat{\rho}[n] = \frac{1}{N - n} \sum_{i=1}^{N-n}\left(x_{i+n} - x_{i}\right)^2
+
+Here :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` is the number of time points in the track, and :math:`x_i` is the track position at time frame :math:`i`.
+This particular estimator is referred to as the Time Averaged MSD (TAMSD), owing its name to the fact that the estimates for each lag are averaged over all time increments.
```
aafkevandenberg, line 74:
```suggestion
This particular estimator is referred to as the Time Averaged MSD (TAMSD), because the estimates for each lag are averaged over all time increments.
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
+    plt.xlim([0, 0.8])
+    ax.view_init(20, -60, 0)
+    ax.set_zticks([])
+    ax.set_xlabel(r'MSD [$\mu m^2$]')
+    ax.set_ylabel('Lag time [s]')
+    plt.tight_layout();
+
+.. image:: figures/msd_distributions.png
+
```
aafkevandenberg, line 121: I would also start a new subsection here 'ensemble MSD'

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+This returns a weighted average of the MSDs coming from all the tracks (where the weight is determined by the number of points that contribute to each estimate).
+If the tracks are of equal length, this weighting will not have an effect (since all the weights will be the same).
+You can plot the ensemble msd as follows::
+
+    ensemble_msd.plot()
+
+.. image:: figures/ensemble_msd.png
+
+MSD estimates should be handled with care.
```
aafkevandenberg, line 135: This paragraph contains an important conclusion and warning on how to handle MSD. Maybe give it a heading such as 'Summary', or 'Conclusion' or some other synonym of TL;DR

aafkevandenberg (commented):
Hi Joep, amazing work. I can see that there's a lot of detailed and technical information that has to be conveyed on this page. Indeed a bit like a review paper.

I did not have time to go through everything, but already added some suggestions.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
+Note that Pylake gives you both an estimate for the diffusion constant, as well as its expected uncertainty and the number of lags used in the computation.
+The uncertainty estimate in this case is based on Equation A1b in :cite:`bullerjahn2020optimal`.
+
+Generalized Least Squares
+^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Another option is to use generalized least squares (GLS).
+Based on the definition of the MSD estimator, it is possible to derive the structure of the covariance matrix :cite:`bullerjahn2020optimal`.
+This covariance matrix can then be used in the estimation procedure to determine the diffusion constant.
```
aafkevandenberg, line 183: I am missing the reasoning as to why this method is useful. What about:

As mentioned above, subsequent data points in an MSD curve are highly correlated. One can account for these correlations by computing the covariance matrix of the MSD values :cite:`bullerjahn2020optimal`. This covariance matrix can then be used in the estimation procedure to determine the diffusion constant.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+
```
aafkevandenberg, line 184: How does this method compare to the one above? Does the method from michalet account for correlations? and do I understand correctly that the GLS method should be combined with a selection of the optimal number of lags?

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,667 @@
...
+Here :math:`\sigma` is the static localization accuracy, :math:`R` is a motion blur constant and :math:`\Delta t` represents the time step.
+With pure diffusive motion (a complete absence of drift) in an isotropic medium, 1-dimensional MSDs can be fitted by the following relation:
+
+.. math::
+
+    \rho[n] = 2 D n \Delta t + offset
+
+Here :math:`D` is the diffusion constant in :math:`um^2/s`, :math:`\Delta t` is the time step, :math:`n` is the step index and the offset is determined by the localization accuracy.
+
+While it may be tempting to use a large number of lags in the diffusion estimation procedure, this actually produces poor estimates of the diffusion constant :cite:`qian1991single,michalet2010mean,michalet2012optimal`.
```
aafkevandenberg, line 160:
```suggestion
While it may be tempting to use a large number of lags in the diffusion estimation procedure, this actually produces poor estimates of the diffusion constant :cite:`qian1991single,michalet2010mean,michalet2012optimal`, because, as mentioned above, the error in the MSD value increases with lag time.

```

docs/theory/diffusion/diffusion.rst
aafkevandenberg: And for OLS you mention the blur constant and localization accuracy, but you don't mention this for the other methods.

aafkevandenberg (commented):
More comments

docs/theory/diffusion/diffusion.rst
aafkevandenberg: The last sentence is optional. I thought it would be nice to list all the experimental artefacts that are accounted for by the various methods in on place.

docs/theory/diffusion/diffusion.rst
JoepVanlier: Idea was to have fully runnable examples. I hope to do the same one day for force calibration. I hope it might empower those who are comfortable with Python to do numerical experiments for their own specific situations, thereby having more eyes on the code and maybe identify additional corner cases we did not foresee.

docs/theory/diffusion/diffusion.rst
JoepVanlier: Oh whoops. I meant to write follows from Einstein's equations for Brownian motion. I was referring to Eq 11 in this: http://users.physik.fu-berlin.de/~kleinert/files/eins_brownian.pdf

I think your equation is missing a `D`? 😅 

Would you be ok with the following?:

The standard deviation of this distribution follows from the theory of Brownian motion [cite]: $\sqrt{2 D \Delta t}$?

Or even just:

The standard deviation of this distribution follows from the diffusion equation: $\sqrt{2 D \Delta t}$?

I guess it's established enough that it doesn't require a specific citation. I'm also not sure it's usually referred to as a law?

docs/theory/diffusion/diffusion.rst
JoepVanlier: But distance travelled *is* a meaningful quantity. A distance is by definition is always positive.

How about:
```
Diffusive particles have an average displacement of zero.
Rather than calculating an average displacement, it makes more sense to consider a form of distance travelled.
In practice, the mean squared displacement is often used.
When dealing with pure diffusive motion (a complete absence of drift) in an isotropic medium, the mean squared displacements follow the following relation:
```

docs/theory/diffusion/diffusion.rst
aafkevandenberg: ah yes, Einstein wrote many equations 😅 
I prefer calling it 'diffusion equation'. A citation is indeed not necessary.

docs/theory/diffusion/diffusion.rst
JoepVanlier: Fair enough. 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: I like your sentence better, but shouldn't there be a `the` before MSD?

As you can see, the Mean Squared Displacement (MSD) gives a simple relation between the diffusion constant and the observed displacements, which is why *the* MSD is often used for diffusion analysis.

Or maybe even:

As you can see, the Mean Squared Displacement (MSD) gives a simple relation between the diffusion constant and the observed displacements, which is why MSD analysis is often used for diffusion analysis.

docs/theory/diffusion/diffusion.rst
JoepVanlier: I was deliberately avoiding that term to keep it light. But if you think it's ok, I'll add it. I like it 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: 👍

JoepVanlier: I think the inclusion of ergodicity can lead to confusion here.

Pure diffusion as described in this chapter is ergodic.

You can have sub-populations with different diffusion constants, but that doesn't change that the diffusion itself is ergodic. If you were able to isolate the two distributions and average a _lot_ of data, you'd see that EAMSD gives the same result as EATAMSD. This is something that is true for pure diffusion and actually something they test for in some types of anomalous diffusion (whether there is ergodicity breaking).

I do like the other changes to the sentence though. I would suggest going with the following instead if you're also ok with that:

`If the goal of the analysis is to find potential differences in diffusion constants between different tracks, the following time-averaged estimator can be used:`

docs/theory/diffusion/diffusion.rst
JoepVanlier: haha, yes, much better

docs/theory/diffusion/diffusion.rst
JoepVanlier: That he did, that he did :D

I will amend it!

docs/theory/diffusion/diffusion.rst
JoepVanlier: Good idea. Fixing...

docs/theory/diffusion/diffusion.rst
JoepVanlier: That's a great idea suggestion!

Reading it back, I wonder if `large lags` is descriptive enough since what is large really depends on your track length.

What do you think of something like the following?

```
Important take-aways
^^^^^^^^^^^^^^^^^^^^

MSD estimates should be handled with care.

#. For a densely sampled track, the number of points contributing to a time-averaged MSD estimate is inversely linearly related to the lag time.
#. Poorly averaged MSDs will show a high variance and should not be assumed to be Gaussian distributed.
#. MSDs for different lags are highly correlated and one should not fit these values as though they were independent data points unless care is taken that it is valid to do so (more on this below).

In addition, there are other experimental parameters, such as motion blur and localization accuracy that have to be accounted for.
In the next sections, we will discuss various methods for diffusion analysis that take these variables into account.
```

docs/theory/diffusion/diffusion.rst
JoepVanlier: Yeah, that sounds a lot better.

In practice, the performance of that method was somewhat disappointing to me. It seems that in most cases, it doesn't perform better than OLS if you just stick to the optimal number of points suggested by Michalet. But it takes a lot longer to compute.

If you include more points than considered optimal, then GLS will perform better than doing OLS with too many points, but I'm not entirely sure how often that is useful. Especially considering the additional constraints GLS includes (tracks may not have gaps).

I deliberately kept this section somewhat short to emphasize the other two methods.

docs/theory/diffusion/diffusion.rst
JoepVanlier: > And for OLS you mention the blur constant and localization accuracy, but you don't mention this for the other methods.

For `GLS` things are the same as `OLS` and we don't support supplying the motion blur manually. It all gets subsumed in that offset. `GLS` is identical to `OLS` in every way except for the fact that we take into account the covariance matrix and don't have a clear procedure for estimating the number of lags to include.

But good point, for `CVE`! I should mention it specifically. I have added the following:

Here $D$ represents the diffusion constant, $\sigma$ is the localization uncertainty standard deviation and $R$ is the motion blur constant. In the current implementation, we assume the motion blur to be negligible for confocal scans (see note below).

In the note (to try and not distract from the main text too much), I show the reasoning for having a motion blur of zero.

```
    The motion blur coefficient :math:`R` is a value between `0` and `1/4` given by

    .. math::

        R = \frac{1}{\Delta t} \int_{0}^{\Delta t}S(t) \left(1 - S(t)\right)dt

    with

    .. math::

        S(t) = \int_{0}^{t} c(t') dt'

    Here, the aperture function is defined as :math:`c(t)`, where :math:`c(t)` represents the fraction of illumination happening over time.
    :math:`c(t)` is normalized w.r.t. area under the curve.
    For a rectangular shutter or exposure window, one obtains :math:`R = 1/6 \frac{t_{exposure}}{line\_time}`.
    In the current implementation we assume the motion blur constant to be negligible (zero) for confocal acquisition, since the time spent over the particle's PSF compared to the line time is very low.

    When estimating both localization uncertainty and the diffusion constant, the motion blur factor has no effect on the estimate of the diffusion constant itself, but it does affect the calculated uncertainties.
    In the case of a provided localization uncertainty, it does impact the estimate of the diffusion constant itself.
```

I have considered providing a ballpark estimate for the motion blur coefficient based on the pixel size, shape of the scan volume and line time; but I feel that should be its own task (since then one might also want to consider the shape of the PSF). All in all, it requires some additional investigation.

We could consider exposing `R` as an optional parameter for the `CVE` method with a default of zero for now, but that would make things harder once we get camera based kymographs that require a higher `R`. My hope is that one day we can just expose an `particle exposure time` on the `Kymo` that will hook straight into this `R` value here.

For camera-based tracking R becomes much more relevant as it begins to approach `1/6`.

docs/theory/diffusion/diffusion.rst
JoepVanlier: 👍

JoepVanlier:
> More comments

thank you :)

docs/theory/diffusion/diffusion.rst
aafkevandenberg: Looks good!

docs/theory/diffusion/diffusion.rst
aafkevandenberg: I also don't feel strong about it, but mean squared displacements is widely used, so I would just use it from the start.

docs/theory/diffusion/diffusion.rst
aafkevandenberg: Good point, I like your suggestion.

docs/theory/diffusion/diffusion.rst
aafkevandenberg: So, the point you want to make in the first item, is that MSD at longer lag times is averaged over fewer data points? And with 'densely sampled track' you mean that there are no gaps and that with gaps, the relationship would not be linear? Is it important to imphasize that? How about:

For larger lag times, MSD values are averaged over fewer lags. This means that the uncertainty of the MSD estimate increases with lag time.

aafkevandenberg: Otherwise I like your suggestion!

docs/theory/diffusion/diffusion.rst
aafkevandenberg: I see, indeed a good idea to keep it short. And I see you also advice OLS in the concluding remarks. Looks good like this.

docs/theory/diffusion/diffusion.rst
aafkevandenberg: Thanks for adding the note and explanation! If the motion blur is also negligible for the ensemble approach then I guess it's fine to not expose R for now. And to wait for the camera based kymos.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+    plt.hist(results["cve"][-1], 30, label="cve");
+    plt.hist(results["ols"][-1], 30, label="ols");
+    plt.title(f"SNR = {snrs[-1]}")
+    plt.ylabel('Probability density')
+    plt.xlabel('Diffusion constant [$\mu$m/s]')
+    plt.xlim([-20, 20])
+    plt.tight_layout()
+
+.. image:: figures/histogram_diff.png
+
```
aafkevandenberg, line 389: Detail: In the figure histogram_diff with high SNR, it would be nice to have CVE on top.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+    snrs = 10**np.arange(-1, 1.1, 0.125)
+    results, variances = test_estimation_methods(simulate_tracks, snrs, methods=methods)
+    for method, estimates in results.items():
+        plot_accuracy(snrs, estimates, variance=variances[method], label=method);
+
+    plt.ylim([-5, 6])
+
+.. image:: figures/cve_ensemble.png
+
+Here we simulated `10` tracks with the same `8` steps as in the case which performed so poorly before.
```
aafkevandenberg, line 508: Reading down to this point, I already forgot what `8` steps performed poorly before.. Better to just spell out what you are doing here.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+.. image:: figures/cve_carryover.png
+
+As expected, the uncertainty lies somewhere between the ensemble estimate and the individual track estimate.
+Note however that we would be getting this precision on a per track basis.
+This means that we could still use this analysis to obtain a distribution of diffusion estimates and see if there are subgroups with different diffusion constants.
+
+MSD-based methods
+^^^^^^^^^^^^^^^^^
+
+For MSD based methods, this same procedure is not optimal.
```
aafkevandenberg, line 545: I'd either remove this sentence, or be more explicit what 'this procedure' is.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+    for method in ("cve ensemble", "ols ensemble", "cve ensemble (filtered)", "ols ensemble (filtered)"):
+        plot_accuracy(snrs, results[method], variance=variances[method], label=method);
+    plt.ylim([-1, 3])
+
+.. image:: figures/filtering_short_tracks.png
+
+The good news is that the ensemble estimates are pretty robust to filtering short tracks.
+This is expected since these estimates already incorporate weighting (assigning lower weight to short tracks).
+The bad news is that the variance for low SNR is still very high, while the method with the most precision at low SNR (OLS) is very biased.
+Unfortunately, what this means is that we either need to settle for ensemble estimates, or accept high variance results.
```
aafkevandenberg, line 703: Really nice that you consider all these different scenarios and illustrate what you have to think about when doing diffusion analysis!

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+
+Concluding remarks
+------------------
+When the SNR and the number of points per track is high, all methods presented here perform well.
+In this case, the recommendation is to use CVE, while carrying over the localization uncertainty from the ensemble estimate.
+Choosing this option will allow the study of diffusion constants for individual tracks with the largest precision.
```
aafkevandenberg, line 709: Now imagine a common scenario where a protein is highly diffusive (high SNR) under one condition, and immobile (low SNR) under another condition. What would you advice? It feels awkward to use different methods for the same protein. Maybe using ensemble OLS for both?

aafkevandenberg (approved):
Some more feedback. I did not have time yet to run the code...

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+
+    When estimating both localization uncertainty and the diffusion constant, the motion blur factor has no effect on the estimate of the diffusion constant itself, but it does affect the calculated uncertainties.
+    In the case of a provided localization uncertainty, it does impact the estimate of the diffusion constant itself.
+
+The performance of these covariance-based estimators can be characterized by its signal to noise ratio (SNR).
+This SNR is defined by:
+
+.. math::
+
+    \mathrm{SNR} = \frac{\sqrt{D \Delta t}}{\sigma}
```
aafkevandenberg, line 266: Maybe the Pylake docs is not the place for this. But it would be great to have a guideline somewhere of what $\sigma$ is for typical experiments. And is $\Delta t$ a self-chosen time step? That way users can compute the SNR for their own experiments.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+Unfortunately, what this means is that we either need to settle for ensemble estimates, or accept high variance results.
+
+Concluding remarks
+------------------
+When the SNR and the number of points per track is high, all methods presented here perform well.
+In this case, the recommendation is to use CVE, while carrying over the localization uncertainty from the ensemble estimate.
+Choosing this option will allow the study of diffusion constants for individual tracks with the largest precision.
+
+When the SNR is low, restrict to long tracks if possible (filter short tracks).
+If the diffusion constants are the same for all tracks and the tracks are roughly equally long, use ensemble estimates based on OLS.
```
aafkevandenberg, line 712: Typically tracks are not roughly equally long, they are exponentially distributed. Unless you cut them off of course. Why is this requirement important?

docs/theory/diffusion/diffusion.rst
aafkevandenberg: Actually, since OLS performs better for SNR<1 and all methods perform well for SNR>1, I would tend to advice using OLS?  And only if computation speed is important and SNR>1 shift to CVE? Or am I missing part of the reasoning why CVE is better?

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,713 @@
...
+            histogram_centers[histogram_centers < view_limit],
+            hist[histogram_centers < view_limit],
+            zs=lag_time,
+            zdir='y',
+            alpha=0.8,
+            width=histogram_centers[1]-histogram_centers[0]
+        )
+
+    plt.xlim([0, 0.8])
+    ax.view_init(20, -60, 0)
```
aafkevandenberg, line 118: I got an error here, and according to[ this ](https://matplotlib.org/stable/api/toolkits/mplot3d/view_angles.html) page, matplotlib does not have the 'roll' option for `a.view.init()`. Maybe just remove the zero?

aafkevandenberg (commented):
I just downloaded the Notebook and ran all the simulations successfully. Nice!
The figures are great and are very instructive.

docs/theory/diffusion/diffusion.rst
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: Yes, I will amend it. Thank you! 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: Fair point! Fixing...

docs/theory/diffusion/diffusion.rst
JoepVanlier: Good point! I've changed it to: 
```
For MSD based methods, simply calculating a weighted average of per-track estimates is not optimal.
```

docs/theory/diffusion/diffusion.rst
JoepVanlier: I'm also not sure if we can give general advice other than (SNR > 1 go with CVE) to be honest. What I would personally do is probably try all options. If they all agree, the situation is simple. If the results obtained with the methods are different, then I think the best thing to do is to try and simulate your own case with the tools provided in this document and see which method will give you the best answer for your particular data.

docs/theory/diffusion/diffusion.rst
JoepVanlier: $\Delta t$ is the kymograph line time. Estimates for $\sigma$ are returned from the CVE estimator as a result.

I agree it's a good thing to include; but I don't feel confident providing those numbers in the docs yet at this point.

docs/theory/diffusion/diffusion.rst
JoepVanlier: Without this requirement the optimal number of points will be estimated incorrectly and the results from the Michalet paper do not necessarily hold anymore. It's an assumption made in the Michalet paper. They recommend restricting diffusive analysis to long tracks and making sure they are (nearly) the same length.

docs/theory/diffusion/diffusion.rst
JoepVanlier: Good catch! Removing. The default is zero anyhow.

docs/theory/diffusion/diffusion.rst
JoepVanlier: I've changed the text to:
```
When the SNR and the number of points per track is high, all methods presented here perform well.
In this case, the recommendation is to use CVE, while carrying over the localization uncertainty from the ensemble estimate.
Choosing this option will allow the study of diffusion constants for individual tracks with the largest precision.

When the SNR is low, restrict to long tracks if possible (refine the tracks and filter short tracks out) and use OLS.

If the uncertainty in the diffusion estimates is still too large, but the expectation is that the diffusion constants for different tracks are the same, consider using ensemble averages and compare ensemble averages obtained with CVE and OLS.
Ensemble estimates perform well down to relatively low SNR.
Note that a extra requirement for OLS is that the diffusion constants are roughly equally long
```

docs/theory/diffusion/diffusion.rst
aafkevandenberg: Yeah, $\sigma$ is a kind of spec, that does not belong in the Pylake docs. It is indeed best to just give instructions on how to estimate this quantity, rather than giving the value.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+.. math::
+
+    \overline{\hat{D}} = \frac{\sum_{m=1}^M N_m \hat{D}_m}{\sum_{m=1}^M N_m}
+
+Here :math:`M` is the number of tracks, :math:`D_m` corresponds to the diffusion constant of track :math:`m` while :math:`N_m` corresponds to the number of data points contributing to its estimate.
+The associated variance of the weighted mean is given by:
+
+.. math::
+
+    \overline{\hat{D}} = \frac{\sum_{m=1}^M N_m \left(\hat{D}_m - \overline{\hat{D}}\right)^2}{\left(M - 1\right) \sum_{m=1}^M N_m}
```
rpauszek, line 479: this has the same variable symbol as the mean above `\overline{\hat{D}}`

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+The uncertainty estimate Pylake returns on the individual CVEs is a little conservative this time (vertical lines), but the ensemble uncertainty estimate is pretty accurate.
+
+If we can safely assume that the localization uncertainty is constant then it is possible to improve the diffusion estimates by computing an ensemble averaged localization error first and then using that estimate when determining the per-track CVE.
+Let's define a function that does just this and compare it to the single step CVE::
+
+    def cve_single_using_ensemble_loc_uncertainty(tracks):
+        ensemble_estimate = tracks.ensemble_diffusion("cve")
+        return tracks.estimate_diffusion(
+            method="cve",
+            # Plug in the ensemble localization uncertainty estimates
```
rpauszek, line 519: comment inside the function call is a bit weird, I'd move it above

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+    plt.figure(figsize=(12, 4))
+    plt.subplot(1, 2, 1)
+    for method in ("ols single", "gls single", "cve single"):
+        plot_accuracy(snrs, results[method], variance=variances[method], label=method);
+    plt.ylim([-1, 3])
+
+    plt.subplot(1, 2, 2)
+    for method in ("cve unweighted", "cve ensemble", "ols ensemble"):
+        plot_accuracy(snrs, results[method], variance=variances[method], label=method);
+    plt.ylim([-1, 3])
```
rpauszek, line 663: might be nice to put some titles here to make it more clear what the two plots are comparing

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+------------------
+When the SNR and the number of points per track is high, all methods presented here perform well.
+In this case, the recommendation is to use CVE, while carrying over the localization uncertainty from the ensemble estimate.
+Choosing this option will allow the study of diffusion constants for individual tracks with the largest precision.
+
+When the SNR is low, but per track estimates are required, restrict to long tracks if possible (refine the tracks and filter short tracks out) and use OLS.
+
+If the uncertainty in the diffusion estimates is still too large, but the expectation is that the diffusion constants for different tracks are the same, consider using ensemble averages and compare ensemble averages obtained with CVE and OLS.
+Ensemble estimates perform well down to relatively low SNR.
+Note that a extra requirement for OLS is that the diffusion constants are roughly equally long.
```
rpauszek, line 715: what do you mean by "equally long"?

rpauszek (changes requested):
Wow, this is fantastic! I also especially appreciate the wonderful technical discussion with @aafkevandenberg; I basically just had to focus on clarity and nit-picky formatting things (you'll see I apparently have a thing for using `\mathrm{}`.

There are quite a few comments but again most are nit-picky. There are a couple however, where I really couldn't understand what was going on an I think it'd be good to clarify these before mergin.

docs/theory/index.rst
```diff
@@ -0,0 +1,18 @@
...
+
+For all of the examples, it is assumed that the following lines precede any other code::
+
+    import numpy as np
+    import matplotlib.pyplot as plt
+
+    import lumicks.pylake as lk
+
+.. toctree::
+    :caption: Segments
```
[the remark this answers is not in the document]
JoepVanlier: Good call. Changing.

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
 
-where :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` is the number of time points in the track, and :math:`x_i` is the track position at time frame :math:`i`.
+If it is safe to assume that the localization variance is the same for all the tracks, then one can achieve a more precise estimate for individual tracks by estimating this quantity at an ensemble level and then using that in the per-track estimation procedure::
 
-What we can see in this definition is that it uses the same data points several times, thereby resulting in a well averaged estimate.
-However, the downside of this estimator is that the calculated values are highly correlated :cite:`qian1991single,michalet2010mean,michalet2012optimal` which needs to be accounted for in subsequent analyses.
+    ensemble_estimate = tracks.ensemble_diffusion("cve")
+    tracks.estimate_diffusion(
+        "cve",
+        # Pass the ensemble localization uncertainties to the method
```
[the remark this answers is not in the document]
JoepVanlier: Sure 👍

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
 
-In the following, we'll use three simulated :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` instances of length 62 based on a diffusion constant of `10.0`.
+For more information on this procedure and how it affects the estimation uncertainties, please refer to the :doc:`theory section on diffusive motion</theory/diffusion/diffusion>`.
 
-With Pylake, we can calculate the MSD from a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` with a single command::
+Mean Squared Displacements
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Mean Squared Displacements (MSD) are a different quantity used to characterize particle motion.
+Purely diffusive processes leads to Mean Squared Displacements that linearly depend on the lag time.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
+MSDs for individual tracks are available through :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.msd`::
 
-    tracks[0].msd()
+    msd = tracks[0].msd()
 
-This returns a tuple of lags and MSD estimates. If we only wish MSDs up to a certain lag, we can provide a `max_lag` argument::
+Which returns a tuple of lags and MSD estimates. If we only wish MSDs up to a certain lag, we can provide a `max_lag` argument::
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
-
 The following properties can be read from :class:`~lumicks.pylake.kymotracker.detail.msd_estimation.EnsembleMSD`:
 
 - `lags` : Lags at which the MSD was computed.
 - `msd` : Mean MSD for each lag.
 - `sem` :  Standard error of the mean corresponding to each MSD.
 - `variance` : Variance of each MSD average.
 - `count` : Number of elements that contributed to the estimate at this lag.
-- `effective_sample_size` : Effective sample size.
+- `effective_sample_size` : Effective sample size used to compute the standard error of the mean.
```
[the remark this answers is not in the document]
JoepVanlier: This is `EnsembleMSD` (e.g. the MSD estimates rather than the diffusion constant estimates based on the MSD estimates), not `EnsembleDiffusion`. Though, maybe revisiting it, maybe we should just constrain this to a link to the API docs (now that we have them) to keep the tutorial focused. I can add a line of text for the CVE as well.

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
-
-By default, this will use the optimal number of lags (which in this case seems to be around 3-4), but once again a `max_lag` parameter can be specified to plot a larger number of lags::
-
-    [track.plot_msd(max_lag=100, marker='.') for track in tracks]
-
-.. image:: figures/kymotracking/msdplot_100_lags.png
-
-It's not hard to see from this graph why taking too many lags results in unacceptably large variances (note how the traces diverge).
+This method is not recommended for very short tracks, as it will produce biased results in such scenarios.
+Care must be taken that while this estimate may be optimal for quantifying an individual track; it may lead to biased estimates when allowed to vary for each individual track.
```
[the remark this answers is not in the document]
JoepVanlier: Right, I did word that kind of poorly. How about the following?

```
It is not recommended to apply Ordinary Least Squares to invididual tracks if they are very short since it will produce biased results in such scenarios.
Care must be taken that while the estimated number of lags may be optimal for quantifying an individual track; it may lead to biased estimates when averaging the results of several tracks with a different number of lags used for each individual track.
```

I did merge it into the warning box (see next comment response).

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
+For more information on this, see the :doc:`theory section for diffusive processes</theory/diffusion/diffusion>`.
 
 .. note::
 
+    This method is not recommended for very short tracks, as it will produce biased results in such scenarios.
+
```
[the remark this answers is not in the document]
JoepVanlier: How about merging the previous bit and this one into:

```
.. warning::

    It is not recommended to apply Ordinary Least Squares to invididual tracks if they are very short since it will produce biased results in such scenarios.
    Care must be taken that while the estimated number of lags may be optimal for quantifying an individual track; it may lead to biased estimates when averaging the results of several tracks with a different number of lags used for each individual track.
    For more information on this, see the :doc:`theory section for diffusive processes</theory/diffusion/diffusion>`.

    The uncertainty estimate and optimal number of lags obtained for Ordinary Least Squares relies on having a successful positional localization for every frame in a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack`.

    If frames are missing, then this method will issue a warning with the suggestion to refine tracks prior to estimation using :ref:`localization_refinement`.
    If this is not possible, please switch to :ref:`diffusion_cve`.
```

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
     The uncertainty estimate and optimal number of lags obtained for Ordinary Least Squares relies on having a successful positional localization for every frame in a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack`.
 
     If frames are missing, then this method will issue a warning with the suggestion to refine tracks prior to estimation using :ref:`localization_refinement`.
     If this is not possible, please switch to :ref:`diffusion_cve`.
 
-.. _diffusion_gls:
+If the diffusion constant can be assumed the same for an ensemble of tracks, it is possible to get a more precise estimate of the diffusion constant by first calculating the MSD for all of them, and then performing only a single fit.
+You can do this with Pylake by using :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.ensemble_diffusion`::
+
+    tracks.estimate_diffusion(method="ols")
```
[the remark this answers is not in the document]
JoepVanlier: Whoops! Yes, it certainly should be 😊

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
+
+For more information on this procedure and its performance, please refer to the :doc:`theory section</theory/diffusion/diffusion>`.
 
 Generalized Least Squares
 ^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Another option is to use generalized least squares :cite:`bullerjahn2020optimal`.
-This method is slower, since it has to solve some implicit equations, but it does not suffer from the large variance when including more lags (since it takes into account the covariance matrix of the MSD)::
+Generalized linear least squares suffers less from including more lags than optimal as it takes into account the covariance matrix of the MSD :cite:`bullerjahn2020optimal`.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,136 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
+This method is slower, since it has to solve some implicit equations. In addition to that, its current implementation cannot cope with unequal track lengths or missing frames.
+It can be selected by choosing `"gls"` for the method::
```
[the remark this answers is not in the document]
JoepVanlier: Oh, something went wrong here. For `gls` there is _only_ the option for the single track case. And the unequal length is not relevant there; only the no missing frames. Great catch, thank you!

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+The displacements of a particle undergoing free diffusion are uncorrelated over time and distributed according to a normal distribution.
+The standard deviation of this distribution follows from the diffusion equation and is given by: :math:`\sqrt{2 D \Delta t}`.
+We can simulate 1D diffusion as:
+
+.. math::
+
+    x_{i+1} = x_i + \sqrt{2 D \Delta t} N(0, 1)
+
+Here :math:`x_i` is the coordinate at time frame :math:`i`, :math:`D` is the diffusion constant, :math:`\Delta t` is the time step and :math:`N(0, 1)` is a sample from a normal distribution with a mean of zero and a standard deviation of 1.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+We will use this functionality to demonstrate the various methods for estimating diffusion constants implemented in Pylake.
+
+Method background
+-----------------
+
+Consider a freely diffusive track, how would one quantify its diffusivity?
+One convenient way would be to consider its positional increments over time.
+As shown in the simulation model, the increments in free diffusion are independent of the current position and depend directly on the diffusion constant.
+Considering that diffusion is a stochastic process, it is important to average over many such increments to obtain a result not subject to a lot of stochastic error.
+The exact details of how to obtain a diffusion constant depend on the method in question and will be discussed below.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+    \rho[n] = \langle\left(x_{i} - x_{i + n}\right)^2\rangle = 2 D n \Delta t
+
+Here the brackets :math:`\langle\ldots\rangle` indicate the expected value, :math:`\rho[n]` corresponds to the mean squared displacement for lag :math:`n`, :math:`\Delta t` is the time step of one lag and :math:`D` is the diffusion constant we seek.
+As you can see, the Mean Squared Displacement (MSD) gives a simple relation between the diffusion constant and the observed displacements, which is why MSD analysis is often used for diffusion analysis.
+
+Time-Averaged Mean Squared Displacements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are multiple ways to estimate this quantity in practice.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+As shown in the simulation model, the increments in free diffusion are independent of the current position and depend directly on the diffusion constant.
+Considering that diffusion is a stochastic process, it is important to average over many such increments to obtain a result not subject to a lot of stochastic error.
+The exact details of how to obtain a diffusion constant depend on the method in question and will be discussed below.
+
+Mean Squared Displacements
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Diffusive particles have an average displacement of zero.
+Rather than calculating an average displacement, it makes more sense to consider a form of distance travelled.
+In practice, the mean squared displacement is often used.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+When dealing with pure diffusive motion (a complete absence of drift) in an isotropic medium, the mean squared displacements follow the following relation:
+
+.. math::
+
+    \rho[n] = \langle\left(x_{i} - x_{i + n}\right)^2\rangle = 2 D n \Delta t
+
+Here the brackets :math:`\langle\ldots\rangle` indicate the expected value, :math:`\rho[n]` corresponds to the mean squared displacement for lag :math:`n`, :math:`\Delta t` is the time step of one lag and :math:`D` is the diffusion constant we seek.
+As you can see, the Mean Squared Displacement (MSD) gives a simple relation between the diffusion constant and the observed displacements, which is why MSD analysis is often used for diffusion analysis.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+Time-Averaged Mean Squared Displacements
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+There are multiple ways to estimate this quantity in practice.
+We can average over different tracks (ensemble averaging) or over time (time averaging).
+If all tracks experience the same diffusion, the ensemble averaged motion of a collection of independent molecules results in the same diffusion estimates as the time-averaged motion; the diffusive process is 'ergodic'.
```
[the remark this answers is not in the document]
JoepVanlier: I've dropped the first sentence. The second sentence is kind of important though, since it provides the justification for time averaging the MSDs. Maybe I should make that more explicit. How about:
```
If all tracks experience the same diffusion, the ensemble averaged motion of a collection of independent molecules results in the same diffusion estimates as the time-averaged motion; the diffusive process is 'ergodic'.
As a consequence, we can use time-averaged estimators to estimate diffusion constants on a per-track basis.
```

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+.. math::
+
+    \hat{\rho}[n] = \frac{1}{N - n} \sum_{i=1}^{N-n}\left(x_{i+n} - x_{i}\right)^2
+
+Here :math:`\hat{\rho}[n]` corresponds to the estimate of the MSD for lag :math:`n`, :math:`N` is the number of time points in the track, and :math:`x_i` is the track position at time frame :math:`i`.
+This particular estimator is referred to as the Time Averaged MSD (TAMSD), because the estimates for each lag are averaged over all time increments.
+Consider the definition of the TAMSD, it can be seen that the displacements are not independent from one another.
+While this definition results in a well-averaged estimate, this reduced uncertainty comes at a cost.
+Estimates obtained using this estimator are highly correlated :cite:`qian1991single,michalet2010mean,michalet2012optimal` which needs to be considered in subsequent analyses (more on this later).
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+With Pylake, we can calculate the MSD for a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` with :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.msd`.
+This returns a tuple of time and MSD values, which we can directly plot::
```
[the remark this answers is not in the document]
JoepVanlier: Correct.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+    plt.xlim([0, 0.8])
+    ax.view_init(20, -60)
+    ax.set_zticks([])
+    ax.set_xlabel(r'MSD [$\mu m^2$]')
+    ax.set_ylabel('Lag time [s]')
+    plt.tight_layout();
+
+Ensemble MSD
+^^^^^^^^^^^^
```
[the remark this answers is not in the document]
JoepVanlier: Whoops.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+Ordinary Least Squares
+^^^^^^^^^^^^^^^^^^^^^^
+
+Real measurements are affected by noise, which leads to localization error.
+In addition to this, depending on the method used to detect the position of the particle, there may also be motion blur.
+These two sources of error manifest themselves as an offset in the MSD curve given by:
+
+.. math::
+
+    offset = 2 \sigma^2 - 4 R D \Delta t
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+Here :math:`\sigma` is the static localization accuracy, :math:`R` is a motion blur constant and :math:`\Delta t` represents the time step.
```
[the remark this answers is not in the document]
JoepVanlier: Uncertainty is better.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+With pure diffusive motion (a complete absence of drift) in an isotropic medium, 1-dimensional MSDs can be fitted by the following relation:
+
+.. math::
+
+    \rho[n] = 2 D n \Delta t + offset
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+Here :math:`D` is the diffusion constant in :math:`um^2/s`, :math:`\Delta t` is the time step, :math:`n` is the step index and the offset is determined by the localization accuracy.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+Defining the displacements as :math:`\Delta x_n = x_n - x_{n + 1}`, the displacement covariance matrix is tri-diagonal :cite:`vestergaard2014optimal,vestergaard2015estimation`:
+
+.. math::
+
+    \begin{align}
+    \langle\left(\Delta x_n\right)^2\rangle & = & 2 D \Delta t + 2 \left(\sigma^2 - 2 R D \Delta t\right) &\\
+    \langle\Delta x_n \Delta x_{n+1}\rangle & = & -\left(\sigma^2 - 2 R D \Delta t \right) &\\
+    \langle\Delta x_n \Delta x_m\rangle & = & 0 &\mathrm{, for }\left| n-m \right| > 1.
+    \end{align}
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+
+    with
+
+    .. math::
+
+        S(t) = \int_{0}^{t} c(t') dt'
+
+    Here, the aperture function is defined as :math:`c(t)`, where :math:`c(t)` represents the fraction of illumination happening before time `t`.
+    :math:`c(t)` is normalized such that :math:`S(0) = 0` and :math:`S(\Delta t) = 1`.
+    For a rectangular shutter or exposure window, one obtains :math:`R = \frac{1}{6} \frac{t_{exposure}}{line\_time}`.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+    In the current implementation we assume the motion blur constant to be negligible (zero) for confocal acquisition, since the time spent over the particle's PSF compared to the line time is very low.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+    In the case of a provided localization uncertainty, it does impact the estimate of the diffusion constant itself.
+
+The performance of these covariance-based estimators can be characterized by its signal to noise ratio (SNR).
+This SNR is defined by:
+
+.. math::
+
+    \mathrm{SNR} = \frac{\sqrt{D \Delta t}}{\sigma}
+
+Here :math:`D` is the diffusion constant, :math:`\Delta t` the time step and :math:`\sigma` the localization uncertainty.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+    plt.ylabel('Probability density')
+    plt.legend()
+    plt.subplot(2, 1, 2, sharex=ax1)
+    plt.hist(results["ols"][-1], 30, label="ols");
+    plt.hist(results["cve"][-1], 30, label="cve");
+    plt.title(f"SNR = {snrs[-1]}")
+    plt.ylabel('Probability density')
+    plt.xlabel('Diffusion constant [$\mu$m/s]')
+    plt.xlim([-20, 20])
+    plt.tight_layout()
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: Moved.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+    snrs = 10**np.arange(-1, 1.1, 0.125)
+    results, variances = test_estimation_methods(simulate_tracks, snrs, methods=methods)
+    for method, estimates in results.items():
+        plot_accuracy(snrs, estimates, variances[method], label=method);
+
+    plt.ylim([-5, 6])
+
+.. image:: figures/ols_bias.png
+
+We can see that while both result in increased precision, one is not accurate at all for low SNRs (the mean deviates a lot from 1).
```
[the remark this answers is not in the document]
JoepVanlier: I'd prefer something like
```
We can see that while both result in increased precision, one is not accurate at all (highly biased) for low SNRs (the mean deviates a lot from 1).
```
since highly biased is basically a synonym for inaccurate. Would you be ok with that as well?

docs/theory/diffusion/diffusion.rst
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: Oh no, that sentence makes no sense 😅 

I meant:
``` 
Note that an extra requirement for OLS is that the diffusive tracks are roughly equally long.
````

File: docs/theory/diffusion/diffusion.rst [hunk 57]
```diff
@@ -0,0 +1,715 @@
...
+
+Here :math:`D` is the diffusion constant in :math:`um^2/s`, :math:`\Delta t` is the time step, :math:`n` is the step index and the offset is determined by the localization accuracy.
+
+While it may be tempting to use a large number of lags in the diffusion estimation procedure, this actually produces poor estimates of the diffusion constant :cite:`qian1991single,michalet2010mean,michalet2012optimal`, because, as mentioned above, the error in the MSD value increases with lag time.
+There exists an optimal number of lags to fit such that the estimation error is minimal.
+This optimal number of lags depends on the ratio between the diffusion constant and the dynamic localization accuracy:
+
+.. math::
+
+    \epsilon_{localization} = \frac{offset}{slope} = \frac{2 \sigma^2 - 4 R D \Delta t}{2 D \Delta t} = \frac{\sigma^2}{D \Delta t} - 2 R
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+This method can be extended to handle tracks that have gaps due to blinking by only considering the successful localizations :cite:`vestergaard2016optimizing`.
+To take this into account, we replace :math:`\Delta t` with :math:`\overline{\Delta t_m}`, where :math:`t_m` indicates the timestep between the successful localizations :math:`m` and :math:`m+1` :cite:`vestergaard2016optimizing`.
+
+If the localization uncertainty is known beforehand, one can derive the following estimator:
+
+.. math::
+
+    \hat{D} = \frac{\overline{\left(\Delta x_n\right)^2} - 2 \sigma^2}{2 (1 - 2 R) \Delta t}
+
+.. note::
```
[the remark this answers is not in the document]
JoepVanlier: I'll make it a separate subsection.

docs/theory/diffusion/diffusion.rst
```diff
@@ -0,0 +1,715 @@
...
+Here :math:`x_i` is the coordinate at time frame :math:`i`, :math:`D` is the diffusion constant, :math:`\Delta t` is the time step and :math:`N(0, 1)` is a sample from a normal distribution with a mean of zero and a standard deviation of 1.
+Our observations of the particle's position are subject to noise.
+In this section, we model this as Gaussian additive noise on the positions in the track.
+
+Note that free diffusion assumes that the space available for diffusion :math:`L` is sufficiently large that the measurement time interval is much smaller than :math:`L / 2D` :cite:`qian1991single`:
+
+To simulate diffusive tracks with observation noise, we can use :func:`~lumicks.pylake.simulation.simulate_diffusive_tracks`::
+
+    tracks = lk.simulation.simulate_diffusive_tracks(diffusion_constant=1.0, steps=40, dt=0.01, observation_noise=0.1, num_tracks=25)
+    tracks.plot()
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
JoepVanlier: Good catch! Changed it to 
```
    \mathrm{var}\left(\overline{\hat{D}}\right) = \frac{\sum_{m=1}^M N_m \left(\hat{D}_m - \overline{\hat{D}}\right)^2}{\left(M - 1\right) \sum_{m=1}^M N_m}
```

lumicks/pylake/simulation/diffusion.py
```diff
@@ -0,0 +1,87 @@
...
+    def get_blank_kymo(self, *, line_time_seconds):
+        if line_time_seconds not in self.kymos:
+            self.kymos[line_time_seconds] = _kymo_from_array(
+                np.zeros((0, 0)), "r", line_time_seconds=line_time_seconds, pixel_size_um=1.0
+            )
+
+        return self.kymos[line_time_seconds]
+
+
+_blank_kymo_pool = _BlankKymoPool()
```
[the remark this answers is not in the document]
JoepVanlier: Good catch! A cache is a much better solution to the same problem.

I've added a little docstring to make sure that the reason for the cache is motivated.
```python
# We use a cache to ensure that KymoLineGroups with the same properties end up referring to the same
# kymograph. This allows them to be added.
@cached(cache={})
```

docs/theory/diffusion/diffusion.rst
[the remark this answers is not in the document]
rpauszek: hmm, it still looks kinda strange, but better. I guess without the specialized LaTeX tools for rendering multi-line equations this is as good as we'll get

rpauszek (approved):
I think it's pretty much ready to go. Just left some very minor suggestions in my comments that I didn't mark resolved, but nothing blocking.

Great work!

docs/tutorial/kymotracking.rst
```diff
@@ -481,187 +481,128 @@ Since this algorithm is specifically looking for curvilinear structures, it can
...
-    [track.plot_msd(max_lag=100, marker='.') for track in tracks]
-
-.. image:: figures/kymotracking/msdplot_100_lags.png
-
-It's not hard to see from this graph why taking too many lags results in unacceptably large variances (note how the traces diverge).
+.. warning::
 
-.. note::
+    It is not recommended to apply Ordinary Least Squares to invididual tracks if they are very short since it will produce biased results in such scenarios.
+    Care must be taken that while the estimated number of lags may be optimal for quantifying an individual track; it may lead to biased estimates when averaging the results of several tracks with a different number of lags used for each individual track.
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, that's better. 👍

docs/tutorial/kymotracking.rst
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/diffusion/diffusion.rst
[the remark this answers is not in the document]
JoepVanlier: Yes, good catch.

docs/tutorial/kymotracking.rst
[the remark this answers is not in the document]
JoepVanlier: Yeah good idea, I'll add a sentence mentioning that.

docs/theory/diffusion/diffusion.rst
JoepVanlier: For typical Kymos that I've seen in house, the effect is very small (the spot size is small compared to the width of the Kymo). I think for the purpose of this document, for now, we just explicitly state that for `CVE` we have it set to be zero.

PR #436, opened 2022-12-12
Update kymo docs to use example dataset

rpauszek:
**Why this PR?**
Everybody loves real data!

Note: while working on this I also found a bug that I introduced on final cleanup for the custom colormaps PR, so that's fixed here (in a separate commit)

[Docs build here](https://lumicks-pylake.readthedocs.io/en/kymo_docs/tutorial/kymographs.html)


docs/tutorial/kymographs.rst
```diff
@@ -5,37 +5,65 @@ Kymographs
...
 
-    kymo = file.kymos["cas9"]
-    kymo.plot(channel="green", aspect="auto", adjustment=lk.ColorAdjustment(0, 5))
+Or access a particular `Kymo` object directly to work with::
+
+    # use the kymo name as a dict key
+    kymo = file.kymos["16"]
+
+    # or get the first kymo from the dict
+    kymo_name, kymo = file.kymos.popitem()
```
JoepVanlier, line 29: I have mixed feelings about adding `.popitem()` here because it can lead to confusion unless you actually explain what it does exactly. Reason being that if they'd do this:

```python
kymos = file.kymos
kymo_name, kymo = kymos.popitem()
```

 It would remove the `kymo` from the dictionary.

docs/tutorial/kymographs.rst
```diff
@@ -5,37 +5,65 @@ Kymographs
...
 .. image:: figures/kymographs/kymo_intro.png
 
-Here we see the :meth:`~lumicks.pylake.kymo.Kymo.plot()` convenience function. The `channel`
-argument accepts the strings "red", "green", "blue", or "rgb". This function accepts keyword
-arguments that are passed to :func:`plt.imshow() <matplotlib.pyplot.imshow()>` internally. Note
-also, the axes are labeled with the appropriate time and position units.
+Note that the axes are labeled with the appropriate time and position units.
+
+The first argument `channel` accepts the strings "red", "green", "blue", or "rgb". We also see that the color limits can be set easily using
+the :class:`~lumicks.pylake.ColorAdjustment` class. The first two arguments act like the `vmin` and `vmax` arguments used with
```
JoepVanlier, line 53: Might be nice to mention that these can also be passed per color? Then again, it's already mentioned in the API docs. Maybe superfluous.

docs/tutorial/kymographs.rst
```diff
@@ -44,13 +72,15 @@ The kymograph can also be exported to TIFF format::
 Kymo data and details
 ---------------------
 
-We can access the raw image data as `numpy` arrays::
+We can access the raw image data as a :class:`numpy.ndarray`::
 
-    rgb = kymo.get_image("rgb")  # matrix with `shape == (h, w, 3)`
-    blue = kymo.get_image("blue")  # single color so `shape == (h, w)`
+    rgb = kymo.get_image("rgb")  # matrix with `shape == (height, width, 3 colors)`
+    blue = kymo.get_image("blue")  # single color so `shape == (height, width)`
```
JoepVanlier, line 78: 🙌

docs/tutorial/kymographs.rst
```diff
@@ -44,13 +72,15 @@ The kymograph can also be exported to TIFF format::
...
 
     # Plot manually
-    plt.imshow(kymo.get_image("green"), aspect="auto", adjustment=lk.ColorAdjustment(0, 5))
+    plt.figure()
+    plt.imshow(kymo.get_image("green"), aspect="auto", vmax=15)
```
JoepVanlier, line 82: Prefer `vmax` for simplicity?

docs/tutorial/kymographs.rst
```diff
@@ -75,116 +105,148 @@ There are also several properties available for convenient access to the kymogra
...
-    kymo.crop_by_distance(6, 24).plot("green")
+    plt.figure()
+    kymo.crop_by_distance(9.5, 26).plot("rgb", aspect="auto", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+    plt.show()
 
 .. image:: figures/kymographs/kymo_cropped.png
 
-Kymographs can also be sliced in order to obtain a specific time range.
-For example, one can plot the region of the kymograph between 114.2 and 164.6 seconds using::
+.. warning::
```
JoepVanlier, line 130: `.. note::` ? You even mention it as a note.

docs/tutorial/kymographs.rst
```diff
@@ -75,116 +105,148 @@ There are also several properties available for convenient access to the kymogra
...
 
-    kymo["114.2s":"164.6s"].plot("green")
+    Note, slicing in time is currently only supported for unprocessed kymographs. If you want to both crop and slice a kymo,
+    the order of operations is important: you need to slice before cropping::
```
JoepVanlier, line 133: I really like that you explicitly mention this order of operations thing.

One minor nitpick though, I find the double `:` looks a bit weird here. Maybe just make `you need to slice before cropping` its own sentence?

docs/tutorial/kymographs.rst
```diff
@@ -75,116 +105,148 @@ There are also several properties available for convenient access to the kymogra
...
+    Also, the accuracy of the calibration is dependent on how the kymo is cropped. If you crop the kymo by visually estimating the
+    bead edges, the resulting position should be taken as approximate.
+
+We can also interactively slice, crop, and calibrate kymographs using :meth:`~lumicks.pylake.kymo.Kymo.crop_and_calibrate`::
+
+    widget = kymo.crop_and_calibrate(channel="rgb", tether_length_kbp=48.502, aspect="auto", adjustment=lk.ColorAdjustment(0, 99.5, mode="percentile"))
 
 .. image:: figures/kymographs/kymo_interactive.png
 
 Simply click and drag the rectangle selector to the desired ROI. After closing the widget, we can access the edited kymograph
```
JoepVanlier, line 194: You didn't edit this, but what does closing the widget mean in a lab context?

docs/tutorial/kymographs.rst
```diff
@@ -75,116 +105,148 @@ There are also several properties available for convenient access to the kymogra
...
+    plt.title("downsampled both")
+
+    plt.tight_layout()
+    plt.show()
 
-.. image:: figures/kymographs/kymo_downsampled_time_and_position.png
+.. image:: figures/kymographs/kymo_downsampled.png
 
 Note however, that not all functionalities are present anymore when downsampling a kymograph. For
 example, if we downsample a kymograph by time, we can no longer access the per pixel timestamps::
```
JoepVanlier, line 246: Maybe mention why?
It's also good to know that downsampling by space still allows more downstream analysis than downsampling by time does.

docs/tutorial/kymographs.rst
```diff
@@ -197,18 +259,23 @@ We can downsample channel data according to the lines in a kymo. We can use
...
+    downsampled.plot(start=force.start, label="downsampled like kymo")
+    plt.legend()
+    plt.show()
+
+.. image:: ./figures/kymographs/force_downsampled_like_kymo.png
 
 There is also a convenience function to plot a kymograph along with a downsampled force trace::
 
-    kymo.plot_with_force("1x", "green")
+    kymo.plot_with_force("1x", "green", adjustment=lk.ColorAdjustment(0, 15))
```
JoepVanlier, line 275: Would be nice to reference this function in the text so they have something to click that will take them to the relevant API docs page.

docs/tutorial/kymographs.rst
```diff
@@ -197,18 +259,23 @@ We can downsample channel data according to the lines in a kymo. We can use
...
 
 This will average the forces over each Kymograph line and plot them in a correlated fashion.
 The function can also take a dictionary of extra arguments to customize the kymograph plot.
```
JoepVanlier, line 278: You have `kwargs` and `kymo_args`. Would be good to link to the API docs so people can quickly see what goes where.

lumicks/pylake/adjustments.py
```diff
@@ -126,7 +126,7 @@ def nothing(cls):
 
 
 def _make_cmap(name, color):
-    LinearSegmentedColormap.from_list(name, colors=[(0, 0, 0), color])
+    return LinearSegmentedColormap.from_list(name, colors=[(0, 0, 0), color])
```
JoepVanlier, line 129: 😊

JoepVanlier (approved):
Looks great! Beautiful data too!

docs/tutorial/kymographs.rst
rpauszek: Ah I was wondering why it wasn't removed. This is what I always use because then I don't have to look up the kymo name first, so I thought it might be nice to add here; but I get your point. Might be nice to add some kind of functionality to get the kymos by index rather than by name, but that's another ticket

docs/tutorial/kymographs.rst
rpauszek: yeah I think leave it for the API docs is fine

docs/tutorial/kymographs.rst
rpauszek: it's using `plt.imshow()` which doesn't have an `adjustment` argument 😆 just one of the many mistakes I've found now being able to actually run the tutorial code

docs/tutorial/kymographs.rst
rpauszek: oops didn't notice that

docs/tutorial/kymographs.rst
rpauszek: yeah, when I originally wrote this and tested the plot I wasn't using `jupyter` at all (surprising, I know) so you have to close the window before the code to grab the new kymo is executed. I just removed the whole sentence

docs/tutorial/kymographs.rst
rpauszek: good point. I also really don't like the implementation of this one (I think I modeled it after FD curve construction). Maybe for a separate PR deprecating this and having instead `kymo.plot_with_channel(Slice, color)`

lumicks/pylake/adjustments.py
rpauszek: 😒

tobiasjj: 🙈

tobiasjj (approved):
LGTM, graphs look nice, text is descriptive and good to follow! One suggestion: Probably split off the interactive section into its own. 🚴

lumicks/pylake/adjustments.py
JoepVanlier: I think the recommendation is there so you actually get a name in call stacks instead of `<lambda>`.

But um, we can add an exception for `E731` if you want :D

PR #438, opened 2022-12-13
kymo/stack: make new plot for video export

JoepVanlier:
**Why this PR?**
We don't want our video export to include things from previous plots.

```python
plt.figure()
plt.plot([1, 2, 3], [1, 2, 3])
f = lk.File("C:\\Dev\\AVI rendering\\20190828-105121 Scan DNA1_CII_incubate_2min.h5")
cs = f.scans.popitem()[1]
cs.export_video("rgb", "d:\\new.gif")
```

Old behaviour:
old

New behaviour:
new

### DIFF
--- a/lumicks/pylake/detail/imaging_mixins.py
+++ b/lumicks/pylake/detail/imaging_mixins.py
@@ -162,7 +162,7 @@ class VideoExport:
             image_handle = plot_func(frame=start_frame + num, image_handle=image_handle)
             return plt.gca().get_children()
 
-        fig = plt.gcf()
+        fig = plt.figure()
         fig.patch.set_alpha(1.0)
         line_ani = animation.FuncAnimation(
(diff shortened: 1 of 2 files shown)


lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -162,7 +162,7 @@ def plot(num):
             image_handle = plot_func(frame=start_frame + num, image_handle=image_handle)
             return plt.gca().get_children()
 
-        fig = plt.gcf()
+        fig = plt.figure()
```
tobiasjj, line 165: 3to6 is close to one, this is worth a 🥳

tobiasjj (approved):
LGTM! 🚴

lumicks/pylake/detail/imaging_mixins.py
JoepVanlier: 😄

PR #440, opened 2022-12-16
Fix some issues with dwelltime bootstrapping

rpauszek:
**Why this PR?**
Originally the issue was opened because trying to plot a bootstrap before sampling raised a cryptic exception. While figuring out how to best deal with this, I noticed some other things that could be improved

- would be better to just get rid of the `DwelltimeModel.bootstrap` attribute altogether to remove statefulness. Here I changed it to a property with a deprecation warning.
- also deprecated `DwelltimeBootstrap.plot()` and renamed to `DwelltimeBootstrap.hist()` in line with `DwelltimeModel`
- made `DwelltimeBootstrap` a frozen data class and added an `extend()` method to append new samples to the distribution. Again trying to get away from statefullness
- now trying to access `DwelltimeModel.bootstrap` before calling `DwelltimeModel.calculate_bootstrap()` raises a `RuntimeError`. This is a stopgap fix until the `bootstrap` property is removed in a future release

[Docs build here](https://lumicks-pylake.readthedocs.io/en/bootstrap_fixups/tutorial/population_dynamics.html#dwell-time-analysis) and [API changes here](https://lumicks-pylake.readthedocs.io/en/bootstrap_fixups/_api/lumicks.pylake.population.dwelltime.DwelltimeBootstrap.html)

### DIFF
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -51,6 +51,25 @@ class DwelltimeBootstrap:
         samples = DwelltimeBootstrap._sample(optimized, iterations)
         return cls(optimized, samples[: optimized.n_components], samples[optimized.n_components :])
 
+    def extend(self, iterations):
+        """Extend the distribution by additional sampling iterations.
+
+        Parameters
+        ----------
+        iterations : int
+            number of iterations (random samples) to add to the bootstrap distribution
+        """
+        new_samples = DwelltimeBootstrap._sample(self.model, iterations)
+        n_components = self.model.n_components
+        new_amplitudes = new_samples[:n_components]
+        new_lifetimes = new_samples[n_components:]
+
+        return DwelltimeBootstrap(
+            self.model,
+            np.hstack([self.amplitude_distributions, new_amplitudes]),
+            np.hstack([self.lifetime_distributions, new_lifetimes]),
+        )
+
     @staticmethod
     def _sample(optimized, iterations):
 
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -7,24 +7,35 @@ import matplotlib.pyplot as plt
 from deprecated.sphinx import deprecated
 
 
-@dataclass
+@dataclass(frozen=True)
 class DwelltimeBootstrap:
     """Bootstrap distributions for a dwelltime model.
 
-    This class is stored in the `DwelltimeModel.bootstrap` attribute
-    and should not be constructed manually.
+    This class should be initialized using :meth:`lk.DwelltimeModel.calculate_bootstrap()
+    <lumicks.pylake.DwelltimeModel.calculate_bootstrap>` and should not be constructed manually.
 
     Attributes
     ----------
-    _samples : np.ndarray
-        array of optimized model parameters for each bootstrap sample pull; shape is
-        [number of parameters, number of samples]
+    model : :class:`~lumicks.pylake.DwelltimeModel`
+        Original model sampled for the bootstrap distribution
+    amplitude_distributions : np.ndarray
+        Array of sample optimized amplitude parameters; shape is [number of components, number of samples]
+    lifetime_distributions : np.ndarray
+        Array of sample optimized lifetime parameters; shape is [number of components, number of samples]
     """
 
-    _samples: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
+    model: "DwelltimeModel"
+    amplitude_distributions: np.ndarray = field(repr=False)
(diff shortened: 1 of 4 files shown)


lumicks/pylake/population/dwelltime.py
```diff
@@ -7,23 +7,32 @@
...
+    amplitude_distributions : np.ndarray
+        Array of sample optimized amplitude parameters; shape is [number of components, number of samples]
+    lifetime_distributions : np.ndarray
+        Array of sample optimized lifetime parameters; shape is [number of components, number of samples]
     """
 
-    _samples: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
+    model: "DwelltimeModel"
+    amplitude_distributions: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
+    lifetime_distributions: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
```
JoepVanlier, line 29: I'd prefer if these didn't have defaults.

The way it is set up now means that they'll have an incorrect shape compared to real data when `n_components > 1` which leads to not being able to rely on it being the correct shape.

One could consider polling `n_components` from the model, but I don't really see a need for creating this class empty often (probably never once `_bootstrap` is deprecated) so maybe just omitting the default and handling it in the single use case for it is best.

lumicks/pylake/population/dwelltime.py
```diff
@@ -198,11 +234,30 @@ def __init__(
             max_observation_time,
             options=self._optim_options,
         )
-        self.bootstrap = DwelltimeBootstrap()
+        self._bootstrap = DwelltimeBootstrap(self)
```
JoepVanlier, line 237: Add the same TODO here as below. If you go with the earlier suggestion, this would become:
```suggestion
        # TODO: remove with deprecation
        self._bootstrap = DwelltimeBootstrap(self, np.empty((n_components, 0)), np.empty((n_components, 0)))
```

lumicks/pylake/population/dwelltime.py
```diff
@@ -7,23 +7,32 @@
...
     """
 
-    _samples: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
+    model: "DwelltimeModel"
+    amplitude_distributions: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
+    lifetime_distributions: np.ndarray = field(default_factory=lambda: np.array([]), repr=False)
 
-    def _sample_distributions(self, optimized, iterations):
+    def __post_init__(self):
+        assert self.amplitude_distributions.shape == self.lifetime_distributions.shape
```
JoepVanlier, line 32: Would be nice to check whether data is consistent with the model provided (i.e. check whether the arrays are two dimensional and whether `n_components` of the model provided corresponds to the shape of dimension `1`).

JoepVanlier (changes requested):
Nice!

I like the informative error and the path towards the better API. I have a few suggestions, mostly around checking some of the data integrity at the door a bit more rather than handling cases downstream.

lumicks/pylake/population/dwelltime.py
rpauszek: good point. Indeed, I think once we remove the `bootstrap` property we can just raise on empty construction. Also added a specific test for the empty instance, just as a double check, I think it's nice to have. We can consider removing this after the property is gone.

 I left the fixups as separate commits for now but I'll squash before merging

lumicks/pylake/population/dwelltime.py
```diff
@@ -98,7 +125,18 @@ def calculate_stats(self, key, component, alpha=0.05):
...
+            "This method has been renamed to more closely match its behavior. Use "
+            ":meth:`DwelltimeBootstrap.hist()` instead."
+        ),
+        action="always",
+        version="0.13.3",
+    )
     def plot(self, alpha=0.05, n_bins=25, hist_kwargs=None, span_kwargs=None, line_kwargs=None):
+        self.hist(alpha, n_bins, hist_kwargs, span_kwargs, line_kwargs)
+
+    def hist(self, alpha=0.05, n_bins=25, hist_kwargs=None, span_kwargs=None, line_kwargs=None):
```
JoepVanlier, line 139: Ah, sorry for not noticing this in my earlier review, but now would be a really good time to make these keyword arguments if it's preferable (with the name change). I imagine that especially the `kwargs` ones might benefit from this for if we ever get more arguments in between?

JoepVanlier (approved):
Looks great!

One last thing that might be worth considering before you merge is whether you want any of the arguments for `hist` to be keyworded. Considering that you're changing the name, I think it might be nice to at least have the `kwargs` arguments be keyworded moving forward.

PR #441, opened 2022-12-19
test_imaging_mixins: ignore RuntimeWarning raised by numpy >= 1.24

tobiasjj:
**Why?**
Numpy >= 1.24 raises a `RuntimeWarning` during casting if an floating point error occurs: 
https://github.com/numpy/numpy/releases/tag/v1.24.0
https://github.com/numpy/numpy/pull/21437

This causes `test_imaging_mixins` to fail:
numpy_1 24_overflow_warning

The affected funtion is `TiffExport.export_tiff()`

**Solution**
Ignore the warning in the tests.

### DIFF
--- a/lumicks/pylake/tests/test_imaging_mixins.py
+++ b/lumicks/pylake/tests/test_imaging_mixins.py
@@ -115,6 +115,10 @@ def test_export_tiff_int(tmp_path):
         export_image16.export_tiff(tmp_path / "float16", dtype=np.float16)
 
 
+@pytest.mark.filterwarnings(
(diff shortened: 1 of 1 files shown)


rpauszek (approved):
🚢

JoepVanlier:
@rpauszek @tobiasjj A bit late to the party, but isn't this warning us about an actual bug though?

Shouldn't the correct behavior be to explicitly clip prior to conversion if clip is `True`? Now the behaviour is undefined (I can't find anything in the numpy docs guaranteeing a particular behavior in the case of overflow). In my quick experiment it wraps around (which is is not what we want I think).

I also notice that the description in the `RuntimeError` is wrong. It mentions `force=True` while the argument is called `clip`.

PR #442, opened 2022-12-19
docs: nbexport: Add support to embed images in markdown cells.

tobiasjj:
# Why?
As a developer I want to include images in the notebooks that can be exported from the docs.
Additionally, I also want `notes` and `warnings` to be (properly) displayed in the exported notebooks.

# How?
* Upon writing the docs one can mark images to be attached to and embedded in the exported notebook with the option `:nbattach:`
* Notes in the notebook are now rendered with a blue box, and warnings are included and rendered with a red box

To see how the notebook looks with the modified `nbexport`, see the following Example:

# Example
## Example source
```rst
.. warning::
    This is alpha functionality. While usable, this has not yet been tested in a large
    number of different scenarios. The API may also still be subject to change.

New NBExport funtionality
=========================

.. only:: html

    :nbexport:`Download this page as a Jupyter notebook <self>`

In this tutorial, we will show you how to properly export images to notebooks.

Attach or not attach images to notebooks
----------------------------------------

Image that should not be attached to the notebook:

.. image:: figures/new_export/noattach.jpg

Image that should be attached to the notebook:

.. image:: figures/new_export/attach.jpg
  :nbattach:

.. note:: 
   Warnigs and notes will be rendered with colored boxes in the exported notebook.
```
[docs output](https://user-images.githubusercontent.com/28353120/208445602-d074cc93-fd8a-4c9f-b057-6707288e5698.PNG)
* **notebook before:**
notebook_no_attach
* **notebook after:**
notebook_attach

### DIFF
--- a/docs/_ext/nbexport.py
+++ b/docs/_ext/nbexport.py
@@ -1,13 +1,15 @@
-import os
 import itertools
+import mimetypes
+import os
 import posixpath
-
-from sphinx import addnodes, roles
-from docutils import nodes, writers
+from base64 import b64encode
 
 import nbformat
+from docutils import nodes, writers
+from docutils.parsers.rst import directives
 from nbconvert.preprocessors import ExecutePreprocessor
 from nbconvert.preprocessors.execute import CellExecutionError
+from sphinx import addnodes, roles
 
 
 def _finilize_markdown_cells(nb):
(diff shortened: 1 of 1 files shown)


docs/_ext/nbexport.py
```diff
@@ -200,6 +230,12 @@ def visit_math_block(self, node):
         self.write_markdown("$$\n{}\n$$\n\n".format(node.astext().strip()))
         raise nodes.SkipNode
 
+    def visit_image(self, node):
+        if "nbattach" in node and node["nbattach"]:
+            filename = os.path.basename(node["uri"])
+            self.write_markdown(f"![{filename}](attachment:{filename})\n")
```
JoepVanlier, line 236: I think you should add two linebreaks here since markdown wants two, otherwise it'll still wrap weirdly:

image

```suggestion
            self.write_markdown(f"![{filename}](attachment:{filename})\n\n")
```

JoepVanlier (approved):
Great! This is very nice to have.

Only a few tiny fixups I would suggest, but otherwise, good to go.

While reviewing this, I noticed the API links in our docs are broken when exporting to notebooks.

I made a ticket on the backlog for this, so no need to worry about it now.

PR #446, opened 2023-01-05
fdfitter: improve docstrings for fd models

JoepVanlier:
**Why this PR?**
This PR fixes three (small) things.

1. The F-d fitter uses the docstring provided in the implementation of the F-d model as `__repr_html__`. Some of these models had no description and therefore no description was provided to the user. This is fixed now.

I considered defining the docstring in the implementation only and reading that docstring from the API side (in `models.py`) but after doing so felt that the added complexity was not really warranted to deduplicate the sections that could be deduplicated. Reason being that for the render in the Jupyter `__repr_html__`, one usually wants to use `$a$` since it renders nicely; while in the docstring for the API side, one wants :math:`a` since that formats correctly to the online documentation.

2. The criterion for the validity of the inextensible worm-like chain was modified to reflect that in reality it depends on the model parameters. This equation comes from Wang et al [1].

3. It should be mentioned that none of the models is particularly valid at very low forces (around 0.1 pN for DNA). There are models that claim to fare better at such low forces, but we do not have them implemented yet (see [2]).

[1] https://www.sciencedirect.com/science/article/pii/S0006349597787800
[2] https://www.sciencedirect.com/science/article/pii/S0006349599772073

Docs build can be found here: 
https://lumicks-pylake.readthedocs.io/en/ms_validity/_api/lumicks.pylake.wlc_marko_siggia_force.html
https://lumicks-pylake.readthedocs.io/en/ms_validity/_api/lumicks.pylake.wlc_marko_siggia_distance.html

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -25,6 +25,11 @@
 * `DwelltimeBootstrap` is now a frozen dataclass.
 * Attempting to access `DwelltimeModel.bootstrap` before sampling now raises a `RuntimeError`; however, see the deprecation note above for proper API to access bootstrapping distributions.
 
+#### Bugfixes
+
+* Fixed issue with model description not being available in Jupyter notebooks for some force-distance models.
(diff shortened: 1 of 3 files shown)


rpauszek (approved):
nice fixups 🚢

PR #448, opened 2023-01-13
Fixup DwelltimeBootstrap histograms

rpauszek:
**Why this PR?**
While testing some other things, I noticed that only up to 2 components are shown for the histogram even if the model components > 2. Also changed so we use the estimated parameter rather than the mean of the bootstrap distribution to mark the plots

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,9 @@
 * Added option to calculate a diffusion estimate based on the ensemble MSDs using the Ordinary Least Squares (OLS) method.
 * Added `DwelltimeBootstrap.extend()` to draw additional samples for the distribution.
 
+#### Bug fixes
+* Fixed bug in `DwelltimeBootstrap.hist()` (previously named `DwelltimeBootstrap.plot()`, see below). Previously, only up to two components were plotted; now all components are plotted appropriately.
+
 #### Deprecations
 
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -14,11 +14,16 @@
 * Added option to calculate a diffusion estimate based on the ensemble MSDs using the Ordinary Least Squares (OLS) method.
 * Added `DwelltimeBootstrap.extend()` to draw additional samples for the distribution.
 
+#### Bug fixes
+* Fixed bug in `DwelltimeBootstrap.hist()` (previously named `DwelltimeBootstrap.plot()`, see below). Previously, only up to two components were plotted; now all components are plotted appropriately.
```
JoepVanlier, line 18:
```suggestion
#### Bug fixes

* Fixed bug in `DwelltimeBootstrap.hist()` (previously named `DwelltimeBootstrap.plot()`, see below). Previously, only up to two components were plotted; now all components are plotted appropriately.
```

lumicks/pylake/population/dwelltime.py
```diff
@@ -180,7 +180,7 @@ def plot_axes(data, key, component, use_index):
             data = self.lifetime_distributions.squeeze()
             plot_axes(data, "lifetime", 0, False)
         else:
-            for component in range(2):
+            for component in range(self.n_components):
```
JoepVanlier, line 197: 👍

JoepVanlier (approved):
Looks good to me.

PR #452, opened 2023-01-20
kymotracking: add bias reduction step into the centroid estimator

JoepVanlier:
**Why this PR?**
Subpixel estimates can be very biased when using a simple centroid estimator when there is appreciable background. The reason is that the subpixel deviation is normalized by the total intensity over area used for the estimate. When the background is high, this results in a contraction of the subpixel correction towards the center.

**How to review this PR?**
I recommend going commit by commit. In the first few commits, I set up the algorithm in pure Python.  The beforelast commit vectorizes the algorithm. This makes it a bit trickier to read, but the performance gains are worthwhile.

**Description**
This PR is an implementation of [this ](https://opg.optica.org/oe/fulltext.cfm?uri=oe-16-18-14064&id=171315) paper. In a nutshell, the paper shows that under mild assumptions the bias is proportional to how far the spot is from the center. Instead of trying to remove the background, they suggest adjusting the window such that the spot is centered (including removing fractional parts of the window).

What you see below is a histogram of 1000 localizations for each of the 40 spot locations tried. The black lines indicate the mean (center) and 1 sigma bounds. The yellow line indicates the true value used for simulation.

image

The variance of this estimator goes up when the window size gets bigger, therefore it is important to choose the window size somewhat appropriate for the spot size. This same effect is not true for Gaussian refinement.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -30,6 +30,7 @@
 
 * Added API for notes, i.e. `file.notes` returns a dictionary of notes.
 * Added option to add a scale bar to by providing a [`lk.ScaleBar()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ScaleBar.html) to plotting or export functions.
+* Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 
 ## v0.13.3 | 2023-01-26
 
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 
 #### Breaking changes
 
+* When performing particle tracking on kymographs, bias correction is now enabled by default; without this correction, kymographs with high background signal will suffer from biased localization estimates. To disable bias correction, specify `bias_correction=False` to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 * Changed several `asserts` to `Exceptions`.
   * Attempting to read `KymoTracks` from a `CSV` file that doesn't have the expected file format will result in an `IOError`.
   * Attempting to extend `KymoTracks` by `KymoTracks` originating from a different `Kymograph` now results in a `ValueError`.
--- a/lumicks/pylake/kymotracker/detail/peakfinding.py
+++ b/lumicks/pylake/kymotracker/detail/peakfinding.py
@@ -97,40 +97,62 @@ class KymoPeaks:
         return coordinates, time_points, peak_amplitudes
 
 
-def bounds_to_centroid_data(left_edge, right_edge):
+def bounds_to_centroid_data(index_array, left_edges, right_edges):
     """Helper function to return selection indices, pixel centers and weights.
 
-    This function generates indices for sampling based on the left-most and right-most
-    bound. It also generates appropriate pixel centers and weights that account for
-    the fact that they are only partial pixels.
+    This function generates indices for sampling based on the left-most and right-most bound. It
+    also generates appropriate pixel centers and weights that account for the fact that they are
+    only partial pixels.
 
     Parameters
     ----------
-    left_edge, right_edge : float
-        Lower and upper pixel edge
+    index_array: np.ndarray
+        2D pixel indices with shape `(N, M)` where `N` is the number of pixels being refined and
+        `M` is the window size used for refinement.
+    left_edges, right_edges : np.ndarray
+        Lower and upper pixel edges
 
     Returns
     -------
     selection : np.ndarray
-        Indices which select which points to use.
+        Boolean mask with shape `(N, M)` which select which points from the window to use.
+        `N` is the number of pixels being refined, while `M` is the window size used for refinement.
     centers : np.ndarray
-        Pixel centers, where the edge pixels are corrected for being partial.
+        Pixel center positions with shape `(N, M)` with edge pixel positions accounting for being
+        partial pixels. `N` is the number of pixels being refined, while `M` is the window size
+        used for refinement.
     weights : np.ndarray
-        Weights that account for down-weighting the edges.
+        Weights with shape `(N, M)` that account for down-weighting the edges.
+        `N` is the number of pixels being refined, while `M` is the window size used for refinement.
     """
-    selection = np.arange(left_edge, np.ceil(right_edge), dtype=int)
-    centers = selection + 0.5
-    weights = np.ones(selection.size)
-    centers[0] = (left_edge + np.floor(left_edge) + 1) / 2
-    centers[-1] = (right_edge + np.ceil(right_edge) - 1) / 2
-    weights[0] = 1.0 - (left_edge - np.floor(left_edge))
-    weights[-1] = 1.0 - (np.ceil(right_edge) - right_edge)
+    left_floor = np.floor(left_edges).astype(int)
+    right_ceil = np.ceil(right_edges).astype(int)
 
-    return selection, centers, weights
+    # Upper bound needs to be rounded inwards
+    right_floor = np.floor(np.nextafter(right_edges, 0)).astype(int)
+
+    # Find which pixels we are still using
+    active = np.logical_and(
+        index_array >= left_floor[:, np.newaxis], index_array < right_ceil[:, np.newaxis]
+    )
+
+    num_points = left_floor.size
+    centers = active * (index_array + 0.5)
+
+    centers[np.arange(num_points), left_floor] = (left_edges + left_floor + 1) / 2
+    centers[np.arange(num_points), right_floor] = (right_edges + right_ceil - 1) / 2
+
+    # Pixels outside the bounds are not used
+    weights = active.astype(float)
+
+    # Fractional pixels are down-weighted.
+    weights[np.arange(num_points), left_floor] = 1.0 - (left_edges - np.floor(left_edges))
+    weights[np.arange(num_points), right_floor] = 1.0 - (np.ceil(right_edges) - right_edges)
+    return active, centers, weights
 
 
-def unbiased_centroid(data, tolerance=1e-3, max_iterations=50, epsilon=1e-8):
-    """Perform an unbiased centroid estimation
+def unbiased_centroid(position, data, tolerance=1e-3, max_iterations=50, epsilon=1e-8):
+    """Perform an unbiased centroid estimation in a vectorized manner
 
     The bias in centroid refinement is proportional to the asymmetry around the spot. To remove this
     bias, we define a virtual window that's symmetric around the spot position. In practice, this
@@ -144,8 +166,12 @@ def unbiased_centroid(data, tolerance=1e-3, max_iterations=50, epsilon=1e-8):
 
     Parameters
     ----------
+    position : np.ndarray
+        Subpixel coordinates obtained with regular centroid method.
     data : np.ndarray
-        1D array
+        2D image array with data used for refinement. It contains a small region around each spot.
+        It has shape `(N, M)`, with `N` the number of pixels being refined, while `M` is the
+        window size used for refinement.
     tolerance : float
         Convergence tolerance
     max_iterations : int
@@ -155,8 +181,8 @@ def unbiased_centroid(data, tolerance=1e-3, max_iterations=50, epsilon=1e-8):
 
     Returns
     -------
-    estimated_position : float
-        Estimated subpixel position of the spot after bias correction. The origin (position 0) of
+    estimated_positions : np.ndarray
+        Estimated subpixel positions of the spot after bias correction. The origin (position 0) of
         the position is defined at the window origin.
 
     References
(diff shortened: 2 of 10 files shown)


lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -184,13 +185,15 @@ def refine_peak_based_on_moment(
...
+        kernel size used for this refinement will be 2 * half_kernel_size + 1.
     max_iter : int
         Maximum number of iterations
     eps : float
-        We add a little offset to the normalization to prevent divisions by zeros on pixels that did not have any photon
-        counts. Eps sets this offset.
+        We add a little offset to the normalization to prevent divisions by zeros on pixels that
+        did not have any photon counts. Eps sets this offset.
+    bias_correction : bool
+        Apply a bias correction at the end of the centroid refinement?
```
alessiamarcolini, line 233: uncommon to read a question in a docstring.

better use something like `whether to apply...`

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -75,13 +75,143 @@ def flatten(self):
...
+        last_position = position.copy()
+
+    # Internally, we used a coordinate system that has pixel centers at 0.5, 1.5 etc. This is
+    # practical because it simplifies the code. Externally, we define pixel centers at 0, 1, 2 etc,
+    # so we subtract a half here.
+    return position - 0.5
+
+
+def _clip_kernel_to_edge(max_half_width, coordinates, dataset_width):
+    """This function finds the number of pixels between the coordinates and the image edge.
```
alessiamarcolini, line 191:
```suggestion
    """Find the number of pixels between the coordinates and the image edge.
```

from https://peps.python.org/pep-0257/

> The docstring is a phrase ending in a period. It prescribes the function or method’s effect as a command (“Do this”, “Return that”), not as a description; e.g. don’t write “Returns the pathname …”.

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -75,13 +75,143 @@ def flatten(self):
...
+
+    Parameters
+    ----------
+    max_half_width : int
+        Desired kernel size. If the kernel completely fits, then this is the returned value.
+    coordinates : np.ndarray
+        1D array of coordinates
+    dataset_width : int
+        Shape of the data matrix to clip against (basically the maximum value for the coordinate)
+    """
```
alessiamarcolini, line 203: missing return section

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -75,13 +75,143 @@ def flatten(self):
...
+    also generates appropriate pixel centers and weights that account for the fact that they are
+    only partial pixels.
+
+    Parameters
+    ----------
+    index_array: np.ndarray
+        2D pixel indices
+    left_edges, right_edges : np.ndarray
+        Lower and upper pixel edges
+    """
```
alessiamarcolini, line 91: missing returns section

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -92,13 +222,15 @@ def refine_peak_based_on_moment(
...
+        kernel size used for this refinement will be 2 * half_kernel_size + 1.
     max_iter : int
         Maximum number of iterations
     eps : float
-        We add a little offset to the normalization to prevent divisions by zeros on pixels that did not have any photon
-        counts. Eps sets this offset.
+        We add a little offset to the normalization to prevent divisions by zeros on pixels that
+        did not have any photon counts. Eps sets this offset.
+    bias_correction : bool
+        Apply a bias correction at the end of the centroid refinement?
```
alessiamarcolini, line 233: missing return section

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -384,9 +396,19 @@ def refine_tracks_centroid(tracks, track_width=None):
...
-        for kymographs calibrated in microns. For kymographs calibrated in kilobase pairs the
-        corresponding value is calculated using 0.34 nm/bp (from duplex DNA).
+        If `None`, the default is 0.35 (half the wavelength of the red limit of the visible
+        spectrum) for kymographs calibrated in microns. For kymographs calibrated in kilobase pairs
+        the corresponding value is calculated using 0.34 nm/bp (from duplex DNA).
+    bias_correction : bool
+        Correct coordinate bias by ensuring that the window is symmetric (see [1]_ for more
+        information). Note that while this increases the variance (reduces precision), it can
+        greatly reduce the bias when there is a high background.
+
```
alessiamarcolini, line 406: missing return section

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -486,6 +509,9 @@ def refine_tracks_gaussian(
         When supplied, the background is not estimated but fixed at this value.
     """
     assert overlap_strategy in ("ignore", "skip", "multiple")
+    assert isinstance(window, int) or np.issubdtype(
+        window.dtype, np.integer
+    ), "window parameter should be integer"
```
alessiamarcolini, line 514: assertions in code that is not testing code should be avoided, and a proper exception should be raised instead

> An assert’s primary function should be to isolate a bug’s root cause and should never be used for standard error handling. While asserts can be beneficial, it’s essential to use them for debugging and tests, and not to control the flow of the logic for the program or application and use them only in safe situations.

nice readings:

https://towardsdatascience.com/2-caveats-to-avoid-when-using-python-assert-statement-8060e123c26a
https://snyk.io/blog/the-dangers-of-assert-in-python/

lumicks/pylake/kymotracker/tests/test_refinement.py
```diff
@@ -82,6 +81,22 @@ def test_refinement_track(loc, inv_sigma=0.3):
     np.testing.assert_allclose(track.coordinate_idx, loc, rtol=1e-2)
 
 
+@pytest.mark.parametrize("loc", [25.3, 25.5, 26.25, 23.6])
+def test_refinement_with_background(loc, inv_sigma=0.3):
```
alessiamarcolini, line 85: why is `inv_sigma` a parameter with a default value? "who" can actually provide the value here?

alessiamarcolini (changes requested):

lumicks/pylake/kymotracker/kymotracker.py
tiagobonetti: I guess we might need a ticket on pylake to clean all assertions up...

JoepVanlier: Huh. TIL. Good to know, thank you @alessiamarcolini .

> I guess we might need a ticket on pylake to clean all assertions up...

Yeah, I was just thinking that as well. I'll pick this up this sprint. Shouldn't take too long.

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -75,6 +75,98 @@ def flatten(self):
...
+    Fractional pixels are dealt with by down-weighting edge pixels and recomputing a new pixel
+    center for those pixels.
+
+    For more information on this method, please refer to [1]_.
+
+    References
+    ----------
+    .. [1] Berglund, A. J., McMahon, M. D., McClelland, J. J., & Liddle, J. A. (2008).
+           Fast, bias-free algorithm for tracking single particles with variable size and
+           shape. Optics express, 16(18), 14064-14075.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -75,6 +75,98 @@ def flatten(self):
...
+    ----------
+    data : np.ndarray
+        1D array
+    tolerance : float
+        Convergence tolerance
+    max_iterations : int
+        Maximum iterations
+    epsilon : float
+        Small value to make sure we don't divide by zero
+    """
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: Good point. Fixed.

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: Ah yes. Bad habit. Thanks for pointing it out. Fixing!

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: So, I was looking at this again. I will make a separate PR for fixing these asserts up in other places. I also dropped the int check I added here. I've made a ticket to look at this particular case in a bit more detail, since the docstring marks it as an int (which is why I added the assert), but from what I can tell, downstream can in principle handle it being float.

lumicks/pylake/kymotracker/tests/test_refinement.py
JoepVanlier: It shouldn't be. Removed now.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: 👍

alessiamarcolini (approved):

changelog.md
```diff
@@ -4,6 +4,7 @@
 
 #### Breaking changes
 
+* When performing particle tracking on kymographs, bias correction is now enabled by default. To disable bias correction, please specify `bias_correction=False` to `lk.track_greedy()` and `lk.refine_tracks_centroid()` if you wish to disable these. Without bias correction, kymographs with background will suffer from biased localization estimates.
```
```suggestion
* When performing particle tracking on kymographs, bias correction is now enabled by default; without this correction, kymographs with high background signal will suffer from biased localization estimates. To disable bias correction, specify `bias_correction=False` to `lk.track_greedy()` and `lk.refine_tracks_centroid()`. ```

File: lumicks/pylake/kymotracker/detail/peakfinding.py [hunk 12]
```diff
@@ -97,13 +97,166 @@ def flatten(self):
...
+    """Helper function to return selection indices, pixel centers and weights.
+
+    This function generates indices for sampling based on the left-most and right-most bound. It
+    also generates appropriate pixel centers and weights that account for the fact that they are
+    only partial pixels.
+
+    Parameters
+    ----------
+    index_array: np.ndarray
+        2D pixel indices
```
rpauszek, line 110: I think this could be a bit clearer. I'm guessing this is for the whole kymograph, not one scan line at a time? Maybe indicated the expected shape like you do in the `Returns` section could help

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -97,13 +97,166 @@ def flatten(self):
...
+    center for those pixels.
+
+    For more information on this method, please refer to [1]_.
+
+    Parameters
+    ----------
+    position : np.ndarray
+        Subpixel coordinates obtained with regular centroid method.
+    data : np.ndarray
+        2D array
```
rpauszek, line 171:
```suggestion
        2D image array
```

lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -97,13 +97,166 @@ def flatten(self):
...
+
+        if np.max(np.abs(position - last_position)) < tolerance:
+            break
+
+        # We change position in place each iteration, so we need to be explicit about the copy
+        last_position = position.copy()
+
+    # Internally, we used a coordinate system that has pixel centers at 0.5, 1.5 etc. This is
+    # practical because it simplifies the code. Externally, we define pixel centers at 0, 1, 2 etc,
+    # so we subtract a half here.
```
rpauszek, line 226: 👍

rpauszek (approved):
looks great, just a few suggestions for clarity

changelog.md
```diff
@@ -30,6 +31,7 @@
 
 * Added API for notes, i.e. `file.notes` returns a dictionary of notes.
 * Added option to add a scale bar to by providing a [`lk.ScaleBar()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ScaleBar.html) to plotting or export functions.
+* Added extra method for centroid refinement that shrinks the window to reduce estimation bias.
```
rpauszek, line 34: this seems a bit oddly worded, at least for me I immediately thought of a method on a class. Maybe instead, we should mention explicitly the added `bias_correction` keyword arg?

JoepVanlier: Good point. I changed it to:
```
```

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: Yeah. Prior to vectorization it was literally one spot for one scan line at a time. After it's every spot at the same time. I changed it to:
```
        2D pixel indices with shape `(N, M)` where `N` is the number of pixels being refined and
        `M` is the window size used for refinement.
```

lumicks/pylake/kymotracker/detail/peakfinding.py
JoepVanlier: I've rephrase it a bit for clarity (also because it took me a few seconds to figure out what it was again):
```
        2D image array with data used for refinement. It contains a small region around each spot.
        It has shape `(N, M)`, with `N` the number of pixels being refined, while `M` is the
        window size used for refinement.
```

PR #453, opened 2023-01-23
Refactor dwell time analysis to remove explicit constraint

rpauszek:
**Why this PR?**

Previously we used the `SLSQP` algorithm which supports constraints in order to ensure fractional amplitudes sum to 1. Now, for a model with `K` components, `K-1` amplitude parameters are passed, with the final term calculated within the log likelihood function.

This allows us to switch to `L-BFGS-B` and has the added benefit that for a single component model, the amplitude parameter is not included in the optimization, but inherently set to 1.

### DIFF
--- a/lumicks/pylake/kymotracker/tests/test_kymotrack.py
+++ b/lumicks/pylake/kymotracker/tests/test_kymotrack.py
@@ -633,10 +633,10 @@ def test_fit_binding_times(blank_kymo):
     tracks = KymoTrackGroup([k1, k2, k3, k4])
 
     dwells = tracks.fit_binding_times(1)
-    np.testing.assert_allclose(dwells.lifetimes, [1.002547])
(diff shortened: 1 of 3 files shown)


JoepVanlier (changes requested):
I think the proposed setup has too much parameter freedom for `N > 2`. For three (or more) components, how do you enforce the non-negativity constraint for `amp3` and up?

With `amp1` and `amp2` between `0` and `1` you can get a negative third amplitude. Quickly trying a three component model fails on this branch, while it seems to work (although you need _a lot_ of data for reliability) on current `main`.

rpauszek:
Figuring out how to enforce the non-negativity constraint on the final amplitude parameter needs some work. Additionally, it's likely we can't easily use the Hessian for uncertainty estimation won't be useful in this case, so no real need to switch to `L-BFGS-S`. We might want to keep the constrained optimizer, but switch to an inequality constraint where the sum of fitted amplitudes is `< 1`

PR #455, opened 2023-01-24
ensemble_msd: forward  kwargs  in `EnsembleMSD.plot()`

JoepVanlier:
**Why this PR?**
It's very minor, but in most cases where we just `.plot()` a single thing we simply forward the `**kwargs`. I found myself essentially remaking this `plot()` function because it wasn't there. Figured I'd better just do the fixup at the source.


rpauszek (approved):
🚢

PR #457, opened 2023-01-30
Update Install.rst

aafkevandenberg:
I added an item to the FAQ addressing a frequently occurring error during installation: https://github.com/conda/conda/issues/11795 I chose to use an older conda version because it is easy to do and it has worked for several users.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -282,6 +282,28 @@ If you have already activated the correct environment, but you still do not see 
 Frequently asked questions
 --------------------------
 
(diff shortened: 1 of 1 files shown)


JoepVanlier (approved):
Yeah, I think for now this is a reasonable solution until this issue stops happening when doing a fresh install using the windows installer. One minor request, maybe rename the commit message to reflect what's being changed. Maybe something like `docs: add workaround for openssl error to faq`

PR #458, opened 2023-01-30
plotting: allow users to add a scale bar to their plots if they want

JoepVanlier:
**Why this PR?**
This request came up multiple times during the user sessions.

I figured I'd play a bit with `AnchoredOffsetBox` on a friday afternoon, and it allowed this to be rolled into a feature relatively painlessly. Because we just use `matplotlib` API, it plays nicely with interactive widgets:

interactive_scale

Plays nicely with movie export:

new

corrstack

And it seems pretty easy to use. Note that adding a scale bar to `plot_correlated` is not yet supported, but it may be added in the future.

Also note that I went relatively minimal on the addition to the docs (since it's so similar to `ColorAdjustment` and the docstring for the function itself contains examples). Also note that `ImageStack` is essentially uncalibrated. I left a little `TODO` there to make sure that we update that once we add the relevant calibration metadata.

Docs build is here: https://lumicks-pylake.readthedocs.io/en/scalebar/
Interesting bit is here: https://lumicks-pylake.readthedocs.io/en/scalebar/_api/lumicks.pylake.ScaleBar.html

### DIFF
--- a/lumicks/pylake/scalebar.py
+++ b/lumicks/pylake/scalebar.py
@@ -94,6 +94,24 @@ class ScaleBar:
         Font size to use for the labels.
     **kwargs
         additional arguments passed to :class:`matplotlib.offsetbox.AnchoredOffsetBox`.
+
+    Examples
+    --------
+    ::
+
+        import lumicks.pylake as lk
+
+        # Loading a kymograph.
+        h5_file = pylake.File("example.h5")
+        _, kymo = h5_file.kymos.popitem()
+
+        # Show default scale bar
+        kymo.plot("green", scale_bar=lk.ScaleBar())
+        plt.show()
+
+        # Show scale bar with a scale of 10 seconds on the x axis
+        kymo.plot("green", scale_bar=lk.ScaleBar(size_x=10))
+        plt.show()
     """
 
     size_x: Optional[float] = None
--- a/lumicks/pylake/scalebar.py
+++ b/lumicks/pylake/scalebar.py
@@ -1,0 +1,138 @@
+from dataclasses import dataclass, field
+from matplotlib.offsetbox import AnchoredOffsetbox, AuxTransformBox, VPacker, HPacker, TextArea
+from typing import Optional, Union
+
+
+def _create_scale_legend(
+    transform,
+    size_x,
+    size_y,
+    label_x,
+    label_y,
+    loc,
+    color,
+    separation,
+    barwidth,
+    fontsize,
+    **kwargs,
+):
+    """Draws a scale legend for the current axis.
+
+    Parameters
+    ----------
+    transform : :class:`matplotlib.transforms.CompositeGenericTransform`
+        The coordinate frame (typically axes.transData).
+    size_x, size_y : float or None
+        Length of the scale bars in data units. `0` or `None` omits the scale.
+    label_x, label_y : str or None
+        Labels for the scale bars. None results in no label on the scale.
+    loc : str or int
+        Position in the containing axis.  Valid locations are 'upper left', 'upper center',
+        'upper right', 'center left', 'center', 'center right', 'lower left', 'lower center',
+        'lower right'.
+    color : Matplotlib color
+        Color to use for the text and labels.
+    separation : float
+        separation between labels and bars in points.
+    bar_width : float or None
+        Width of the scale bar(s)
+    fontsize : float or None
+        Font size to use for the labels.
+    **kwargs
+        additional arguments passed to :class:`matplotlib.offsetbox.AnchoredOffsetBox`."""
+    from matplotlib.patches import Rectangle
+
+    bars = AuxTransformBox(transform)
+    bar_args = {"ec": color, "lw": barwidth, "fc": None}
+
+    if size_x:
+        bars.add_artist(Rectangle(xy=(0, size_y), width=size_x, height=0, **bar_args))
+
+    if size_y:
+        bars.add_artist(Rectangle(xy=(0, 0), width=0, height=size_y, **bar_args))
+
+    textprops = {"color": color, "fontsize": fontsize}
+    if size_x and label_x:
+        xlabel = TextArea(label_x, textprops=textprops)
+        bars = VPacker(children=[bars, xlabel], align="center", pad=0, sep=separation)
+
+    if size_y and label_y is not None:
+        ylabel = TextArea(label_y, textprops=textprops)
+
+        if size_x and label_x is not None:
+            # Add a dummy label with the same text such that the label is properly centered.
+            # Without this dummy element, the label ends up in the middle of the full vertical
+            # height of the scales and the text underneath.
+            dummy = TextArea(label_x, textprops={**textprops, "visible": False})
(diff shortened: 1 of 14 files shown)


lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+    fontsize,
+    **kwargs,
+):
+    """Draws a scale legend for the current axis.
+
+    Parameters
+    ----------
+    transform : :class:`matplotlib.transforms.CompositeGenericTransform`
+        The coordinate frame (typically axes.transData).
+    size_x, size_y : float
```
alessiamarcolini, line 24:
```suggestion
    size_x, size_y : float or None
```

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+        Length of the scale bars in data units. `0` or `None` omits the scale.
+    label_x, label_y : str or None
+        Labels for the scale bars. None results in no label on the scale.
+    loc : position in the containing axis
+        Valid locations are 'upper left', 'upper center', 'upper right', 'center left',
+        'center', 'center right', 'lower left', 'lower center', 'lower right'.
+    color : color
```
alessiamarcolini, line 31: there should be the type here

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+
+    Parameters
+    ----------
+    transform : :class:`matplotlib.transforms.CompositeGenericTransform`
+        The coordinate frame (typically axes.transData).
+    size_x, size_y : float
+        Length of the scale bars in data units. `0` or `None` omits the scale.
+    label_x, label_y : str or None
+        Labels for the scale bars. None results in no label on the scale.
+    loc : position in the containing axis
```
alessiamarcolini, line 28:
```suggestion
    loc : str
```

alessiamarcolini: maybe `position in the containing axis` can go below

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+    return AnchoredOffsetbox(loc, child=bars, frameon=False, **kwargs)
+
+
+@dataclass
+class ScaleBar:
+    """Draws a scale legend for the current axis.
+
+    Parameters
+    ----------
+    size_x, size_y : float
```
alessiamarcolini, line 78:
```suggestion
    size_x, size_y : float or None
```

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+        Width of the scale bars in data units. `0` or `None` omits the scale.
+    label_x, label_y : str or None
+        Labels for the scale bars. None results in no label on the scale.
+    loc : position in the containing axis
+        Valid locations are 'upper left', 'upper center', 'upper right', 'center left',
+        'center', 'center right', 'lower left', 'lower center', 'lower right'.
+    color : color
```
alessiamarcolini, line 85: same as before

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+
+    size_x: float or None = None
+    size_y: float or None = None
+    label_x: str or None = None
+    label_y: str or None = None
+    loc: str or int = "upper right"
+    color: str = "white"
+    separation: float or None = 2.0
+    barwidth: float or None = None
+    fontsize: float or None = None
```
alessiamarcolini, line 123: `float or None` is not actually a valid type annotation

you can use either `Optional[str]` or `str | None` (Python 3.10+)

lumicks/pylake/scalebar.py
```diff
@@ -0,0 +1,153 @@
...
+    kwargs: dict = field(default_factory=dict)
+
+    def _attach_scale_bar(self, axis, size_x, size_y, x_unit, y_unit):
+        """Attach scale bar to an axis
+
+        axis : :class:`matplotlib.pyplot.axis`
+            Axis to attach scale bar to
+        size_x, size_y : float
+            Length of the scale bars in data units unless the user specified a custom one. Note that
+            `0` or `None` omits the scale.
```
alessiamarcolini, line 133: it's not so clear that the parameter set at initialization time has priority over the one passed here

alessiamarcolini (commented):
just comments about docstrings and type annotations

lumicks/pylake/scalebar.py
JoepVanlier: Good catch! 👍

lumicks/pylake/scalebar.py
JoepVanlier: This is actually an interesting one. I wasn't sure what to put here. Basically, the type should be any color `matplotlib` can handle. When I look at [their](https://github.com/matplotlib/matplotlib/blob/c23ccdde6f0f8c071b09a88770e24452f2859e99/lib/matplotlib/colors.py#L271) [code](https://github.com/matplotlib/matplotlib/blob/c23ccdde6f0f8c071b09a88770e24452f2859e99/lib/matplotlib/lines.py#L1060), I see them just referring to these as `color` never specifying concrete types for them.

Currently, I think they are a union of `arraylike` and `str`, but to me, that seems less instructive to the user than `color`. It would also mean that if `mpl` supported additional types for their colors in the future, we would have to update this docstring.

What do you think?

lumicks/pylake/scalebar.py
JoepVanlier: Whoops, yes, definitely. Good catch!

lumicks/pylake/scalebar.py
JoepVanlier: Good catch.

lumicks/pylake/scalebar.py
JoepVanlier: Good catch, thanks! We still support older versions than 3.10, so I'll go with `Optional[]` and `Union[]`.

lumicks/pylake/scalebar.py
JoepVanlier: Good catch! I've tried to rephrase it a bit to:
```
These values are only used if the user did not explicitly set one during initialization.
```

lumicks/pylake/scalebar.py
alessiamarcolini: yeah that's an interesting one :D
looks like it can be a string or a tuple: https://matplotlib.org/stable/tutorials/colors/colors.html

 at this point I would put `Matplotlib color` as they say [here](https://github.com/matplotlib/matplotlib/blob/c23ccdde6f0f8c071b09a88770e24452f2859e99/lib/matplotlib/colors.py#L271)

JoepVanlier: Yeah, I agree it's probably good to be explicit, since it's not our own definition.

lumicks/pylake/scalebar.py
JoepVanlier: 👍 

Changed to `Matplotlib color`

PR #462, opened 2023-02-03
other: assertions to exceptions (part 2)

JoepVanlier:
Inspired by @alessiamarcolini comment on [this PR](https://github.com/lumicks/pylake/pull/452), I decided to remove all the `assert`-based input validation we are doing since these can easily be disabled in production and generally a well named exception is more informative. These only cover issues that were improper behaviour before, so it shouldn't affect anyone that was using the API correctly; but the changes are technically breaking, so it's a good idea to do it before `1.0`.

This is PR 2 of N with changes pertaining to this. While the first one is very specific for `FdFitting`, this one is a bit more of a hodgepodge. I tried to keep the commits small to make it easier to review, but I will squash it down a bit before merging.

I did a few drive-by fixups on some docstrings as well, but that's not the main focus of this PR.

I also noticed that in some cases, we weren't testing the correct thing. Please review commit by commit and see the commit message for more details.

Docs build here: https://lumicks-pylake.readthedocs.io/en/assert_other

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -18,7 +18,9 @@
   * FdFitter: When adding data to a fit, adding data with more than one dimension will raise a `ValueError`.
   * FdFitting: Attempting to evaluate a parameter trace with `lk.parameter_trace()` for a parameter that is not part of the model now results in a `ValueError`.
   * FdFitting: Attempting to compute a parameter trace while providing an incomplete set of parameters will now result in a `ValueError`.
-  
+  * Attempting to use `Slice.downsampled_over()` or `Slice.downsampled_like()` with timestamps ranges or another channel that doesn't overlap with the channel now produces a `ValueError`.
+  * Attempting to construct a `TimeSeries` where the length of the timestamp array is not equal to the length of the data array results in a `ValueError`.
+
 #### New features
 
 * Added API for notes, i.e. `file.notes` returns a dictionary of notes.
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@
   * FdFitting: Attempting to compute a parameter trace while providing an incomplete set of parameters will now result in a `ValueError`.
   * Attempting to use `Slice.downsampled_over()` or `Slice.downsampled_like()` with timestamps ranges or another channel that doesn't overlap with the channel now produces a `ValueError`.
   * Attempting to construct a `TimeSeries` where the length of the timestamp array is not equal to the length of the data array results in a `ValueError`.
+  * `FdEnsemble` alignment now produces a `ValueError` if fewer than 2 datasets are provided.
 
 #### New features
 
--- a/changelog.md
+++ b/changelog.md
@@ -22,6 +22,7 @@
   * Attempting to construct a `TimeSeries` where the length of the timestamp array is not equal to the length of the data array results in a `ValueError`.
   * `FdEnsemble` alignment now produces a `ValueError` if fewer than 2 datasets are provided.
   * Image reconstruction now raises a `ValueError` if the length of the data and infowave are not equal.
+  * Plotting: When creating a plot providing `axes` and an `image_handle`, a `ValueError` is raised when those `axes` do not belong to the `image_handle` provided. 
 
 #### New features
 
--- a/changelog.md
+++ b/changelog.md
@@ -24,6 +24,7 @@
   * Image reconstruction now raises a `ValueError` if the length of the data and infowave are not equal.
   * Plotting: When creating a plot providing `axes` and an `image_handle`, a `ValueError` is raised when those `axes` do not belong to the `image_handle` provided. 
   * Widefield: Attempting to open multiple `TIFF` as a single ImageStack will now raise a `ValueError` if the alignment matrices of the individual `TIFF` are different.
+  * PowerSpectrum: Attempting to replace the power spectral values of a `PowerSpectrum` using `with_spectrum` using a vector of incorrect length will raise a `ValueError`.
 
 #### New features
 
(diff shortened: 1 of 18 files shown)


lumicks/pylake/channel.py
```diff
@@ -431,7 +431,7 @@ def downsampled_like(self, other_slice, reduce=np.mean):
         timestamps = other_slice.timestamps
         delta_time = np.diff(timestamps)
 
-        if self._src.start >= timestamps[-1] or self._src.stop <= timestamps[0] - delta_time[0]:
+        if self._src.start >= timestamps[-1] or self._src.stop <= timestamps[0]:
             raise RuntimeError("No overlap between range and selected channel.")
```
JoepVanlier, line 435: Option: we could also consider just returning an empty `Slice` in this case.

JoepVanlier, line 435: Option: we could also consider just returning an empty Slice in this case.

lumicks/pylake/detail/widefield.py
```diff
@@ -333,16 +341,24 @@ def verify_stack_similarity(self, other):
...
+                raise ValueError(
+                    "Alignment matrices must be the same for stacks to be merged. The following "
+                    f"alignment matrices were found in one stack but not the other {diff_keys}."
+                )
+
+            # Check the actual matrices we have here
+            for channel, mat in self._alignment_matrices.items():
+                if mat != other._alignment_matrices[channel]:
+                    raise ValueError(
+                        f"Alignment matrices must be the same for stacks to be merged. The "
```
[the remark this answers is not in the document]
alessiamarcolini: why is flake8 not complaining?
it looks it's not actually running :D

lumicks/pylake/detail/widefield.py
```diff
@@ -450,7 +466,8 @@ def __init__(self, matrix=None):
 
     def __mul__(self, mat):
         """Perform matrix multiplication such that `self * mat == np.matmul(self, mat)`."""
-        assert isinstance(mat, TransformMatrix), "Operands must be of type `TransformMatrix`."
+        if not isinstance(mat, TransformMatrix):
+            raise TypeError("Operands must be of type `TransformMatrix`.")
```
alessiamarcolini, line 470:
```suggestion
            raise TypeError(f"Operands must be of type `TransformMatrix`, got `{type(mat)}`.")
```

alessiamarcolini (changes requested):

lumicks/pylake/detail/widefield.py
```diff
@@ -323,7 +323,15 @@ def __init__(self, tiff_file, align_requested):
             self._alignment = Alignment(align_requested, AlignmentStatus.missing, False)
 
     def verify_stack_similarity(self, other):
-        """Verifies that the metadata for these images reflects a compatible image"""
+        """Verifies that the metadata for these images reflects a compatible image
+
+        Parameters
+        ----------
+        self : ImageDescription
+            Image metadata.
```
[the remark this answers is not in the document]
JoepVanlier: Whoops 😊 . Good catch!

lumicks/pylake/detail/widefield.py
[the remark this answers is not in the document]
JoepVanlier: Well that's interesting. I'm sure I've seen `flake8` warnings in our CI before. I wonder what broke. I'll investigate this in a separate PR.

lumicks/pylake/channel.py
```diff
@@ -264,9 +271,11 @@ def downsampled_over(self, range_list, reduce=np.mean, where="center"):
         if not isinstance(range_list, list):
             raise TypeError("Did not pass timestamps to range_list.")
 
-        assert len(range_list[0]) == 2, "Did not pass timestamps to range_list."
-        assert self._src.start < range_list[-1][1], "No overlap between range and selected channel."
-        assert self._src.stop > range_list[0][0], "No overlap between range and selected channel"
+        if len(range_list[0]) != 2:
+            raise RuntimeError("Did not pass timestamps to range_list.")
```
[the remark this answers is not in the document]
JoepVanlier: I agree. It's an error on my part.

I am wondering if we're being overly restrictive here though.

I did not want to piggyback that change in this PR (since there are many small changes), but I think that for something to be `iterable`, and returning a `start` and `stop` value at each iteration would actually be sufficient for this function to work. Doing such a hard check against a `list` doesn't feel very pythonic to me.

changelog.md
```diff
@@ -2,6 +2,17 @@
 
 ## v1.0.0 | t.b.d.
 
+#### Breaking changes
+
+* Changed several `asserts` to `Exceptions`.
+  * Attempting to use `Slice.downsampled_over()` or `Slice.downsampled_like()` with timestamps ranges or another channel that doesn't overlap with the channel now produces a `RuntimeError`.
```
tobiasjj, line 8:
```suggestion
  * Attempting to use `Slice.downsampled_over()` with timestamps ranges or another channel that doesn't overlap with the channel now produces a `ValueError`.
  * Attempting to use `Slice.downsampled_over()` without timestamps ranges now produces a `ValueError`.
```

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -111,8 +111,27 @@ def num_samples(self):
...
 
     def with_spectrum(self, power, num_points_per_block=1):
-        """Return a copy with a different spectrum"""
-        assert len(power) == len(self.power), "Power has incorrect length"
+        """Return a copy with a different spectrum
+
+        Parameters
+        ----------
+        power : np.ndarray
+            Vector of power spectral values
```
tobiasjj, line 119:
```suggestion
            Vector of power spectral values.
```

tobiasjj (changes requested):
Only nitpicks: One missing changelog entry and one missing dot.

changelog.md

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: 👍

tobiasjj (approved):
LGTM! 🚴

PR #463, opened 2023-02-03
kymotracker: change assertions to exceptions (part 3)

JoepVanlier:
Inspired by @alessiamarcolini comment on https://github.com/lumicks/pylake/pull/452, I decided to remove all the assert-based input validation we are doing since these can easily be disabled in production and generally a well named exception is more informative. These only cover issues that were improper behaviour before, so it shouldn't affect anyone that was using the API correctly; but the changes are technically breaking, so it's a good idea to do it before 1.0.

This is PR 3 of N with changes pertaining to this. This one focuses on things related to kymotracking. I tried to keep the commits small to make it easier to review, but I will squash it down a bit before merging.

I did a few drive-by fixups on some docstrings as well, but that's not the main focus of this PR.

I also noticed that in some cases, we weren't testing the correct thing. Please review commit by commit and see the commit message for more details.

Docs build here: https://lumicks-pylake.readthedocs.io/en/assert_kymo


lumicks/pylake/kymotracker/detail/peakfinding.py
```diff
@@ -45,9 +45,22 @@ def reset_assignment(self):
...
+        time_points : np.ndarray
+            Time points (in frame indices) of detected peaks
+        peak_amplitudes : np.ndarray
+            Peak amplitudes of detected peaks
+        """
+        if len(time_points) == 0:
+            raise ValueError("You need to provide at least one time point")
+
+        if any(len(time_points) != len(x) for x in (coordinates, peak_amplitudes)):
+            raise ValueError("Number of time points, coordinates and peak amplitudes must be equal")
```
alessiamarcolini, line 63:
```suggestion
            raise ValueError(f"Number of time points ({len(time_points}), coordinates ({len(coordinates)}) and peak amplitudes ({peak_amplitudes}) must be equal")
```

lumicks/pylake/kymotracker/tests/test_io.py
```diff
@@ -189,3 +189,10 @@ def test_csv_version(version, read_with_version):
     for j, track in enumerate(imported_tracks):
         np.testing.assert_allclose(track.time_idx, data["time (pixels)"][j])
         np.testing.assert_allclose(track.coordinate_idx, data["coordinate (pixels)"][j])
+
+
+@pytest.mark.parametrize("filename", ["csv_bad_format.csv", "csv_unparseable.csv"])
+def test_bad_csv(filename, tmpdir_factory, blank_kymo):
+    with pytest.raises(IOError, match="Invalid file format!"):
+        file = Path(__file__).parent / f"./data/{filename}"
```
alessiamarcolini, line 197:
```suggestion
        file = Path(__file__).parent / "data" / filename
```

lumicks/pylake/population/dwelltime.py
```diff
@@ -29,9 +29,16 @@ class DwelltimeBootstrap:
     lifetime_distributions: np.ndarray = field(repr=False)
 
     def __post_init__(self):
-        assert self.amplitude_distributions.shape[1] == self.lifetime_distributions.shape[1]
-        assert self.amplitude_distributions.shape[0] == self.model.n_components
-        assert self.lifetime_distributions.shape[0] == self.model.n_components
+        if self.amplitude_distributions.shape[1] != self.lifetime_distributions.shape[1]:
+            raise ValueError(
+                "Number of amplitude samples should be the same as number of lifetime samples."
```
alessiamarcolini, line 34:
```suggestion
                f"Number of amplitude samples ({self.amplitude_distributions.shape[1]}) should be the same as the number of lifetime samples ({self.lifetime_distributions.shape[1]})."
```

alessiamarcolini (changes requested):

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -78,11 +78,26 @@ def import_kymotrackgroup_from_csv(filename, kymo, channel, delimiter=";"):
...
+
+    Returns
+    -------
+    kymotrack_group : KymoTrackGroup
+        Tracked kymograph lines that were stored in the file.
+
+    Raises
+    ------
+    IOError
+        If the file format is not as we expect.
```
tobiasjj, line 90:
```suggestion
    IOError
        If the file format is not as expected.
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -630,8 +630,22 @@ def calculate_localization_error(frame_lags, msd):
         frame lags to include.
     msd : array_like
```
tobiasjj, line 631: As you changed all the other docstring types, probably this one, too?
```suggestion
    msd : np.ndarray
```

tobiasjj (approved):
Only nitpicks. LGTM! 🚴

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: I think `array_like` is good enough here. It produces the exact same result with any `array_like`.

In the return argument documentation we returned an `ndarray` (there we know what we return).
And in the other function args, we immediately applied an operation on them without invoking `np.asarray` first (which would result in incorrect behaviour and therefore the input there had to be `np.ndarray`).

PR #464, opened 2023-02-08
track_lines: two bug fixes and an improvement

JoepVanlier:
*Why this PR?*
This PR fixes a number of bugs present in `track_lines`.

Note, this PR builds on top of https://github.com/lumicks/pylake/pull/452 since it needs the bias corrected refinement.

1. `track_lines` typically returned one more track than requested.
2. `track_lines` could return non-integer time indices, which made very little sense (since they were x-coordinates along the image axis, rather than actual acquisition time points). It could also return _multiple_ points per kymograph line.
3. Because `track_lines` really looked for line structures rather than act tailored to kymographs, the endpoints were pretty arbitrary. Reason being that if a line moves downward near the end, you just get the last point on the line-like structure, even though you should get a point around the middle of that frame as the true position. This PR makes that behaviour better by performing a bias-corrected centroid refinement step at the end. While, I know that in most cases a composable API is preferable, I think that in this case, adding the centroid refinement as a core step in the algorithm helps prevent user error.

See below for an example where I deliberately made it hit into all possible failure modes described above:

track_lines_old
Note the multiple points on the right most frame and how it is off-center.

If we interpolate to integer, we get something better, but we still see that the last points are incorrect:
track_lines_new_no_refinement

Finally, with all the fixes in place, we get a good result:
track_lines_new

With these changes, it becomes a better tracking algorithm than it was:

image

I think more improvement is likely possible (but not urgent for now). For instance, X and Y are filtered with the same kernel width; despite being dimensions with a very different meaning. It seems likely that future improvements could be achieved by looking into that further.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -33,6 +33,10 @@
 * Added option to add a scale bar to by providing a [`lk.ScaleBar()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ScaleBar.html) to plotting or export functions.
 * Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 
+#### Bug fixes
+
+* Fixed incorrect behaviour in `lk.track_lines()` by interpolating back to integer frame times. Prior to this change, `lk.track_lines()` would provide a subpixel accurate position along the time axis of the kymograph as well. However, this position was specified with respect to the coordinate system of the image, rather than actual acquisition times. As such, it would produce incorrect results when performing downstream analysis that rely on the time corresponding to an actual time. Note that `lk.track_greedy()` is not affected. 
+
 ## v0.13.3 | 2023-01-26
 
 #### New features
--- a/changelog.md
+++ b/changelog.md
@@ -37,7 +37,7 @@
 #### Bug fixes
 
 * Fixed incorrect behaviour in `lk.track_lines()` by interpolating back to integer frame times. Prior to this change, `lk.track_lines()` would provide a subpixel accurate position along the time axis of the kymograph as well. However, this position was specified with respect to the coordinate system of the image, rather than actual acquisition times. As such, it would produce incorrect results when performing downstream analysis that rely on the time corresponding to an actual time. Note that `lk.track_greedy()` is not affected.
-* Fixed bug in `lk.track_lines()` which made it return one more line than requested through the parameter `max_lines`.
+* Fixed bug in `lk.track_lines()` where one extra line was returned rather than the number requested through the parameter `max_lines`.
 
 ## v0.13.3 | 2023-01-26
 
--- a/changelog.md
+++ b/changelog.md
@@ -5,7 +5,7 @@
 #### Breaking changes
 
 * When performing particle tracking on kymographs, bias correction is now enabled by default; without this correction, kymographs with high background signal will suffer from biased localization estimates. To disable bias correction, specify `bias_correction=False` to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
-* Made `lk.track_lines()` perform bias-corrected centroid refinement after tracking to improve localization accuracy. Note that the old behaviour can be recovered by passing `refine=False`.
+* `lk.track_lines()` now performs bias-corrected centroid refinement after tracking to improve localization accuracy. Note that the old behaviour can be recovered by passing `refine=False`.
 * Changed several `asserts` to `Exceptions`.
   * Attempting to read `KymoTracks` from a `CSV` file that doesn't have the expected file format will result in an `IOError`.
   * Attempting to extend `KymoTracks` by `KymoTracks` originating from a different `Kymograph` now results in a `ValueError`.
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 #### Breaking changes
 
 * When performing particle tracking on kymographs, bias correction is now enabled by default; without this correction, kymographs with high background signal will suffer from biased localization estimates. To disable bias correction, specify `bias_correction=False` to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
+* Made `lk.track_lines()` perform bias-corrected centroid refinement after tracking to improve localization accuracy. Note that the old behaviour can be recovered by passing `refine=False`.
 * Changed several `asserts` to `Exceptions`.
   * Attempting to read `KymoTracks` from a `CSV` file that doesn't have the expected file format will result in an `IOError`.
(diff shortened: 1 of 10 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -251,6 +285,7 @@ def track_lines(
     continuation_threshold=0.005,
     angle_weight=10.0,
     rect=None,
+    refine=True,
```
rpauszek, line 288: Considering that skipping this step leads to bad results, and we already automatically do this refinement for `track_greedy` I don't think we need to have this option. It keeps things simpler and more consistent I think.

 We can always add it back if needed in the future, specifically if we start using this algorithm for 2D tracking or something like that.

lumicks/pylake/kymotracker/tests/test_lines_algorithm.py
```diff
@@ -139,3 +140,72 @@ def test_lines_algorithm_input_validation(kymo_integration_test_data):
...
+
+    _|_|_|_|_|_|_|_|_|_|_|
+    _|_|_|X|_|_|_|_|_|_|_|
+    _|_|_|_|X|_|_|_|_|_|_|
+    _|_|_|_|_|X|_|_|_|_|_|
+    _|_|_|_|_|_|X|_|_|_|_|
+    _|_|_|_|_|_|_|X|_|_|_|
+    _|_|_|_|_|_|_|_|_|_|_|
+    _|_|_|_|_|_|_|X|_|_|_|
+    _|_|_|_|_|_|_|_|_|_|_|
```
rpauszek, line 181: beautiful

rpauszek (changes requested):
Nice to see this algorithm get closer to being useful. Mostly minor suggestions, but I do think it's fine here to not have the refinement be optional

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -242,6 +242,40 @@ def track_greedy(
...
+    interpolated_coordinate : np.ndarray
+    """
+
+    def round_down(x):
+        """Rounds to integer, but consistently rounds halves down."""
+        return int(np.ceil(x - 0.5))
+
+    def round_up(x):
+        """Rounds to integer, but consistently rounds halves up."""
+        return int(np.floor(x + 0.5))
```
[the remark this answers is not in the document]
JoepVanlier: You think?

I don't feel particularly strongly about it, but personally the following is less readable to me:
```python
    # Rounds integer. Here a value of 0.5 for the lower bound is consistently rounded down, while 
    # a value of 0.5 for the upper bound is consistently rounded up.
    new_time_idx = np.arange(int(np.floor(time_idx[0] + 0.5)), int(np.ceil(time_idx[-1] - 0.5)) + 1)
```

I left it as a separate fixup commit so we can still decide.

changelog.md
```diff
@@ -33,6 +34,11 @@
...
 * Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 
+#### Bug fixes
+
+* Fixed incorrect behaviour in `lk.track_lines()` by interpolating back to integer frame times. Prior to this change, `lk.track_lines()` would provide a subpixel accurate position along the time axis of the kymograph as well. However, this position was specified with respect to the coordinate system of the image, rather than actual acquisition times. As such, it would produce incorrect results when performing downstream analysis that rely on the time corresponding to an actual time. Note that `lk.track_greedy()` is not affected.
+* Fixed bug in `lk.track_lines()` which made it return one more line than requested through the parameter `max_lines`.
```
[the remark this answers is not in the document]
JoepVanlier: Looks better to me.

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: I think you need fewer points around the line when you don't do refinement, which could be a benefit. But I suppose we don't really need that right now. Amended!

lumicks/pylake/kymotracker/tests/test_lines_algorithm.py
JoepVanlier: pingu-art

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: One thing to note is that one test was using this algorithm without refinement. That test seems to find some spurious single point peaks now (that could easily be filtered with `filter_lines`). We could fix this behaviour in a future PR. I can also make an extra commit to add a `min_length` parameter to this one.

lumicks/pylake/kymotracker/kymotracker.py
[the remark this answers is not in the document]
rpauszek: I was thinking more along the lines of 
```
new_time_idx = np.arange(
    int(np.floor(time_idx[0] + 0.5)),  # consistently rounds halves down
    int(np.ceil(time_idx[-1] - 0.5)),  # consistently rounds halves up
)
```
which to me seems about as readable, but without the extraneous function calls

lumicks/pylake/kymotracker/kymotracker.py
rpauszek: we end up with single point lines from `track_greedy` so to me it's not a problem. I had suggested adding a `min_length` arg to `track_greedy` but we decided against it, I forget why though. But here, it's just consistent behaviour

lumicks/pylake/kymotracker/kymotracker.py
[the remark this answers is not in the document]
JoepVanlier: Ah, yeah, makes sense. That's quite a bit better. It missed the `+1` though, so I added that. Unfortunately it breaks the symmetry.

```
    new_time_idx = np.arange(
        int(np.floor(time_idx[0] + 0.5)),  # consistently rounds halves down
        int(np.ceil(time_idx[-1] - 0.5)) + 1,  # consistently rounds halves up and adds one
    )
```

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Right. I'll leave it as is.

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -300,8 +299,8 @@ def track_lines(
     subpixel minimum and angle between the successive normal vectors is computed. The candidate
     with the lowest score is then selected.
 
-    If desired (`refine=True`), each spot is subsequently refined by performing a bias-corrected
-    centroid optimization according to [2]_.
+    Each spot is subsequently refined by performing a bias-corrected centroid optimization
+    according to [2]_.
```
rpauszek, line 303: I think the information here can be combined into line 289. As is, it seems a bit repetitive

changelog.md
```diff
@@ -33,6 +34,11 @@
...
 * Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 
+#### Bug fixes
+
+* Fixed incorrect behaviour in `lk.track_lines()` by interpolating back to integer frame times. Prior to this change, `lk.track_lines()` would provide a subpixel accurate position along the time axis of the kymograph as well. However, this position was specified with respect to the coordinate system of the image, rather than actual acquisition times. As such, it would produce incorrect results when performing downstream analysis that rely on the time corresponding to an actual time. Note that `lk.track_greedy()` is not affected.
+* Fixed bug in `lk.track_lines()` one extra line was returned rather than the number requested through the parameter `max_lines`.
```
rpauszek, line 40:
```suggestion
* Fixed bug in `lk.track_lines()` where one extra line was returned rather than the number requested through the parameter `max_lines`.
```

rpauszek (approved):
looks good, just 2 minor comments on phrasing

PR #466, opened 2023-02-15
force_calibration: refactor the docs

JoepVanlier:
**Why this PR?**
The docs were getting unwieldly. The tutorial delved too deeply into the theory and there was no theory section to speak of.

This PR attempts four things:
- Split the more equation heavy parts of the theory away from the tutorial so that the tutorial can really be focused on how to use the calibration module rather than all the underlying theory.
- Added a small preface section to the theory that aims to build a little intuition before delving into the math.
- Base the tutorial on real files such that the examples for a regular calibration, a height calibration and a fast sensor calibration are now executable (the raw data is in the package that we intend to upload to Zenodo). Especially the section with the fast sensor calibration will be nice to motivate this model for users that are still on the fence whether they should use hydro for large beads.
- While doing this, I found a bug in the Jupyter notebook rendering. In `note` and `warning` items, hyperlinks were broken (because they should be specified in html once you go into a `<div>` environment). I kept it as a separate commit.

Of note:
- In the hope of getting nice vector images, I originally tried to replace the PNGs with SVGs, but the rendering looked terrible on Chrome, so I reverted it back.

Docs link [here](https://lumicks-pylake.readthedocs.io/en/shuffle_force_calibration/tutorial/force_calibration.html) and [here](https://lumicks-pylake.readthedocs.io/en/shuffle_force_calibration/theory/force_calibration/force_calibration.html).


docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-Note that `pylake` always returns the bulk drag coefficient :math:`\gamma_0`.
+    force_model = lk.PassiveCalibrationModel(bead_diameter, hydrodynamically_correct=False, distance_to_surface=5)
 
-Axial Calibration
------------------
+To summarize, the workflow can be visualized as follows:
 
-For calibration in the axial direction, no hydrodynamically correct theory exists.
-In this case, one should use a Lorentzian with a specific correction term :cite:`schaffer2007surface`:
+.. image:: figures/force_calibration/surface_calibration_workflow.png
```
agom, line 162: I think a visual representation of the of the bead/surface would be good to explain which method to use in which situation, something like this
Screenshot 2023-02-28 at 12 06 45
We could also use it in the bluelake docs.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-Note however, that this makes using the hydrodynamically correct model critical, as the Lorentzian doesn't actually capture the data very well.
-This is illustrated in the figure below.
-Here we see the same power spectrum (acquired on a fast detector) fitted with three different models.
+- The diameter of the bead :math:`d` in microns.
+- The dynamic viscosity :math:`\eta` in Pascal seconds.
+- The distance to the surface :math:`h` in microns.
 
-.. image:: figures/force_calibration/fast_hydro.png
+The viscosity in turn depends strongly on the local temperature around the bead, which is typically poorly known.
+In active calibration, an extra peak with a known amplitude is introduced into the power spectrum which allows the trap to be calibrated without prior knowledge of the drag coefficient.
```
rpauszek, line 189:
```suggestion
The viscosity in turn depends strongly on the local temperature around the bead, which is typically poorly known.

In active calibration, we oscillate the stage with a known frequency and amplitude. This introduces an extra peak in the power spectrum which allows the trap to be calibrated without the assumptions of the theoretical drag coefficient.
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-------------------
+    f = lk.File("near_surface_active_calibration.h5")
+    volts = decalibrate(f.force1x)
+    bead_diameter = f.force1x.calibration[0]["Bead diameter (um)"]
+    # Calibration performed at 1.04 * bead_diameter
+    distance_to_surface = 1.04 * bead_diameter
 
-For certain applications, passive force calibration, as described above, is not sufficiently accurate.
-Using active calibration, the accuracy of the calibration can be improved, because active calibration uses fewer assumptions than passive calibration.
+Extract the nanostage data::
```
rpauszek, line 204:
```suggestion
First we need to extract the nanostage data which is used to determine the driving amplitude and frequency::
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
 
-When performing passive calibration, we base our calculations on a theoretical drag coefficient.
-This theoretical drag coefficient depends on parameters that are only known with limited precision: the diameter of the bead and the viscosity.
-This viscosity in turn depends strongly on the local temperature around the bead, which is typically poorly known.
+    driving_data = f["Nanostage position"]["X"]
 
-During active calibration, the trap or nanostage is oscillated sinusoidally.
-These oscillations result in a driving peak in the force spectrum.
-Using power spectral analysis, the force can then be calibrated without prior knowledge of the drag coefficient.
+For an active calibration, we should see an oscillation here.
```
rpauszek, line 208:
```suggestion
For data acquired with active calibration in Bluelake, this will be a sinusoidal oscillation.
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-The physical spectrum is then given by a thermal part (like before) and an active part:
+Instead of using the :class:`~lumicks.pylake.PassiveCalibrationModel` presented in the previous section, we now use the :class:`~lumicks.pylake.ActiveCalibrationModel`.
+We also need to provide the sample rate at which the data was acquired, and a rough guess for the driving frequency.
+`pylake` will find an accurate estimate of the driving frequency based on this initial estimate (provided that it is close enough)::
 
-.. math::
+    active_model = lk.ActiveCalibrationModel(driving_data.data, volts.data, driving_data.sample_rate, bead_diameter, driving_frequency_guess=38, distance_to_surface=distance_to_surface)
 
-    P^\mathrm{thermal}(f) = \frac{D}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)}
+To check the determined frequency, we can look at the determined driving frequency::
```
rpauszek, line 222:
```suggestion
We can check the determined driving frequency with::
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+    show_peak = lk.calculate_power_spectrum(volts.data, sample_rate=78125, num_points_per_block=5, fit_range=(10, 23000))
+    show_peak.plot()
 
-If we use the basic Lorentzian model, then the theoretical power (integral over the delta spike) corresponding to the driving input is given by :cite:`tolic2006calibration`:
+.. image:: figures/force_calibration/calibration_peak.png
 
-.. math::
+The driving peak is clearly visible in the spectrum.
+Let's calculate the power spectrum we'll use for fitting.
+It is important to _not_ include the driving peak when doing this (the default will only go up to 100 Hz)::
```
rpauszek, line 237:
```suggestion
Next let's calculate the power spectrum we'll use for fitting.
It is important to *not* include the driving peak when doing this (the default will only go up to 100 Hz)::
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
 
-:math:`\Delta f` refers to the width of one spectral bin.
-Here the thermal contribution that needs to be subtracted is obtained from fitting the thermal part of the spectrum using the passive calibration procedure from before.
-The desired positional calibration is then:
+Analogous to the passive calibration procedure, we can specify `hydrodynamically_correct=True` if we wish to use the hydrodynamically correct theory here.
+Especially for bigger beads this is highly recommended (more on this later).
 
-.. math::
+More convenient calibration
+---------------------------
```
rpauszek, line 252: I don't know there's a good way to do this, but I feel this section could be almost at the top. It'd be nice right away to show users that they don't *need* to go through the hassle of setting up the models manually. 

Perhaps move the first half of this section toward the top as a TLDR-ish part and then take the 2nd half here comparing different models as a "Comparison" section here still?

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
 
-    R_d = \sqrt{\frac{W_\mathrm{physical}}{W_\mathrm{measured}}} \quad \mathrm{[m/V]}
+For convenience, `pylake` offers :func:`~lumicks.pylake.calibrate_force()` which performs the entire calibration procedure in a single function call::
```
rpauszek, line 254: sometimes you use "Pylake" and sometimes `pylake`; I guess we should make a decision on what we want to use and be consistent throughout the docs (at a later time) but we can start with this tutorial

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+    0.11662183772410809
 
-We can now use this model to fit the power spectrum::
+.. note::
 
-    calibration = lk.fit_power_spectrum(power_spectrum, active_model)
-    calibration
+    When hydrodynamic correction is enabled, the height parameter impacts both passive and active calibration.
+    The reason for this is that when using the more accurate theory, the expected shape of the power spectrum depends on the distance to the surface.
+    For more information on this see the :doc:`theory section on force calibration</theory/force_calibration/force_calibration>` section.
```
rpauszek, line 304: this seems a bit strange. In the first sentence you talk about hydrodynamic correction, but in the 2nd distance to surface; the connection isn't really clear

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-And that's all there is to it.
+Fast Sensors
+------------
 
-Analogously, we can specify `hydrodynamically_correct=True` if we wish to use the hydrodynamically correct theory here.
-This fits the thermal part with the hydrodynamically correct power spectrum and also uses a hydrodynamically correct model for the peak:
+By default, the models used always incorporate a filtering effect.
+This filtering effect occurs because a fraction of the light is not measured instantaneously by most detectors.
+Fast detectors have the ability to respond much faster to incoming light resulting in no visible filtering effect in the frequency range we are fitting.
+This means that for a fast detector, we do not need to include a filtering effect in our model.
```
rpauszek, line 312: If you go with my suggestion for adding an intro section, we don't need the explanation of the filtering effect here.
```suggestion
Fast detectors have the ability to respond much faster to incoming light resulting in no visible filtering effect in the frequency range we are fitting.
This means that for a fast detector, we do not need to include a filtering effect in our model.
```
We might also want to explicitly state that this depends on the particular hardware in the C-Trap (it's implied here, but let's make it super clear).

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+We can omit this effect by passing `fast_sensor=True` to the calibration models or to :func:`~lumicks.pylake.calibrate_force()`.
+Note however, that this makes using the hydrodynamically correct model critical, as the Lorentzian doesn't actually capture the data very well.
```
rpauszek, line 314: I think this situation needs a bit more explanation; I don't quite get why hydro is more important and if it's dependent on active vs passive

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+
+.. image:: figures/force_calibration/fast_sensors.png
+
+Note how the fit with the Lorentzian model with the filtering effect enabled seems to fit the data quite well.
+However, the apparent quality of a fit can be deceiving.
+When the `fast_sensor` flag is enabled, it can be seen that the Lorentzian model doesn't actually describe the data.
+Switching to the hydrodynamically correct model results in a superior fit to the power spectrum.
+
+In the case of the Lorentzian with a detector filtering effect, the fitting procedure has used the parameters of the filter to fit some of the high frequency attenuation.
+With the diode model disabled, we obtain a very biased fit.
```
rpauszek, line 351: here are you using "Lorentzian" to refer to passive calibration? If so, I think we should be consistent and just use passive calibration throughout

rpauszek (changes requested):
This is looking really great; definitely better to have the theory parts split off rather than intermingled. There's quite a few suggestions here for, IMO, some different organization to give a better flow and reading experience.

So far I've gone through the tutorial part, I'll do a separate set of comments for the theory section next

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-    volts = (force_slice - offset) / response
-
-Force calibration models are fit to power spectra. To compute a power spectrum from our data::
-
-    power_spectrum = lk.calculate_power_spectrum(volts.data, sample_rate=volts.sample_rate)
+Force calibration refers to computing the conversion factors needed to convert the raw voltages recorded by the position sensitive detectors to actual forces and displacements.
+The values we wish to calculate are:
 
-This function returns a power_spectrum which we can plot::
+- Trap stiffness :math:`\kappa`, which tells us how stiff the trap is.
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, good call, I'd say it's a little stronger than just related to though.

Perhaps the following?
```
Trap stiffness :math:`\kappa`, which reflects how strongly a bead is held by a trap
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+- Force response :math:`R_d`, which converts from voltage to force.
+- Distance response :math:`R_f`, which converts from voltage to distance.
 
-    power_spectrum.plot()
+This tutorial will focus on performing force calibration with Pylake.
```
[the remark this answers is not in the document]
JoepVanlier: Yes, good point. Added.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+This tutorial covers calibration using Pylake.
+Our starting point will be data acquired with Bluelake, which is already calibrated.
+Therefore, the first step is to undo the calibration applied by Bluelake to get the raw voltage signals.
+Let's load the dataset::
 
-Additional parameters can be specified to change the amount of downsampling applied (`num_points_per_block`), the range over which to compute the spectrum (`fit_range`) or to exclude specific frequency ranges from the spectrum (`excluded_ranges`)::
+    f = lk.File("passive_calibration.h5")
 
-    power_spectrum = lk.calculate_power_spectrum(volts.data, sample_rate=volts.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=[(700, 800), (14500, 14600)])
+We grab the chunk of data that was used for calibration last time::
```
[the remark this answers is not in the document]
JoepVanlier: I don't think this is correct actually. It contains only one calibration section. The zeroth is always the calibration that was active prior to the slice (so that the voltages can be recovered). There is only one actual calibration here.

Still, I think making this more explicit like you have is good. So how about the following?

```
The force timeline in this file contains a section used for calibration.
We can see calibrations relevant to this file by inspecting the `calibration` attribute for the entire force :class:`~lumicks.pylake.channel.Slice`.
This is a list of dictionaries containing all of the relevant calibration parameters::

    print(len(f.force1x.calibration))
    print(f.force1x.calibration[1])

The first calibration always refers to the calibration that was active at the start of the slice (usually from a calibration that occurred before the start of the slice).
Our goal here is to recalibrate the data in the calibration section (corresponding to `calibrations[1]`).
First we grab the chunk of data that was used for that calibration in Bluelake::
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
 
-    D_\mathrm{physical} = \frac{k_B T}{\gamma_0} \quad \mathrm{[m^2/s]}
+.. image:: figures/force_calibration/bl_dictionary.png
 
-Here :math:`k_B` is the Boltzmann constant and :math:`T` is the local temperature in Kelvin. Comparing this to its measured counterpart in volts squared per second provides us with the desired calibration factor:
+Here, they are listed as `Exclusion range 0 (min.) (Hz)`, `Exclusion range 0 (max.) (Hz)` etc.
+To reproduce the result obtained with Bluelake, these should be excluded from the power spectrum::
 
-.. math::
+    power_spectrum = lk.calculate_power_spectrum(volts.data, sample_rate=volts.sample_rate, fit_range=(1e2, 23e3), num_points_per_block=2000, excluded_ranges=([19348, 19668], [24308, 24548]))
```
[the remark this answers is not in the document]
JoepVanlier: I was also wondering about that actually. I will fix it up.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+Considering that a noise peak may be very narrow, it is beneficial to lower the number of points per block temporarily to find the exact exclusion range.
+After determination of this exclusion range, the number of points per block can be increased again.
 
-.. math::
+Passive calibration
+-------------------
 
-    \gamma_0 = 3 \pi \eta d \quad \mathrm{[kg/s]}
+In literature, passive calibration is often referred to as thermal calibration.
+It involves fitting a physical model to the power spectrum obtained in the previous step.
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, I kept it generic on purpose, since I wanted to keep the terminology to a minimum. It's also not just a Lorentzian. It's a Lorentzian times a filtering function in the simplest case. I'd rather rephrase it a bit in the fast sensor section to not mention the Lorentzian, but refer to it as the simple model. What do you think?

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-Note that in `pylake` we actually use `microns` to specify the bead diameter for convenience.
+The most important of these is the bead diameter (in microns).
+Let's use the bead diameter found in the calibration performed in Bluelake.
 
-We use the bead diameter found in the calibration performed in Bluelake.
-You can optionally also provide a viscosity (in Pa/s) and temperature (in degrees Celsius)::
+You can optionally also provide a viscosity (in Pa/s) and temperature (in degrees Celsius).
+To find the viscosity of water at a particular temperature, Pylake uses :func:`~lumicks.pylake.viscosity_of_water` which implements the model presented in :cite:`huber2009new`.
+When omitted, this function will automatically be used to look up the viscosity of water for that particular temperature
+Note that for experiments that use a different medium than water, the viscosity at the experimental temperature should explicitly be provided.
```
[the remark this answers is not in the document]
JoepVanlier: Yes.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
-
-Faxen's law
------------
-
-The hydrodynamically correct model presented in the previous section works well when the bead center is at least 1.5 times the radius above the surface.
+.. note::
 
-When going closer, the drag effect becomes stronger than the frequency dependent effects and better models to approximate the local drag exist.
+    One thing to note is that when using the hydrodynamically correct model, the equation for the drag does include the viscosity and bead diameter.
+    However, they now appear in a term which already amounts to a small correction therefore the impact of any errors in these is reduced :cite:`tolic2006calibration`.
```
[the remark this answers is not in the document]
JoepVanlier: Yes. I think we can actually omit this statement. It is not really relevant for the tutorial, and more for the theory section.

docs/tutorial/force_calibration.rst
JoepVanlier: 🙌

docs/tutorial/force_calibration.rst
JoepVanlier: Good point. I doubt with every new instance of Pylake in the docs. I'll make it consistent.

docs/tutorial/force_calibration.rst
JoepVanlier: Yes. I agree. I think inverting the argument may be better here. How about:
```
    When fitting with the hydrodynamically correct model, the `distance_to_surface` parameter impacts the expected shape of the power spectrum.
    Consequently, when this model is selected, this parameter affects both passive and active calibration.
    For more information on this see the :doc:`theory section on force calibration</theory/force_calibration/force_calibration>` section.
```

docs/tutorial/force_calibration.rst
JoepVanlier: No, I use Lorentzian to mean the simple model. I think we should actual hold off on calling it a Lorentzian in the tutorial (and save that for the theory section). I've started referring to it as simple model.

I have also updated the description a bit, and added a question. While I normally try to avoid adding questions like this in explanations, I think in this case it is warranted:

```
Note how the power spectral fit with the simple model seems to fit the data quite well as long as we also include the filtering effect.
However, the apparent quality of a fit can be deceiving.
Considering that this dataset was acquired on a fast sensor, we should omit the filtering effect.
When the `fast_sensor` flag is enabled, it can be seen that the simple model doesn't actually describe the data.
Switching to the hydrodynamically correct model results in a superior fit to the power spectrum.

So what is happening here? Why did the first fit look good?
When we fit the power spectrum with the simple model and include the filtering effect, the fitting procedure uses the parameters that characterize the filter to fit some of the high frequency attenuation.
With the filtering effect disabled, we obtain a very biased fit because the model fails to fit the data.
```

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+    W_\mathrm{physical} = \frac{A^2}{2\left(1 + \frac{f_c^2}{f_\mathrm{drive}^2}\right)} \quad \mathrm{[m^2]}
+
+Subtracting the thermal part of the spectrum, we can determine the same quantity experimentally.
+
+.. math::
+
+    W_\mathrm{measured} = \left(P_\mathrm{measured}^\mathrm{total}(f_\mathrm{drive}) - P_\mathrm{measured}^\mathrm{thermal}(f_\mathrm{drive})\right) \Delta f \quad \mathrm{[V^2]}
+
+:math:`\Delta f` refers to the width of one spectral bin.
```
rpauszek, line 332:
```suggestion
where :math:`\Delta f` refers to the width of one spectral bin.
```

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+Analogously to passive calibration, there is also a hydrodynamically correct theory for active calibration which should be used when inertial forces cannot be neglected.
+This involves fitting the thermal spectrum with the hydrodynamically correct power spectrum discussed earlier, but also requires using a hydrodynamically correct model for the peak:
+
+.. math::
+
+    P_\mathrm{hydro}^\mathrm{active}(f) = \delta \left(f - f_\mathrm{drive}\right) \frac{\left(A f_\mathrm{drive} \left|\gamma / \gamma_0\right|\right)^2}{2 \left(\left(f_{c,0} + f \mathrm{Im}(\gamma/\gamma_0) - f^2/f_{m, 0}\right)^2 + \left(f \mathrm{Re}(\gamma / \gamma_0)\right)^2\right)}
+
+We can also include a distance to the surface like before.
+This results in an expression for the drag coefficient :math:`\gamma` that depends on the distance to the surface.
+This drag coefficient is given by the same equations as listed in the section on the hydrodynamically correct model.
```
rpauszek, line 357:
```suggestion
This results in an expression for the drag coefficient :math:`\gamma` that depends on the distance to the surface which is given by the same equations as listed in the section on the hydrodynamically correct model.
```

rpauszek (commented):
Really nice work on the theory section! Plenty of suggestions, but here they're all pretty minor

docs/tutorial/force_calibration.rst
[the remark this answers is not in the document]
rpauszek: maybe could work. You use "Lorentzian" a lot in the theory section too. So either replacing the term everywhere or defining it somewhere I'd be ok with

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
 
-More convenient calibration
-^^^^^^^^^^^^^^^^^^^^^^^^^^^
+High corner frequencies
+-----------------------
 
-For convenience, we also provide a function named :func:`~lumicks.pylake.calibrate_force` which executes the entire calibration procedure::
+In specific situations, the filtering effect of the position sensitive detector can cause issues when calibrating.
+The power spectrum of the bead in the optical trap is characterized by a diffusion constant and a corner frequency.
+The filtering effect is characterized by a constant that reflects the fraction of light that is transmitted instantaneously and a corner frequency (referred to as the diode frequency to be able to differentiate).
```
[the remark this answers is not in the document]
JoepVanlier: I like it. I'd prefer actually having part of the section you propose here at the top but with a few changes.
The actual models should not be explained in the tutorial. Those are in the theory section. Therefore I would like to suggest the following amendment (basically refer them to the theory section for the models and the tut for their use):

```
To calibrate the optical traps we fit a power spectrum of the voltage measured by a position sensitive diode (PSD) to a theoretical model. The measured voltage (and thus the shape of the power spectrum) is mainly determined by

1. The Brownian motion of the bead within the trap.
2. The instantaneous response of the PSD to the incident light.

The contribution to the power spectrum by the bead in the optical trap is characterized by a diffusion constant and a corner frequency.
The second contribution refers to a filtering effect where the PSD becomes less sensitive to changes in signal at high frequencies.
This filtering effect is characterized by a constant that reflects the fraction of light that is transmitted instantaneously and a corner frequency (this is referred to as the diode frequency to be able to differentiate).
The various theoretical models that can be used to fit these data are described in the :doc:`theory section on force calibration</theory/force_calibration/force_calibration>`, while their use is described below.

```

docs/tutorial/force_calibration.rst
JoepVanlier: I felt the same way when writing this but held off on moving it because I didn't want to fracture the introduction and use of the convenience function. I do agree though, and will attempt to move it up.

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+    P_{diffusion}(f) = \frac{D}{\pi^2 f^2}
+
+We can plot what this looks like for different diffusion constants::
+
+    def set_plot_style():
+        plt.xscale("log")
+        plt.yscale("log")
+        plt.xlabel("Frequency [Hz]")
+        plt.ylabel("Amplitude [$\mu m^2$/s]");
+        plt.legend()
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, I droped it. I get a bit triggerhappy with defining these. :D

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+Passive calibration is also often referred to as thermal calibration and involves calibration without moving the trap or stage.
+In passive calibration, the Brownian motion of the bead in the trap is analyzed in order to find calibration factors for both the positional detection as well as the force.
+
+To fit passive calibration data, we will use a model based on a number of publications by the Flyvbjerg group :cite:`berg2004power,tolic2004matlab,hansen2006tweezercalib,berg2006power`.
+This model can be found in :class:`~.PassiveCalibrationModel`.
+The most basic form of passive calibration starts by fitting the following equation to the power spectrum:
+
+.. math::
+
+    P(f) = \frac{D_\mathrm{measured}}{\pi ^ 2 \left(f^2 + f_c ^ 2\right)} g(f, f_\mathrm{diode}, \alpha) \quad \mathrm{[V^2/Hz]}
```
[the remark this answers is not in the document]
JoepVanlier: Hmm. Personally, I quite like having them there, since it makes it more explicit what is in what units (which is important since we're talking about calibration factors here). I wonder if we can improve it with some formatting changes. Though I wouldn't know what.

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+where :math:`D_\mathrm{measured}` corresponds to the diffusion constant (in `volts`), :math:`f` the frequency and :math:`f_c` the corner frequency.
```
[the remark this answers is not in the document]
JoepVanlier: No, it is `V^2/s`. I meant to state that it's in voltage as opposed to a unit of length, but writing it like this is indeed unhelpful. Changing it to `V^2/s`.

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+    g(f, f_\mathrm{diode}, \alpha) = \alpha^2 + \frac{1 - \alpha ^ 2}{1 + (f / f_\mathrm{diode})^2}
+
+Here :math:`\alpha` corresponds to the fraction of the signal response that is instantaneous, while :math:`f_\mathrm{diode}` characterizes the frequency response of the diode.
+Note that not all sensor types require this second term.
+
+To convert the parameters obtained from this spectral fit to a trap stiffness, the following is computed:
+
+.. math::
+
+    \kappa = 2 \pi \gamma_0 f_c \quad \mathrm{[N/m]}
```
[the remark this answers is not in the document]
JoepVanlier: Personally, I like 'em by the side, since it makes what goes with what very clear imho. We could consider increasing the spacing; but it might start looking weird at some point as well.

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Here :math:`\kappa` then represents the final trap stiffness.
```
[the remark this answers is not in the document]
JoepVanlier: Yes XD

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+Moving closer than this limit, we fall back to a model that more accurately describes the change in drag at low frequencies, but neglects the frequency dependent effects.
+
+For lateral calibration, the following approximation is typically used :cite:`schaffer2007surface`:
+
+.. math::
+
+    \gamma_\mathrm{faxen}(R/l) = \frac{\gamma_0}{
+        1 - \frac{9R}{16l} + \frac{1R^3}{8l^3} - \frac{45R^4}{256l^4} - \frac{1R^5}{16l^5}
+    }
```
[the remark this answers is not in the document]
JoepVanlier: Added: The model that we use in this case is Faxen’s law for the approximate drag on a sphere near a surface under creeping flow conditions.

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+        1.0
+        - \frac{9R}{8l}
+        + \frac{1R^3}{2l^3}
+        - \frac{57R^4}{100l^4}
+        + \frac{1R^5}{5l^5}
+        + \frac{7R^{11}}{200l^{11}}
+        - \frac{1R^{12}}{25l^{12}}
+    }
+
+This model deviates less than 0.1% from Brenner's exact formula for :math:`l/R >= 1.1` and less than 0.3% over the entire range of :math:`l` :cite:`schaffer2007surface`:.
```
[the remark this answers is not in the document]
JoepVanlier: Yes, I agree.

```
Similarly as for the lateral component, we will fall back to a model that describes the change in drag at low frequencies.
However, while we had a simple expression for the lateral drag as a function of distance, no simple closed-form equation exists for the axial dimension.
Brenner et al provide an exact infinite series solution :cite:`brenner1961slow`.
Based on this solution :cite:`schaffer2007surface` derived a simple equation which approximates the distance dependence of the axial drag coefficient.
```

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+    P^\mathrm{active}(f) = \frac{A^2}{2\left(1 + \frac{f_c^2}{f_\mathrm{drive}^2}\right)} \delta(f - f_\mathrm{drive})
+
+Added together, these give rise to the full power spectrum:
+
+.. math::
+
+    P^\mathrm{total}(f) = P^\mathrm{thermal}(f) + P^\mathrm{active}(f)
+
+Since we know the driving amplitude, :math:`A`, we know how the bead reacts to the driving motion and we can observe this response in the power spectral density (PSD), we can use this relation to determine the positional calibration.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+- The first is to split the time series into windows of equal length, compute the power spectrum for each chunk of data and averaging these. This procedure is referred to as windowing.
+- The second is to calculate the spectrum for the full dataset and downsampling the obtained spectrum in the spectral domain by averaging adjacent bins according to :cite:`berg2004power`. This procedure is referred to as blocking.
+
+We use the blocking method for spectral averaging.
+This allows us to reject noise peaks at high resolution prior to blocking.
+Note however, that the error incurred by this blocking procedure depends on :math:`n_b`, the number of points per block, :math:`\Delta f`, the spectral resolution and inversely on the corner frequency :cite:`berg2004power`.
+
+Setting the number of points per block too low would result in a bias from insufficient averaging :cite:`berg2004power`.
+In practice, one should use a high number of points per block (:math:`n_b >> 100`), unless a very low corner frequency precludes this.
```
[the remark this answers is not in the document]
JoepVanlier: It did! :)

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+- Trap stiffness :math:`\kappa`, which tells us how stiff the trap is.
+- Force response :math:`R_d`, which converts from voltage to force.
+- Distance response :math:`R_f`, which converts from voltage to distance.
 
-    power_spectrum.plot()
+This tutorial will focus on performing force calibration with Pylake.
+For the theoretical background on the physics of force calibration, please refer to :doc:`theory section on force calibration</theory/force_calibration/force_calibration>`.
 
-.. image:: figures/force_calibration/force_calibration_blocked_spectrum.png
+Undoing the previous calibration
```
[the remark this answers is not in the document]
JoepVanlier: I have moved part of it up, and included your description.

I did add a little warning box in this section though to at least alert users that there could be an issue there when `f_c` approaches `f_diode`.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,375 @@
...
+The corner frequency of the physical spectrum can be found in the results as `f_c` and depends on the laser power and bead size (smaller beads resulting in higher corner frequencies) .
+The corner frequency of the filtering effect can be found in the results as `f_diode` (which stands for diode frequency) and depends on the incident intensity :cite:`berg2003unintended`.
+When these two frequencies get close, they cannot be determined reliably anymore.
+The reason for this is that the effect of one can be compensated by the other.
+When working with small beads or at high laser powers, it is important to verify that the corner frequency `f_c` does not approach the frequency of the filtering effect `f_diode`.
 
-    fit = lk.calibrate_force(volts.data, bead_diameter, temperature, sample_rate=volts.sample_rate, active_calibration=True, hydrodynamically_correct=True, fit_range=(10e2, 20e3))
+Sometimes, the filtering effect has been characterized independently.
+In that case, the arguments `fixed_diode` and `fixed_alpha` can be passed to :func:`~lumicks.pylake.calibrate_force()` to fix these parameters to their predetermined values.
```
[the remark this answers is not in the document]
JoepVanlier: I've added a warning box at the top regarding the `f_c`, `f_diode` thing.

docs/tutorial/force_calibration.rst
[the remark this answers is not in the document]
JoepVanlier: I've explicitly adding a sentence defining it in the theory section. I think using terminology in the section where people are interested in theory is more acceptable, no?

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,357 @@
...
+
+    \gamma(f, R/l) = \frac{\gamma_\mathrm{stokes}(f)}{1 - \frac{9}{16}\frac{R}{l}\left(1 - \left((1 - i)/3\right)\sqrt{\frac{f}{f_{\nu}}} + \frac{2}{9}\frac{f}{f_{\nu}}i - \frac{4}{3}(1 - e^{-(1-i)(2l-R)/\delta})\right)} \quad \mathrm{[kg/s]}
+
+Where :math:`\delta = R \sqrt{\frac{f_{\nu}}{f}}` represents the aforementioned penetration depth, :math:`R` corresponds to the bead radius and :math:`l` to the distance from the bead center to the nearest surface.
+
+While these models may look daunting, they are all available in `pylake` and can be used by simply providing a few additional arguments to the :class:`~.PassiveCalibrationModel`.
+It is recommended to use these equations when less than 10% systematic error is desired :cite:`tolic2006calibration`.
+
+The figure below shows the difference between the hydrodynamically correct spectrum and the regular Lorentzian for various bead sizes.
+It can be seen that for large bead sizes and higher trap powers the differences can be substantial.
```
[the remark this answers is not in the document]
JoepVanlier: Updated.

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,440 @@
...
+This is a list of dictionaries containing all of the relevant calibration parameters::
 
-We can calibrate the position by considering the diffusion of the bead:
+    print(len(f.force1x.calibration))
+    print(f.force1x.calibration[1])
 
-.. math::
+The first calibration always refers to the calibration that was active at the start of the slice (usually from a calibration that occurred before the start of the slice).
+Our goal here is to recalibrate the data in the calibration section (corresponding to `calibrations[1]`).
+First we grab the chunk of data that was used for that calibration in Bluelake::
```
rpauszek, line 56: yeah I still find this confusing. I think it could help to use the terms "calibration measurement" and "calibration applied to ....", part of the confusion here I think is just re-used terminology so being more explicit in this case I think is helpful.

But what really made it clear to me was plotting things out. could add a code block to make a figure like this:

image

the code I used is as follows (but I dunno if it's still too complicated to put in the tutorial, but I don't see a significantly better way to clean it up)

```
force = f.force1x

# final timestamp in data relative to first timestamp, in seconds
stop = (force.timestamps[-1] - force.timestamps[0]) * 1e-9

# measurement timestamp ranges for each calibration relative to first timestamp, in seconds
calibrations = [
    (np.array([cal["Start time (ns)"], cal["Stop time (ns)"]]) - force.timestamps[0]) * 1e-9
    for cal in force.calibration
]

plt.figure()
force.plot()
plt.axvspan(*calibrations[0], alpha=0.5, fc="tab:blue", label="calibration[0] measurement")
plt.axvspan(calibrations[0][1], calibrations[1][0], alpha=0.2, fc="tab:blue", label="calibration[0] applied")
plt.axvspan(*calibrations[1], alpha=0.5, fc="tab:orange", label="calibration[1] measurement")
plt.axvspan(calibrations[1][1], stop, alpha=0.2, fc="tab:orange", label="calibration[0] applied")

plt.legend()
plt.margins(x=0)
# plt.xlim(0, stop)  # uncomment this line to zoom in to data
    
plt.tight_layout()
plt.show()
```

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,440 @@
...
-To fit this model to the data, you can now invoke::
+The rest of this tutorial illustrates the various steps involved when performing such a calibration.
 
-    calibration = lk.fit_power_spectrum(power_spectrum, force_model)
-    calibration
+Obtaining the power spectrum
+----------------------------
 
-This will produce a table with your fitted calibration parameters.
+Next, we need the power spectrum to fit. To compute a power spectrum from our data we can invoke :func:`~lumicks.pylake.calculate_power_spectrum()`::
```
rpauszek, line 96: considering the change above, "next" doesn't make sense in this context. We could do something like:
```suggestion
To use the more manual lower-level API, we first need the power spectrum to fit. To compute a power spectrum from our data we can invoke :func:`~lumicks.pylake.calculate_power_spectrum()`::
```

docs/theory/force_calibration/force_calibration.rst
```diff
@@ -0,0 +1,363 @@
...
+
+    plt.xlabel("Frequency [Hz]")
+    plt.ylabel("Amplitude [$\mu m^2$/s]");
+    plt.legend()
+
+    plt.subplot(1, 2, 2)
+    diffusion, corner_freq = 1000, 1000
+    for corner_freq in [1000, 5000, 10000]:
+        line, = plt.loglog(
+            f, diffusion / (np.pi * (f**2 + corner_freq**2)), label=f"corner_freq={corner_freq} Hz"
```
rpauszek, line 79: I think here for the label we should either use "corner frequency" or `$f_c$` for consistency

rpauszek (changes requested):
Definitely getting better. There's just a couple extra suggestions that I think will really help improve clarity

docs/tutorial/force_calibration.rst
JoepVanlier: hydro-guide
Courtesy of Thana :)

docs/tutorial/force_calibration.rst
```diff
@@ -1,380 +1,440 @@
...
-Passive calibration is also often referred to as thermal calibration.
-This model can be found in :class:`~.PassiveCalibrationModel`. It is calibrated by fitting the following equation to the power spectrum:
+1. The Brownian motion of the bead within the trap.
+2. The instantaneous response of the PSD to the incident light.
 
-.. math::
+The contribution to the power spectrum by the bead in the optical trap is characterized by a diffusion constant and a corner frequency.
+The second contribution refers to a filtering effect where the PSD becomes less sensitive to changes in signal at high frequencies.
+This filtering effect is characterized by a constant that reflects the fraction of light that is transmitted instantaneously and a corner frequency (this is referred to as the diode frequency to be able to differentiate).
+The various theoretical models that can be used to fit these data are described in the :doc:`theory section on force calibration</theory/force_calibration/force_calibration>`, while their use is described below.
```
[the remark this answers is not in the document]
JoepVanlier: Added.

docs/tutorial/force_calibration.rst
JoepVanlier: While I like the idea of having this in the docs, I fear it'll distract from the force calibration part of the docs, because people might get hung up on the complexity.

Instead I made the following image, which I propose we use instead (along with some further minor improvements on the wording, differentiating calibration from measurement).

calibrations

rpauszek (approved):
🙌 🚀 good to go!

PR #468, opened 2023-02-23
`KymoTrackGroup`: more options for finding and removing tracks

JoepVanlier:
**Why this PR?**
It adds some small convenience features.

- It allows a `KymoTrack` to be removed from the group with `.remove()`.
- It allows `KymoTrackGroup` to be indexed with a boolean array. This can be convenient when one wants to remove a whole bunch of elements at the same time.
- It removes `_concatenate_tracks`, which was unused internal api.
- Added private API for looking up a particular track in the group by it's python `id`.

Relevant section of the docs build [here](https://lumicks-pylake.readthedocs.io/en/no_track_uuids/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.html#lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup)

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -35,6 +35,7 @@
 * Added option to add a scale bar to by providing a [`lk.ScaleBar()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ScaleBar.html) to plotting or export functions.
 * Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
 * Added `KymoTrackGroup.remove()` to remove a `KymoTrack` from a `KymoTrackGroup`.
+* Allow boolean array indexing (e.g. `tracks[[False, False, True]]`) and indexing with arrays of indices (e.g. `tracks[[1, 3]]`) for `KymoTrackGroup`. See the [API documentation](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.html#lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup) for more information.
 
 #### Bug fixes
 
--- a/changelog.md
+++ b/changelog.md
@@ -28,6 +28,7 @@
   * Widefield: Attempting to open multiple `TIFF` as a single ImageStack will now raise a `ValueError` if the alignment matrices of the individual `TIFF` are different.
   * PowerSpectrum: Attempting to replace the power spectral values of a `PowerSpectrum` using `with_spectrum` using a vector of incorrect length will raise a `ValueError`.
 * When removing tracks with the kymotracking widget, only tracks that are entirely in the selection rectangle will be removed. Prior to this change, any tracks intersecting with the selection rectangle would be removed.
+* Added checks which enforce `KymoTracks` in a `KymoTrackGroup` to be unique. Extending a `KymoTrackGroup` with `KymoTrack` instances that are already part of the group will now result in a `ValueError`. Similarly, constructing a new `KymoTrackGroup` with duplicate `KymoTracks` will also produce a `ValueError`.
 
 #### New features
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -627,6 +627,21 @@ def __add__(self, other):
         new_group.extend(other)
         return new_group
 
+    def _get_track_by_id(self, python_id):
```
alessiamarcolini, line 667: would it make sense to turn this into a `__contains__` method (and the parameter is the track obj instead of the id) so you can use the `in` operator?

JoepVanlier: I think it would make a lot of sense to also add a `__contains__` method for convenience (and I will add it).

This method in particular has something to do with communication with another software where we can't store the entire object for other implementational reasons. This is also why I chose to make it private.

JoepVanlier: Actually,  come to think of it, I don't think we actually need a `__contains__` dunder. The default suffices for this container already (since we defined `__iter__` and just iterating over the elements and against them should suffice).

alessiamarcolini: makes sense! 👍

alessiamarcolini (approved):

mwilkowski (approved):
nice!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -624,6 +664,31 @@ def __add__(self, other):
...
+        ----------
+        python_id : int
+            python identity
+
+        Returns
+        -------
+        track : KymoTrack or None
+        """
+        tracks = [track for track in self._src if id(track) == python_id]
+        return tracks[0] if tracks else None
```
mikhas, line 680: Is multiple tracks having the same Python id a valid use case even?

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -116,15 +117,88 @@ def test_kymotrackgroup(blank_kymo):
...
+
+
+def test_kymotrackgroup_lookup(blank_kymo):
+    k1 = KymoTrack(np.array([1, 2, 3]), np.array([2, 3, 4]), blank_kymo, "red")
+    k2 = KymoTrack(np.array([2, 3, 4]), np.array([3, 4, 5]), blank_kymo, "red")
+    k3 = KymoTrack(np.array([3, 4, 5]), np.array([4, 5, 6]), blank_kymo, "red")
+
+    tracks = KymoTrackGroup([k1, k2, k3])
+    tracks._get_track_by_id(id(k1)) is k1
+    tracks._get_track_by_id(id(k2)) is k2
```
mikhas, line 170: Can we test with a KymoTrackGroup that contains the same track multiple times, or a track that is shared between at least two KymoTrackGroups?

mikhas (dismissed):

lumicks/pylake/kymotracker/tests/test_kymotrack.py
JoepVanlier: The former should be explicitly excluded now.

Anything specific you would like to test with the track that is shared between two groups?

changelog.md
```diff
@@ -28,12 +28,15 @@
...
 
 #### New features
 
 * Added API for notes, i.e. `file.notes` returns a dictionary of notes.
 * Added option to add a scale bar to by providing a [`lk.ScaleBar()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ScaleBar.html) to plotting or export functions.
 * Implemented bias correction for centroid refinement that shrinks the window to reduce estimation bias. This bias correction can be toggled by passing a `bias_correction` argument to [`lk.track_greedy()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.track_greedy.html#) and [`lk.refine_tracks_centroid()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.refine_tracks_centroid.html).
+* Added `KymoTrackGroup.remove()` to remove a `KymoTrack` from a `KymoTrackGroup`.
+* Allow boolean array indexing (e.g. `tracks[[False, False, True]]`) and indexing with arrays of indices (e.g. `tracks[[1, 3]]`) for `KymoTrackGroup`. See the [API documentation](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.html#lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup) for more information.
```
rpauszek, line 39: 🙌

PR #469, opened 2023-02-28
docs: refer to the faq earlier

JoepVanlier:
**Why this PR?**

Based on feedback we found that people tend to read over the reference to the FAQ at the end of the installation instructions. So now, it's presented at the top, in a little warning box up front.

We also explicitly instruct users to install a specific version of conda on windows to work around the OpenSSL issue.

Docs build here: https://lumicks-pylake.readthedocs.io/en/installation_warning/install.html#installation

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -3,6 +3,12 @@ Installation
 
 Pylake can be installed on Windows, Linux and Mac and requires Python 3.7 or newer.
 
+.. note::
+
+    While we strive to make our instructions as easy to follow as possible, sometimes things can go wrong.
+    Should anything return an error while installing `pylake`, please refer to the `Frequently asked questions`_ for troubleshooting options.
+    If your error message is not listed there, please feel free to open an issue `on our GitHub repository <https://github.com/lumicks/pylake/issues>`_ and we will have a look at it.
+
 Installation using Anaconda
 ---------------------------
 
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -3,6 +3,12 @@ Installation
 
 Pylake can be installed on Windows, Linux and Mac and requires Python 3.7 or newer.
 
+.. note::
+
+    While we strive to make our instructions as easy to follow as possible, sometimes things can go wrong.
+    Should anything return an error while installing `pylake`, please refer to the `Frequently asked questions`_ for troubleshooting options.
+    If your error message is not listed there, please feel free to open an issue and we will have a look at it.
```
rpauszek, line 10:
```suggestion
    If your error message is not listed there, please feel free to open an issue `on our GitHub repository <https://github.com/lumicks/pylake/issues>`_ and we will have a look at it.
```

docs/install.rst
```diff
@@ -21,7 +27,13 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
...
 #. Create a new environment for Pylake named pylake by invoking the following command::
 
-    conda create -n pylake
+    conda create -n pylake conda=4.9
+
+.. note::
+
+    On Windows, there are issues with OpenSSL that preclude a successful installation on the latest version of conda.
+    This is why it is recommended to create an environment specifically with conda version `4.9` for now.
+    See `OpenSSL Error`_ for more information.
```
rpauszek, line 36: hmm, this seems to break the numbering in the ordered list. I think it can be solved by indenting the note block ([from here](https://sublime-and-sphinx-guide.readthedocs.io/en/latest/lists.html#code-images-and-other-content-inside-lists)

image

rpauszek (changes requested):
I think it helps to have this right at the top now, definitely more user friendly. I built the docs and noticed there's an issue with the list numbering caused by the new note block, but I think it's an easy fix

rpauszek (approved):
🚢

PR #470, opened 2023-03-02
Fixup GMM documentation

rpauszek:
**Why this PR?**
I noticed there was an error in the `Return` section `GaussianMixtureModel.extract_dwell_times()`. I also took the opportunity to make some other general improvements to the other docstrings for the class

[Docs build here](https://lumicks-pylake.readthedocs.io/en/docstring_oops/_api/lumicks.pylake.GaussianMixtureModel.html#lumicks.pylake.GaussianMixtureModel)

### DIFF
--- a/lumicks/pylake/population/mixture.py
+++ b/lumicks/pylake/population/mixture.py
@@ -5,9 +5,13 @@ from .dwelltime import _dwellcounts_from_statepath
 
 
 def as_sorted(fcn):
-    """Decorator to return results sorted according to mapping array."""
+    """Decorator to return results sorted according to mapping array.
 
-    def wrapper(self, *args, **kwargs):
+    To be used as a method decorator in a class that supplies an index
+    mapping array via the `._map` attribute.
+    """
+
+    def wrapper(self, *args, **kwargs) -> np.ndarray:
         result = fcn(self, *args, **kwargs)
         return result[self._map]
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/population/mixture.py
```diff
@@ -17,22 +21,23 @@ def wrapper(self, *args, **kwargs):
 
 
 class GaussianMixtureModel:
-    """A wrapper around scikit-learn's GMM.
+    """A wrapper around :class:`sklearn.mixture.GaussianMixture`.
```
JoepVanlier, line 24: :)

lumicks/pylake/population/mixture.py
```diff
@@ -17,22 +21,23 @@ def wrapper(self, *args, **kwargs):
...
-    to state mean in order to facilitate comparison of models with different number of states
+    This model accepts a 1D array as training data. *The state parameters are sorted according
+    to state mean* in order to facilitate comparison of models with different number of states
     or trained on different datasets. As the current implementation is designed to specifically
-    handle 1D data, model parameters are also returned as 1D arrays (np.squeeze() is applied to the results)
-    so that users do not have to be concerned with the shape of the output results.
+    handle 1D data, model parameters are also returned as 1D arrays (:func:`numpy.squeeze()` is applied to
+    the results) so that users do not have to be concerned with the shape of the output results.
 
     Parameters
```
JoepVanlier, line 32: Would it make sense to mention that this functionality is in alpha here? Or are we considering this API to be mature enough already? (Considering we have it marked as alpha in the tutorial).

lumicks/pylake/population/mixture.py
```diff
@@ -161,12 +161,12 @@ def extract_dwell_times(self, trace, *, exclude_ambiguous_dwells=True):
         return dwell_times
 
     @property
-    def bic(self):
+    def bic(self) -> float:
         """Bayesian Information Criterion."""
```
JoepVanlier, line 165: Might be nice to expand the docstring for these information criteria a little bit if we're touching this up anyway?

JoepVanlier (approved):
LGTM

lumicks/pylake/population/mixture.py
rpauszek: no, I want to keep it alpha because we haven't really worked with it much yet. good point to put it here too

PR #472, opened 2023-03-03
Stackymo

JoepVanlier:
**Why this PR?**
This is preparation work for `Kymo.from_stack()` which is going to be the next PR.
We've had the metadata to read pixel sizes for a while now. This exposes it with the same API as `Scan` and `Kymo`.

I've also made sure that the plots now use this information (analogously to `Scan` and `Kymo`). When the metadata is unavailable, these end points return `None` and plots are uncalibrated.

Calibrated:
image

Calibrated (single color):
image

Uncalibrated:
image

I tested movie export for this feature by manual testing.

P.S. I also found a small bug that messed up the `name` field of the stack, resulting in a bad title, so I fixed that as well. I added a regression test for it too.


changelog.md
```diff
@@ -45,6 +45,7 @@
 
 * Fixed incorrect behaviour in `lk.track_lines()` by interpolating back to integer frame times. Prior to this change, `lk.track_lines()` would provide a subpixel accurate position along the time axis of the kymograph as well. However, this position was specified with respect to the coordinate system of the image, rather than actual acquisition times. As such, it would produce incorrect results when performing downstream analysis that rely on the time corresponding to an actual time. Note that `lk.track_greedy()` is not affected.
 * Fixed bug in `lk.track_lines()` where one extra line was returned rather than the number requested through the parameter `max_lines`.
+* Fixed regression that caused [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html). See [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html) to incorrectly store the stack name (resulting in an extra `('` in front of the name in plots). This regression was introduced in `v0.13.0` when enabling multi-file support.
```
rpauszek, line 49: I think you're clauses are in the wrong order 😆 can't make a suggestion here for some reason, but should be

```
* Fixed regression that caused [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html) to incorrectly store the stack name (resulting in an extra `('` in front of the name in plots). This regression was introduced in `v0.13.0` when enabling multi-file support. See [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html)
```

rpauszek (approved):
Looks good, and it's really nice to have the API consistent with `Scan`. Just see the changelog oopsie 😏 

I see what you mean about the last commit, but I'm ok with it since we have workable functionality. I think we need to revisit positional calibrations for all of the image classes anyways as I think there's a lot to be unified; we can refactor the internal stuff then

PR #473, opened 2023-03-07
Docs fixups

rpauszek:
**Why this PR?**
- ensure piezotracking docs work with example data. In order to push through for release I only did minor fixups; more extensive revisions can be done at a later time
- use `plt.figure()` and `plt.show()` in all plotting calls. If users download a tutorial and use an interactive backend they'd have to add these manually to get plots to display below the proper cell. This adds a bit of extra boilerplate in the tutorials, but I think worth it so the notebook "just works" when users download
- fixup some of the example data filenames. these will be replaced with downloads from zenodo at release


alessiamarcolini (approved):

JoepVanlier (approved):
Looks good. One minor request before merging. I didn't notice last time, but it's probably a good idea to avoid the word fixup in the commit messages if you can.

PR #475, opened 2023-03-08
Enforce keyword arguments for tracking

rpauszek:
**Why this PR?**
Motivated by renaming `vel` argument to `velocity`. However, since there are many arguments and their meaning can't easily be inferred from the values passed, decided to make the breaking change for `v1.0` and enforce keyword use.


JoepVanlier (changes requested):
Don't forget `_get_default_parameters`. It still has `"vel"` as entry.

rpauszek:
> Don't forget `_get_default_parameters`. It still has `"vel"` as entry.

it's changed, line 795. or am I missing another place?

JoepVanlier:
> > Don't forget `_get_default_parameters`. It still has `"vel"` as entry.
> 
> it's changed, line 795. or am I missing another place?

Nah, that was it.

JoepVanlier (approved):
LGTM

PR #477, opened 2023-03-08
`ImageStack` -> `Kymo`

JoepVanlier:
**Why this PR?**
It's valuable for users to be able to generate `Kymos` from their camera stacks.

Turns this:
image

Into this:
image

Doing nothing more than this:

```python
stack_tether = stack.define_tether((6.94423, 4.22381), (20.47474,  4.08063))
kymograph = stack_tether.to_kymo(adjacent_lines=5)
```

This requires a re-assessment on how we treat motion blur, as camera images typically have far more motion blur, resulting in biased estimates if not correctly taken into account during specific downstream analyses (such as diffusion estimation using `cve`).

This PR introduces a new property on the `Kymo` which keeps track of this motion blur constant.

- For camera's computing the motion blur is relatively straightforward. It should simply be `(1/6) * exposure / line_time`.
- For `Kymo`'s generated from 2D `Scans` we will eventually also need a particular motion blur constant that is also different. This is more complicated since the exposure window will have gaps (gaps can be worse than having the shutter continuously open).
- For downsampled kymos we also need different handling.
- For regular confocal images we neglect the motion blur (`R=0`) as its effect on a 1D `Scan` is very small. We could consider incorporating this in the future as an improvement.

In this PR we specify the motion blur constant as `None` for non-standard `Kymo`s. We will add their correct definitions as we go, since every addition needs some testing.

Note that I will merge the docs update into the PR introducing the functionality prior to merging.

Docs build [here](https://lumicks-pylake.readthedocs.io/en/to_kymo/tutorial/imagestack.html#constructing-a-kymograph-from-an-image-stack) and [here](https://lumicks-pylake.readthedocs.io/en/to_kymo/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.to_kymo).

### DIFF
--- a/lumicks/pylake/image_stack.py
+++ b/lumicks/pylake/image_stack.py
@@ -5,6 +5,7 @@ import numpy as np
 import numpy.typing as npt
 from deprecated.sphinx import deprecated
 
+from .kymo import Kymo, _kymo_from_image_stack
 from .adjustments import no_adjustment
 from .detail.image import make_image_title
 from .detail.imaging_mixins import FrameIndex, TiffExport, VideoExport
@@ -217,6 +218,81 @@ class ImageStack(FrameIndex, TiffExport, VideoExport):
         """
         data = self._src.with_tether(np.asarray((point1, point2)) / self._pixel_calibration_factors)
         return self.from_dataset(data, self.name, self._start_idx, self._stop_idx)
+
+    def to_kymo(self, half_window, reduce=np.sum) -> Kymo:
+        """Convert this :class:`ImageStack` to a :class:`~lumicks.pylake.kymo.Kymo` by sampling
+        along the tether.
+
+        Parameters
+        ----------
+        half_window : int
+            Number of pixels adjacent to the tether on either side to average over. The kymograph
+            will be calculated from the pixel values reduced to a one pixel line using the
+            downsampling function provided in :func:`reduce`.
+        reduce : callable
+            The :mod:`numpy` function which is going to reduce multiple samples into one. The
+            default is :func:`np.sum <numpy.sum>`.
+
+        Returns
+        -------
+        Kymo
+
+        Raises
+        ------
+        ValueError
+            If the frame rate or exposure time are not constant.
+        ValueError
+            If the image stack does not have a tether defined.
+        ValueError
+            If the number of adjacent lines are negative.
+        ValueError
+            If the number of adjacent lines exceed the image dimensions.
+
+        Notes
+        -----
+        In order to convert a :class:`ImageStack` to a :class:`~lumicks.pylake.kymo.Kymo`, a tether
+        needs to be defined first. See the example below for more information.
+
+        Examples
+        --------
+        ::
+
+            import lumicks.pylake as lk
+            import matplotlib.pyplot as plt
+
+            stack = lk.ImageStack("camera_recording.tiff")
+
+            # Plot the stack
+            plt.figure()
+            stack.plot()
+
+            # Define tether by a pair of X, Y coordinates.
+            stack = stack.define_tether((7.19, 4.07), (22.74, 4.15))
+
+            # Plot the tether to verify.
+            stack.plot_tether()
+            kymo = stack.to_kymo(half_window=3)
+
+            # Plot the kymograph just created
+            plt.figure()
+            kymo.plot()
+
+            # Or alternatively, use the interactive widget to define the tether graphically
+            %matplotlib widget  # Enable interactive widgets
+            stack = lk.ImageStack("camera_recording.tiff")
+
+            # Opens an interactive widget. See help(lk.ImageStack.crop_and_rotate) for more info.
+            widget = stack.crop_and_rotate()
+            kymo = widget.image.to_kymo(half_window=3)
+            kymo.plot()
+        """
+        return _kymo_from_image_stack(
+            self,
+            half_window,
+            self.pixelsize_um[0] if self.pixelsize_um else None,
+            self.name,
+            reduce,
+        )
 
     def get_image(self, channel="rgb"):
         """Get image data for the full stack as an `np.ndarray`.
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -876,12 +876,12 @@ def _cve(
     dt : float
         Time step
     blur_constant : float
-        Motion blur coefficient. Should be between 0 and 1/4.
+        Motion blur coefficient. Should be between 0 and 1/4 or `np.nan` if not clearly defined.
 
         The normalized shutter function is defined as :math:`c(t)`, where :math:`c(t)` represents
         whether the shutter is open or closed. :math:`c(t)` is normalized w.r.t. area. For no
-        motion blur, :math:`c(t) = \delta(t_{exposure})`, whereas for a constantly open shutter it is
-        defined as :math:`c(t) = 1 / \Delta t`.
+        motion blur, :math:`c(t) = \delta(t_{\mathrm{exposure}})`, whereas for a constantly open
+        shutter it is defined as :math:`c(t) = 1 / \Delta t`.
 
         With :math:`c(t)` defined as before, the motion blur constant is defined as:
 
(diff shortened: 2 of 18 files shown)


lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -876,12 +876,12 @@ def _cve(
     dt : float
         Time step
     blur_constant : float
-        Motion blur coefficient. Should be between 0 and 1/4.
+        Motion blur coefficient. Should be between 0 and 1/4 or `np.nan` if not clearly defined.
```
rpauszek, line 879: in the docstring for `Kymo.motion_blur` you say it should be between 0 and 1/6 but here 0 and 1/4 ??

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -876,12 +876,12 @@ def _cve(
...
 
         The normalized shutter function is defined as :math:`c(t)`, where :math:`c(t)` represents
         whether the shutter is open or closed. :math:`c(t)` is normalized w.r.t. area. For no
-        motion blur, :math:`c(t) = \delta(t_{exposure})`, whereas for a constantly open shutter it is
-        defined as :math:`c(t) = 1 / \Delta t`.
+        motion blur, :math:`c(t) = \delta(t_{exposure})`, whereas for a constantly open shutter it
```
rpauszek, line 883:
```suggestion
        motion blur, :math:`c(t) = \delta(t_\mathrm{exposure})`, whereas for a constantly open shutter it
```

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -917,6 +919,13 @@ def _cve(
     With :math:`D` the diffusion constant and :math:`\sigma`
     This method requires SNR > 1. Below SNR = 1 it degrades rapidly [12]_.
 
+    Warns
+    -----
+    RuntimeWarning
+        if choice is cve and the kymograph these lines were tracked on doesn't have its motion
```
rpauszek, line 925:
```suggestion
        if choice is 'cve' and the kymograph these lines were tracked on doesn't have its motion
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -535,13 +543,23 @@ def estimate_diffusion(
...
-            # We hardcode the blur constant for confocal for now (no motion blur)
+            try:
+                blur = self._kymo.motion_blur_constant
+            except NotImplementedError:
+                warnings.warn(
+                    RuntimeWarning(
+                        "Motion blur cannot be taken into account for this type of Kymo. As a "
+                        "consequence, not all estimates will be available."
+                    )
+                )
```
rpauszek, line 554: "not all estimates" is pretty vague. it only affects the localization uncertainty estimates, not the diffusion estimates correct?

rpauszek (changes requested):
It'll be great to finally have this as public API and I'm much appreciated by users with widefield systems.

Besides some minor nitpicks the main things I pointed out are:
- default to `reduce=np.sum` for consistency with other imaging methods we have
- clarification on what exactly is not available when there's no motion blur

lumicks/pylake/image_stack.py
```diff
@@ -290,6 +291,82 @@ def define_tether(self, point1, point2):
...
+        Parameters
+        ----------
+        adjacent_lines : int
+            Number of lines adjacent to the tether to average over. The kymograph will be
+            calculated from the pixel values reduced to a one pixel line using the downsampling
+            function provided in :func:`reduce`.
+        reduce : callable
+            The :mod:`numpy` function which is going to reduce multiple samples into one. The
+            default is :func:`np.mean <numpy.mean>`, but :func:`np.sum <numpy.sum>` could also be
+            appropriate.
```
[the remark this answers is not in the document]
JoepVanlier: Agreed.

lumicks/pylake/image_stack.py
```diff
@@ -290,6 +291,82 @@ def define_tether(self, point1, point2):
...
+    def to_kymo(self, adjacent_lines, reduce=np.mean) -> Kymo:
+        """Convert this :class:`ImageStack` to a :class:`~lumicks.pylake.kymo.Kymo` by sampling
+        along the tether.
+
+        Parameters
+        ----------
+        adjacent_lines : int
+            Number of lines adjacent to the tether to average over. The kymograph will be
+            calculated from the pixel values reduced to a one pixel line using the downsampling
+            function provided in :func:`reduce`.
```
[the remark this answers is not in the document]
JoepVanlier: After discussing this for a bit, `half_window` it is. :)

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: It's not inconsistent.

1/6 `t_exposure` / `t_frame` is what you get for a continuously open shutter, but you can do worse than that.

If you look at the integral, you'll find that the worst you can get is a two pulse shutter per frame. So basically taking to snapshots at the start and end of the frame, which results in 1/4.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Depends. If you pass in no localization variance, it affects the `stderr` for the diffusion constant and the `localization_variance`.  However, if you've passed in a `localization_variance`, it affects the diffusion constant itself as well.

JoepVanlier: I have specified it more precisely now, and raise in the case where the diffusion constant is also unavailable (see last commit, which I kept separate).

lumicks/pylake/kymotracker/detail/msd_estimation.py
rpauszek: ok yeah, it is a bit confusing here but I agree could be just as confusing/distracting to have all this extra information that isn't particularly relevant to the data we expect. We can always update if we start getting questions about this

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -506,12 +506,26 @@ def estimate_diffusion(
...
         Warns
         -----
         RuntimeWarning
-            if choice is cve and the kymograph these lines were tracked on doesn't have
-            `kymo.motion_blur_constant` defined. As a result, some uncertainty estimates may not be
-            available. Estimates that are unavailable are returned as `np.nan`.
+            if choice is `cve` and the kymograph these lines were tracked on doesn't have its motion
+            blur constant clearly defined. As a result, the localization variance and standard
+            error for the diffusion constant will not be available. Estimates that are unavailable
+            are returned as `np.nan`.
```
rpauszek, line 528: I would use same as above
```
if `method == "cve"` and the source kymograph does not have a clearly defined motion blur constant
```

rpauszek (approved):
Definitely much more informative for what can and cannot be done if the motion blur is not defined. Just added some simple re-wording suggestions if you think they sound better

Great work! 👏

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -546,6 +560,14 @@ def estimate_diffusion(
             try:
                 blur = self._kymo.motion_blur_constant
             except NotImplementedError:
+                if localization_variance:
+                    raise ValueError(
+                        "Cannot compute diffusion constant reliably for a kymograph where the "
+                        "motion blur constant is not clearly defined and the localization variance"
+                        "is provided. Omit the localization variance to calculate a diffusion "
+                        "constant."
```
rpauszek, line 568:
```
Cannot compute diffusion constant reliably for a kymograph that does not
have a clearly defined motion blur constant and the localization variance is provided. 
Omit the localization variance to calculate a diffusion constant.
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -506,12 +506,26 @@ def estimate_diffusion(
...
 
+        Raises
+        ------
+        ValueError
+            if choice is `cve` but the `KymoTrack` has fewer than 3 time intervals defined.
+        ValueError
+            if choice is `cve`, the kymograph these lines were tracked on doesn't have its motion
+            blur constant clearly defined and a localization variance is passed in. As a result, the
+            diffusion constant cannot be reliably calculated. The solution is then to not provide a
+            localization uncertainty.
```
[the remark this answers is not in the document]
JoepVanlier: Much nicer :)

PR #479, opened 2023-03-09
Bump python and mpl minimum versions

rpauszek:
**Why this PR?**
So we can take advantages of some newer features and previous minimum requirements were quite old.

Notes:
- tested with a fresh virtual environment according to our installation instructions (except using `pip` to install from branch) and works fine
- had issues with the old format of the `.readthedocs.yml` not supporting Python 3.9; switched to the newer `version 2` format which is recommended anyways. A slight issue is I had to include all requirements from `setup.py` in `docs/requirements.txt` in addition to extra docs related packages in order for the build to succeed on RtD
- sadly because of the unresolved issue with `OpenSSL` errors from newer `conda` versions, we will not in fact be bringing on the pattern matching just yet 😞


docs/requirements.txt
```diff
@@ -1,9 +1,19 @@
...
+scipy>=1.1, <2
+matplotlib>=3.5
+tifffile>=2020.9.30
+tabulate>=0.8.8, <0.9
+opencv-python-headless>=3.0
+cachetools>=3.1
+deprecated>=1.2.8
+scikit-learn>=0.18.0
+scikit-image>=0.17.2
+tqdm>=4.27.0
```
JoepVanlier: Don't do this. Install the package instead (that'd be more similar to how it was previously handled and doesn't duplicate these dependencies). Hang on, I'll push a commit. If you agree, you can just squash it into the previous one (it also undoes the changes to the reqs).

setup.py
```diff
@@ -54,7 +54,7 @@ def read(filename):
         "h5py>=3.4, <4",
         "numpy>=1.20, <2",
         "scipy>=1.1, <2",
-        "matplotlib>=2.2",
```
JoepVanlier, line 57: Considering that there are no real changes other than the requirements, I'd personally prefer the updated reqs to just be one commit.

Also, gridspec here we gooo :D

JoepVanlier (changes requested):
I don't like duplicating the requirements. I added a commit with a possible alternative.

.readthedocs.yaml
```diff
@@ -0,0 +1,17 @@
+# build:
+#   image: latest
+# python:
+#   version: "3.9"
+#   pip_install: true
+# requirements_file: docs/requirements.txt
```
JoepVanlier: Why leave commented stuff?

docs/requirements.txt
rpauszek: definitely much nicer to not do this. I could've sworn I tried what you added in your commit, but maybe just not that particular combination of everything. At any rate, thanks for figuring this out

.readthedocs.yaml
rpauszek: rebase mishap

JoepVanlier (approved):
Lookin' good!

shipit

PR #489, opened 2023-03-31
dwelltime: Improve numerics dwell time analysis

JoepVanlier:
**Why this PR?**
This PR is number three in a story where we introduce more reliable uncertainty estimation.

This PR in particular provides three small improvements which we need to have things running smoothly:
- It provides the analytical gradient for the dwell time model. This improves performance when fitting dwelltime models. The performance improvement is typically a factor two or more).
- It adds a bound constraint on the estimated parameters to avoid numerical issues that occur when amplitudes or lifetimes go to zero.
- Finally, it suppresses a warning which is annoying to users and warns about something that doesn't impact the problem we are solving in a meaningful way.

One thing that might be worth noting prior to review is that I did notice all the reciprocals appearing in the gradient. I have a separate branch in which I changed the parametrization (and the associated gradient) to use rates `1/lifetime`, or use `log(amplitudes)` and `log(rates)`.

To my disappointment, in the cases I tested, these transformations don't actually improve performance or affect the results much at all. That's why I left them out of this PR. They also don't relax the requirement on the parameter bounds.

That said, the other improvements made in this PR _did_ make a difference.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,8 @@
 
 * Improve API description of the channel data in a default `FdCurve`.
 * Improved fitting performance dwell time analyses by implementing analytic gradient.
+* Added lower and upper bounds when fitting dwell time models. This prevents numerical issues that occur when approaching these extremes. Prior to this change, over-parameterized models could encounter numerical issues due to divisions by (near) zero. After this change, amplitudes are constrained to stay between `1e-9` and `1.0 - 1e-9`, while lifetimes are constrained to remain between `1e-8` and `1e8`.
+* Provided better initial guess for dwell time models with three or more components. Prior to this change, the dwelltime models involving three components or more would get an initial guess with an average lifetime beyond the mean of the observations.
 
 ## v1.0.0 | 2023-03-14
 
--- a/changelog.md
+++ b/changelog.md
@@ -5,7 +5,7 @@
 #### New features
 
 * Added [`DwelltimeModel.profile_likelihood()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.DwelltimeModel.html#lumicks.pylake.DwelltimeModel.profile_likelihood) for calculating profile likelihoods for dwell time analysis. These can be used to determine confidence intervals and assess whether the model has too many parameters (i.e., if a model with fewer components would be sufficient).
-  
+
 #### Bug fixes
 
 * Fixed bug that could lead to mutability of `Kymo` and `Scan` through `.get_image()` or `.timestamps`. This bug was introduced in `0.7.2`. Grabbing a single color image from a `Scan` or `Kymo` using `.get_image()` returns a reference to an internal image cache. This numpy array was write-able and modifying data in the returned image would result in future calls to `.get_image()` returning the modified data rather than the original `Kymo` or `Scan` image. Timestamps obtained from `.timestamps` were similarly affected. These arrays are now returned non-writeable.
@@ -13,6 +13,7 @@
 #### Improvements
 
 * Improve API description of the channel data in a default `FdCurve`.
+* Improved fitting performance dwell time analyses by implementing analytic gradient.
 
 ## v1.0.0 | 2023-03-14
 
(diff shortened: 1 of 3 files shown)


lumicks/pylake/population/dwelltime.py
```diff
@@ -761,6 +768,60 @@ def label_maker(a, t, n):
...
+    dtotal_damp = (sum_components * dlognorm_damp + np.exp(components) * dlogamp_damp) / total_denom
+    dtotal_dtau = (
+        sum_components * dlognorm_dtau + np.exp(components) * dlogtauterm_dtau
+    ) / total_denom
+    unsummed_gradient = np.vstack((dtotal_damp, dtotal_dtau))
+
+    return -np.sum(unsummed_gradient, axis=1)
+
+
 def exponential_mixture_log_likelihood_components(
```
rpauszek, line 825: I think for "public" functions like these it might be a good time to rename them with the leading `_` and denote them as internal. I don't think this is a breaking change since we don't expose these through the package `__init__.py` and the module is specifically marked alpha functionality; more of a safeguard for our future selves

lumicks/pylake/population/dwelltime.py
```diff
@@ -761,6 +768,60 @@ def label_maker(a, t, n):
...
+    max_observation_time : float
+        maximum observation time in seconds
+    """
+    amplitudes, lifetimes = np.reshape(params, (2, -1))
+    # We clip the amplitude lower bound. An amplitude of 1e-14 is negligible.
+    amplitudes = np.clip(amplitudes[:, np.newaxis], 1e-14, np.inf)
+    lifetimes = lifetimes[:, np.newaxis]
+    t = t[np.newaxis, :]
+
+    t_min, t_max = min_observation_time, max_observation_time
```
rpauszek, line 793: Based on my comment below, I think we could just use the short names as the actual arguments (here and in the other internal functions) so we don't end up bloating the code with statements like this. I think the names are still descriptive enough, and we can keep the more verbose names in the public constructor.

lumicks/pylake/population/dwelltime.py
```diff
@@ -967,7 +1032,8 @@ def _exponential_mle_optimize(
 
     if initial_guess is None:
         initial_guess_amplitudes = np.ones(n_components) / n_components
-        initial_guess_lifetimes = np.mean(t) * np.arange(1, n_components + 1)
+        fractions = np.arange(1, n_components + 1)
+        initial_guess_lifetimes = np.mean(t) * n_components * fractions / np.sum(fractions)
```
rpauszek, line 1036: how did you come up with this? is it some standard way of making initial guesses?

lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -4,15 +4,17 @@
...
     _dwellcounts_from_statepath,
     DwelltimeBootstrap,
     _handle_amplitude_constraint,
     _exponential_mle_optimize,
+    exponential_mixture_log_likelihood,
+    _exponential_mixture_log_likelihood_gradient
 )
 
 
-@pytest.mark.filterwarnings("ignore:Values in x were outside bounds")
```
rpauszek, line 15: huzzah 🥳

rpauszek (approved):
Taken with the previous 2 PRs, this is a significant improvement to the feature, great work. There are some minor suggestions that aren't necessarily in scope of this PR, but minor things that could be wrapped in and help a bit here. Otherwise, good to go!

lumicks/pylake/population/dwelltime.py
```diff
@@ -761,6 +768,60 @@ def label_maker(a, t, n):
         plt.tight_layout()
 
 
+def _exponential_mixture_log_likelihood_gradient(
```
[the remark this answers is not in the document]
JoepVanlier: Gradient to me signifies that its a scalar function whereas Jacobian is the derivative of a vector valued function. 
Anyways, I don't feel that strongly about it, so we can change it.

lumicks/pylake/population/dwelltime.py
JoepVanlier: Not really, but I figured that the weighted average of the means should be the total mean to make sure it doesn't get really large when the number of components is large.

changelog.md
```diff
@@ -5,14 +5,18 @@
 #### New features
 
 * Added [DwelltimeModel.profile_likelihood()](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.DwelltimeModel.html#lumicks.pylake.DwelltimeModel.profile_likelihood) for calculating profile likelihoods for dwell time analysis. These can be used to determine confidence intervals and assess whether the model has too many parameters (a smaller model with fewer components would be sufficient).
-  
+* Improved fitting performance dwell time analyses by implementing analytic gradient.
+
```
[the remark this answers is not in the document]
JoepVanlier: yes :D

PR #490, opened 2023-04-05
docs: improve force calibration docs

JoepVanlier:
**Why this PR?**
These improvements to the docs were made based on feedback.

The aim is three-fold:
1. To make the connection between te diffusion equation and the full optical trap calibration equation a little bit more clear. Before, it appeared as though the drag only appeared once we got to the trapping part (but it was always there in the diffusion equation).
2. To provide some guidance on the direction of the biases when you average too little, or use an unsuitable model.
3. To sketch the calibration setup a little bit, showing a schematic of where the force/displacement detection is taking place.

Docs build: https://lumicks-pylake.readthedocs.io/en/fixup_force_calib_docs/theory/force_calibration/force_calibration.html

### DIFF
--- a/docs/theory/force_calibration/force_calibration.rst
+++ b/docs/theory/force_calibration/force_calibration.rst
@@ -5,7 +5,30 @@ Force Calibration
 
     :nbexport:`Download this page as a Jupyter notebook <self>`
 
-Force calibration refers to computing the conversion factors needed to convert the raw voltages recorded by the position sensitive detectors to actual forces and displacements.
+Why is force calibration necessary?
+-----------------------------------
+
+Optical tweezers typically measure forces and displacements by detecting deflections of a trapping laser by a trapped bead.
+These deflections are measured at the back focal plane of the beam using position sensitive detectors (PSDs).
+
+.. image:: figures/back_focal.png
+  :nbattach:
+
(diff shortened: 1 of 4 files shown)


docs/theory/force_calibration/force_calibration.rst
```diff
@@ -240,17 +288,31 @@ Faxen's law
...
 
 .. math::
 
     \gamma_\mathrm{faxen}(R/l) = \frac{\gamma_0}{
         1 - \frac{9R}{16l} + \frac{1R^3}{8l^3} - \frac{45R^4}{256l^4} - \frac{1R^5}{16l^5}
     } \tag{$\mathrm{kg/s}$}
 
+At frequency zero, the frequency dependent model used in the previous section, reproduces this model up to and including its second order term in :math:`R/l`.
+It is however, a lower order model and when we go really close to the surface, the accuracy goes down quickly.
+The model predictions at frequency zero deviate strongly from the higher order model as shown in the figure below.
```
rpauszek, line 304:
```suggestion
It is, however, a lower order model and the accuracy decreases rapidly as the distance between the bead and surface become very small.
The figure below shows how the model predictions at frequency zero deviate strongly from the higher order model:
```

rpauszek (approved):
nice improvements. just left some suggestions for re-phrasing

PR #491, opened 2023-04-06
tutorial notebooks: minor fixes

JoepVanlier:
**Why this PR?**
I was preparing to update the notebooks we have on Harbor to not use deprecated API. I noticed a few small errors in the deprecation warning and added some crosslinking in the kymotracking such that the notebook doesn't have to duplicate the entire algorithm description.

I also added a small commit which adds some more descriptions as to what is in an `FdCurve` as per customer request.

Docs build here: https://lumicks-pylake.readthedocs.io/en/minor_fixups/

### DIFF
--- a/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
+++ b/docs/examples/cas9_kymotracking/cas9_kymotracking.rst
@@ -36,7 +36,7 @@ Download the kymograph data
 
 The kymograph data is stored on Zenodo, a general-purpose open-access repository developed under the European OpenAIRE program and operated by CERN.
 Zenodo allows researchers to deposit data sets, research software, reports, and any other research related digital artifacts and allows them to be referenced by a Digital Object Identifier (DOI).
-We can download the kymograph we need directly from Zenodo using Pylake.
+We can download the kymograph we need directly from Zenodo using the function :func:`~lumicks.pylake.download_from_doi`.
 Since we don't want it in our working folder, we'll put it in a folder called `"test_data"`::
 
     filenames = lk.download_from_doi("10.5281/zenodo.4247279", "test_data")
@@ -100,7 +100,8 @@ This will provide us with a view of the kymograph and some dials to tune in algo
 custom `axis_aspect_ratio` which determines our field of view. This input is not necessary, but provides a better
 view of our data.
 
-We will be using the greedy algorithm. The `threshold` should typically be chosen somewhere between the expected
+We will be using the greedy algorithm. For more information on how it works, please refer to the :ref:`Pylake kymotracking tutorial<track_greedy>`.
+The `threshold` should typically be chosen somewhere between the expected
 baseline photon count and the photon count of a true track (note that you can see the local photon count between square
 brackets while hovering over the kymograph). The `track width` should roughly be set to the expected spot size (in the spatial dimension) of a
 track. The `window` should be chosen such that small gaps in a track can be overcome, but not so large that spurious
@@ -144,7 +145,7 @@ bead. This binding should be omitted from the analysis::
 
 .. image:: kymowidget.png
 
-One last thing to note is that we assigned the `KymoWidgetGreedy` to the variable `kymowidget`. That means that from
+One last thing to note is that we assigned the :class:`~lumicks.pylake.nb_widgets.kymotracker_widgets.KymoWidgetGreedy` to the variable `kymowidget`. That means that from
 this point on, we can interact with it through the handle name `kymowidget`.
 
 Exporting from the widget results in a file that contains the track coordinates in pixels and real units.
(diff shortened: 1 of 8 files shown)


docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -36,7 +36,7 @@ Download the kymograph data
 
 The kymograph data is stored on Zenodo, a general-purpose open-access repository developed under the European OpenAIRE program and operated by CERN.
 Zenodo allows researchers to deposit data sets, research software, reports, and any other research related digital artifacts and allows them to be referenced by a Digital Object Identifier (DOI).
-We can download the kymograph we need directly from Zenodo using Pylake.
+We can download the kymograph we need directly from Zenodo using the function :func:`~lumicks.pylake.download_from_doi` Pylake.
```
rpauszek, line 39:
```suggestion
We can download the kymograph we need directly from Zenodo using the function :func:`~lumicks.pylake.download_from_doi`.
```

docs/examples/cas9_kymotracking/cas9_kymotracking.rst
```diff
@@ -156,9 +157,9 @@ This sums the photon counts from `pixel_position - sampling_width` to (and inclu
 Analyzing the results
 ---------------------
 
-The tracks are available in `kymowidget.tracks`, which returns a `KymoTrackGroup` object. The group is a customized list of :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` objects
-which contain lists of position and time coordinates for each tracked particle. These can be accessed with the `KymoTrack.position` and
-`KymoTrack.seconds` properties, respectively. Let's grab the longest track we found, and have a look at its position over time::
+The tracks are available in `kymowidget.tracks`, which returns a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup` object.
+The group is a customized list of :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` objects which contain lists of position and time coordinates for each tracked particle.
+These can be accessed with the :attr:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.position` and :attr:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.seconds` properties, respectively. Let's grab the longest track we found, and have a look at its position over time::
```
rpauszek, line 162:
```suggestion
The tracks are available from the :attr:`~lumicks.pylake.nb_widgets.KymoWidgetGreedy.tracks` property, which returns a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup` object.
This is a customized list of :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack` objects which in turn contain lists of position and time coordinates for each tracked particle.
These coordinates can be accessed with the :attr:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.position` and :attr:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.seconds` properties, respectively. Let's grab the longest track we found, and have a look at its position over time::
```

lumicks/pylake/fdcurve.py
```diff
@@ -12,12 +12,17 @@
...
+    r"""An FD curve exported from Bluelake
 
     By default, the primary force and distance channels are `downsampled_force2`
     and `distance1`. Alternatives can be selected using :func:`FdCurve.with_channels()`.
     Note that it does not modify the FD curve in place but returns a copy.
 
+    Note
+    ----
+    The default channel `downsampled_force2` contains the total force on trap 2, e.g.
+    :math:`\sqrt{F_{2, x}^2 + F_{2, y}^2}`
```
rpauszek, line 24:
```suggestion
    The default channel `downsampled_force2` is the total force on trap 2, e.g.
    :math:`\sqrt{F_{2, x}^2 + F_{2, y}^2}`
```

lumicks/pylake/fdcurve.py
```diff
@@ -26,6 +31,21 @@ class FdCurve(DownsampledFD):
...
+        import lumicks.pylake as lk
+        f = lk.File("fdcurves.h5")
+
+        # Show FdCurve "1"
+        fd_curve = f.fdcurves["1"]
+        fd_curve.plot()
+
+        # Make a new F,d curve based on the `start` and `stop` time of FdCurve "1", but only include
+        # the x-component of the force on trap 2
+        fd_curve = fd_curve.with_channels("2x", distance="1")
```
rpauszek, line 48: I wonder if it's better to move the example section before attributes. Only because, in the actual rendering at first glance it's difficult to tell that the example is part of the class docstring
image

rpauszek (approved):
Looks good, left a few minor suggestions for re-phrasing. I also noticed (now that the API stub is built) it would be good to use a proper `Returns` section in the docstring for `download_from_doi`. So the docstring would be

```
    """Download files from a Zenodo DOI (i.e. 10.5281/zenodo.#######) 

    Returns
    --------
    list :  list of downloaded filenames.

    Parameters
    ...
```

rpauszek:
ah, also just noticed the figures need to be updated with the new data

lumicks/pylake/fdcurve.py
JoepVanlier: Yeah, that's a fair point. I swapped it around.

JoepVanlier:
> ah, also just noticed the figures need to be updated with the new data

The Cas9 one still uses the old one, since that's what we use on Harbor (it is also on Zenodo already). Considering that this ticket was mostly about getting rid of deprecated API and updating the Harbor notebooks to make use of the latest API, I'd rather not piggyback updating the dataset now.

PR #493, opened 2023-04-13
kymotrackgroup: add `_flip` operation

JoepVanlier:
**Why this PR?**
We want some private API to be able to flip a kymograph.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -173,6 +173,22 @@ class KymoTrack:
             self._channel,
         )
 
+    def _flip(self, kymo):
+        """Return a copy flipped vertically.
+
+        Parameters
+        ----------
+        kymo : lumicks.pylake.Kymo
+            Flipped kymograph
+        """
+
+        return KymoTrack(
+            self.time_idx,
+            self._localization._flip(self._kymo.pixelsize[0] * (self._kymo.pixels_per_line - 1)),
+            kymo,
+            self._channel,
+        )
+
     def with_offset(self, time_offset, coordinate_offset):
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -173,6 +173,22 @@ def _with_coordinates(self, time_idx, localization):
...
+        kymo : lumicks.pylake.Kymo
+            Flipped kymograph
+        """
+
+        return KymoTrack(
+            self.time_idx,
+            self._localization._flip(self._kymo.pixelsize[0] * (self._kymo.pixels_per_line - 1)),
+            kymo,
+            self._channel,
+        )
```
rpauszek, line 190: The more I think about it, I wonder if we need this; specifically does it make sense to flip one track? The use case is really if you want to flip the image, in which case you'd want to flip all the associated tracks. From that, I'd argue it'd be better to get rid of this method and just move this code into a list comprehension at the group level.

What do you think?

rpauszek (approved):
Looks good, and nice use of `replace`, I didn't know about that one.

I approved, but consider the comment I left about whether we really need the method on the `KymoTrack` also, just on the group might be good enough and even less error-prone

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: I agree that we don't need it at the track level. However, the feature is easier to test and reason about this way and I think this set up may be less complicated in the long run. Implementing this only on the group would involve touching private variables of the track (`_localization`). Right now it's simple, but as it gets more complex we may forget things if we spread out direct access of internals.

PR #494, opened 2023-04-13
`sample_from_image`: fix rounding issue

JoepVanlier:
**Why this PR?**
This PR fixes a bug that would round the center of the window to the incorrect pixel when using `KymoTrack.sample_from_image()`. Coming from the kymotracker, pixel coordinates are defined with their origin at the center of the pixel, whereas this function incorrectly assumed that the origin was on the left side of the pixel. The relative effect of this bug is larger for small windows.

Note that I use `int(x + 0.5)` rather than `int(round(x))` because I figured it'd make more sense to round consistent to slicing (define a pixel from `x - 0.5` up to (but not including) `x + 0.5`), whereas `round` uses banker's rounding (rounds to nearest even). In real-world situations this likely makes very little difference for us.

I realized that this bug was present while working on https://github.com/lumicks/pylake/pull/493 . I also want to note that despite fixing this bug, you will still not get the exact same sampling result if you flip the Kymo and redo the analysis (due to tiny rounding errors for localizations that happen to be exactly at the pixel edge).

Ideally, I think it would be better to allow for subpixel windows (similarly to how we handle bias-corrected centroid refinement), but I think we should discuss the implications of introducing that a bit more first. For now, I think fixing this bug is a clear improvement that can be done independently now.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
 #### Bug fixes
 
 * Fixed bug that could lead to mutability of `Kymo` and `Scan` through `.get_image()` or `.timestamps`. This bug was introduced in `0.7.2`. Grabbing a single color image from a `Scan` or `Kymo` using `.get_image()` returns a reference to an internal image cache. This numpy array was write-able and modifying data in the returned image would result in future calls to `.get_image()` returning the modified data rather than the original `Kymo` or `Scan` image. Timestamps obtained from `.timestamps` were similarly affected. These arrays are now returned non-writeable.
+* Fixed a bug that would round the center of the window to the incorrect pixel when using `KymoTrack.sample_from_image()`. Pixels are defined with their origin at the center of the pixel, whereas this function incorrectly assumed that the origin was on the edge of the pixel. The effects of this bug are larger for small windows.
 
 #### Improvements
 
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -9,6 +9,7 @@
 #### Bug fixes
 
 * Fixed bug that could lead to mutability of `Kymo` and `Scan` through `.get_image()` or `.timestamps`. This bug was introduced in `0.7.2`. Grabbing a single color image from a `Scan` or `Kymo` using `.get_image()` returns a reference to an internal image cache. This numpy array was write-able and modifying data in the returned image would result in future calls to `.get_image()` returning the modified data rather than the original `Kymo` or `Scan` image. Timestamps obtained from `.timestamps` were similarly affected. These arrays are now returned non-writeable.
+* Fixed a bug that would round the center of the window to the incorrect pixel when using `KymoTrack.sample_from_image()`. Pixels are defined with their origin at the center of the pixel, whereas this function incorrectly assumed that the origin was on the left side of the pixel. The effects of this bug are larger for small windows.
```
rpauszek, line 12: bottom or top side of the pixel depending on how you plot, left is temporal
```suggestion
* Fixed a bug that would round the center of the window to the incorrect pixel when using `KymoTrack.sample_from_image()`. Pixels are defined with their origin at the center of the pixel, whereas this function incorrectly assumed that the origin was on the edge of the pixel. The effects of this bug are larger for small windows.
```

rpauszek (approved):
:shipit:

PR #495, opened 2023-04-14
kymotracker: add `split` action

JoepVanlier:
**Why this PR?**
Nothing too exciting, but requested more than once, this PR adds a split action to the kymotracker.

split_tracks

Note that I piggybacked a few improvements for the localization data on this PR, which allows us to not lose the Gaussian refinement metadata if it exists.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 #### New features
 
 * Added [`DwelltimeModel.profile_likelihood()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.DwelltimeModel.html#lumicks.pylake.DwelltimeModel.profile_likelihood) for calculating profile likelihoods for dwell time analysis. These can be used to determine confidence intervals and assess whether the model has too many parameters (i.e., if a model with fewer components would be sufficient).
+* Added functionality to split tracks in the Kymotracking widget.
 
 #### Bug fixes
 
--- a/changelog.md
+++ b/changelog.md
@@ -20,6 +20,7 @@
 * Provided better initial guess for dwell time models with three or more components. Prior to this change, the dwelltime models involving three components or more would get an initial guess with an average lifetime beyond the mean of the observations.
 * Silenced false positive warnings about the optimization violating the bound constraints. Prior to this change, dwell time analysis would produce warnings about the optimizer exceeding bounds. These excursions beyond the bound are very small and have no effect on the model.
 * Parameters corresponding to Gaussian refinement are now preserved when two refined tracks are concatenated.
+* Set the minimum value for the `Min length` slider in the Kymotracking widget to `2`.
 
 ## v1.0.0 | 2023-03-14
 
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@
 * Added lower and upper bounds when fitting dwell time models. This prevents numerical issues that occur when approaching these extremes. Prior to this change, over-parameterized models could encounter numerical issues due to divisions by (near) zero. After this change, amplitudes are constrained to stay between `1e-9` and `1.0 - 1e-9`, while lifetimes are constrained to remain between `1e-8` and `1e8`.
 * Provided better initial guess for dwell time models with three or more components. Prior to this change, the dwelltime models involving three components or more would get an initial guess with an average lifetime beyond the mean of the observations.
 * Silenced false positive warnings about the optimization violating the bound constraints. Prior to this change, dwell time analysis would produce warnings about the optimizer exceeding bounds. These excursions beyond the bound are very small and have no effect on the model.
+* Parameters corresponding to Gaussian refinement are now preserved when two refined tracks are concatenated.
 
 ## v1.0.0 | 2023-03-14
 
--- a/lumicks/pylake/kymotracker/detail/localization_models.py
+++ b/lumicks/pylake/kymotracker/detail/localization_models.py
@@ -1,5 +1,5 @@
 import numpy as np
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, replace, fields
 
 __all__ = ["LocalizationModel", "GaussianLocalizationModel"]
 
@@ -29,6 +29,21 @@ class LocalizationModel:
             Size of the kymograph in physical units.
         """
         return self.with_position(size - self.position)
+
+    def __add__(self, other):
+        if other.__class__ is not self.__class__:
+            raise TypeError(
+                f"Incompatible localization models {self.__class__.__name__} and "
+                f"{other.__class__.__name__}."
+            )
+
+        init_kwargs = {
+            f.name: np.hstack((getattr(self, f.name), getattr(other, f.name))) for f in fields(self)
+        }
+        return self.__class__(**init_kwargs)
+
+    def __getitem__(self, item):
+        return self.__class__(**{f.name: getattr(self, f.name)[item] for f in fields(self)})
 
 
 @dataclass(frozen=True)
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -277,6 +277,30 @@ class KymoTrack:
         interpolated_coord = np.interp(interpolated_time, self.time_idx, self.coordinate_idx)
         return self._with_coordinates(interpolated_time, interpolated_coord)
 
+    def _split(self, node):
+        """Split track.
+
+        Splits a track at `node`. Returns two tracks if successful.
+
+        Parameters
+        ----------
+        node : int
+            Node index to split at
+
+        Raises
+        ------
+        ValueError
+            If asked to split at a point that would result in an empty track.
+        """
+        node = np.clip(node, 0, len(self))
+        before = self._with_coordinates(self.time_idx[:node], self._localization[:node])
+        after = self._with_coordinates(self.time_idx[node:], self._localization[node:])
+
+        if not before or not after:
+            raise ValueError("Invalid split point. This split would result in an empty track")
+
+        return before, after
+
     def sample_from_image(self, num_pixels, reduce=np.sum):
         """Sample from image using coordinates from this KymoTrack.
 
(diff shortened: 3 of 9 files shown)


changelog.md
```diff
@@ -5,6 +5,7 @@
 #### New features
 
 * Added [`DwelltimeModel.profile_likelihood()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.DwelltimeModel.html#lumicks.pylake.DwelltimeModel.profile_likelihood) for calculating profile likelihoods for dwell time analysis. These can be used to determine confidence intervals and assess whether the model has too many parameters (i.e., if a model with fewer components would be sufficient).
+* Added functionality to split lines to the Kymotracking widget.
```
rpauszek, line 8:
```suggestion
* Added functionality to split tracks in the Kymotracking widget.
```

changelog.md
```diff
@@ -18,6 +19,7 @@
 * Added lower and upper bounds when fitting dwell time models. This prevents numerical issues that occur when approaching these extremes. Prior to this change, over-parameterized models could encounter numerical issues due to divisions by (near) zero. After this change, amplitudes are constrained to stay between `1e-9` and `1.0 - 1e-9`, while lifetimes are constrained to remain between `1e-8` and `1e8`.
 * Provided better initial guess for dwell time models with three or more components. Prior to this change, the dwelltime models involving three components or more would get an initial guess with an average lifetime beyond the mean of the observations.
 * Silenced false positive warnings about the optimization violating the bound constraints. Prior to this change, dwell time analysis would produce warnings about the optimizer exceeding bounds. These excursions beyond the bound are very small and have no effect on the model.
+* Parameters corresponding to Gaussian refinement are now preserved when two refined lines are concatenated.
```
rpauszek, line 22:
```suggestion
* Parameters corresponding to Gaussian refinement are now preserved when two refined tracks are concatenated.
```

lumicks/pylake/kymotracker/detail/localization_models.py
```diff
@@ -30,6 +30,21 @@ def _flip(self, size):
...
         return self.with_position(size - self.position)
 
+    def __add__(self, other):
+        if other.__class__ is not self.__class__:
+            raise TypeError(
+                f"Incompatible localization models {self.__class__.__name__} and "
+                f"{other.__class__.__name__}."
+            )
+
+        init_args = {
```
rpauszek, line 40: nitpick: `init_kwargs`?

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -273,6 +277,18 @@ def interpolate(self):
...
+        """Split track
+
+        Parameters
+        ----------
+        node : int
+            Node index to split at
+        """
+        return self._with_coordinates(
+            self.time_idx[:node], self._localization[:node]
+        ), self._with_coordinates(self.time_idx[node:], self._localization[node:])
```
rpauszek, line 290: nitpick: I don't really like this enforced formatting. Maybe clearer to just make two variables and then return both rather than all in one shot

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -764,12 +780,23 @@ def _flip(self):
...
+        Parameters
+        ----------
+        track : KymoTrack
+            Track to split
+        split_node : int
+            Index of the node in the track to split at
+        """
+        first_half, second_half = track._split(split_node)
+        self._src[self._src.index(track)] = first_half
+        self._src.append(second_half)
```
rpauszek, line 795: I wonder would it make more sense to insert the "new" track right after the first half of the original one? Basically if you start with `[track1, track2, track3]` and split on `track2` end up with `track1, track2a, track2b, track3]`. I think there are arguments for both. This way you still preserve the spatial order of the tracks (but that's not really guaranteed on our side anyway once you start messing the things), whereas your proposal would preserve the indexing of the list.

I could be convinced either way, just some food for thought

lumicks/pylake/nb_widgets/detail/mouse.py
```diff
@@ -52,12 +54,17 @@ def set_active(self, state):
...
             )
+
             self._callback_ids["release"] = self._axes.get_figure().canvas.mpl_connect(
                 "button_release_event", lambda event: self.button_release(event)
             )
-            self._callback_ids["motion"] = self._axes.get_figure().canvas.mpl_connect(
-                "motion_notify_event", lambda event: self.handle_motion(event)
-            )
+
+            if self._drag_callback:
```
rpauszek, line 62: why do we need this extra `if` now?

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -169,6 +170,29 @@ def get_nearest(x, y):
...
+
+            nodes = get_node_info()
+            distance, idx = get_nearest(event.x, event.y)
+            if distance < cutoff_radius:
+                track_index, node_index, seconds, _ = nodes[idx]
+
+                # Explicit copy to ensure current state pushed to undo stack on assignment.
+                tracks = copy(self.tracks)
+
+                # If we click _beyond_ the note, cut there! Without adding the check on seconds
```
rpauszek, line 187:
```suggestion
                # If we click _beyond_ the node, cut there! Without adding the check on seconds
```

rpauszek (approved):
looks good. left some minor suggestions/corrections. Also some thoughts on where we want to put the "new" track, but I can go either way

mikhas:
Nice work! So it splits the track between the two data points? What if I split between the first & second data point (or last & penultimate)? Or more general: What if the split results in tracks that suddenly don't make the `min_len` cut?

lumicks/pylake/kymotracker/detail/localization_models.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Personally, I see more room for error by performing that insertion here. One might not realize they're performing an insertion, while iterating on a list and selectively splitting and then get errors.

JoepVanlier:
> So it splits the track between the two data points? 
Yes.

> What if I split between the first & second data point (or last & penultimate)?
You get a single point track, which is invisible in the kymotracker. This is a good point.

> Or more general: What if the split results in tracks that suddenly don't make the `min_len` cut?
I would argue that a pass would have to be added to filter tracks `< min_length` after this split operation on the group.

rpauszek:
> > So it splits the track between the two data points?
> 
> Yes.
> 
> > What if I split between the first & second data point (or last & penultimate)?
> 
> You get a single point track, which is invisible in the kymotracker. This is a good point. I would argue split should either throw or be a `nullopt` if one of the sides ends up empty or a single point.
> 
> > Or more general: What if the split results in tracks that suddenly don't make the `min_len` cut?
> 
> I would argue that a pass would have to be added to filter tracks `< min_length` after this split operation on the group.

`min_length` is only enforced when you call `filter_tracks` so as far as the widget goes, this specific case should be handled there (not in `KymoTrackGroup`). I would then say probably `raise` if a `0` length track results, but allow for a length of `1`, since you can get these from `track_greedy()` anyways, and then need to specifically call `filter_tracks` afterwards.

JoepVanlier:
> > > So it splits the track between the two data points?
> > 
> > 
> > Yes.
> > > What if I split between the first & second data point (or last & penultimate)?
> > 
> > 
> > You get a single point track, which is invisible in the kymotracker. This is a good point. I would argue split should either throw or be a `nullopt` if one of the sides ends up empty or a single point.
> > > Or more general: What if the split results in tracks that suddenly don't make the `min_len` cut?
> > 
> > 
> > I would argue that a pass would have to be added to filter tracks `< min_length` after this split operation on the group.
> 
> `min_length` is only enforced when you call `filter_tracks` so as far as the widget goes, this specific case should be handled there (not in `KymoTrackGroup`). I would then say probably `raise` if a `0` length track results, but allow for a length of `1`, since you can get these from `track_greedy()` anyways, and then need to specifically call `filter_tracks` afterwards.

I added a `min_length` parameter to the split function at the group level. When tracks are being dropped, it now issues a warning which gets propagated to the widget. That didn't quite sit well with me for the edge cases where an empty track is returned however.

Just to compare, I added two options to ponder. The first is to just return fewer tracks from `split` when one is empty. This then gets filtered, and issues a warning to the user. Personally, I would find this confusing.

The second is to raise when one side is truly empty and then catch that in `_split_tracks()` at the group level. I think the latter is better, since it allows us to just handle the end points as "no operation took place", which I think is less confusing to users, than getting a warning but seeing no change on screen.

What do you think @mikhas @rpauszek ?

mikhas:
The warning in the first option feels odd because -- as you said yourself -- you already decided to handle the case gracefully, yet still issue a warning. So the question then becomes whether the warning is necessary.

In the second option, you say silently ignoring the edge cases, even _if_ they could be handled somewhat gracefully, from the side of the library implementer, is never OK. Seeing how the edge cases are at times difficult to reason about in an UI, I probably prefer the explicit nature of this option, even if it is slightly more inconvenient to use, API-wise.

So, what I ask myself now is whether KymoTracksGroup operations have grown esoteric enough that the use of context managers is warranted?

JoepVanlier:
> The warning in the first option feels odd because -- as you said yourself -- you already decided to handle the case gracefully, yet still issue a warning. So the question then becomes whether the warning is necessary.

Fair point. Especially considering it is an explicit argument to filter small tracks, it isn't.

> In the second option, you say silently ignoring the edge cases, even _if_ they could be handled somewhat gracefully, from the side of the library implementer, is never OK.

It is a private function whose sole purpose is to do something in the widget. Saying that though, did convince me that the logic should probably live in the widget.

> Seeing how the edge cases are at times difficult to reason about in an UI, I probably prefer the explicit nature of this option, even if it is slightly more inconvenient to use, API-wise.

Right. Explicit exceptions it is. I'll just let it bubble up. You can refer to the code in the kymotracker to see how it should be used in that case.

> So, what I ask myself now is whether KymoTracksGroup operations have grown esoteric enough that the use of context managers is warranted?

I don't really see how a context manager would be beneficial in this situation.

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: good point, yeah let's keep it as is

rpauszek (approved):
🚢

changelog.md
```diff
@@ -19,6 +20,8 @@
 * Added lower and upper bounds when fitting dwell time models. This prevents numerical issues that occur when approaching these extremes. Prior to this change, over-parameterized models could encounter numerical issues due to divisions by (near) zero. After this change, amplitudes are constrained to stay between `1e-9` and `1.0 - 1e-9`, while lifetimes are constrained to remain between `1e-8` and `1e8`.
 * Provided better initial guess for dwell time models with three or more components. Prior to this change, the dwelltime models involving three components or more would get an initial guess with an average lifetime beyond the mean of the observations.
 * Silenced false positive warnings about the optimization violating the bound constraints. Prior to this change, dwell time analysis would produce warnings about the optimizer exceeding bounds. These excursions beyond the bound are very small and have no effect on the model.
+* Parameters corresponding to Gaussian refinement are now preserved when two refined tracks are concatenated.
+* Set minimum `KymoTrack` length to `2` points.
```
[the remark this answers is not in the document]
JoepVanlier: Yes!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -764,12 +792,35 @@ def _flip(self):
         flipped_kymo = self._kymo.flip()
         return KymoTrackGroup([track._flip(flipped_kymo) for track in self])
 
+    def _split_track(self, track, split_node, min_length):
```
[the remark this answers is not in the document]
JoepVanlier: I think in this case, especially since it's internal, it's better to not give it a default to force the user to think about it.

PR #497, opened 2023-04-17
mock_confocal: don't explicitly write SampleFormat.

JoepVanlier:
**Why this PR?**
We don't need to store the `SampleFormat` (which conveys the `datatype`) ourselves.

`tifffile` already infers this from the `numpy` `dtype` and sets this themselves in write [here](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L2726)

It used to ignore it if it was already there, but in the most recent version it produces a [warning](https://github.com/cgohlke/tifffile/blob/master/tifffile/tifffile.py#L118). These warnings were visible when running the benchmark:

```
<tifffile.TiffWriter 'bench_tiff.tiff'> not writing extratag 339
<tifffile.TiffWriter 'bench_tiff.tiff'> not writing extratag 339
<tifffile.TiffWriter 'bench_tiff.tiff'> not writing extratag 339
```


tobiasjj (approved):
LGTM! 🚴

PR #498, opened 2023-04-21
Enable `KymoTrackGroup` with multiple source `Kymo`s

rpauszek:
**Why this PR?**
Enables global analysis across multiple kymos for various downstream analyses. The idea is that each kymo can be tracked separately and then you add all of the resulting groups together. This gives a single interface for downstream analyses regardless of single or multiple kymos tracked.

**Notes**
- Ensemble diffusion enabled for `cve` under all cases
- Ensemble diffusion for `ols` and ensemble MSD checks first if all line times and pixel sizes of the source kymos are the same; otherwise throw
- Exporting to CSV and flipping disabled for more than one source kymo
- Certain features currently raise `NotImplementedError` with this PR; will be implemented next in separate PRs
    - refinement algorithms (centroid and gaussian)
        - idea is to break combined group into single source groups, refine, and then rebuild to the original group
    - dwelltime analysis
        - idea is to allow `min` and `max` acquisition times for each dwell so that normalization constants can be calculated properly
    - `_histogram_binding_profile`
        - necessitates a change in the way time binning is performed to allow for unequal line times. This is maybe even a bug fix since currently if the number of scan lines is not a multiple of the number of frames requested per bin, the leftover scan lines are ignored in the result.
- Started a new test file to specifically test the source kymographs, including (hopefully) all possible combinations of conditions
- Tried to break this down into many commits with each only covering a certain feature or functionality; will squash it down a bit before merging

### DIFF
--- a/lumicks/pylake/kymotracker/detail/msd_estimation.py
+++ b/lumicks/pylake/kymotracker/detail/msd_estimation.py
@@ -1028,7 +1028,7 @@ def estimate_diffusion_cve(
     )
 
 
-def ensemble_cve(kymotracks):
+def ensemble_cve(kymotracks, calculate_localization_var=True):
     """Calculate ensemble-based CVE.
 
     Determines the weighted average of the mean and standard error of the CVE estimates.
@@ -1040,6 +1040,16 @@ def ensemble_cve(kymotracks):
     ----------
     kymotracks : lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup
         Group of kymotracks
+    calculate_localization_var : bool
+        If localization variance can be calculated. Must be False if source kymographs
+        do not have the same line times or pixel sizes.
+
+    Warns
+    -----
+    RuntimeWarning
+        if `method == "cve"` and the source kymographs do not have the same line times
+        or pixel sizes. As a result, the localization variance and variance of the localization
+        variance not be available. Estimates that are unavailable are returned as `np.nan`.
 
     References
     ----------
@@ -1066,8 +1076,11 @@ def ensemble_cve(kymotracks):
     ensemble_mean, ensemble_mean_var = mean_var(
         np.array([c.value for c in cve_based]), counts, counts_sum
     )
-    ensemble_localization_var, var_of_localization_var = mean_var(
-        np.array([c.localization_variance for c in cve_based]), counts, counts_sum
+
+    ensemble_localization_var, var_of_localization_var = (
+        mean_var(np.array([c.localization_variance for c in cve_based]), counts, counts_sum)
+        if calculate_localization_var
+        else (np.nan, np.nan)
     )
 
     return DiffusionEstimate(
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1,3 +1,4 @@
+import itertools
 from copy import copy
 from sklearn.neighbors import KernelDensity
 from .detail.msd_estimation import *
@@ -20,12 +21,21 @@ def export_kymotrackgroup_to_csv(filename, kymotrack_group, delimiter, sampling_
     sampling_width : int or None
         If, this will sample the source image around the kymograph track and export the summed intensity with
         the image. The value indicates the number of pixels in either direction to sum over.
+
+    Raises
+    ------
+    NotImplementedError
+        If group contains tracks from more than one source kymograph.
     """
     if not kymotrack_group:
         raise RuntimeError("No kymograph tracks to export")
 
+    kymotrack_group._validate_single_source(
+        "Exporting a group with tracks from more than a single source kymograph"
+    )
+
     time_units = "seconds"
-    position_units = kymotrack_group._kymo._calibration.unit
+    position_units = kymotrack_group._calibration_info["unit"]
 
     idx = np.hstack([np.full(len(track), idx) for idx, track in enumerate(kymotrack_group)])
     coords_idx = np.hstack([track.coordinate_idx for track in kymotrack_group])
@@ -703,25 +713,85 @@ class KymoTrackGroup:
 
     def __init__(self, kymo_tracks):
         self._src = kymo_tracks
-        if self:
-            if len(set(kymo_tracks)) != len(kymo_tracks):
-                raise ValueError(
-                    "Some tracks appear multiple times. The provided tracks must be unique."
-                )
+        self._kymos = self._validate_compatible_sources()
 
-            self._validate_single_source(kymo_tracks)
+    def _validate_compatible_sources(self, additional_tracks=()):
+        """Check that source kymos for all tracks (including in self) are compatible.
 
-    def _validate_single_source(self, kymo_tracks):
-        kymos = set([track._kymo._id for track in kymo_tracks])
-        channels = set([track._channel for track in kymo_tracks])
+        Parameters
+        ----------
+        additional_tracks : KymoTrackGroup, optional
+            Additional tracks to be added to the current instance.
 
-        if len(kymos) > 1:
-            raise ValueError("All tracks must have the same source kymograph.")
+        Returns
+        -------
+        set
+            Set of source kymograph instances
+        """
+        tracks = list(itertools.chain(self, additional_tracks))
+        if not len(tracks):
+            return ()
 
+        if len(set(tracks)) != len(tracks):
+            raise ValueError(
+                "Cannot extend this KymoTrackGroup with a KymoTrack that is already part of the group"
+                if additional_tracks
+                else "Some tracks appear multiple times. The provided tracks must be unique."
+            )
+
+        channels = {track._channel for track in tracks}
         if len(channels) > 1:
             raise ValueError("All tracks must be from the same color channel.")
 
-        return next(iter(kymos)), next(iter(channels))
+        kymos = tuple({track._kymo: None for track in tracks}.keys())
+        if len(calibrations := set([kymo._calibration.unit for kymo in kymos])) > 1:
+            raise ValueError(
+                f"All tracks must be calibrated in the same units, got {calibrations}."
+            )
+
+        return kymos
+
+    def _validate_single_source(self, method_description):
+        if (n_kymos := len(self._kymos)) > 1:
+            raise NotImplementedError(
+                f"{method_description} is not supported. "
+                f"This group contains tracks from {n_kymos} source kymographs."
+            )
+
+    def _validate_single_linetime_pixelsize(self):
+        """Check that image acquisition attributes (scan line times and pixel sizes)
+        are the same for all source kymos.
+
+        Returns
+        -------
+        bool
+            If validity criteria are met (all source kymographs have the same line times and pixel sizes)
+        str
+            Error message to be raised if validity criteria are not met
+        """
+
+        line_times = {kymo.line_time_seconds for kymo in self._kymos}
+        pixel_sizes = {kymo.pixelsize_um[0] for kymo in self._kymos}
+
+        line_times_err_msg = (
+            ""
+            if len(line_times) == 1
+            else (
+                f"All source kymographs must have the same line times, got {sorted(line_times)} seconds."
+            )
+        )
+
+        px_size_err_msg = (
+            ""
+            if len(pixel_sizes) == 1
+            else (
+                "All source kymographs must have the same pixel sizes, "
+                f"got {sorted(pixel_sizes)} {self._calibration_info['unit']}."
+            )
+        )
+
+        if line_times_err_msg or px_size_err_msg:
+            raise ValueError(" ".join([line_times_err_msg, px_size_err_msg]))
 
     def __iter__(self):
         return self._src.__iter__()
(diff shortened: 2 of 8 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -703,25 +704,55 @@ class KymoTrackGroup:
...
+        if not len(tracks):
+            return dict(
+                kymos=set(),
+                line_times_seconds=set(),
+                pixel_sizes=set(),
+                channel=None,
+                calibration_unit=None,
+                calibration_unit_label=None,
+            )
+            # return SourceAttributes()
```
JoepVanlier: ?

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -537,85 +537,6 @@ def test_empty_binding_histogram():
...
-    # test empty result
-    tracks_empty = KymoTrackGroup([])
-    assert len(tracks_empty) == 0
-
-    tracks_empty = filter_tracks(tracks_a, 5)
-    assert len(tracks_empty) == 0
-
-    with pytest.raises(
-        RuntimeError,
-        match=re.escape("No kymo associated with this empty group (no tracks available)"),
```
JoepVanlier, line 577: I think we're missing a test for this in the new variant.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -703,25 +704,38 @@ class KymoTrackGroup:
...
-            self._validate_single_source(kymo_tracks)
+    def _validate_compatible_sources(self, *additional_tracks):
+        """Check that source kymos for all tracks (including in self) are compatible.
 
-    def _validate_single_source(self, kymo_tracks):
-        kymos = set([track._kymo._id for track in kymo_tracks])
-        channels = set([track._channel for track in kymo_tracks])
+        Parameters
+        ----------
+        *additional_tracks : KymoTrackGroup
```
JoepVanlier, line 723: Why are you passing this as `*args`? Do we need support for multiple here? Wouldn't a default parameter with an empty tuple suffice?

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1262,6 +1291,7 @@ def ensemble_diffusion(self, method, *, max_lag=None):
         if method == "cve":
             return ensemble_cve(self)
```
JoepVanlier, line 1304: I think that while this one will work, we should probably remove the localization uncertainty estimate for "mixed" image acquisition `Kymo`s, since these will depend on the `Kymo` in question. The risk is otherwise that the user gets one number and proceeds to plug this into the per-track estimations.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -703,25 +713,73 @@ class KymoTrackGroup:
...
+        kymos = {track._kymo for track in tracks}
+        if len(calibrations := set([kymo._calibration.unit for kymo in kymos])) > 1:
+            raise ValueError(
+                f"All tracks must be calibrated in the same units, got {calibrations}."
+            )
+
+        return kymos
+
+    def _validate_single_source(self, method_description):
+        if (n_kymos := len(self._kymos)) > 1:
```
JoepVanlier, line 750: :walrus:

JoepVanlier (changes requested):
Looks good. I left a few comments here and there, but I think we'll find out more as we start filling out the blanks.

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -225,8 +225,9 @@ def test_kymotrack_group_extend(blank_kymo):
 
     tracks = KymoTrackGroup([k1, k2, k3])
 
-    duplicate_error = "Cannot extend this KymoTrackGroup with a KymoTrack that is already " \
-                      "part of the group"
+    duplicate_error = (
+        "Cannot extend this KymoTrackGroup with a KymoTrack that is already " "part of the group"
```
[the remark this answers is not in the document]
rpauszek: oops I think that was a `black` mishap

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -969,7 +969,7 @@ def plot(self, *, show_outline=True, show_labels=True, axes=None, **kwargs):
             track.plot(show_outline=show_outline, show_labels=False, axes=ax, **kwargs)
 
         if show_labels:
-            ax.set_ylabel(f"position ({self._kymo._calibration.unit_label})")
+            ax.set_ylabel(f"position ({next(iter(self._kymos))._calibration.unit_label})")
```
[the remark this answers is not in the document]
rpauszek: yeah, my idea was to get rid of `._kymo` once we have the other `NotYetImplemented` features done. I think in the context of potentially multiple source kymos this property could be confusing (or at least could lead to subtle bugs if we don't really think about it).

But considering how much we look up the unit or label, I added a `_calibration_info` property that returns a `dict` with the unit and label.

The other option would be to wrap the functionality of `_validate_single_kymo` into `._kymo`, basically raise if you try to get the property and there's more than one source. Preference?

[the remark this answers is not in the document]
JoepVanlier: Ah okay. Fair enough! Yeah, maybe that is better.

Well, given that most cases I saw were cases where you don't care about which `Kymo` you get. If you raise in those cases, then it doesn't make much sense to have the function anymore.

Just check at the door here then and make sure you don't end up with `StopIteration` for an empty group.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -755,8 +765,7 @@ def _validate_acquisition_attributes(self):
             ""
             if len(line_times) == 1
             else (
-                "All source kymographs must have the same line times, "
-                f"got {line_times} seconds."
+                "All source kymographs must have the same line times, " f"got {line_times} seconds."
```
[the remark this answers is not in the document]
rpauszek: black again

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: I think just my knee jerk reaction to not using iterators as default kwargs, but since `tuple` is immutable works just as well

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -1040,6 +1040,16 @@ def ensemble_cve(kymotracks):
     ----------
     kymotracks : lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup
         Group of kymotracks
+    valid_localization_var : bool
+        If localization variance can be calculated. Must be False if source kymographs
+        do not have the same line times or pixel sizes.
```
JoepVanlier, line 1045: `calculate_localization_var`?

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -1066,9 +1076,20 @@ def mean_var(x, count, count_sum):
...
-    )
+
+    if not valid_localization_var:
+        warnings.warn(
+            RuntimeWarning(
+                "Localization variances cannot be reliably calculated for an ensemble of "
+                "tracks from kymographs with different line times or pixel sizes."
+            ),
+            stacklevel=2,
+        )
```
JoepVanlier, line 1087: To me, having this warning here, seems a bit smelly because the message contains details that you shouldn't have at this level. I would make it optional to calculate it here, but put the warning upstream.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1300,17 +1308,24 @@ def ensemble_diffusion(self, method, *, max_lag=None):
             diffusion constant will not be available. If only one track is available, the standard
             error on the diffusion constant will also not be available. Estimates that are
             unavailable are returned as `np.nan`.
+        RuntimeWarning
+            if `method == "cve"` and the source kymographs do not have the same line times
+            or pixel sizes. As a result, the localization variance and variance of the localization
+            variance not be available. Estimates that are unavailable are returned as `np.nan`.
```
JoepVanlier, line 1325: 🙌

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -776,8 +783,9 @@ def _validate_single_linetime_pixelsize(self):
             )
         )
 
-        if line_times_err_msg or px_size_err_msg:
-            raise ValueError(" ".join([line_times_err_msg, px_size_err_msg]))
+        is_valid = not (line_times_err_msg or px_size_err_msg)
+        err_msg = "" if is_valid else " ".join([line_times_err_msg, px_size_err_msg])
+        return is_valid, err_msg
```
JoepVanlier, line 793: nitpick (feel free to ignore): This seems a bit C-style to me. Not a big deal, but could just throw and catch it if you can handle it?

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: 👍

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -764,9 +764,7 @@ def _validate_single_linetime_pixelsize(self):
         line_times_err_msg = (
             ""
             if len(line_times) == 1
-            else (
-                "All source kymographs must have the same line times, " f"got {line_times} seconds."
-            )
+            else ("All source kymographs must have the same line times, got {line_times} seconds.")
```
JoepVanlier: Not an `f-string`. Would be nice to also add a test for this.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1300,6 +1300,12 @@ def ensemble_diffusion(self, method, *, max_lag=None):
         max_lag : int
             Maximum number of lags to include when using the ordinary least squares method (OLS).
 
+        Raises
+        ------
+        ValueError
+            if `method == "ols"` and the source kymographs do not have the same line times
+            or pixel sizes.
+
```
JoepVanlier, line 1313: 👍

JoepVanlier (changes requested):
Almost there I think.

lumicks/pylake/kymotracker/detail/msd_estimation.py
rpauszek: hah that was what I originally did, but looking at the rest of the API it seemed that the warnings/errors were triggered at the low level API so I changed it. I'll switch it back

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -703,25 +713,84 @@ class KymoTrackGroup:
...
+                if additional_tracks
+                else "Some tracks appear multiple times. The provided tracks must be unique."
+            )
 
+        channels = {track._channel for track in tracks}
         if len(channels) > 1:
             raise ValueError("All tracks must be from the same color channel.")
 
-        return next(iter(kymos)), next(iter(channels))
+        kymos = {track._kymo for track in tracks}
```
JoepVanlier, line 746:
```suggestion
        kymos = {track._kymo: None for track in tracks}
```

Just a thought, but it might be worthwhile to enforce insertion order to prevent surprises.

We could use a dummy dictionary as a poor man's insertion-ordered set. For most intents and purposes it'll work the same, but we get a guaranteed insertion order, which might prevent surprises. A concrete example, when working with a non-identifiable model, and pulling out the tracks in set order, I noticed that results were different (the order of the residual seems to matter).

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: That's how it was originally but then it seemed to get messy in `ensemble_diffusion`; but I think I can live with this solution

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: good catch. I was playing fast and loose with the regex for the actual values (would accept anything between the `{}`) because `set` is not ordered. I sorted it now so that we can guarantee the exact message, but without some extra formatting, we end up with `got [0.5, 1] seconds` instead of `got {0.5, 1} seconds`. I guess here it's not necessary to stick to the strict `set` repr in the message if you're ok with that

JoepVanlier: Right. Regarding the random order of `_kymos`, I wonder if that's not gonna be more of a headache than its worth (see later comment). I predict that this pattern will come up frequently and it may be wise to nip it in the bud now.

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: hmm, seems a bit odd that the order of the residual matters? but we can do this, like you said just to avoid any surprises. I think I came up with a pretty compact yet readable way to basically make a `OrderedTupleFromDictFromSet` 😆 that's surprisingly still pretty readable. Added benefit: we can get rid of all the `next(iter(set))` nonsense

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -776,7 +776,9 @@ def _validate_single_linetime_pixelsize(self):
         line_times_err_msg = (
             ""
             if len(line_times) == 1
-            else (f"All source kymographs must have the same line times, got {sorted(line_times)} seconds.")
+            else (
+                f"All source kymographs must have the same line times, got {sorted(line_times)} seconds."
```
JoepVanlier, line 780: sorting is no longer needed if you go with the `tuple` approach?

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Great!

lumicks/pylake/kymotracker/kymotrack.py
rpauszek: actually you still need it because you're still making a `set` to make sure that you have a single value (even if it comes from 2 different kymos). So `._kymos` is `tuple` but what you're checking is still a (unordered) `set`

JoepVanlier: Ah right, of course.

JoepVanlier (approved):
LGTM!

200w

PR #499, opened 2023-04-24
Update install.rst: Add cv2 error to faq

aafkevandenberg:
I included a method to resolve the error that cv2 cannot be found. This solution is the first suggestion on the opencv website, https://pypi.org/project/opencv-python/, and worked for at least one customer.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -304,6 +304,18 @@ If you have already activated the correct environment, but you still do not see 
 Frequently asked questions
 --------------------------
 
+**When importing lumicks.pylake on Windows I get an error with cv2**
+
+The full error message is::
+
+    ImportError: DLL load failed while importing cv2: The specified procedure could not be found.
+    
+To resolve this error, follow these steps:
+    
+* Remove Anaconda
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -304,6 +304,18 @@ If you have already activated the correct environment, but you still do not see
...
 
+**When importing lumicks.pylake on Windows I get an error with cv2**
+
+The full error message is::
+
+    ImportError: DLL load failed while importing cv2: The specified procedure could not be found.
+    
+To resolve this error, follow the following steps:
+    
+* Remove the current Pylake installation. If you are not using Anaconda for anything else but Pylake, you can simply remove Anaconda in this step.
```
JoepVanlier, line 315: Are we sure removing only Pylake is sufficient?

aafkevandenberg: No.... I will just say to remove Anaconda then.

JoepVanlier (approved):
Looks good to go. Hope this resolves most of these issues in the field. 🐍

docs/install.rst
JoepVanlier: 👍

PR #500, opened 2023-04-25
profile_likelihood: make sure plots work in the `n_components=1` case

JoepVanlier:
**Why this PR?**
Found this plotting bug while working on the generalization of dwelltime analysis.

Not sure how this happened or why it wasn't covered, but luckily it was caught before this feature was ever released. Added a test so that this case is covered as well.

I left the test and fix as separate commits so you can verify that it does indeed hit the bug. I'll squash them down to one commit before merging.


rpauszek (approved):
😅 🚀

PR #501, opened 2023-04-25
dwelltime analysis: enabling per observation pdf support limits for the `DwelltimeModel`

JoepVanlier:
**Why this PR?**
This model forms the backend for the dwell time analysis on a `KymoLineGroup`.

To enable global analysis, we need to take into account that observations coming from different kymographs may have had different support (different range of possible observation times). For example, the minimum observable dwelltime depends on the line time of a kymograph.

One thing that's unexpectedly a little tricky is how to handle plotting. When dealing with two different observation limits, it is entirely possible that one covers a bin only partially. This means that we have to take into account both the weighting inside the pdf, but also the effect discretization has on the model.

Not taking the partial bin effect into account:
image

Taking both into account:
image

If you discretize less, you can see what's really happening here:
image

For now, I just evaluate the PDF at a larger number of points. Basically, for every bin I sample a set number of points. The `10000` default is somewhat arbitrary, but it doesn't take very long at all, and should be fine enough for all purposes.

### DIFF
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -467,16 +467,24 @@ class DwelltimeModel:
         physically relevant parameter.*
     n_components : int
         number of components in the model.
-    min_observation_time : float
-        minimum experimental observation time
-    max_observation_time : float
-        maximum experimental observation time.
+    min_observation_time : float or numpy.ndarray
+        minimum experimental observation time. This can either be a scalar or a numpy array with
+        the same length as `dwelltimes`.
+    max_observation_time : float or numpy.ndarray
+        maximum experimental observation time. This can either be a scalar or a numpy array with
+        the same length as `dwelltimes`.
     tol : float
         The tolerance for optimization convergence. This parameter is forwarded as the `ftol`
         argument to :func:`scipy.optimize.minimize(method="SLSQP") <scipy.optimize.minimize()>`.
     max_iter : int
         The maximum number of iterations to perform. This parameter is forwarded as the `maxiter`
         argument to :func:`scipy.optimize.minimize(method="SLSQP") <scipy.optimize.minimize()>`.
+
+    Raises
+    ------
+    ValueError
+        if `min_observation_time` or `max_observation_time` is not a scalar and is not the same
+        length as `dwelltimes`.
     """
 
     def __init__(
@@ -490,6 +498,13 @@ class DwelltimeModel:
         max_iter=None,
         use_jacobian=True,
     ):
+        for time, msg in ((min_observation_time, "minimum"), (max_observation_time, "maximum")):
+            if not np.isscalar(time) and not (time.size == dwelltimes.size and time.ndim == 1):
+                raise ValueError(
+                    f"Size of {msg} observation time array ({time.size}) must be equal to that "
+                    f"of dwelltimes ({len(dwelltimes)})."
+                )
+
         self.n_components = n_components
         self.dwelltimes = dwelltimes
         self.use_jacobian = use_jacobian
@@ -800,13 +815,17 @@ def _exponential_mixture_log_likelihood_jacobian(params, t, t_min, t_max):
     t = t[np.newaxis, :]
 
     exp_bound_difference = np.exp(-t_min / lifetimes) - np.exp(-t_max / lifetimes)
-    norm_factor = 1.0 / np.sum(amplitudes * exp_bound_difference)
+    norm_factor = 1.0 / np.sum(amplitudes * exp_bound_difference, axis=0)
 
     # Calculate derivatives of the normalization constant
     dnorm_damp = -(norm_factor**2) * exp_bound_difference
     dlognorm_damp = (1.0 / norm_factor) * dnorm_damp
 
-    max_bound = t_max * np.exp(-t_max / lifetimes) if np.all(t_max / lifetimes < 1e10) else 0
+    max_div_lifetimes = t_max / lifetimes
+    valid = max_div_lifetimes < 1e10
+    max_bound = np.zeros(max_div_lifetimes.shape)
+    max_bound[valid] = (t_max * np.ones(lifetimes.shape))[valid] * np.exp(-max_div_lifetimes[valid])
+
     exp_bound_difference2 = max_bound - t_min * np.exp(-t_min / lifetimes)
     dnorm_dtau = norm_factor**2 * (amplitudes / (lifetimes**2)) * exp_bound_difference2
     dlognorm_dtau = (1.0 / norm_factor) * dnorm_dtau
@@ -981,7 +1000,10 @@ def _exponential_mle_bounds(n_components, min_observation_time, max_observation_
     return (
         *[(1e-9, 1.0 - 1e-9) for _ in range(n_components)],
         *[
-            (max(min_observation_time * 0.1, 1e-8), min(max_observation_time * 1.1, 1e8))
+            (
+                max(np.min(min_observation_time) * 0.1, 1e-8),
+                min(np.max(max_observation_time) * 1.1, 1e8),
+            )
             for _ in range(n_components)
         ],
     )
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -797,7 +797,7 @@ class DwelltimeModel:
         else:
             raise ValueError("spacing must be either 'log' or 'linear'")
 
-        bins = scale(*limits, n_bins)
+        bins = scale(*limits, n_bins + 1)
         hist_kwargs = {"facecolor": "#cdcdcd", "edgecolor": "#aaaaaa", **(hist_kwargs or {})}
         component_kwargs = {"marker": "o", "ms": 3, **(component_kwargs or {})}
         fit_kwargs = {"color": "k", **(fit_kwargs or {})}
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -319,11 +319,21 @@ class DwelltimeBootstrap:
         n_data = optimized.dwelltimes.size
         samples = np.empty((optimized._parameters.size, iterations))
         for itr in range(iterations):
-            sample = np.random.choice(optimized.dwelltimes, size=n_data, replace=True)
+            # Note: we have per-dwelltime limits, so we need to resample these too. Form an array
+            # with dwelltimes, min limits, and max limits as columns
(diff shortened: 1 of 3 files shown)


lumicks/pylake/population/dwelltime.py
```diff
@@ -669,12 +693,27 @@ def pdf(self, x):
...
+                * np.exp(
+                    _exponential_mixture_log_likelihood_components(
+                        self.amplitudes,
+                        self.lifetimes,
+                        x,
+                        *limits,
+                    )
+                )
+                for limits, count in zip(unique_limits, counts)
+            ]
```
rpauszek, line 712: I think we discussed this briefly, but I'm not sure how I feel about changing the functionality of `.pdf()`. This implementation is specifically for plotting over the histogram, but there are still use cases for evaluating the PDF at a single point rather than over a discrete bin; which by definition at that point you're either below or above any particular observation limit.

Would it maybe make sense to keep the original PDF method, and add this as a specific separate method (probably private)?

lumicks/pylake/population/dwelltime.py
```diff
@@ -744,7 +788,11 @@ def hist(
         component_kwargs = {"marker": "o", "ms": 3, **(component_kwargs or {})}
         fit_kwargs = {"color": "k", **(fit_kwargs or {})}
 
-        components = self.pdf(centers)
+        # We average over a relatively fine mesh to make sure our PDF estimates are good when not
+        # all t_min and t_max are the same.
+        n_sample = max(3, int_steps // n_bins)
+        time_points = np.hstack([np.linspace(t1, t2, n_sample) for t1, t2 in zip(bins, bins[1:])])
+        components = np.mean(self.pdf(time_points).reshape(self.n_components, -1, n_sample), axis=2)
```
rpauszek, line 795: if you go with the suggestion of implementing a separate discretized PDF, could potentially wrap this in there too to keep the full implementation together

lumicks/pylake/population/dwelltime.py
```diff
@@ -800,13 +848,17 @@ def _exponential_mixture_log_likelihood_jacobian(params, t, t_min, t_max):
...
 
     # Calculate derivatives of the normalization constant
     dnorm_damp = -(norm_factor**2) * exp_bound_difference
     dlognorm_damp = (1.0 / norm_factor) * dnorm_damp
 
-    max_bound = t_max * np.exp(-t_max / lifetimes) if np.all(t_max / lifetimes < 1e10) else 0
+    max_div_lifetimes = t_max / lifetimes
+    valid = max_div_lifetimes < 1e10
+    max_bound = np.zeros(max_div_lifetimes.shape)
+    max_bound[valid] = (t_max * np.ones(lifetimes.shape))[valid] * np.exp(-max_div_lifetimes[valid])
```
rpauszek, line 860: could save on some parentheses to make this a bit cleaner
```suggestion
    max_bound[valid] = np.full(lifetimes.shape, t_max)[valid] * np.exp(-max_div_lifetimes[valid])
```

lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -31,6 +31,44 @@ def test_likelihood(exponential_data):
...
+    [
+        (np.array([0.2, 0.3, 0.6, 1.2]), 0.1, 1.4, -0.604631649126058),
+        (np.array([0.2, 0.3, 0.6, 1.2]), np.array([0.1, 0.1, 0.3, 0.3]), 1.3, 0.25784386656710745),
+        (np.array([0.2, 0.3, 0.6, 1.2]), 0.1, np.array([0.5, 0.5, 1.3, 1.3]), 1.1932900895106002),
+        (
+            np.array([0.2, 0.3, 0.6, 1.2]),
+            np.array([0.1, 0.1, 0.3, 0.3]),
+            np.array([0.5, 0.5, 1.3, 1.3]),
+            1.7155483581674074,
+        ),
```
rpauszek, line 45: I'd be fine wrapping this with a `#fmt: off`

rpauszek (changes requested):
Looks really nice, mostly I just left some minor wording/commenting suggestions. The one thing to take a look at before approving though is about changing the functionality for the `.pdf()` method. See what you think

lumicks/pylake/population/dwelltime.py
JoepVanlier: I'm not sure what you mean by "which by definition at that point you're either below or above any particular observation limit.".

Do you have a concrete use case for it? And which observation limits would you use in that case?

This one isn't specific for histogram. The added complexity is needed for the fact that there can be multiple observation limits.

The old histogram also contained a bug (maybe this needs a changelog entry as well actually 🤔), in the sense that it actually (incorrectly) returned nonzero values beyond the observation limits (meaning that technically, given that you've already taken the observation limits into account in the normalization, it would be incorrectly normalized).

lumicks/pylake/population/dwelltime.py
```diff
@@ -467,16 +476,24 @@ class DwelltimeModel:
...
-    min_observation_time : float
-        minimum experimental observation time
-    max_observation_time : float
-        maximum experimental observation time.
+    min_observation_time : float or numpy.ndarray
+        minimum experimental observation time. This can either be a single value or a numpy
+        array with the same length as `dwelltimes`.
+    max_observation_time : float or numpy.ndarray
+        maximum experimental observation time. This can either be a single value or a numpy
+        array with the same length as `dwelltimes`.
```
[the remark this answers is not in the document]
JoepVanlier: Yes. 👍

lumicks/pylake/population/dwelltime.py
```diff
@@ -467,16 +476,24 @@ class DwelltimeModel:
...
         argument to :func:`scipy.optimize.minimize(method="SLSQP") <scipy.optimize.minimize()>`.
     max_iter : int
         The maximum number of iterations to perform. This parameter is forwarded as the `maxiter`
         argument to :func:`scipy.optimize.minimize(method="SLSQP") <scipy.optimize.minimize()>`.
+
+    Raises
+    ------
+    ValueError
+        if the number of minimum or maximum observation times does not equal one or the number of
+        data points provided in the `dwelltimes` argument.
```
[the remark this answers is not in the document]
JoepVanlier: It certainly is, thanks!

lumicks/pylake/population/dwelltime.py
JoepVanlier: That's not the same though. `t_max` can either be an array or scalar.

lumicks/pylake/population/tests/test_dwelltimes.py
JoepVanlier: 👍

lumicks/pylake/population/dwelltime.py
rpauszek: oh yeah that's the whole point of this PR 😖

lumicks/pylake/population/dwelltime.py
JoepVanlier: I split the discretized PDF into its own function. While doing this, I actually noticed that the number of bins requested is not actually the number of bins you get. I fixed this small issue in a separate commit.

lumicks/pylake/population/dwelltime.py
```diff
@@ -319,11 +319,20 @@ def _sample(optimized, iterations) -> np.ndarray:
         n_data = optimized.dwelltimes.size
         samples = np.empty((optimized._parameters.size, iterations))
         for itr in range(iterations):
-            sample = np.random.choice(optimized.dwelltimes, size=n_data, replace=True)
+            # Note: we have per-dwelltime limits, so we need to resample these too.
+            to_resample = np.vstack(
+                list(np.broadcast(optimized.dwelltimes, *optimized._observation_limits))
+            )
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/population/dwelltime.py
rpauszek: yeah I was confusing this with the discretized PDF needed for histogramming (which you addressed in the next comment). It is weird because this PDF is not generalizable (it's set to the fraction of data points with each min/max combo used to train it) ... but there's no real better way to do it, it's just a strange scenario. Realistically based on expected acquisition parameters and biological lifetimes, most people won't even notice this behavior so it's fine to keep it as is.

lumicks/pylake/population/dwelltime.py
```diff
@@ -669,12 +694,27 @@ def pdf(self, x):
...
-            _exponential_mixture_log_likelihood_components(
-                self.amplitudes, self.lifetimes, x, *self._observation_limits
-            )
+        limits = np.vstack(list(np.broadcast(*self._observation_limits)))
+        unique_limits, counts = np.unique(limits, axis=0, return_counts=True)
+
+        pdfs = np.asarray(
+            [
+                (count / sum(counts))
+                * np.logical_and(x >= limits[0], x < limits[1])
```
rpauszek, line 703: do we still need this bit since the components will be `0` outside the bounds when returned from `_exponential_mixture_log_likelihood_components`?

lumicks/pylake/population/dwelltime.py
```diff
@@ -695,6 +735,26 @@ def plot(
...
+        ----------
+        bins : numpy.ndarray
+            Monotonically increasing array of bin edges.
+        int_steps : int
+            To obtain accurate pdf values when plotting, we sample the pdf at a finer resolution
+            than the histogram bins. The number of steps we evaluate per bin is defined as:
+            max(3, int_steps // n_bins). Default: 10000.
+        """
+        # We average over a relatively fine mesh to make sure our PDF estimates are good when not
+        # all t_min and t_max are the same.
```
rpauszek, line 751: do we need this since it's basically in the docstring now?

lumicks/pylake/population/dwelltime.py
```diff
@@ -723,6 +784,10 @@ def hist(
             scaling for the x-axis; when `None` default is "linear"
         yscale : {"log", "linear", None}
             scaling for the y-axis; when `None` default is same as `bin_spacing`
+        int_steps : int
+            To obtain accurate pdf values when plotting, we sample the pdf at a finer resolution
+            than the histogram bins. The number of steps we evaluate per bin is defined as:
+            max(3, int_steps // n_bins). Default: 10000.
```
rpauszek, line 790: do we want to expose this to the users? I think it'd actually be more confusing (because most people won't notice this weird mixed observation range stuff). I would say let's just give it a default for us internally and if we hear of issues we can expose it then.

PR #502, opened 2023-04-25
kymotrackgroup: multi-kymo dwell time analysis

JoepVanlier:
**Why this PR?**
The generalization of dwell time analysis to using multiple minimum and maximum observation times introduced in [this](https://github.com/lumicks/pylake/pull/501) PR was meant to support global analysis on `KymoTrackGroup`s containing lines originating from multiple kymographs. This PR hooks up these two.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1064,6 +1064,76 @@ class KymoTrackGroup:
             KymoTrackGroup([track for track in self if track._kymo == kymo]) for kymo in self._kymos
         ]
 
+    @staticmethod
+    def _extract_dwelltime_data_from_groups(groups, exclude_ambiguous_dwells):
+        """Compute data needed for dwelltime analysis from a dictionary of KymoTrackGroups.
+
+        This data consists of dwelltimes and observation limits per track. Note that dwelltimes of zero
+        are automatically dropped.
+
+        Parameters
+        ----------
+        groups : iterable of KymoTrackGroup
+            An iterable which provides a sequence of KymoTrackGroup. Note that each group can only
+            have one `Kymo` associated with it.
+        exclude_ambiguous_dwells : bool
+            Determines whether to exclude dwelltimes which are not exactly determined. If `True`,
+            tracks which start in the first frame or end in the last frame of the kymograph are not
+            used in the analysis, since the exact start/stop times of the binding event are not
+            definitively known.
+
+        Returns
+        -------
+        dwelltimes : numpy.ndarray
+            Dwelltimes
+        min_obs : numpy.ndarray
+            List of minimum observation times extracted from the kymos
+        max_obs : numpy.ndarray
+            List of maximum observation time
+        removed_zeros : bool
+            Whether zeroes were dropped
+
+        Raises
+        ------
+        ValueError
+            if one of the KymoTrackGroups has more than one `Kymo` associated with it
+        """
+        removed_zeros = False
+
+        def extract_dwelltime_data(group):
+            nonlocal removed_zeros
+
+            if len(group._kymos) > 1:
+                raise ValueError("This group has more than one Kymo associated with it.")
+
+            tracks = (
+                filter(KymoTrack._check_ends_are_defined, group)
+                if exclude_ambiguous_dwells
+                else group
+            )
+            dwelltimes_sec = np.array([track.seconds[-1] - track.seconds[0] for track in tracks])
+            nonzero_dwelltimes_sec = dwelltimes_sec[dwelltimes_sec > 0]
+            removed_zeros = removed_zeros or len(nonzero_dwelltimes_sec) != len(dwelltimes_sec)
+
+            # Gracefully handle empty groups
+            if nonzero_dwelltimes_sec.size == 0:
+                return np.empty((3, 0))
+
+            min_observation_time = np.min(nonzero_dwelltimes_sec)
+            max_observation_time = group[0]._image.shape[1] * group[0]._line_time_seconds
+
+            return np.vstack(
+                list(
+                    np.broadcast(nonzero_dwelltimes_sec, min_observation_time, max_observation_time)
+                )
+            ).T
+
+        dwelltimes, min_obs, max_obs = np.hstack(
+            [extract_dwelltime_data(tracks) for tracks in groups]
+        )
+
+        return dwelltimes, min_obs, max_obs, removed_zeros
+
     def fit_binding_times(
         self, n_components, *, exclude_ambiguous_dwells=True, tol=None, max_iter=None
     ):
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1072,20 +1149,19 @@ def fit_binding_times(
             The maximum number of iterations to perform. This parameter is forwarded as the `maxiter` argument
             to `scipy.minimize(method="L-BFGS-B")`.
         """
-        self._validate_single_source("Dwelltime analysis")
+        if not self:
```
alessiamarcolini, line 1152: is this a shortcut to `if len(self) < 1` ?

if so, it looks a bit weird and hard to read if someone is not familiar with the codebase

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -7,6 +7,70 @@
...
+    ------
+    ValueError
+        if one of the KymoTrackGroups has more than one `Kymo` associated with it
+    """
+    removed_zeros = False
+
+    def extract_dwelltime_data(group):
+        nonlocal removed_zeros
+
+        if len(group._kymos) > 1:
```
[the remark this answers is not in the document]
JoepVanlier: We have no hard guidelines on touching private API that belongs to Pylake itself. We mostly decide this on a case-by-case basis.

While I'm aware that that is the ideal situation, in this case, the classes involved here are user-facing classes and we don't want to expose implementational details to the user.

We have quite a few of these functions that we consider "public" for internal use, but which are actually protected because we want to be able to change them at any time without the risk that outside users may end up depending on them. In addition, we also don't really want to pollute the namespace with them.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -7,6 +7,70 @@
...
+            raise RuntimeError("This group has more than one Kymo associated with it.")
```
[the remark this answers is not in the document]
JoepVanlier: Good catch, thanks!

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Isn't this a pretty common empty check for a list-like class though? I don't mind changing it to `if not len(self):` if you think that is more clear.

But I'd rather not go with the `< 1`. Afaik, length dunders cannot return negative values, so the `< 1` would make me wary whether something special is happening here.

alessiamarcolini (approved):
LGTM

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -7,6 +7,70 @@
 from ..population.dwelltime import DwelltimeModel
 
 
+def _extract_dwelltime_data_from_groups(groups, exclude_ambiguous_dwells):
```
rpauszek, line 10: I don't really like this at the module level. I think it's distracting, especially for a function that's only used in one place. I get that it makes sense here in that it doesn't use `self`, but I think I'd prefer it as a method on `KymoTrackGroup` and then can call `_tracks_by_kymo()` internally rather than passing in `groups`.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1052,6 +1116,19 @@ def save(self, filename, delimiter=";", sampling_width=None):
...
+
+        Returns
+        -------
+        dict of KymoTrackGroup
+            returns a dictionary where the keys are Kymos which the associated tracks
+        """
+        return {
+            kymo: KymoTrackGroup([track for track in self if track._kymo == kymo])
+            for kymo in self._kymos
+        }
```
rpauszek, line 1130: Considering what will be needed for supporting the refinement functions and that you only ever use `.values()` from the returned `dict` I think it'd be better to just return a `list` here.

For the refinement we'll need to also return the indices that the tracks were originally at so we can reconstruct the group in the same order. So leaving it this way you'd return something like `{kymo: (KymoTrackGroup, list)}` which is just kinda unnecessarily messy

lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
```diff
@@ -255,10 +256,109 @@ def test_multisource_not_implemented(kymos, coordinates):
...
+
+    for key in track_dictionary.keys():
+        tracks_dict = track_dictionary[key]
+        tracks_raw = tracks[key]
+        assert len(tracks_dict) == len(tracks_raw)
+        for track_dict, track_raw in zip(tracks_dict, tracks_raw):
+            id(track_dict) == id(track_raw)
+
+
+def test_multi_kymo_dwell(kymos):
```
rpauszek, line 282: I think I had kept these kind of tests (and the next one) in `test_kymotrack.py` next to the ones that essentially test the same functionality for single source. I'd honestly rather not keep adding to that file (especially considering the length of the file) but I think it's better to keep tests for the same functionality in the same location rather than split across different files.

rpauszek (changes requested):
Yay for global dwelltime analysis. Just a couple things that I think should be moved around, but relatively easy changes

lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
JoepVanlier: Yeah, fair enough. I will move them.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Yeah, fair enough. Initially I was doubting whether this function should actually live here, or in the dwelltime module, but now that it's here, I agree it'd be nicer to stick it in the class where it belongs.

I'd prefer to keep it a static method though. If all its doing is looking up the tracks immediately with `_tracks_by_kymo()`, I'd rather have that a bit more explicit. It also makes the test a bit more readable imho.

changelog.md
```diff
@@ -8,6 +8,7 @@
 * Added functionality to split tracks in the Kymotracking widget.
 * Added functionality to provide minimum and maximum observation times per data point to `DwelltimeModel` and `DwelltimeBootstrap`.
 * Enabled adding `KymoTrackGroup`s with tracks from different source kymographs. *Note that certain features (e.g., `KymoTrackGroup.ensemble_msd()` and `KymoTrackGroup.ensemble_diffusion("ols")`) require that all source kymographs have the same pixel size and scan line times.*
+* Added support for dwell time analysis for `KymoTrackGroup` with multiple source `Kymo`s.
```
rpauszek, line 11: I wonder if we actually need this? Considering the feature to allow for multiple source kymographs is introduced in the same version, this *just works* rather than being a special case, no?

rpauszek (approved):
Great! just a thought on the changelog, but other than that :shipit:

changelog.md
JoepVanlier: Probably no need indeed.

PR #503, opened 2023-05-05
Refinement for multiple source kymos

rpauszek:
**Why this PR?**
Because we want to be able to refine no matter what.

Basically I implemented a kind of recursion where if you have multiple source kymographs, you split into groups with a single source, refine, then reconstruct to the original group order.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -11,6 +11,7 @@ from .detail.peakfinding import (
 from .detail.localization_models import GaussianLocalizationModel
 import numpy as np
 import warnings
+from itertools import chain
 
 __all__ = [
     "track_greedy",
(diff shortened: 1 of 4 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -486,7 +494,20 @@ def refine_tracks_gaussian(
...
+                group,
+                window,
+                refine_missing_frames,
+                overlap_strategy,
+                initial_sigma=initial_sigma,
+                fixed_background=fixed_background,
+            )
+            for group in groups
+        ]
+        return KymoTrackGroup([track for _, track in sorted(zip(chain(*indices), chain(*groups)))])
```
JoepVanlier, line 510: Nice use of `chain`! It'll also make it work nicely with empty groups.

Just a thought, considering you use this pattern twice, could consider doing something like:
```python
def _apply_to_group(tracks, func, *args, **kwargs):
    groups, indices = tracks._tracks_by_kymo()
    groups = [func(group, *args, **kwargs) for group in groups]
    return KymoTrackGroup([track for _, track in sorted(zip(chain(*indices), chain(*groups)))])
```
Would make it a little easier to test that for some more cases. Then again, we could save that for later, if we end up using the pattern more often.

JoepVanlier (changes requested):
Looks great!

I like that the way this is set up, we only call the refinement functions once per `Kymo`.

There is one subtlety to consider, which is that if you have two tracked regions close to each-other, and you use the option to fit overlapping Gaussians together, tracks that were close can now end up being fitted together, whereas before they couldn't. I don't think this is a problem, but it's something to keep in mind in case questions come up.

The only thing I would like to see is to make the test a little more non-trivial, so that we don't accidentally break this sometime in the future.

JoepVanlier (approved):
Looks good to me.

PR #506, opened 2023-05-09
Gaussian refinement: Don't allow track swaps

JoepVanlier:
**What's the problem?**
While working with some of the multimer data, I noticed frequent swaps when performing Gaussian refinement with tracks close together.

What can happen is that a signal dips below the detection range, in which case it can easily happen that tracks change identity. The way we tried to prevent this is by sorting going into the fitting algorithm, and sorting the results before putting them back into the tracks. However, this method can fail when you consider track interpolation over low signal regions followed by refinement. What happens in these cases is that one of the peaks is missing; therefore causing reassignment of the other tracks.

If you choose the range large, you can quickly get beauties like this:
Screenshot 2023-05-05 at 15 26 20

The solution I propose here is not ideal; but I do think it beats the alternative. Basically, I would propose adding some additional optimization bounds for the peak positions between each peak position, making sure the peaks can not change order.

In addition, I am including a better amplitude estimate that is based on the photon counts found in each range.

See below for a comparison between the initial guesses for the two methods (real data):
old_vs_new

Instead of patching the existing method, I have decided to go with implementing this as a new fitting mode so that we ensure backward compatibility. We could consider deprecating the old one as well.

Note: I rebased this branch on `plot_gaussian`, because it helps to be able to plot things without having to mess with internals.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -511,7 +511,7 @@ def refine_tracks_gaussian(
         How to deal with frames in which the fitting window of two
         :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack`'s overlap.
 
-        - 'multiple' : fit the peaks simultaneously.
+        - 'multiple' : fit the peaks simultaneously (deprecated, use `simultaneous` instead).
         - 'simultaneous' : fit the peaks simultaneously with estimation bounds between peaks.
         - 'ignore' : do nothing, fit the frame as-is (ignoring overlaps).
         - 'skip' : skip optimization of the frame; remove from returned :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack`.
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -520,7 +521,7 @@ def refine_tracks_gaussian(
         Fixed background parameter in photons per second.
         When supplied, the background is not estimated but fixed at this value.
     """
-    if overlap_strategy not in ("ignore", "skip", "multiple"):
+    if overlap_strategy not in ("ignore", "skip", "multiple", "simultaneous"):
```
rpauszek, line 524: I think it would be good to add a deprecation warning if "multiple" is used. It's kinda weird to have 2 options with one of them being buggy, so I assume we'll want to remove this in the next major release?

rpauszek (approved):
Looks good, just see my comment about deprecating the "multiple" option

PR #507, opened 2023-05-10
flake8: B028 got renamed to B907. See link in the description

tobiasjj:
https://github.com/PyCQA/flake8-bugbear/pull/333

This PR adjusts the ignore rule of flake8, accordingly.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -29,6 +29,6 @@ ignore =
     F403,F405,
     # Don't complain about f"'{x}'" quotes. `!r` is not particularly readable, especially not for
     # cases where one would have to resort to f"{str(x)!r}"
-    B028,
+    B907,
 
 show-source = True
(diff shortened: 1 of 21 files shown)


JoepVanlier (commented):
Hmm... I'm on the fence whether to change all these warnings now or wait for the next major version to do it. While it is a bugfix, it is a breaking change since changing the `stacklevel` to `2` will cause certain warnings to change in existing code (and potentially on any other CI that uses `pylake`).

End users can explicitly silence these however.

JoepVanlier:
I wonder if doing something similar as `pandas` does would make sense for this: https://github.com/pandas-dev/pandas/blob/49851bbb723f758bbdbd983fe0cb36c2bc5d1b83/pandas/util/_exceptions.py#L30

Basically, they find the first stack level that isn't in pandas and use that one.

PR #508, opened 2023-05-10
Finalize global binding profile

rpauszek:
**Why this PR?**
For now we're going to specifically disallow plotting binding profiles for `KymoTrackGroup`s that have more than one source kymo. The usefulness of such a plot is questionable and there are a ton of corner cases that could cause problems and/or could be dealt with in multiple ways (none of which are ideal). We can revisit in the future depending on user feedback


JoepVanlier (approved):
LGTM!

PR #509, opened 2023-05-11
kymotrack/kymotrackgroup: allow plotting the localization fit

JoepVanlier:
**Why this PR?**
It is good for users to be able to verify their localization fits, to make sure nothing went wrong. This PR adds two methods to be able to plot them quickly.

peak2

Why is it important to plot this? Well, if a user chooses a window that's too small, they might inadvertently find themselves in the situation where the baseline eats up a large chunk of the amplitude, such as here:

image

A similar situation would happen if you take a window that is too large, but fail to track one of the peaks inside the window.

I intend to add something to the tutorial about this feature (and how to verify that your fits are good etc), but I will keep that for the next PR since there are two PRs with changes that impact the plots in that section.

### DIFF
--- a/lumicks/pylake/detail/utilities.py
+++ b/lumicks/pylake/detail/utilities.py
@@ -30,6 +30,51 @@ def first(iterable, condition=lambda x: True):
     return next(x for x in iterable if condition(x))
 
 
+def replace_key_aliases(dictionary, *key_alias_lists):
+    """Replace key aliases by a specific one.
+
+    This function is especially useful for validating user-supplied plotting kwargs against
+    pylake-supplied default plotting parameters. For instance, a user could supply either
+    `color="red"` or `c="red"` to specify the line color.
+
+    Parameters
+    ----------
+    dictionary : dict
+        Python dictionary.
+    *key_alias_lists : list
+        Each argument is a list of key aliases. When a key in this list is found, it is converted
+        to the first alias for that key in the list.
+
+    Returns
+    -------
+    dict
+        Dictionary with keys replaced by their first alias.
+
+    Raises
+    ------
+    ValueError
+        If multiple keys are provided for the same alias.
+
+    Examples
+    --------
+    ::
+
+        def plot_line(x, y, **kwargs):
+            default_kwargs = {"color": "red"}
+            plt.plot(x, y, **default_kwargs | replace_key_aliases(kwargs, ["color", "c"])
+    """
+    for aliases in key_alias_lists:
+        used_keys = [key for key in aliases if key in dictionary.keys()]
+        if len(used_keys) > 1:
+            raise ValueError(
+                f"Multiple keys provided which are aliases for same property: {used_keys}"
+            )
+        elif len(used_keys) == 1:
+            dictionary[aliases[0]] = dictionary.pop(used_keys[0])
+
+    return dictionary
+
+
 def unique(input_list):
     unique_list = []
     [unique_list.append(x) for x in input_list if x not in unique_list]
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1,6 +1,7 @@
 import itertools
 from copy import copy
 from sklearn.neighbors import KernelDensity
+from ..detail.utilities import replace_key_aliases
 from .detail.msd_estimation import *
 from .detail.localization_models import LocalizationModel, CentroidLocalizationModel
 from .detail.peakfinding import _sum_track_signal
@@ -332,6 +333,58 @@ class KymoTrack:
     def _pixelsize(self):
         """Kymograph (spatial) pixel size in physical units."""
         return self._kymo.pixelsize[0]
+
+    def _model_fit(self, node_idx):
+        """Extract fitted model from localization model at particular node if available"""
+        num_coords = len(self)
+        node_idx = num_coords + node_idx if node_idx < 0 else node_idx
+        if not 0 <= node_idx < num_coords:
+            raise IndexError(
+                f"Node index is out of range of the KymoTrack. Kymotrack has length {num_coords}."
+            )
+
+        pixelsize, num_pixels = self._kymo.pixelsize[0], self._kymo.pixels_per_line
+        coords = np.arange(0.0, num_pixels * pixelsize, 0.1 * pixelsize)
+        model = self._localization.evaluate(coords, node_idx, pixelsize)
+
+        return coords, model
+
+    def plot_fit(self, node_idx, *, fit_kwargs=None, data_kwargs=None, show_data=True):
+        """Plot the localization model
+
+        Plots the model fit for a particular localization. Note that currently, this fit is only
+        available for Gaussian refinement.
+
+        Parameters
+        ----------
+        node_idx : int
+            Index of the track node to plot (corresponding to a specific scan line).
+        fit_kwargs : dict
+            Dictionary containing arguments to be passed to the fit plot.
+        data_kwargs : dict
+            Dictionary containing arguments to be passed to the data plot.
+        show_data : bool
+            Plot the data.
+
+        Raises
+        ------
+        NotImplementedError
+            if localization was not performed using Gaussian localization
+        IndexError
+            if requesting a fit for an index out of bounds of the KymoTrack.
+        """
+        model_fit = self._model_fit(node_idx)  # Verifies whether model exists for this index
+        aliases = ["color", "c"]
+        if show_data:
+            plt.plot(
+                np.arange(0.0, self._kymo.pixels_per_line) * self._kymo.pixelsize[0],
+                self._image[:, self.time_idx[node_idx]],
+                **{"color": "#c8c8c8"} | replace_key_aliases(data_kwargs or {}, aliases),
+            )
+
+        plt.plot(*model_fit, **{"color": "C0"} | replace_key_aliases(fit_kwargs or {}, aliases))
+        plt.xlabel(f"Position [{self._kymo._calibration.unit_label}]")
+        plt.ylabel("Photon counts [#]")
 
     def _check_ends_are_defined(self):
         """Checks if beginning and end of the track are not in the first/last frame."""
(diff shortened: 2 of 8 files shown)


aafkevandenberg (commented):
Very nice feature! I tried it and it works well. I never gave the refinement much thought, it is really insightful to see how different choices affect the gaussian fit.

I don't really like `frame_idx` as `frame` is used for a 2D scan Frame. `Line` is already been used for the tracks. How about `time_idx`?

It would indeed be important to add an explanation to the docs of what you are looking at exactly, what a good fit looks like, and what you can change in the refinement to improve the fit. And also explain how a bad fit affects the downstream analysis.

JoepVanlier:
> Very nice feature! I tried it and it works well. I never gave the refinement much thought, it is really insightful to see how different choices affect the gaussian fit.

👍 

> I don't really like `frame_idx` as `frame` is used for a 2D scan Frame. `Line` has already been used for the tracks. How about `time_idx`?

Ahh, I see.

The reason I went with `frame` on the `KymoTrackGroup` is because it is actually analogous to a Scan `frame`. It is just a 1D scan frame. We also use the notion of frames for Kymos in a few other docstrings (see for example `fit_binding_times`).

For the `KymoTrack`, I wonder if `time_idx` could lead to confusion. A track may skip time points (e.g. time being `0`, `dt`, `2dt`, `3dt`, `5dt` for instance). In that case, `time_idx=4` would be time point `5dt`. This is actually also an issue for the `frame_idx`. So maybe on the `KymoTrack` it should be called `node_idx` to make explicit that it's just the track node we're talking about.

For `KymoTrackGroup` we could go with `frame_idx` or `time_idx`. If you feel strongly about `time_idx` I don't mind changing it for the group specifically.

> It would indeed be important to add an explanation to the docs of what you are looking at exactly, what a good fit looks like, and what you can change in the refinement to improve the fit. And also explain how a bad fit affects the downstream analysis.

Yeah. That's a good point. I think we should probably have dedicated section for localization methods (the kymotracking chapter is too big already as it is). I think it would be good to do that in a separate PR, since it would be nice to expand it a little with some simulations as well. I can make a ticket for it.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1070,6 +1127,122 @@ def plot(self, *, show_outline=True, show_labels=True, axes=None, **kwargs):
...
+            plt.figure()
+            def plot_fit(frame):
+                plt.cla()  # Clear the current axis
+                refined.plot_fit(frame, show_track_index=True)
+
+            widgets.interact(plot_fit, frame=widgets.IntSlider(0, max=kymo.shape[1] - 1));
+        """
+        self._validate_single_source("plot_fit")
+        if not self._kymos:
+            raise ValueError("No tracks in the group")
```
rpauszek, line 1214: elsewhere when we test for empty group and raise we have a different error message (which is leftover from before we added multi-source functionality so there wouldn't be a breaking change). Might be good to keep it consistent.

In `_validate_single_source` we use
```
if not self:
    raise RuntimeError("No kymo associated with this empty group (no tracks available)")
```
however, in `_calibration_info` we have the same message but raise an `IndexError` 😖

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1070,6 +1127,122 @@ def plot(self, *, show_outline=True, show_labels=True, axes=None, **kwargs):
...
+
+        kymo_data = kymo.get_image(self._channel)[:, frame_idx]
+        data_kwargs = {"color": "#c8c8c8", **(data_kwargs or {})}
+        plt.plot(np.arange(kymo.pixels_per_line) * kymo.pixelsize[0], kymo_data, **data_kwargs)
+
+        # Put the numeric values above the baseline
+        y_coord = np.min(kymo_data) + 0.05 * (np.max(kymo_data) - np.min(kymo_data))
+        for track_idx, track_frame_idx in self._tracks_in_frame(frame_idx):
+            self[track_idx].plot_fit(
+                track_frame_idx, show_data=False, show_legend=False, fit_kwargs=fit_kwargs
```
rpauszek, line 1232: why do we need `show_legend=False`?

rpauszek:
> Very nice feature! I tried it and it works well. I never gave the refinement much thought, it is really insightful to see how different choices affect the gaussian fit.
> 
> I don't really like `frame_idx` as `frame` is used for a 2D scan Frame. `Line` has already been used for the tracks. How about `time_idx`?
> 
> It would indeed be important to add an explanation to the docs of what you are looking at exactly, what a good fit looks like, and what you can change in the refinement to improve the fit. And also explain how a bad fit affects the downstream analysis.

I don't really mind `frame_idx` since I think of the scan lines as analogous to frames of a `Scan` but I see your point. Also got me thinking more about the different behaviors between `KymoTrack` and `KymoTrackGroup`. 

For the track the index is referring to the specific node, so if you scroll through you don't see plots for frames that aren't localized. Whereas for `KymoTrackGroup` you will see the data for frames even if there are no tracks.  So semantically it might be best to use something like `node_idx` for `KymoTrack` and `any(frame_idx, time_idx, scan_line_idx)` for `KymoTrackGroup`

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -273,6 +273,63 @@ def _pixelsize(self):
         """Kymograph (spatial) pixel size in physical units."""
         return self._kymo.pixelsize[0]
 
+    def _model_fit(self, coordinate_idx):
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, in retrospect `node_idx` makes more sense, since it communicates that we are really talking about the detected nodes of the track here.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -273,6 +273,63 @@ def _pixelsize(self):
...
+        num_coords = len(self)
+        coordinate_idx = num_coords + coordinate_idx if coordinate_idx < 0 else coordinate_idx
+        if not 0 <= coordinate_idx < num_coords:
+            raise IndexError(
+                f"Coordinate index is out of range of the KymoTrack. Kymotrack has length "
+                f"{num_coords}."
+            )
+
+        pixelsize, num_pixels = self._kymo.pixelsize[0], self._kymo.pixels_per_line
+        coords = np.arange(0.0, num_pixels * pixelsize, 0.1 * pixelsize)
```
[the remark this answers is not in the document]
JoepVanlier: Good idea. I'll make a ticket on the backlog for it; since when doing this, it'd also be nice to change it in other places at the same time. I feel when adding the kymo cropping offset would be a good time to take this in that same sprint.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Actually, I suspect this isn't needed anymore, since _validate_single_source handles that now (when I first made this branch, that wasn't checked there yet).

That does mean this particular exception wasn't tested though. Added a test case now.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Premature optimization. Didn't want to constantly set the `xlabels`, but I just tested a bit and it doesn't matter much at all. I dropped the `show_legend` arg now.

lumicks/pylake/detail/utilities.py
```diff
@@ -30,6 +30,39 @@ def first(iterable, condition=lambda x: True):
     return next(x for x in iterable if condition(x))
 
 
+def replace_key_aliases(dictionary, *key_alias_lists):
+    """Replace key aliases by a specific one.
+
```
rpauszek, line 35: I think it would be useful to add a bit more explanation here as the function is generic (which I like) but a bit abstract-sounding, at least from the argument names. Maybe something like:
```suggestion
    """Replace key aliases by a specific one.
    
       This function is especially useful for validating user-supplied plotting kwargs
       against pylake-supplied default plotting parameters. For instance, a user could 
       supply either `color="red"` or `c="red"` to specify the line color. 

```

lumicks/pylake/detail/utilities.py
```diff
@@ -30,6 +30,39 @@ def first(iterable, condition=lambda x: True):
...
+    Returns
+    -------
+    dict
+        Dictionary with keys replaced by their first alias.
+
+    Raises
+    ------
+    ValueError
+        If multiple keys are provided for the same alias.
+    """
```
rpauszek, line 53: alternative or in conjunction with my previous comment, could include an example for extra clarity
```suggestion
    Raises
    ------
    ValueError
        If multiple keys are provided for the same alias.
        
    Example
    --------
    def plot_line(x, y, **kwargs):
        default_kwargs = {"color": "red"}
        plt.plot(x, y, **default_kwargs | replace_key_aliases(kwargs, ["color", "c"])
    """
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -333,6 +334,58 @@ def _pixelsize(self):
...
+    def plot_fit(self, node_idx, *, fit_kwargs=None, data_kwargs=None, show_data=True):
+        """Plot the localization model
+
+        Plots the model fit for a particular localization. Note that currently, this fit is only
+        available for Gaussian refinement.
+
+        Parameters
+        ----------
+        node_idx : int
+            Index of the coordinate to plot.
```
rpauszek, line 361: this still doesn't sound correct to me
```suggestion
            Index of the track node to plot (corresponding to a specific scan line).
```

rpauszek (approved):
great! pretty much good to go 🚀  just left some minor wording suggestions

lumicks/pylake/detail/utilities.py
JoepVanlier: Yeah, that's a good point. I bet our future selves would be thankful for that :D

lumicks/pylake/detail/utilities.py
JoepVanlier: I think the example syntax is a little different with the `::`, but good idea, I will add it.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Oi. I thought I had gotten all of them. Thanks for catching that.

PR #510, opened 2023-05-11
diffusion: include gaussian refinement in warning

JoepVanlier:
**Why this PR?**
If we mention that centroid refinement is an option, we should also mention gaussian refinement. This omission was noticed by Aafke.


JoepVanlier:
I merged this into the release PR.

PR #511, opened 2023-05-11
fitting: fix incorrect model being used for `dsdna`

JoepVanlier:
**Why this PR?**
One of the API calls returns an incorrect model. This bug was introduced in Pylake v0.13.2 ( https://github.com/lumicks/pylake/pull/381 ).


tobiasjj (approved):
My bad! Good you fixed it. LGTM 🚴

PR #513, opened 2023-05-15
sample_from_image: ensure backward compatibility

JoepVanlier:
**Why this PR?**
We need to make sure we don't include any breaking bug fixes in a minor release.


lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -47,6 +49,15 @@ def __init__(
...
             and values for the tracking algorithm parameters
+        correct_origin : bool, optional
+            Use the correct pixel origin when summing track intensities when saving the file.
+            When saving the tracks, the widget stores the sum of the intensities around the tracks.
+            This sampling routine had a bug in previous versions of Pylake. Kymotracks are defined
+            with the origin of each image pixel defined at the center. Earlier versions of the
+            routine that samples photon counts around the track had a bug which assumed the origin
+            at the edge of the pixel. Setting this flag to `True` produces the correct behavior.
+            The default is set to `None` which results in a warning, while `False` reproduces the
+            old behavior.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

JoepVanlier:
> While it still seems weird to me to keep buggy behavior as default because of semver, I think this is handled relatively gracefully. The docstrings and warnings are nice and descriptive, just had some suggestions for further clarity.

Thank you!

> I wonder, since they're large chunks of texts and duplicated often should we define these as a variable and use `fstring` to insert where needed? Does that work for docstrings? I guess probably one definition in `kymotrack.py` and one in `kymotracker_widgets.py` since they're slightly different use cases.

Unfortunately, [f-strings cannot be used as docstrings](https://docs.python.org/3/reference/lexical_analysis.html#formatted-string-literals) unless you programmatically define them by assigning to `__doc__`, but that seems a bit smelly for this purpose.

> One thing that definitely needs to be added before merging though is updating the tutorial page

Good point! I've added some description of this to the tutorials.

docs/tutorial/kymotracking.rst
```diff
@@ -316,14 +316,20 @@ Extracting summed intensities
...
 
 Sometimes, it can be desirable to extract pixel intensities in a region around our kymograph track. We can quite easily
-extract these using the method :func:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image`. For instance,
-if we want to sum the pixels in a 11 pixel area around the longest kymograph track, we can invoke::
+extract these using the method :func:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image`.
+
+Prior to version 1.1.0 the method :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image` had a bug that assumed the origin of a pixel to be at the edge rather than the center of the pixel.
+Consequently, the sampled window could frequently be off by one pixel.
+The old (incorrect) behavior is maintained until the next major release (2.0) to ensure backward compatibility.
+It is recommended to include the argument `correct_origin=True` which results in using the correct origin.
```
rpauszek, line 324:
```suggestion
.. warning::
    Prior to version `1.1.0` the method :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image` had a bug that assumed the 
    origin of a pixel to be at the edge rather than the center of the pixel.
    Consequently, the sampled window could frequently be off by one pixel.
    The old (incorrect) behavior is maintained until the next major release (version `2.0`) to ensure backward 
    compatibility. It is recommended to include the argument `correct_origin=True` which results in using the correct origin.
```

docs/tutorial/nbwidgets.rst
```diff
@@ -266,10 +266,13 @@ You can easily pan horizontally by clicking and dragging left or right.
 
 You can optionally also pass algorithm parameters when opening the widget::
 
-    lk.KymoWidgetGreedy(kymo, "green", axis_aspect_ratio=2, min_length=4, pixel_threshold=9, window=6, sigma=0.2)
+    lk.KymoWidgetGreedy(kymo, "green", axis_aspect_ratio=2, min_length=4, pixel_threshold=9, window=6, sigma=0.2, correct_origin=True)
 
 .. image:: figures/nbwidgets/kymotracker2.png
 
+.. note::
+    When saving tracks to a file, the widget samples a region around the tracked lines to obtained a summed intensity. Prior to version 1.1.0 the method :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image` had a bug that assumed the origin of a pixel to be at the edge rather than the center of the pixel. Consequently, the sampled window could frequently be off by one pixel. To get the correct behavior and silence this warning, specify `correct_origin=True` when opening the kymotracking widget. The old (incorrect) behavior is maintained until the next major release to ensure backward compatibility. To silence this warning use `correct_origin=False`.
```
rpauszek, line 274: maybe warning? for extra emphasis?

```suggestion
.. warning::
    When saving tracks to a file, the widget samples a region around the tracked lines to obtained a summed intensity. Prior to version `1.1.0` the method :meth:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image` had a bug that assumed the origin of a pixel to be at the edge rather than the center of the pixel. Consequently, the sampled window could frequently be off by one pixel. To get the correct behavior and silence this warning, specify `correct_origin=True` when opening the kymotracking widget. The old (incorrect) behavior is maintained until the next major release to ensure backward compatibility. To silence this warning use `correct_origin=False`.
```

rpauszek (approved):
looks good, 2 little suggestions for wording in the docs

PR #514, opened 2023-05-15
`KymoTrackGroup._kymo` only valid for single source kymo

rpauszek:
**Why this PR?**
`KymoTrackGroup._kymo` is still used in many places where we require a single source kymograph. Here, added an extra check that raises if `KymoTrackGroup._kymos` has more than a single item in order to avoid silent bugs where the criterion isn't met.

_**Update**_
Something weird happened with pushing from a new branch. Closing this and opening new PR

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -845,10 +845,12 @@ class KymoTrackGroup:
 
     @property
     def _kymo(self):
-        try:
-            return self[0]._kymo
-        except IndexError:
+        if len(self._kymos) == 1:
+            return self._kymos[0]
+        elif len(self._kymos) == 0:
             raise RuntimeError("No kymo associated with this empty group (no tracks available)")
+        else:
+            raise RuntimeError("Group contains tracks from multiple source kymographs.")
 
     @property
     def _channel(self):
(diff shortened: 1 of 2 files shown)


JoepVanlier (changes requested):
I don't think we should enforce this at this level. I think by how we set this up, validity has to be checked upstream from this function.

This breaks multi-kymo `ensemble_ols` (which I now notice isn't tested), where we have multiple kymos (and it is fine to use this property), but we only check that select properties are the same.

JoepVanlier:
> I see your point, but I think it's actually 2 different scenarios, and this one is more descriptive for certain things. My thinking was that a valid return from `._kymo` is used iff there's really 1 source kymo and that's required (flipping). Other instances that can have more than 1 kymo but require the same acquisition settings can read from the 1st instance in `._kymos`.
>
> Although conversely, we already use `_validate_single_source` when it's really required so 😖 I guess the risk is that in the future we could forget to use this validation and just rely on `._kymo` by mistake which could lead to a silent bug. I do agree I'd rather not have `._kymos[0]` all over the place.
> 
> What do you think?

Maybe we should drop the property all-together and make it explicit that you always have to think of the fact that there can be more than one or there could be zero. I guess the `[0]` is a bit ugly, but it is explicit, and it will lead to a comment nearby every time we do it to justify it (which in this case is a good thing).

It doesn't require that many changes to get this property removal to work.

It would also be nice to add a test for the multi-kymo `ensemble_ols` considering that we just found that it is untested.

PR #517, opened 2023-05-17
changelog: order changelog by priority, fix links

JoepVanlier:
*Why this PR?*
Would be nice to release Pylake v1.1

Changelog here: https://lumicks-pylake.readthedocs.io/en/prep-1.1/changelog.html . Note that the links won't work until the tag is out. If you want to see what they point to, change `v1.1` in the url to `prep-1.1`.


rpauszek (approved):
🎉 🚢

PR #520, opened 2023-05-31
Pytest: fix check whether a test is marked to be skipped

tobiasjj:
# Why?
The code that decides whether to skip a test checked whether one of the "preflight" or "slow" options was present in `Item.keywords`. This could cause tests to be skipped incorrectly if one of these options was passed as an argument to a parameterised test. This commit fixes the problem by directly checking for the existence of a marker with a name equal to one of the options using `Item.iter_markers()`.


JoepVanlier (approved):
Good point! This could've led to some confusion if that ever happened indeed.

PR #522, opened 2023-06-06
deps: drop `opencv` in favor of `skimage` for affine transformations

JoepVanlier:
**Why this PR?**
We only use `opencv` for affine transformations. Considering that several other packages provide this functionality, we can also use one of the ones we already include. I did notice that there are small differences in the affine transforms; but considering we're not doing quantitative analysis with them yet, it should be fine. I suspect that this has to do with some speed optimizations OpenCV does (I note that the API docs have a difference between `INTER_LINEAR` and `INTER_LINEAR_EXACT` whereas the latter is not available for `warp_affine`).

We could also just use raw `scipy` for this (as `scikit-image` just calls theirs), but personally, I didn't like that one as much because of the double transpose needed on the `scipy` API.

I have also reduced the scope of a few imports that we rarely use.

opencv
OpenCV 4.7.0

scikit-image
scikit-image 0.20.0

scipy
scipy 1.10.1

### DIFF
--- a/lumicks/pylake/detail/widefield.py
+++ b/lumicks/pylake/detail/widefield.py
@@ -1,7 +1,6 @@
 import numpy as np
 import re
 import json
-import cv2
 import tifffile
 import warnings
 import enum
@@ -533,11 +532,17 @@ class TransformMatrix:
         Parameters
         ----------
         theta: float
-            angle of rotation
+            angle of rotation in degrees (counter-clockwise)
         center: np.ndarray
             (x, y) point, center of rotation
         """
-        return cls(cv2.getRotationMatrix2D(tuple(center), theta, scale=1.0))
+        from skimage.transform import EuclideanTransform
+
+        tf_shift = EuclideanTransform(translation=tuple(-c for c in center))
+        tf_rotate = EuclideanTransform(rotation=-np.deg2rad(theta))
+        tf_shift_back = EuclideanTransform(translation=center)
+        rotation = tf_shift + tf_rotate + tf_shift_back
+        return cls(rotation.params[:-1, :])
 
     @classmethod
     def translation(cls, x, y):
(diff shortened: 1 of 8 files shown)


lumicks/pylake/detail/widefield.py
```diff
@@ -533,11 +532,17 @@ def rotation(cls, theta, center):
         Parameters
         ----------
         theta: float
-            angle of rotation
+            angle of rotation in degrees (counter-clockwise)
         center: np.ndarray
             (x, y) point, center of rotation
         """
-        return cls(cv2.getRotationMatrix2D(tuple(center), theta, scale=1.0))
+        from skimage.transform import SimilarityTransform
```
tobiasjj, line 539: Out of interest: Is there a specific reason to no use [`EuclidianTransform`](https://scikit-image.org/docs/stable/api/skimage.transform.html#skimage.transform.EuclideanTransform)?

lumicks/pylake/detail/widefield.py
```diff
@@ -562,24 +567,20 @@ def warp_image(self, data):
...
-        return np.stack(image, axis=2).squeeze()
+        return transform.warp(
+            data,
+            self.invert().matrix,
+            order=1,  # Linear interpolation
+            mode="constant",
+            cval=0.0,  # borderValue
+            clip=True,
+            preserve_range=True,
+        )
```
tobiasjj, line 583: 👍

tobiasjj (approved):
Great! I like the descriptive API of scikit and the more accurate results of the rotation. LGTM 🚴

lumicks/pylake/detail/widefield.py
JoepVanlier: Good point. We're not using scaling, so let's switch it over to `Euclidian`.

PR #523, opened 2023-06-07
widgets: Fix parameter description for minimum length

mikhas:
Reducing the min length parameter in the kymotracker widget does not, in fact, reduce noise but rather increase it. Fixed the extended description to state that _increasing_ the min length parameter does have a de-noising effect.


JoepVanlier (changes requested):
Looks good, but since this is a bugfix, it should target the release branch (`release/1.1`), no?

mikhas:
> Looks good, but since this is a bugfix, it should target the release branch (`release/1.1`), no?

Done!

JoepVanlier (approved):
LGTM, thanks!

PR #525, opened 2023-06-09
deps: import rarely used dependencies on demand

JoepVanlier:
**Why this PR?**
There are some dependencies we use relatively rarely. This PR makes sure we only import them as needed by importing them where they are used.

Note that I also added a changelog entry for a previous bugfix that didn't get one.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,12 @@
 # Changelog
 
+## v1.1.1 | t.b.d.
+
+#### Other changes
+
+* Fixed parameter description for minimum length. It now correctly reads that increasing this parameter reduces tracking noise.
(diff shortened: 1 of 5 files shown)


dean0x7d (commented):
Nice! Can you also move the `cv2` import? It still exists on `release/1.1`.

JoepVanlier:
> Nice! Can you also move the `cv2` import? It still exists on `release/1.1`.

ah right, of course!

changelog.md
```diff
@@ -1,5 +1,12 @@
 # Changelog
 
+## v1.1.1 | t.b.d.
+
+#### Other changes
+
+* Fixed parameter description for minimum length. It now correctly reads that increasing this parameter reduces tracking noise.
+* Don't import packages we rarely use immediately. Rather specifically import functionality from `scikit-image` and `scikit-learn` when we actually use it.
```
dean0x7d, line 8:
```suggestion
* Rarely used but heavy packages like `opencv`, `scikit-image`, and `scikit-learn` are no longer loaded eagerly with `import lumicks.pylake`. Rather, they are loaded on-demand the first time that a feature needs them. This makes importing `pylake` itself faster and mitigates potential third-party issues (e.g. this makes the first issue from the [F.A.Q.](https://lumicks-pylake.readthedocs.io/en/v1.1.0/install.html#frequently-asked-questions) less severe as it no longer affects all users).
```

dean0x7d (approved):
Wonderful! Thanks!

PR #531, opened 2023-06-26
docs: add logical mask example

JoepVanlier:
**Why this PR?**
This has come up in support twice now. Might as well add a sentence or two to show people how to combine different criteria when boolean array indexing the data. Also noticed some deprecated import location in `scipy` that I fixed up.

Docs build here: https://lumicks-pylake.readthedocs.io/en/add_logic_example/tutorial/fdfitting.html#adding-data-to-the-fit

### DIFF
--- a/docs/tutorial/fdfitting.rst
+++ b/docs/tutorial/fdfitting.rst
@@ -174,6 +174,14 @@ when loading the data from `DNA/Lc` to `DNA/Lc_RecA`::
     distance2 = fd2.d[mask2].data
     fit.add_data("RecA", force2, distance2, params={"DNA/Lc": "DNA/Lc_RecA"})
 
+Multiple criteria can be combined into a single mask by using `numpy`'s logical operators.
+For example, to select all forces up to `30` pN and ensure a distance larger than `1.5` micron::
+
+    example_mask = np.logical_and(fd1.f.data <= 30, fd1.d.data > 1.5)
+
+Adding a third condition that the distance has to be `< 4` micron can be done as follows (note the double brackets)::
+
+    example_mask = np.logical_and.reduce((fd1.f.data <= 30, fd1.d.data > 1.5, fd1.d.data < 4.0))
 
 Setting parameter bounds
 ************************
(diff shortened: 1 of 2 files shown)


aafkevandenberg (approved):
Nice, very useful addition! I think just giving the succinct notation when compairing 3 conditions would be sufficient. That form is also easier to extent to even more conditions.

docs/tutorial/fdfitting.rst
```diff
@@ -174,6 +174,14 @@ when loading the data from `DNA/Lc` to `DNA/Lc_RecA`::
...
     fit.add_data("RecA", force2, distance2, params={"DNA/Lc": "DNA/Lc_RecA"})
 
+Multiple criteria can be combined into a single mask by using `numpy`'s logical operators.
+For example, if we want to select all forces up to `30` pN, and only include distances between `1.5` and `4.0` micron, we could create the mask by invoking::
+
+    example_mask = np.logical_and(np.logical_and(fd1.f.data <= 30, fd1.d.data > 1.5), fd1.d.data < 4.0)
+
+Or more succinctly (note the double brackets)::
+
+    example_mask = np.logical_and.reduce((fd1.f.data <= 30, fd1.d.data > 1.5, fd1.d.data < 4.0))
```
aafkevandenberg, line 184: Multiple criteria can be combined into a single mask by using `numpy`'s logical operators.
For example, if we want to select all forces up to `30` pN and distance data larger than '1.5' micron: 

    example_mask = np.logical_and(d1.f.data <= 30, fd1.d.data > 1.5)

Adding a third condition that the distance has to be < 4 micron (note the double brackets)::

    example_mask = np.logical_and.reduce((fd1.f.data <= 30, fd1.d.data > 1.5, fd1.d.data < 4.0))

JoepVanlier: Good point! Amending!

lumicks/pylake/force_calibration/detail/driving_input.py
```diff
@@ -1,5 +1,5 @@
 import numpy as np
-import scipy.signal
+import scipy.signal.windows
```
tobiasjj, line 2: 👍

tobiasjj (approved):
LGTM! 🚴

PR #532, opened 2023-06-30
Generate colormaps from wavelengths

rpauszek:
**Why this PR?**
The original idea came up during a feedback session with a customer who pointed out that "blue channel" data is plotted as blue (corresponding to the excitation color), however the actual emission color is green (red-shifted). Taking it a bit further, I found a paper with equations to approximate RGB values from wavelengths in the visible spectrum

*note: I started up the what's new section for 1.2.0 already here. I think it's easier to build this up as we go along rather than right before release; the intro text reads "sneak peak for 1.2.0" so we'd just need to change this to "has been released" in the final PR*

[docs build here](https://lumicks-pylake.readthedocs.io/en/wavelength_cmap/)

### DIFF
--- a/lumicks/pylake/adjustments.py
+++ b/lumicks/pylake/adjustments.py
@@ -1,6 +1,7 @@
 import numpy as np
 import matplotlib as mpl
 from matplotlib.colors import LinearSegmentedColormap
+from skimage.color import xyz2rgb
 from dataclasses import make_dataclass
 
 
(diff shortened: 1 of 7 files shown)


lumicks/pylake/adjustments.py
```diff
@@ -125,6 +125,54 @@ def nothing(cls):
 no_adjustment = ColorAdjustment.nothing()
 
 
+def wavelength_to_cie(wavelength):
```
JoepVanlier, line 128: `wavelength_to_cie_xyz` or `wavelength_to_xyz`?

lumicks/pylake/adjustments.py
```diff
@@ -157,13 +205,33 @@ class _ColorMaps(
...
+
+        Parameters
+        ----------
+        wavelength: int
+            wavelength to approximate maximum color from
+
+        References
+        ----------
+        .. [1] Chris Wyman, Peter-Pike Sloan, Peter Shirley. "Simple Analytic Approximations to the
+               CIE XYZ Color Matching Functions" Journal of Computer Graphics Techniques (2013) 2, 1-
```
JoepVanlier, line 224: Did this get truncated? `1-`?

lumicks/pylake/tests/test_image.py
```diff
@@ -194,3 +194,38 @@ def test_no_adjust():
...
+        (488, [[0, 0, 0, 1],[0, 0.31309926, 0.37314145, 1],[0, 0.62375244, 0.74336774, 1]]),
+        (590, [[0, 0, 0, 1], [0.50196078, 0.34896242, 0, 1], [1, 0.69519856, 0, 1]]),
+        (650, [[0, 0, 0, 1], [0.44213329, 0, 0, 1], [0.8808124,  0, 0, 1]]),
+        (850, [[0, 0, 0, 1], [0, 3.34046174e-10, 0, 1], [0, 6.65482613e-10, 0, 1]]),
+        # fmt: on
+    ],
+)
+def test_wavelength_to_cmap(wavelength, ref):
+    cmap = colormaps.from_wavelength(wavelength)
+    print(cmap([0, 0.5, 1]))
```
JoepVanlier, line 230: Leftover print statement?

JoepVanlier (changes requested):
Nice find. Only thing I'm not sure about is whether we need to include the `xyz_to_srgb` in our codebase considering that it is functionality already covered by `scikit-image`.

lumicks/pylake/adjustments.py
```diff
@@ -125,6 +125,54 @@ def nothing(cls):
...
+
+    xyz = [
+        calculate_component(**conversion_coefficients[key])
+        for key in conversion_coefficients.keys()
+    ]
+    return np.hstack(xyz)
+
+
+def cie_to_rgb(xyz):
+    """Convert XYZ CIE color components to sRGB with gamma correction."""
```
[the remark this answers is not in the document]
rpauszek: good catch, hadn't even thought to look if this conversion was available

lumicks/pylake/adjustments.py
rpauszek: oops

lumicks/pylake/adjustments.py
```diff
@@ -173,5 +229,10 @@ def __str__(self):
     def rgb(self):
         return None
 
+    def from_wavelength(self, wavelength):
+        xyz = wavelength_to_xyz(wavelength)
+        rgb = xyz2rgb(xyz.reshape([1, 1, 3])).squeeze()
```
JoepVanlier, line 234: Should still convert to linear rgb :)

JoepVanlier (approved):
:shipit:

PR #534, opened 2023-07-10
ci: close figures before issuing another plot

JoepVanlier:
**Why this PR?**
Just noticed that CI wasn't passing on another PR. `mpl` started warning us about layout changes when `tight_layout()` gets called twice (which is good). Probably better to just close the figure after we're done with it in the test, so that state doesn't propagate to the next plot in the test.

### DIFF
--- a/lumicks/pylake/piezo_tracking/tests/test_piezo_tracking.py
+++ b/lumicks/pylake/piezo_tracking/tests/test_piezo_tracking.py
@@ -1,5 +1,6 @@
 import pytest
 import numpy as np
+import matplotlib.pyplot as plt
 from lumicks.pylake.channel import Slice, Continuous, TimeSeries
 from lumicks.pylake.piezo_tracking.piezo_tracking import (
     DistanceCalibration,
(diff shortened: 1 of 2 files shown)


mikhas (commented):
Feels a bit random to see `plt.close(...)` to clean up all the resources acquired by `plt` but no corresponding function to acquire the resources in the first place. Are these convenience functions such as `distance_calibration.plot()` at risk of leaking resources then?

JoepVanlier:
> Feels a bit random to see `plt.close(...)` to clean up all the resources acquired by `plt` but no corresponding function to acquire the resources in the first place. Are these convenience functions such as `distance_calibration.plot()` at risk of leaking resources then?

[We close all plots at the end of every test](https://github.com/lumicks/pylake/blob/daaa1cf691c9becb3be08b1d1434f3484b9c0948/lumicks/pylake/conftest.py#L89) but not within a test. I could have also split the test in two, but I didn't want to litter the entire test file with tiny tests.

How to deal with open plots is up to the user. This is also what they'd expect.

PR #536, opened 2023-07-14
file: add ability to crop to time interval when using `save_as`

JoepVanlier:
**Why this PR?**
Being able to export to a smaller time interval helps to cut down on file size. It also came up through support.

While implementing this, I noticed a few missing APIs that made this cumbersome, and decided to add those. These were:
- `PointScan` had no slicing
- `Scan` didn't handle empty scans gracefully.
- It was not possible to access confocal objects by path.

I mitigated these three issues each in their own commit before implementing the saving.


changelog.md
```diff
@@ -8,13 +8,17 @@
...
 * Added `PowerSpectrum.identify_peaks()` method to the [`PowerSpectrum`](https://lumicks-pylake.readthedocs.io/en/v1.0.0/_api/lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum.html) class. This method uses probability to identify peaks in the spectrum that are not due to the movement of beads in an optical trap. This is beta functionality. While usable, this has not yet been tested in a large number of different scenarios. The API can still be subject to change without any prior deprecation notice!
+* Added support for accessing `Kymo`, `Scan` and `PointScan` by path (e.g. `file["Kymograph"]["my_kymo"]` or `file["Kymograph/my_kymo"]`).
+* Added support for slicing `PointScan`.
+* Added the ability to specify a cropping region when exporting to an h5-file using `file.save_as(filename, crop_time_range=(starting_timestamp, ending_timestamp))`. Note that this interval has to be specified as a tuple of timestamps (in nanoseconds).
```
rpauszek, line 13: the note part seems a bit extraneous here

lumicks/pylake/channel.py
```diff
@@ -575,6 +575,25 @@ def from_dataset(dset, y_label="y", calibration=None):
             calibration=calibration,
         )
 
+    def to_dataset(self, f, name, **kwargs):
+        """Save this to an h5 dataset
+
+        Parameters
+        ----------
+        f : h5py.Group or
```
rpauszek, line 583: `or h5py.File` ?

also `f` seems not so great. maybe `parent`?

lumicks/pylake/detail/h5_helper.py
```diff
@@ -2,42 +2,97 @@
...
+        out_file.create_dataset(name, data=node)
+        out_file[name].attrs.update(node.attrs)
+
+    if not crop_time_range:
+        write_node()
+    else:
+        # Override time ranges. Items know how to crop themselves.
+        try:
+            start, stop = (
+                getattr(lk_file[name][slice(crop_time_range[0], crop_time_range[1])], field)
```
rpauszek, line 18:
```suggestion
                getattr(lk_file[name][slice(*crop_time_range)], field)
```

lumicks/pylake/detail/h5_helper.py
```diff
@@ -2,42 +2,97 @@
...
+    crop_time_range=None,
+    verbose=False,
+):
     """Write a modified h5 file to disk.
 
     Parameters
     ----------
-    h5_file : h5py.File
-        loaded h5 file
+    lk_file : lk.File
```
rpauszek, line 46: isn't renaming positional arguments breaking?

lumicks/pylake/detail/h5_helper.py
```diff
@@ -2,42 +2,97 @@
...
         Output file name.
     compression_level : int
         Compression level for gzip compression.
     omit_data : str or iterable of str, optional
         Which data sets to omit. Should be a set of h5 paths.
+    crop_time_range : tuple of np.int64
+        Specify a time interval to crop to (tuple of a start and stop time). Interval must be
+        specified in nanoseconds since epoch (the same format as timestamps).
+    verbose : bool, optional.
+        Produce verbose output. Default: False.
```
rpauszek, line 58:
```suggestion
        Print verbose output. Default: False.
```

lumicks/pylake/detail/h5_helper.py
```diff
@@ -2,42 +2,97 @@
...
-                    # Non-numerical data doesn't support compression
-                    out_file.create_dataset(name, data=node)
+                    write_cropped_metadata(lk_file, out_file, name, node, crop_time_range, verbose)
                 else:
                     # Numerical data can benefit a lot from compression
-                    out_file.create_dataset(
-                        name, data=node, compression="gzip", compression_opts=compression_level
-                    )
+                    if crop_time_range:
+                        sliced = lk_file[name][slice(crop_time_range[0], crop_time_range[1])]
```
rpauszek, line 77:
```suggestion
                        sliced = lk_file[name][slice(*crop_time_range)]
```

lumicks/pylake/detail/h5_helper.py
```diff
@@ -2,42 +2,97 @@
...
+                    else:
+                        out_file.create_dataset(
+                            name, data=node, compression="gzip", compression_opts=compression_level
+                        )
+                        out_file[name].attrs.update(node.attrs)
             else:
                 out_file.create_group(f"{name}")
-
-            out_file[name].attrs.update(node.attrs)
+                out_file[name].attrs.update(node.attrs)
```
rpauszek, line 95: why the change in scope here? is this handled within the nasty `if else if else else iff andelse` block?

lumicks/pylake/group.py
```diff
@@ -12,25 +12,32 @@ class Group:
         The underlying h5py group object
     """
 
-    def __init__(self, h5py_group):
+    def __init__(self, h5py_group, *, lk_file):
```
rpauszek, line 15: could you add some docstring for `lk_file` please, it took me a while to figure out what was going on

lumicks/pylake/file.py
```diff
@@ -53,24 +52,24 @@ def _check_file_format(self):
...
-                ("Calibration", "force1x.calibration"),
-                ("Marker", "markers"),
-                ("FD Curve", "fdcurves"),
-                ("Kymograph", "kymos"),
-                ("Scan", "scans"),
-                ("Note", "notes"),
-            ]
-        )
+        # List of fields which should not be accessed directly from the h5 and their actual API
+        # accessors. Note that top level fields are automatically printed when print is invoked.
```
rpauszek, line 56: maybe add to this the new behavior (only warns if the output class isn't defined) also maybe add a note about the output class (2nd item in tuple) I didn't really know immediately what it was there for

lumicks/pylake/scan.py
```diff
@@ -154,10 +154,10 @@ def __getitem__(self, item):
             scan[1, 3, 5]   # Error, integer indices are not allowed for the spatial dimensions.
             scan[1, 3:5:2]  # Produces an error, steps are not allowed when slicing.
 
-            stack["1s":"5s"]  # Gets the scan frames from first to the fifth second
-            stack[:"-5s"]  # Gets the scan frames up to the last 5 seconds
+            scan["1s":"5s"]  # Gets the scan frames from first to the fifth second
+            scan[:"-5s"]  # Gets the scan frames up to the last 5 seconds
```
rpauszek, line 158: 👍

rpauszek (changes requested):
looks pretty good and this will be really nice to have. just a couple things (mostly documentation) for adding clarity, especially for things we rarely look at (`group.py`)

lumicks/pylake/group.py
JoepVanlier: Good point! I will add some and the reason it needs to be provided.

changelog.md
JoepVanlier: Removed!

lumicks/pylake/channel.py
JoepVanlier: Yeah, fair enough! Fixed.

lumicks/pylake/channel.py
```diff
@@ -686,6 +705,26 @@ def __len__(self):
...
+            location to save to.
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/detail/h5_helper.py
JoepVanlier: Good catch!

lumicks/pylake/detail/h5_helper.py
JoepVanlier: It is, but this is private API.

lumicks/pylake/detail/h5_helper.py
JoepVanlier: 👍

lumicks/pylake/detail/h5_helper.py
JoepVanlier: 👍

lumicks/pylake/detail/h5_helper.py
JoepVanlier: The change in scope is needed because we want to override these properties in the `Scan/Kymo` case unfortunately (see `write_cropped_metadata` where we first call the `update` and then write our custom ones).

And I'm not a huge fan of the `if/then/else` bonanza either, but it seemed better than the alternatives. Maybe it would become slightly more readable if I drop the numerical branch into a separate function. I've split it into its own function in a separate (optional) last commit. What do you think? Better or worse?

rpauszek: yeah in my original comment I wasn't necessarily suggesting to fixup the giant block; I know sometimes it's necessary. But I do like it as a separate function now that I see it

lumicks/pylake/file.py
```diff
@@ -53,24 +52,26 @@ def _check_file_format(self):
...
-                ("FD Curve", "fdcurves"),
-                ("Kymograph", "kymos"),
-                ("Scan", "scans"),
-                ("Note", "notes"),
-            ]
-        )
+        # List of h5 fields for which custom API exists. Note that top level fields (provided as
+        # the first argument in the tuple) are automatically printed when print is invoked. For
+        # fields where an actual class exists, the second element in the tuple is used to
+        # instantiate an element in that Group (using its `from_dataset` method).
```
rpauszek, line 58: just a suggestion if you think it's warranted 
```suggestion
        # List of h5 fields for which custom API exists. Note that top level fields (provided as
        # the first argument in the tuple) are automatically printed when print is invoked. For
        # fields where an actual class exists, the second element in the tuple is used to
        # instantiate an element in that Group (using its `from_dataset` method). Otherwise
        # a FutureWarning is raised.
```

rpauszek (approved):
Looks great! If you feel like it you could add this already to the docs What's New section, but we'll do a review of that before releasing anyway, so up to you

lumicks/pylake/file.py
JoepVanlier: It's good to be explicit about it indeed. I've added it.

PR #537, opened 2023-07-14
msvst: stabilized wavelets

JoepVanlier:
This is an implementation of the Multi-Scale Variance Stabilizing Transform detailed in the following papers:
- Zhang, B., Fadili, M. J., Starck, J. L., & Olivo-Marin, J. C. (2007, September). Multiscale variance-stabilizing transform for mixed-Poisson-Gaussian processes and its applications in bioimaging. In 2007 IEEE International Conference on Image Processing (Vol. 6, pp. VI-233). IEEE.
- Zhang, B., Fadili, J. M., & Starck, J. L. (2008). Wavelets, ridgelets, and curvelets for Poisson noise removal. IEEE Transactions on image processing, 17(7), 1093-1108.

Which was the one of the best performers in:
Smal, I., Loog, M., Niessen, W., & Meijering, E. (2009). Quantitative comparison of spot detection methods in fluorescence microscopy. IEEE transactions on medical imaging, 29(2), 282-301.

It is intended to become one of the image preprocessing options prior to tracking. Note that it would _not_ be used for localization estimation (as it would generate bias), merely for peak detection.

Some 2D examples:
image
Ground truth

image
Noisy (poisson noise with very high background).

image
Filtered.

Note that your mileage on Kymo's may vary as this gets rid of a constant background (not actual fluorescence detections).

image
Effect on real kymograph tracks.

Left as to do in a future PR is to use a decomposed form for the 2D case and provide a public API for this.


lumicks/pylake/kymotracker/detail/denoising.py
```diff
@@ -0,0 +1,412 @@
...
+
+
+def calculate_vst_coefficients(kernel):
+    """Calculate optimal transform coefficients and expected standard deviation.
+
+    We calculate b and c for the following transform:
+
+        Z = b sign(Y + c) * sqrt(Y + c)
+
+    Using the coefficients from Eq. 10 and b above Eq. 12 we estimate a transform such that
```
rpauszek, line 86: it would be useful here to state which paper the specific eq numbers are from. I don't think you need to repeat the full citation again, maybe just `(Zhang 2008)` would be good enough? still would be really nice if we could have a global reference list for the module 😞 

on closer looks, also seems like the eq numbers might be wrong? I think you're using slightly different terminology for the variables or am I missing something?

lumicks/pylake/kymotracker/detail/denoising.py
```diff
@@ -0,0 +1,412 @@
...
+    return raw_pvalues <= p_cutoff
+
+
+class MultiScaleVarianceStabilizingTransform:
+    """Calculates the required coefficients for determining the variance stabilizing transform (VST)
+
+    The VST is a generalization of the Anscombe transform which can convert Poisson distributed.
+    noise to approximately Gaussian distributed noise. For Poisson distributed data, the variance
+    scales with the mean. After the transform the distribution is roughly Gaussian with a variance
+    of 1. This transformation allows you to apply statistics that rely on normality post-transform.
```
rpauszek, line 186: 👍

lumicks/pylake/kymotracker/detail/denoising.py
```diff
@@ -0,0 +1,412 @@
...
+            Indicate that the detail coefficients were stabilized and this transform
+            must be inverted.
+        """
+        output_image = sum(detail_coefficients + [remainder])
+        if not stabilize:
+            return output_image
+
+        return inverse_variance_stabilizing_transform(output_image, self._coefficients[0])
+
+    def process_image(self, image, false_detection_rate=0.1, verbose=False):
```
rpauszek, line 271: `filter_image()`? In general I don't really like using "process", it's pretty vague sounding

lumicks/pylake/kymotracker/detail/denoising.py
```diff
@@ -0,0 +1,412 @@
...
+        output_image = self._reconstruct_image(detail_coefficients, remainder, stabilize=True)
+
+        # Enforce positivity. Error model states that negative values don't make sense. Yet they
+        # can still occur because of the overcompleteness of the IUWT and the fact that we
+        # modified values in the detail coefficients directly.
+        output_image[output_image < 0] = 0
+
+        return output_image, significant
+
+    def regularized_reconstruction(
```
rpauszek, line 312: nitpick, would be nice to have a verb here. `filter_regularized()`? `reconstruct_regularized()`?

rpauszek (approved):
so I'll be honest, most of the math in here is way over my head (at least without a week of reading the source material) but looks great from the standpoint of code quality. I especially appreciate the detailed docstrings. 

Let's go ahead and merge so we can start testing out how we might want to expose this as public functionality

lumicks/pylake/kymotracker/detail/denoising.py
```diff
@@ -0,0 +1,412 @@
...
+    current = np.asarray([[1.0]])
+    convolved_kernels = [current]
+    for k in kernels:
+        current = scipy.signal.convolve(current, k, mode="full")
+        convolved_kernels.append(current)
+
+    return convolved_kernels
+
+
+def generate_bspline_kernels(num_kernels):
```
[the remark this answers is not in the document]
JoepVanlier: It's the only way to use the wavelets right now. First you generate the kernels and then feed them to the MSVST class. Something like:

```python
kernels = generate_bspline_kernels(4)
msvst = MultiScaleVarianceStabilizingTransform(kernels, two_dimensional=True)

detail_simple, detail_layers = msvst.process_image(data, false_detection_rate=0.1, verbose=False)

detail = msvst.regularized_reconstruction(data, false_detection_rate=0.1, num_iter=100, remove_background=True, verbose=False)
```

PR #538, opened 2023-07-17
Update install.rst: update installation instructions for China in faq

aafkevandenberg:
The installation instructions for China in the faq were missing "jupyter_client<8"

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -473,6 +473,6 @@ Install pip in the activated environment by invoking::
 
 Then install Pylake as follows::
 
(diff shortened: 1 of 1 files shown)


JoepVanlier (commented):
Jupyter is normally only installed with `pip` if you install with `[notebook]`.

If we really always need to include this, I'm wondering whether it's better to just include this extra requirement in our `setup.py` under `extras_require`. That way we wouldn't need special installation instructions for `pip` at all.

aafkevandenberg:
Including this in `setup.py` would indeed be easier.

JoepVanlier:
Should be solved by https://github.com/lumicks/pylake/pull/546

PR #539, opened 2023-07-17
ci: add option to freeze reference data

JoepVanlier:
**Why this PR?**
For another PR I have in flight, I noticed I was about to hardcode a bunch of matrices as data. Instead, it seems better to store these as a file. This PR is supposed to make this a bit more convenient. It is intended as a convenience function for small amounts of data (if we go to bigger datasets we will likely need something else in addition to this).

The function can be used by importing the fixture `reference_data` and then polling the reference data like:

```python
np.testing.assert_allclose(test_data, reference_data("test_data_name", test_data))
```

If the test data does not exist yet, the test data is generated when this function is called. If it does exist, but is different then it raises an error. You can then choose to delete the reference file (which will result in a regeneration) or calling `pytest` with `--update_reference_data` (which will force an update).

On CI, the tests are run with `--strict_reference_data` which disallows regeneration (so that you don't accidentally forget to check in your reference data).

### DIFF
--- a/lumicks/pylake/conftest.py
+++ b/lumicks/pylake/conftest.py
@@ -1,6 +1,8 @@
 import importlib
 import warnings
 import pytest
+import os.path
+import numpy as np
 import matplotlib.pyplot as plt
 
 
@@ -9,6 +11,15 @@ def pytest_addoption(parser):
         parser.addoption(
             f"--run{option}", action="store_true", default=False, help=f"run {option} tests"
         )
+    parser.addoption(
+        "--update_reference_data", action="store_true", default=False, help="Update test data"
+    )
+    parser.addoption(
+        "--strict_reference_data",
+        action="store_true",
+        default=False,
+        help="Treat missing test data as error",
+    )
 
 
 def pytest_collection_modifyitems(config, items):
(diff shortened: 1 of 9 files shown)


dean0x7d (commented):
Nice! Great addition to the test tooling.

lumicks/pylake/conftest.py
```diff
@@ -93,3 +105,51 @@ def configure_mpl():
...
+    --------
+    ::
+
+        def test_freezing(reference_data):
+            test_data = np.array([[1, 2, 3], [1, 2, 3]])
+            np.testing.assert_allclose(test_data, reference_data("test_data", test_data))
+    """
+
+    def get_reference_data(test_name, reference_data):
+        calling_function = inspect.stack()[1][3].replace("test_", "")
```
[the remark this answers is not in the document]
JoepVanlier: Ah yes, that's a lot nicer than reading it from the stack! Thanks for the suggestion!

I like having the option to have the parametrization be pulled in automatically, but I would also like to keep the option around to force a specific filename entirely. In some cases, I would like to be able to keep what I was testing readable at the directory tree level (hence wanting to be able to force it).

I went with something somewhat similar to the replacement you did, with an optional `test_name` argument (which still suffixes the parametrization) and an optional `file_name` if you want to force the file name entirely.

lumicks/pylake/conftest.py
```diff
@@ -93,3 +104,99 @@ def configure_mpl():
...
+        output_fn = request.node.name.replace("test_", "") if file_name is None else file_name
+
+        # Overwrite only the identifier part Test[par1-par2-par3] -> mytest[par1-par2,par3].
+        if test_name is not None:
+            output_fn = output_fn.replace(calling_function, test_name)
+
+        calling_path_name = os.path.dirname(request.fspath)
+
+        reference_data_path = os.path.join(calling_path_name, "reference_data", calling_function)
+        reference_data_fn = os.path.join(reference_data_path, f"{output_fn}.npz")
```
dean0x7d, line 185: Nitpick: I was a bit confused since I'm used to `fn` being short for `function`, not `filename`. E.g. `fn` is a keyword meaning `function` in Rust, and a common API name in the C++ `std` library. Consider `ref_data_filename` instead.

dean0x7d (approved):
I think this will definitely make writing tests easier. Looks great from my side!

lumicks/pylake/conftest.py
JoepVanlier: Good to know. Will do! 👍

PR #545, opened 2023-08-07
Fun with css

rpauszek:
**Why this PR?**
Friday afternoon project because the code blocks within admonitions were bugging me (the gray just didn't look nice on the colored backgrounds) and made the links bold so they stand out more. Decided it would be nice to have the code crossrefs formatted a bit too. 

Before adding that, I took the opportunity to clean up the CSS. Split the single `custom.css` into 4 files separated by concern. ~~Also discovered nesting which make the code a lot nicer IMO~~ Unfortunately nesting isn't supported for Firefox so we're stuck with ugly code for a while.

Kept separate commits so you can see the process I went through to (hopefully) not break any of the formatting we already had, but will squash this down before merge (probably 2 commits, cleanup and additions)

Before
image
After
image

Before
image
After
image

[docs build here](https://lumicks-pylake.readthedocs.io/en/fun_with_css/index.html)


docs/_static/admonitions.css
```diff
@@ -0,0 +1,47 @@
...
+}
+
+.rst-content .important {
+    border-radius: 3px;
+    background-color: var(--lightorange);
+}
+
+.rst-content .important a {
+    color: var(--darkorange);
+}
```
JoepVanlier, line 74: eolrip

docs/_static/api.css
```diff
@@ -0,0 +1,25 @@
...
+    width: 100%;
+    border-radius: 3px;
+}
+
+.rst-content dl[class]:not(.option-list):not(.field-list):not(.footnote):not(.glossary):not(.simple) dl:not(.field-list)>dt {
+    border-left: solid 3px var(--lightgray) !important;
+    border-top: 0px !important;
+    background-color: var(--lightestgray) !important;
+    padding-right: 25px;
+}
```
JoepVanlier, line 26: same here

JoepVanlier (changes requested):
It's nice to separate it into different blocks and I really like the subtle color changes to the main docs (making the links more in line with the box), but it seems that there is some issue rendering warning boxes in the API docs?

image

docs/_static/admonitions.css
rpauszek: the formatter in vs code removes blank lines at the end of files. I left it because if I make changes in the future this will keep happening. can add them if you think it's better

JoepVanlier: Ah okay, then leave it as is. It's not like we look at these files every day anyhow.

PR #546, opened 2023-08-07
deps: enforce upper limit `jupyter_client` and `pyzmq`

JoepVanlier:
**Why this PR?**
To work around a bug in `notebook` we need to pin the upper limit of `jupyter_client` and `zeromq` (see https://github.com/jupyter/notebook/issues/6748).

Until now, we've been asking users to add these extra constraints manually, but forgetting it can lead to broken interactive notebook functionality. It's unnecessary friction when it comes to installing `pylake`. Instead of adding these lines to the pip [installation instructions](https://github.com/lumicks/pylake/pull/538), it's better to just add it to the requirements for the time being.

When we release on `conda`, we should also pin the upper limit in the recipe there.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -51,12 +51,7 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
 
     .. code-block::
 
-        conda install lumicks.pylake "jupyter_client<8"
-
-    .. note::
-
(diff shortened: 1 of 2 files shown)


agom (approved):
🧷

rpauszek (approved):
🚢

PR #547, opened 2023-08-07
Start using `isort` to complement `black` with import sorting/formatting

dean0x7d:
I noticed that `pylake` wasn't using this yet and I figured you might find it useful. It's a very nice thing to complement `black` formatting.


JoepVanlier (approved):
Makes sense to me.

PR #548, opened 2023-08-07
Mark file APIs with `os.PathLike` type hint

dean0x7d:
This is a very minor addition. My IDE showed a warning when I was passing a `pathlib.Path` to `pylake` since the type hint was only `str`. These days, most APIs accept anything that's `PathLike` instead of strict `str`.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -16,7 +16,7 @@ def _read_txt(file, delimiter):
 
     Parameters
     ----------
-    file : str or StringIO
+    file : str | os.PathLike | StringIO
         path to a file to read, or handle to read from directly
     delimiter : str
         Delimiter to use.
@@ -70,7 +70,7 @@ def export_kymotrackgroup_to_csv(
 
     Parameters
     ----------
-    filename : str
+    filename : str | os.PathLike
         Filename to output KymoTrackGroup to.
     kymotrack_group : KymoTrackGroup
         Kymograph tracks to export.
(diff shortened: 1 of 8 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -70,7 +70,7 @@ def export_kymotrackgroup_to_csv(
 
     Parameters
     ----------
-    filename : str
+    filename : str | os.PathLike
```
JoepVanlier, line 73: Elsewhere, we've always been writing `or` in full rather than `|`.

JoepVanlier (approved):
Huh, TIL. Good to know. There are a few more places where we declare a parameter type to be a `str`, where I think it could be a `os.PathLike`:

`detail/h5_helper.py` second parameter of `write_h5`.
`kymotracker/kymotrack.py`, first parameter in `_read_txt(file, delimiter).
`kymotracker/detail/sequence.py`: first parameter in `read_genbank`.

dean0x7d:
@JoepVanlier Thanks for pointing those out. I've added the type hints there as well!

lumicks/pylake/kymotracker/kymotrack.py
dean0x7d: Over the last few years, a lot of work has been put into standardizing the Python type hinting. Various docstring formats used to have their own conventions but it's nice to adopt the new standard way. There's a lot of good info and links here (note the 3 links near the top, including the cheatsheet): https://docs.python.org/3/library/typing.html

Specifically for this case, `Union[A, B, C]` was initially set up, but it has since been refined to the `A | B | C` syntax which is pretty nice. Similarly, `Optional[T]` became `T | None`.

PR #549, opened 2023-08-07
dwelltime: enforce minimum time

JoepVanlier:
**Why this PR?**
The minimum observation time for the discrete dwell time model is the discretization timestep. Since our `DwelltimeModel`, uses a default minimum time of zero, I see a potential risk for misuse here. I think it's better to just raise if we see this being provided.

I also did a fixup on the test here, since I noticed that some of the cases didn't abide this constraint (shows how easy it is to miss it). I also noticed that I accidentally disabled the profile test earlier (so I reverted that).

### DIFF
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -504,6 +504,10 @@ class DwelltimeModel:
     ValueError
         if `min_observation_time` or `max_observation_time` is not a scalar and is not the same
         length as `dwelltimes`.
+    ValueError
+        if `discretization_time` is provided and is bigger than the `min_observation_time`. A
+        minimum observation time shorter than the discretization step does not make sense under
+        this model.
     """
 
     def __init__(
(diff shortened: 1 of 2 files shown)


lumicks/pylake/population/dwelltime.py
```diff
@@ -542,6 +546,14 @@ def __init__(
                         f"the number of dwell times provided ({dwelltimes.size})."
                     )
 
+            # A minimum time shorter than the discretization time makes no sense (cannot be
+            # observed under this model)
+            rel_tolerance = 1e-6
+            if np.any(discretization_timestep > (1.0 + rel_tolerance) * min_observation_time):
+                raise ValueError(
+                    "The discretization timestep cannot be larger than the minimum observable time."
+                )
```
rpauszek, line 555: Could add some more info as we've been doing lately with error messages
```suggestion
            if np.any(discretization_timestep > (1.0 + rel_tolerance) * min_observation_time):
                raise ValueError(
                    f"The discretization timestep ({discretization_timestep:0.2e} sec) cannot be larger "
                    f"than the minimum observable time ({min_observation_time:0.2e} sec)."
                )
```
Might want to think about the units there, could be nicer to show in usec but then it doesn't really match what the user input. Not sure

rpauszek (approved):
good catch

lumicks/pylake/population/dwelltime.py
JoepVanlier: I was doing that originally, but one, both or none of these can actually be an array rather than a number. What should I show in that case, the first violation?

PR #550, opened 2023-08-08
Speed up `import lumicks.pylake` by streamlining `scipy` imports

dean0x7d:
This follows one simple rule: At module scope, always use `import scipy`, never `imports scipy.<foo>` nor `from scipy import <foo>`.

Why: `scipy.<foo>` works anyway following `import scipy` and it will lazily import heavy submodules. Using `imports scipy.<foo>` forces the submodules to be imported eagerly.

Results: On my machine, cold import (first import following installation) goes down from 11.4 seconds to 3.9 seconds. Hot import goes down from 0.7 to 0.4 seconds.

Measuring: I followed the import time profiling method [described here](https://stackoverflow.com/a/51300944) plus a `pipenv` for reproducible cold import measurements:
```
mkdir tmp && cd tmp
pipenv --rm && pipenv shell
pip install <path/to/pylake>
python -X importtime -c "import lumicks.pylake" 2> pylake-cold.log
python -X importtime -c "import lumicks.pylake" 2> pylake-hot.log
tuna pylake-cold.log
tuna pylake-hot.log
```

#### Cold import: BEFORE

Screenshot 2023-08-08 at 10 02 59

#### Cold import: AFTER

Screenshot 2023-08-08 at 10 03 14

#### Hot import: BEFORE

Screenshot 2023-08-08 at 10 03 56

#### Hot import: AFTER

Screenshot 2023-08-08 at 10 04 08

### DIFF
--- a/lumicks/pylake/population/tests/test_dwelltimes.py
+++ b/lumicks/pylake/population/tests/test_dwelltimes.py
@@ -607,7 +607,7 @@ def test_dwelltime_exponential_no_free_params(monkeypatch):
         raise StopIteration
 
     with monkeypatch.context() as m:
-        m.setattr("lumicks.pylake.population.dwelltime.minimize", stop)
+        m.setattr("scipy.optimize.minimize", stop)
 
         def quick_fit(fixed_params):
             return _exponential_mle_optimize(
--- a/lumicks/pylake/adjustments.py
+++ b/lumicks/pylake/adjustments.py
@@ -1,8 +1,8 @@
 from dataclasses import make_dataclass
 
 import numpy as np
+import skimage
 import matplotlib as mpl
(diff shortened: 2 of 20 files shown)


JoepVanlier (approved):
LGTM!

lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -607,7 +607,7 @@ def stop(*args, **kwargs):
         raise StopIteration
 
     with monkeypatch.context() as m:
-        m.setattr("lumicks.pylake.population.dwelltime.minimize", stop)
```
JoepVanlier, line 610: interesting 👀

PR #551, opened 2023-08-08
docs: add section on robust fitting in force calibration tutorial

rjmoerland:
**Why?**
As a follow-up on #528 I have added a section on robust fitting and automated peak identification to the force calibration tutorial. See https://lumicks-pylake.readthedocs.io/en/robert-robust-fitting-example/tutorial/force_calibration.html#robust-fitting

Note that I updated the images of the force calibrations to reflect the fact that there is a new line stating the loss function in the results. I'm not sure this warrants a new image, but it's there just in case.

TO DO:
- [x] Upload / hand over test data (14 MB, is that OK?)

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -150,7 +150,7 @@ To reproduce the result obtained with Bluelake, these should be excluded from th
 
 Note that exclusion ranges are excluded *prior* to downsampling.
 Considering that a noise peak may be very narrow, it is beneficial to lower the number of points per block temporarily to find the exact exclusion range.
-After determination of this exclusion range, the number of points per block can be increased again.
+After determination of this exclusion range, the number of points per block can be increased again. However, also see `Robust fitting`_ for an automated peak identification routine.
 
 Passive calibration
 -------------------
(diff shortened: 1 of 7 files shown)


docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+    plt.title(
+        f'Skewed fit: $f_c$ = {fit.results["fc"].value:.1f}, $D$ = {fit.results["D"].value:.4f}, $f_d$ = {fit.results["f_diode"].value:.1f}'
+    )
+    plt.show()
+
+.. image:: figures/force_calibration/power_spectrum_noise_peak.png
+
+Notice how the tail of the model is skewed towards the peak, in order to reduce the least-squares error. The free parameters to fit the diode filter contribution are 'abused' to reduce the error between the model and the outlier.
+
+Now do a robust fit. We do this by specifying a loss function in the function `fit_power_spectrum()`. For least-squares fitting, the loss function is `'gaussian'`, which is the default if nothing is specified. However, if we specify `'lorentzian'`, a robust fitting routine will be used instead. Because `bias_correction` and robust fitting are mutually exclusive, we need to explicitly turn it off::
```
JoepVanlier, line 499:
```suggestion
Now do a robust fit. We do this by specifying a loss function in the function :func:`~lumicks.pylake.fit_power_spectrum()`. For least-squares fitting, the loss function is `'gaussian'`, which is the default if nothing is specified. However, if we specify `'lorentzian'`, a robust fitting routine will be used instead. Because `bias_correction` and robust fitting are mutually exclusive, we need to explicitly turn it off::
```

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+
+Notice how the model now follows the power spectrum nearly perfectly. The value for `f_diode` has increased significantly, now that it is not abused to reduce the error induced by the outlier. 
+
+This example shows that a robust fitting method is less likely to choke on outliers in the power spectrum data. It is therefore a fair question why one would not use it all the time? First of all, robust fitting leads to a small bias in the fit results for which pylake has no correction. Least-squares fitting also leads to a bias, but this can be corrected with `bias_correction=True`. Secondly, for least-squares fitting, a large pool of literature is available to estimate the errors in the fit of the free parameters, which are implemented in the least-squares fitting routines that Pylake uses. These error estimates are not implemented for robust fitting, and as such, the fit results will show `nan` for the error estimates after a robust fit. However, as will be shown below, the robust fitting results may be used as a start to identify outliers automatically, in order to exclude these from a second, regular least-squares, fit.
```
JoepVanlier, line 516:
```suggestion
This example shows that a robust fitting method is less likely to fail on outliers in the power spectrum data. It is therefore a fair question why one would not use it all the time? Robust fitting leads to a small bias in the fit results for which Pylake has no correction. Least-squares fitting also leads to a bias, but this bias is known (:cite:`norrelykke2010power`) and can be corrected with `bias_correction=True`. Secondly, for least-squares fitting, methods exist to estimate the expected standard errors in the estimates of the free parameters, which are implemented in the least-squares fitting routines that Pylake uses :cite:`press1990numerical`. These error estimates are not implemented for robust fitting, and as such, the fit results will show `nan` for the error estimates after a robust fit. However, as will be shown below, the robust fitting results may be used as a start to identify outliers automatically, in order to exclude these from a second, regular least-squares, fit. 
```

I would actually love to implement profile likelihood for these calibrations as well, since the standard errors obtained from these asymptotic methods can be deceptive when dealing with a pathological case.

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+
+Automated spurious peak detection
+---------------------------------
+
+We will continue the tutorial with the results of the previous section. If you did not yet do that part of the tutorial, please go back and execute the code examples in that section. 
+
+We still have the power spectrum `ps` that was created without blocking or windowing. Here we will use it to identify the peak and automatically obtain frequency exclusion ranges. We will use the method `identify_peaks()` of the `PowerSpectrum` object in order to do so. This method takes a function that accurately models the power spectrum as a function of frequeny, in order to normalize it. It then identifies peaks based on the likelihood of encountering a peak of a certain magnitude in the resulting data set. If we have a "good fit", then the easiest way to get that function is to distill it from our fit parameters and the model that was used during fitting::
```
JoepVanlier, line 523: Doublecheck whether this crosslinking works (and I didn't misspell the class's location), but it'd be nice to just crosslink to the method `identify_peaks`, then they get a clickable link.
```suggestion
We still have the power spectrum `ps` that was created without blocking or windowing. Here we will use it to identify the peak and automatically obtain frequency exclusion ranges. We will use the method :meth:`~lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum.identify_peaks()` in order to do so. This method takes a function that accurately models the power spectrum as a function of frequency, in order to normalize it. It then identifies peaks based on the likelihood of encountering a peak of a certain magnitude in the resulting data set. If we have a "good fit", then the easiest way to get that function is to extract it from our fit parameters and the model that was used during fitting::
```

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+
+    params = [v.value for k, v in fit.results.items() if k in ['fc', 'D', 'f_diode', 'alpha']]
+    model_fun = lambda f: fit.model(f, *params)
```
JoepVanlier, line 526: For now this is fine, but it really makes me wonder whether we should have an API to return the solution parameters directly on the result object, since there is a risk these parameters change, or someone wants to fit a model without diode, or a different diode model in the future. It'd also be much more convenient.

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+    plt.figure()
+    fit_no_peak.plot()
+    plt.title(
+        f'Least squares (ex. peaks): $f_c$ = {fit_no_peak.results["fc"].value:.1f}, $D$ = {fit_no_peak.results["D"].value:.4f}, $f_d$ = {fit_no_peak.results["f_diode"].value:.1f}'
+    )
+    plt.show()
+
+.. image:: figures/force_calibration/power_spectrum_no_noise_peak.png
+
+Notice that no skewing occurs, and that the values of `fc`, `D` and `f_diode` are now closer to values found via robust fitting in the section above.
```
JoepVanlier, line 565: 🥳

JoepVanlier (changes requested):
Wonderful. I have made a few small suggestions.

I do think it might be nice to either expose a `solution_params` or a lambda to the model where they have already been filled in on the `FitResult` to make setting up the model for `identify_peaks` easier, but we can do that in a future PR.

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
 In that case, the arguments `fixed_diode` and `fixed_alpha` can be passed to :func:`~lumicks.pylake.calibrate_force()` to fix these parameters to their predetermined values.
+
+Robust fitting
+--------------
+
+So far, we have been using least-squares fitting routines for force calibration. In that case, we assume that the error in the power at each fequency is distributed according to a Gaussian distribution. The act of blocking or windowing of the power spectrum ensures that this assumption is close enough to the truth so that the fit gives accurate estimates of unknown parameters. Occasionally, the power spectrum might show a spurious noise peak. Such a peak is an outlier in the expected behavior of the spectrum and therefore interferes with the assumption of having a Gaussian error distribution. As a result, the fit is skewed. In those cases, it can be beneficial to do a so-called robust fit. When a robust fit is performed, one assumes that the probability of encountering one or multiple outliers is non-negligible. By taking this into account during fitting, the fit can be made more robust to outliers in the data, which explains the name. An example illustrates the idea.
```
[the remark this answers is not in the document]
rjmoerland: Gladly accepting all the corrections - they make it all the better.

docs/tutorial/force_calibration.rst
rjmoerland: I agree, and honestly I'm not 100% convinced about the 'statistical backing' figure we get right now.

docs/tutorial/force_calibration.rst
rjmoerland: Yeah I agree. This feels kind of 'hacky'. I'd say it would be nice to have a fit result into which you can immediately plug a frequency, and get a power density. TBC

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+        axis.legend()
+        axis.set_title(title)
+        axis.set_xlabel('Frequency [Hz]')
+    ax[1].set_xlim(frequency_exclusions[0][0] - 1.0, frequency_exclusions[-1][1] + 1.0)
+    ax[1].set_xscale('linear')
+    ax[0].set_ylabel('Power [V$^2$/Hz]')
+    plt.suptitle('Identified peaks')
+    plt.show()
+
+Finally, we can do a least-squares fit, but in thise case we will filter out the frequency ranges that contain peaks. Because we use a least-squares method, we get error estimates on the fit parameters, and bias in the fit result can be corrected. The default values of `loss_function='gaussian'` and `bias_correction=True` ensure least-squares fitting and bias correction, so we do not need to specify them::
```
rjmoerland, line 549:
```suggestion
Finally, we can do a least-squares fit, but in this case we will filter out the frequency ranges that contain peaks. Because we use a least-squares method, we get error estimates on the fit parameters, and bias in the fit result can be corrected. The default values of `loss_function='gaussian'` and `bias_correction=True` ensure least-squares fitting and bias correction, so we do not need to specify them::
```

docs/tutorial/force_calibration.rst
rjmoerland: Works! 🥳

JoepVanlier (approved):
Looks great! Just update the DOI and then from my side it's good to go I think.

I think Ray also still wanted to have a peek at it, so maybe hold off on merging until he has also had a chance to look at it.

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+    model_fun = lambda f: fit.model(f, *params)
+
+If there are no spurious peaks, then normalizing the unblocked power spectrum results in random numbers with an exponential distribution with a mean value of 1. The chance of encountering increasingly larger numbers decays exponentially, and this fact is used by `identify_peaks()`::
+
+    frequency_exclusions = ps.identify_peaks(model_fun, peak_cutoff=20, baseline=1)
+
+The parameter `peak_cutoff` is taken as the minimum magnitude of any value in the normalized power spectrum in order to be concidered a peak. The default value is 20, and it corresponds to a chance of about 2 in a billion of a peak of magnitude 20 or larger occuring naturally in a data set. If a peak is found with this or a higher magnitude, the algorithm then expands the range to the left and right until the first time the power spectrum drops below the value `baseline`. The frequencies at which this occurs end up as the lower and higher frequency of a frequency exclusion range. As such, the value of `baseline` controls the width of the frequency exclusion range. We can visualize the excluded peaks as follows::
```
rpauszek, line 532:
```suggestion
The parameter `peak_cutoff` is taken as the minimum magnitude of any value in the normalized power spectrum in order to be considered a peak. The default value is 20, corresponding to a chance of about 2 in a billion of a peak of magnitude 20 or larger occurring naturally in a data set. If a peak is found with this or a higher magnitude, the algorithm then expands the range to the left and right until the first time the power spectrum drops below the value `baseline`. The frequencies at which this occurs end up as a frequency exclusion range. As such, the value of `baseline` controls the width of the frequency exclusion range. We can visualize the excluded peaks as follows::
```

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+            axis.plot(ps.frequency[to_plot], ps.power[to_plot], 'r', label=f'peak {idx}')
+        axis.legend()
+        axis.set_title(title)
+        axis.set_xlabel('Frequency [Hz]')
+    ax[1].set_xlim(frequency_exclusions[0][0] - 1.0, frequency_exclusions[-1][1] + 1.0)
+    ax[1].set_xscale('linear')
+    ax[0].set_ylabel('Power [V$^2$/Hz]')
+    plt.suptitle('Identified peaks')
+    plt.show()
+
```
rpauszek, line 548: I think you forgot to include the figure `identify_peaks.png` here

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+Finally, we can do a least-squares fit, but in this case we will filter out the frequency ranges that contain peaks. Because we use a least-squares method, we get error estimates on the fit parameters, and bias in the fit result can be corrected. The default values of `loss_function='gaussian'` and `bias_correction=True` ensure least-squares fitting and bias correction, so we do not need to specify them::
```
rpauszek, line 549:
```suggestion
Finally, we can do a least-squares fit, but in this case we will filter out the frequency ranges that contain peaks. Because we use a least-squares method, we get error estimates on the fit parameters and bias in the fit result can be corrected (these options are turned one with the default values of `loss_function="gaussian"` and `bias_correction=True`, so we do not need to specify them)::
```

docs/tutorial/force_calibration.rst
```diff
@@ -464,3 +464,102 @@ When working with small beads or at high laser powers, it is important to verify
...
+Use a passive calibration model using the hydrodynamically correct model. Then perform a least-squares fit and plot the result::
+
+    model = lk.PassiveCalibrationModel(4.4, temperature=25.0, hydrodynamically_correct=True)
+    fit = lk.fit_power_spectrum(ps_blocked, model)
+
+    plt.figure()
+    fit.plot()
+    plt.title(
+        f'Skewed fit: $f_c$ = {fit.results["fc"].value:.1f}, $D$ = {fit.results["D"].value:.4f}, $f_d$ = {fit.results["f_diode"].value:.1f}'
+    )
```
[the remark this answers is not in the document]
rjmoerland: You caught me, I'm not a fan of `black` ;)

docs/tutorial/force_calibration.rst
rjmoerland: Thanks!

docs/tutorial/force_calibration.rst
rjmoerland: Indeed!

PR #552, opened 2023-08-08
black: enable `black` formatting for tests

JoepVanlier:
**Why this PR?**
If we're gonna be doing spring cleaning, might as well do all the tests now and just have `black` be active everywhere from now on. We can always selectively turn it off if we really need some bulky matrices.

I quickly went over all the changes and there was nothing particularly egregious anymore that wasn't already covered by explicitly disabling the formatter locally.

### DIFF
--- a/lumicks/pylake/kymotracker/tests/test_stitching.py
+++ b/lumicks/pylake/kymotracker/tests/test_stitching.py
@@ -16,7 +16,6 @@ def test_distance_line_to_point():
 
 
 def test_stitching(blank_kymo, blank_kymo_track_args):
-
     segment_1 = KymoTrack([0, 1], [0, 1], *blank_kymo_track_args)
     segment_2 = KymoTrack([2, 3], [2, 3], *blank_kymo_track_args)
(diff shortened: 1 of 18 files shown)


lumicks/pylake/kymotracker/tests/test_stitching.py
```diff
@@ -52,14 +51,14 @@ def test_stitching(blank_kymo, blank_kymo_track_args):
 
     # Check whether the alignment has to work in both directions
     # - and - should connect
-    track1, track2 = KymoTrack(
-        [0, 1], [0, 0], *blank_kymo_track_args), KymoTrack([2, 3], [0, 0], *blank_kymo_track_args
+    track1, track2 = KymoTrack([0, 1], [0, 0], *blank_kymo_track_args), KymoTrack(
+        [2, 3], [0, 0], *blank_kymo_track_args
```
rpauszek, line 55: could just define these each on their own line (for below too)

rpauszek (approved):
:shipit:

PR #553, opened 2023-08-08
Further speed up `import lymicks.pylake` and add regression test

dean0x7d:
This is following up on https://github.com/lumicks/pylake/pull/550 with further speed improvements and more importantly a regression test and ensures we don't accidentally end up eagerly loading a third-party module that's on the lazy list.

The speedups: Cold import goes down from 3.9 seconds to 1.1 seconds. Hot import goes from 0.4 to 0.15 seconds.

#### Cold import: BEFORE

Screenshot 2023-08-08 at 10 03 14

#### Cold import: AFTER

Screenshot 2023-08-08 at 16 46 51

#### Hot import: BEFORE

Screenshot 2023-08-08 at 10 04 08

#### Hot import: AFTER

Screenshot 2023-08-08 at 16 47 19


JoepVanlier (approved):
Looks good!

lumicks/pylake/tests/test_import_time.py
```diff
@@ -26,3 +27,33 @@ def test_disabling_capturing(report_line):
     report_line(
         f"Module import time: {np.mean(times):.2f} +- {np.std(times):.2f} seconds (N={repeats})"
     )
+
+
+def test_lazy_imports():
```
JoepVanlier, line 32: I like that it's pinned by explicitly testing for it. Otherwise I would definitely see this being forgotten at some point.

PR #555, opened 2023-08-10
deps: update `pylake` dependencies

JoepVanlier:
**Why this PR?**
We rely on some functionalities that depend on newer versions of `numpy` (`dtype` argument for hvstack) and lazy imports for `scipy`.

The lazy import ensures that this works:
```python
import scipy

scipy.optimize.curve_fit  # Prior to 1.9.0 this would fail.
```

Note that `scipy.optimize.constants` seems to have been forgotten in the initial release for lazy imports, but is added in [`1.11.1`](https://github.com/scipy/scipy/commit/855bc76864ea1e577bedb854a5d9aef97a149a86). The latter is probably still a bit too new.

The last thing is a temporary upper limit for the version of `jupyter notebook`. The reason we are introducing this is because we've been seeing issues with the `%matplotlib notebook` not working on several computers (resulting in an error that `IPython is not defined`) and with `%matplotlib ipympl` the kymotracker widget doesn't show up properly (needs further investigation).

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -4,7 +4,6 @@ from functools import partial
 from dataclasses import dataclass
 
 import numpy as np
-import scipy
 
 from .detail.drag_models import faxen_factor, brenner_axial
 from .detail.power_models import (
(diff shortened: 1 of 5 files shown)


rpauszek (approved):
🚀

dean0x7d (approved):
👌

PR #559, opened 2023-08-15
kymotrackgroup: test round trip from old csv

JoepVanlier:
**Why this PR?**
Turns out that you can't resave a `csv` file when it's been opened until you call `lk.filter_tracks()` because of the `None` entries.

### DIFF
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -139,7 +139,21 @@ def export_kymotrackgroup_to_csv(
             ),
         )
 
-    store_column("minimum_length (-)", "%d", minimum_length)
+    if None not in minimum_length:
+        store_column("minimum_length (-)", "%d", minimum_length)
+    else:
+        warnings.warn(
+            RuntimeWarning(
+                "Loaded tracks have no minimum length metadata defined (CSV exported with "
+                "pylake <  1.2.0) therefore some analyses may not be possible with this file. To "
+                "remedy this error and add the required metadata, use `lk.filter_tracks()` on the "
+                "tracks from the widget with the desired minimum track length and resave "
+                "using the `KymoTrackGroup.save()` method. Note that the minimum length should be "
+                "greater or equal to the minimum length used for the original tracking. For more "
+                "information refer to "
+                "https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html#migrating-old-track-files "
+            )
+        )
 
     version_header = f"Exported with pylake v{__version__} | track coordinates v3\n"
     header = version_header + delimiter.join(column_titles)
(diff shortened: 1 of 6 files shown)


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -139,7 +139,16 @@ def store_column(column_title, format_string, new_data):
             ),
         )
 
-    store_column("minimum_length (-)", "%d", minimum_length)
+    if not (None in minimum_length):
```
dean0x7d, line 142: More idiomatic placement of `not`?
```suggestion
    if None not in minimum_length:
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -139,7 +139,16 @@ def store_column(column_title, format_string, new_data):
...
+        store_column("minimum_length (-)", "%d", minimum_length)
+    else:
+        warnings.warn(
+            RuntimeWarning(
+                "This KymotrackGroup has no minimum length. Some analyses may not be possible with "
+                "this file. To get rid of this error and add the metadata, call "
+                "`lk.filter_tracks()` on this file with a given minimum track length."
+            )
+        )
```
dean0x7d, line 151: 1. I think you should promote this to an error. `np.savetxt()` will produce an error anyway. It's a bit weird to get a warning and then immediately an error. The warning text will get lost in there.
2. Wording suggestion: You mention a file in the text but the user just has a track group. Maybe say that this error happens for tracks imported from older files that did not contain all metadata.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Yeah, that's better.

dean0x7d (approved):
👌

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -139,7 +139,21 @@ def store_column(column_title, format_string, new_data):
...
+    else:
+        warnings.warn(
+            RuntimeWarning(
+                "Kymograph tracks have no minimum length defined. This means you loaded a file "
+                "from an older version of pylake (before 1.2.0). Some analyses may not be possible "
+                "with this file. To get rid of this error and add the metadata, call "
+                "`lk.filter_tracks()` on this file with a known minimum track length and resave "
+                "it. Note that this minimum length should be larger than the minimum length used "
+                "when tracking these tracks. Please refer to "
+                "https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html#migrating-old-track-files "
```
rpauszek, line 153:
```suggestion
                "Loaded tracks have no minimum length metadata defined (CSV exported with pylake <  1.2.0)
                "therefore some analyses may not be possible "
                "with this file. To remedy this error and add the required metadata, use "
                "`lk.filter_tracks()` on the tracks from the widget with the desired minimum track length and resave "
                "using the `KymoTrackGroup.save()` method. Note that the minimum length should be greater or "
                "equal to the minimum length used for the original tracking. For more information refer to "
                "https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html#migrating-old-track-files "
```

rpauszek (approved):
Suggested some different wording for the warning, but otherwise I think it's fine

PR #560, opened 2023-08-17
docs: update install instruction with latest conda

JoepVanlier:
**Why this PR?**
We pinned `conda` to an older version [earlier](https://github.com/lumicks/pylake/commit/11d383378bdbeea49c7ae61c4614668bafbc563b) to work around an OpenSSL issue. It seems that on newer `conda` versions, this issue is fixed. However, when creating a new conda environment from an older version _without_ specifying a version, it typically still installs an older version that does have the OpenSSL issue. The newer versions however, do download and extract packages in parallel which is much faster, so I would suggest keeping the explicit version in the installation instructions, but bumping it to the latest one.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -29,7 +29,7 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
 
     .. code-block:: python
 
-        conda create -n pylake conda=4.9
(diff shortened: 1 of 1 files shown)


aafkevandenberg (approved):
I already tested this, works like a charm!

PR #561, opened 2023-08-18
Docs diffusion intro

aafkevandenberg:
This introductions to diffusion analysis summarizes the recommended methods for diffusion analysis in Pylake and informs the user on important steps in diffusion analysis. It is a kind of guide to the Pylake tutorial for a C-Trap user that is not so experienced with diffusion analysis and equations.

Please, check whether the content is accurate, and whether you think a biology user can follow everything.
Also, let me know if an important step in diffusion analysis is missing.

Doc can be viewed here: https://lumicks-pylake.readthedocs.io/en/docs_diffusion_intro/theory/diffusion/intro.html

### DIFF
--- a/docs/theory/diffusion/intro.rst
+++ b/docs/theory/diffusion/intro.rst
@@ -1,0 +1,196 @@
+Introduction
+============
+
+This chapter gives an introduction to diffusion and an overview of the recommended methods for diffusion
+analysis available in Pylake. Rather than on math or syntax, this page focusses on the underlying ideas,
+providing an introduction for anyone who wants to start analyzing a kymograph with a diffusive protein.
+
+Target search
+-------------
+
+Proteins diffusing through a cell often have to search for a target on DNA. A long-standing question
+in biology is how proteins find their target on the huge genome through random, diffusive motion.
+The first hypothesis was that proteins diffuse through the cytoplasm (in 3D), until they find their target.
+Berg and von Hippel were the first to demonstrate that the Lac repressor finds its target faster than
+can be explained by 3D diffusion alone. Now, many experiments have shown that proteins often alternate
+between diffusion in 3D (in the cytoplasm) and in 1D (along DNA). The combination of 1D and 3D diffusion
+is referred to as 'facilitated diffusion' :cite:`Berg1976, Berg1981` and is more efficient than 3D diffusion alone.
+
+.. image:: figures/target_search.png
+
+1D diffusion can directly be visualized by holding a force-stretched DNA between optically trapped beads
+and observing a fluorescently labelled protein bind and move along this DNA. Pylake has many functions
+implemented for analyzing data from such an assay. This chapter focusses on the analysis of kymographs
+with a protein diffusing along DNA, as in the figure below.
+
+.. image:: figures/1D_diffusion.png
+
+The diffusion coefficient
+-------------------------
+
+A purely diffusive protein on DNA has the same probability to move to the right as to the left, therefore
+the average displacement is zero. Over time, the protein will explore larger and larger regions of DNA.
+The *diffusion coefficient (D)* quantifies how fast a protein explores its environment.
+For example, the orange track in the figure below has a higher diffusion coefficient than the blue track.
+The diffusion coefficient has units of :math:`length^{2}/s`. For 1D diffusion along DNA, D is often expressed
+in :math:`µm^{2}/s` or :math:`kbp^{2}/s` :cite:`Kochaniak2009`.
+
+.. image:: figures/diffusive_tracks.PNG
+
+
+Methods for obtaining the diffusion coefficient
+-----------------------------------------------
+
+The first step for analyzing a diffusive protein from a kymograph, is to track each binding event,
+which is further discussed below. The resulting tracks give the positions of each bound protein over time.
+These tracks can then be analyzed further to determine the diffusion coefficient. Below we discuss
+two methods for diffusion analysis that are available in Pylake.
+
+Mean Squared Displacement analysis
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A common method for determining the diffusion coefficient, is to first compute the mean squared
+displacement (MSD) as a function of lag time for each track. For purely diffusive motion, the first part
+of the MSD curve is linear, with the slope proportional to the diffusion coefficient. For 1D diffusion
+the MSD is given by: :math:`MSD = 2D\Delta t`. The diffusion coefficient can be obtained by fitting a
+linear function to the MSD curve, as illustrated in the figure below (simulated data):
+
+.. image:: figures/MSD_analysis.png
+
+Note that for the MSD curve, the 'lag time' (aka 'time interval') is given on the x-axis, rather than 'time'.
+To compute the MSD for a certain lag time, you scan over a track, one data point at a time and compute
+the displacement squared from that time until that time + the lag time. So each point on the MSD curve,
+is an average of displacements squared for a specific lag time, see :ref:`this section <MSD Eqn>`
+for the equation for the MSD.
+
+The largest possible lag time is the duration of a track and the shortest possible lag is the time resolution,
+the line time of the kymograph. The image above does not show the complete MSD curve; the longest possible lag
+time would be 2500 seconds, but the MSD curve is only given up to a lag time of 5 seconds. The full MSD curve looks as follows:
+
+.. image:: figures/MSD_full.png
+
+As you can see, the full MSD curve becomes less and less linear for larger lag times. The non-linear
+shape of the MSD curve for large lag times is caused by:
+
+#. The MSD at shorter time lags is averaged over more lags and therefore more accurate
+
+#. Consecutive data points in the MSD are highly correlated
+
+The MSD curve for larger lags is clearly not suitable for a linear fit. An important step in MSD analysis,
+is therefore to determine which number of lags is optimal for a linear fit and obtaining the diffusion coefficient.
+Taking too many lags, may result in an MSD curve that is not fully linear, while taking too little lags
+may not give enough data points for a good estimate of the diffusion coefficient. There are various methods
+for determining the optimal number of lags. The method recommended in Pylake automatically determines the optimal
+number of lags for you :cite:`michalet2012optimal` (named 'OLS' in Pylake).
+
+Covariance Based Estimator
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Covariance Based Estimator (:ref:`'CVE' <diffusion_cve>`) in Pylake is a method for determining
+the diffusion coefficient that does not rely on the MSD :cite:`vestergaard2014optimal,vestergaard2016optimizing,Tanasie2022`.
+This method directly computes the diffusion coefficient from the displacements; there is no fitting
+involved. For the full equation, see :ref:`this section <CVE Eqn>`.
+
+Comparing CVE and MSD analysis
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When tracks are long and diffusion clearly dominates over noise, CVE and MSD analysis perform equally well.
+The advantage of MSD analysis, is that it can be used to quantify anomalous diffusion (further discussed below).
+A disadvantage of MSD analysis is that it can have a small bias when applied to very short tracks, or when
+the diffusion coefficient is very small (when a method is biased, the diffusion coefficient obtained
+via that method deviates from the real diffusion constant). CVE on the other hand, is an unbiased method
+for determining the diffusion coefficient, but is only used for analysis of free (non-anomalous) diffusion.
+For a more detailed comparison between the performance of CVE and MSD analysis, see :ref:`here<comparing diffusion estimators>`.
+
+Ensemble Based Estimate
+^^^^^^^^^^^^^^^^^^^^^^^
+
+The estimate of the diffusion coefficient can further be improved by using an ensemble estimate.
+For an ensemble estimate using MSD, Pylake averages the MSD for each lag time and combines them into one MSD curve.
+This MSD can then be fitted to obtain the diffusion coefficient. For CVE, Pylake computes the average
+diffusion coefficient of all tracks, where each track is weighted by the number of data points.
+An example of how to compute the ensemble estimate using CVE is given :ref:`here <diffusion_cve>` and
+a comparison between using single and ensemble estimates can be found :ref:`here<ensemble diffusion>`.
+
+Tracking a diffusive protein
+----------------------------
+
+Pylake has built-in tracking algorithms to track binding events on a kymograph over time.
+For more details on how to use the tracking algorithm in Pylake, see the :doc:`Kymotracking tutorial</tutorial/kymotracking>`
+and specifically :ref:`the section on diffusion <kymotracking-diffusion>` regarding diffusion analysis
+on kymographs. After tracking, it is possible to further improve the estimate of the positions and time
+coordinates of the tracks through refinement.
+
+Refinement
+^^^^^^^^^^
+The tracking algorithm in the Pylake kymotracker does not always move from one pixel to the next, but
+sometimes skips a few pixels within a track. In the figure below for example, the track (white) sometimes
+skips one or multiple pixels. The missing frames can be added through refinement. Further, refinement
+can slightly improve the position estimate of a track. (The default tracking algorithm already has subpixel accuracy).
+The orange line shows the update of the coordinates after refining the white track.
+
+*Note that refining missing frames is recommended for MSD analysis, but not necessary for CVE.*
+MSD analysis works best when the time between all data points is the same.
+
+.. image:: figures/refinement.png
+
+Pylake has :ref:`two methods on refinement<localization_refinement>`:
+
+1) **Centroid refinement**
+
+The method 'Centroid refinement' can be used to refine missing frames after tracking. The disadvantage
+of this method, is that it does not perform well when tracks are close together or with high background.
+At the moment, this method is the default in the kymotracker, because it is fast.
+
+2) **Gaussian refinement**
+
+Gaussian refinement fits a Gaussian function to improve the position estimate of a track and can also
+refine missing frames. The advantage of this method is that it is better at refining tracks that are
+close together or have a high background. Refining missing frames is not the default. When doing MSD
+analysis, you have to activate the refinement of missing frames, by setting `refine_missing_frames = True`
+(see section on :ref:`gaussian refinement<kymotracker_gaussian_refinement>`). After refinement, it is
+good practise to inspect the refined peaks. This can be done using
+:meth:`KymoTrack.plot_fit()<lumicks.pylake.kymotracker.kymotrack.KymoTrack.plot_fit>` or
+:meth:`KymoTrackGroup.plot_fit()<lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup.plot_fit>` and
+is illustrated :doc:`here</whatsnew/1.2.0/1_2_0>`.
+
+
+Miscellaneous
+-------------
+
+Negative diffusion coefficients
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+Depending on the method used for diffusion analysis, a diffusion coefficient of a single track may
+sometimes be negative. For example, when a diffusion coefficient is low, some diffusion coefficients
+are above and some below zero. The negative values do not have a biological meaning by themselves,
+but the average of all the diffusion coefficients from different tracks should still give a good,
+positive-valued estimate of the diffusion coefficient (provided that you have enough tracks, and that
+the protein is diffusing freely). The negative data points should not be removed from the dataset,
+otherwise you get a bias in the estimate of the diffusion coefficient :cite:`michalet2012optimal`.
+
+Confined diffusion
+^^^^^^^^^^^^^^^^^^
+If many diffusive proteins bind to the DNA at once and they are not able to pass each other, the proteins
+will confine each others motion, resulting in sub-diffusive or confined diffusive behavior. If you are
+interested in free diffusion of a protein, it is therefore best to keep the density of protein binding
+events low enough, such that the proteins don't meet often. When the DNA is held between two beads in
+an optical tweezer experiment, the beads can also confine the motion of a diffusive protein if the protein
+diffuses close to the beads. If you want to exclude the effect of the beads, you can consider cutting
+the tracks to exclude the part where the protein reaches the bead. Another approach is to account for
+the presence of the beads when analyzing the data, but this option is not available in Pylake at the moment.
+
+Fluctuating DNA
+^^^^^^^^^^^^^^^
+Though the DNA in an optical tweezer experiment is constrained, it still fluctuates. The motion of the
+DNA sets a lower limit on what diffusion coefficient can be observed for a protein diffusing
+along the DNA. A typical approach is to measure the diffusion coefficient of a static protein or
+marker and use this as a reference for the minimal observable diffusion coefficient (see for example
+:cite:`Kaczmarczyk2022`).
+
+Anomalous Diffusion
+^^^^^^^^^^^^^^^^^^^
+If a protein is hindered, for example by obstacles, it can not diffuse freely anymore, and the MSD
+curve will look different. A protein that has some motor activity will also have a
+differently shaped MSD curve. In both cases, we would refer to the motion as 'anomalous diffusion'.
+At the moment, Pylake does not have functionality for analyzing anomalous diffusion.
+Usually, anomalous diffusion is analyzed by looking at the shape of the MSD curve.
--- a/docs/refs.bib
+++ b/docs/refs.bib
@@ -1,3 +1,62 @@
+@article{Berg1976,
+  title = {Association kinetics with coupled diffusional flows: Special application to the lac repressor-operator system},
+  journal = {Biophysical Chemistry},
+  volume = {4},
+  number = {4},
+  pages = {367-381},
+  year = {1976},
+  issn = {0301-4622},
+  doi = {https://doi.org/10.1016/0301-4622(76)80017-8},
+  author = {Otto G. Berg and Clas Blomberg}
+ }
+
+@article{Berg1981,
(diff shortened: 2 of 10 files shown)


docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+    If the DNA is held between two beads in an optical tweezer experiment, the beads can also confine the motion of a diffusive protein, if the protein diffuses far enough. 
+    If you are interested in free diffusion of a protein, it is best to keep the density of protein binding events low enough, so the proteins don't meet often. 
+    This will help reducing the number binding events, but not when the motion is so wide that the protein ends up on the beads. If you want to exclude the effect of the beads, you can consider cutting the track 
+    to exclude the part where the protein reaches the bead.
+
+- **How do I account for the motion of the DNA molecule itself?**
+    Though the DNA in an optical tweezer experiment is constrained, it still is fluctuating. The motion of the DNA sets a lower limit on what diffusion coefficient can be observed for a protein diffusing 
+    along the DNA. To find out this lower limit, you can use a static molecule as a reference. The motion of the DNA also can lead to a bias in the estimation of the diffusion coefficient :cite:`vestergaard2014optimal`.`
+    (SHOULD WE MENTION THE BIAS HERE, HOW BIG IS THE BIAS?)
```
arturrsl, line 132: I would say the MSD of a static molecule gives a flat graph (there's a small slope in the first few tau points). the flat line crosses the y-axis at certain level and this could be the estimate of the bias.

docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+
+- **How do I analyze anomalous diffusion using Pylake?**
+    If a protein is hindered, for example by obstacles, it can not diffuse freely anymore, and the MSD curve will look different. Similarly, a protein that has some motor activity will have a 
+    differently shaped MSD curve. In both cases, we would refer to the motion as 'anomalous diffusion'. At the moment, Pylake does not have any functionality for analyzing anomalous diffusion.
+    Usually, anomalous diffusion is analyzed by looking at the shape of the MSD curve.
```
arturrsl, line 137: It is actually easy to implement in Pylake

Fitting the MSD plot with a power law D∆t^α results in:
 α > 1 for directed motion,
 α ≤ 1 for constrained diffusion
 α = 1 for pure diffusion

docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+The estimate of the diffusion coefficient can further be improved by using an ensemble estimate. 
+In Pylake, functions are implemented that compute the average diffusion coefficient of all tracks, where each track is weighted by the number of data  points. 
+(REF TO PAGE, mention that for MSD, tracks should be approximately equally long?)
+
+Considerations when tracking a diffusive particle
+--------------------------------------------------
+
+Refinement
+^^^^^^^^^^
+Depending on the tracking algorithm and parameters, tracked lines might not have an explicit position value at each time point.
```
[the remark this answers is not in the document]
aafkevandenberg: I updated the comment, is it more clear now?

docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+    coefficient (provided that you have enough tracks, and that the protein is diffusing freely). The negative data points *should not* be removed from the dataset, otherwise you get a bias in your estimate. 
+
+- **What happens to the diffusion coefficient if a protein bumps into another protein or against other obstacles?**
+    If there are many diffusive proteins binding to the DNA at once and they are not able to pass each other, the proteins will confine the motion, resulting in sub-diffusive or confined diffusive behavior.
+    If the DNA is held between two beads in an optical tweezer experiment, the beads can also confine the motion of a diffusive protein, if the protein diffuses far enough. 
+    If you are interested in free diffusion of a protein, it is best to keep the density of protein binding events low enough, so the proteins don't meet often. 
+    This will help reducing the number binding events, but not when the motion is so wide that the protein ends up on the beads. If you want to exclude the effect of the beads, you can consider cutting the track 
+    to exclude the part where the protein reaches the bead.
+
```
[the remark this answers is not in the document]
aafkevandenberg: Not ignoring it :p

I write that the protein concentration should be 'low enough, such that proteins don't meet often', so this aligns with you saying that crowding only starts to matter once protein bounce into each other a lot.

And since the effect of the beads cannot be accounted for in the analysis in Pylake at the moment, I don't want to go into it too much. I'll just mention that it is possible to account for the beads during analysis. Your explanation about flattening makes a lot of sense. But yes, a reference would be nice...

docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,142 @@
...
+Covariance Based Estimator
+^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+The Covariance Based Estimator (:ref:`'CVE' <diffusion_cve>`) in Pylake is a method for determining the diffusion coefficient that does not rely on the MSD :cite:`vestergaard2014optimal,vestergaard2016optimizing`. 
+(CITE STIGLER) This method directly computes the diffusion coefficient from the displacements.
+
+Comparing CVE and MSD analysis
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Since there is no fitting involved, CVE is fast and unbiased (unbiased means that the more data you get, the closer the average gets to the real diffusion constant).
```
JoepVanlier, line 78: I really like that you're adding some more user-friendly documentation.

Sorry about the drive-by comment, I know you said it wasn't ready for review yet, but I noticed this and wanted to leave a quick comment:

- This sentence makes it seem like fitting causes the bias but it doesn't. The bias in the MSD mostly comes from the fact that it is very difficult to design an unbiased estimator based on that quantity (because of the non-linearity applied when estimating it and the correlation structure involved when you estimate it with time averaging). I would also argue that fitting doesn't necessarily make things slow (Since the regular MSD is a very basic linear fit, which is virtually instantaneous. It's the calculation of the MSD itself that costs time because of the calculation of all the measurement pairs).

- The second thing is that the definition of bias presented here is not the statistical definition of bias. What you describe sounds more like consistency (that as `N->inf`, `theta_est -> theta`). I'm honestly not sure whether the MSD is a consistent estimator either, but it might be.

Statistical bias is a property of finite samples. It means that if you were to make a distribution of estimates based on finite samples, the expectation (average) of that distribution would not be the true population value. It is a property for a specific `N`.

A good example of a biased but consistent estimator is the sample variance estimator. This is when you estimate the variance plugging in the mean from the same sample and you don't take into account the Bessel correction. For an infinite number of samples, it will converge to the true value, but if you take the average of many samples computed like this, the one with the Bessel correction will produce the correct result, while the one without will reveal its bias.

A good example of an unbiased but consistent estimator of the population mean would be just taking the first element of the sample. It's clearly unbiased (the average over many of these estimates based on samples from the population would result in the correct estimate), but it will never converge as the sample size for a single estimate goes to infinity.

I would just mention that the CVE is fast and unbiased and leave it at that. I feel bias, consistency and precision could almost be their own theory pages.

docs/introduction/diffusion_intro/diffusion_intro.rst
aafkevandenberg: hmm, yes, I guess a static molecule would undergo a kind of subdiffusive motion. At small distances, the DNA can diffusive freely, but at larger distances, it will feel the confinement from the beads. How this affects the MSD of a moving protein depends on how floppy the DNA is and how that compares to the time resolution and the diffusion coefficient of the protein of interest. Vestergaard PRE 2014 discusses various limits, but does not explain the bias in an intuitive way, so I don't know a good way to summarize this for now. Most papers take the approach you did, to use the fluctuation of a static protein as a lower limit. I think I'll just say that this is a common thing to do.

docs/introduction/diffusion_intro/diffusion_intro.rst
JoepVanlier: It's not quite that simple. We deliberately do not offer that functionality yet since a naive power law fit of the MSDs with an exponent can easily lead to misinterpretation.

It is very easy to get something that looks like anomalous diffusion when it actually isn't (even when simulating pure free diffusion with observation noise). See for example:

Martin, D. S., Forstner, M. B., & Käs, J. A. (2002). Apparent subdiffusion inherent to single particle tracking. Biophysical journal, 83(4), 2109-2117.
Kepten, E., Weron, A., Sikora, G., Burnecki, K., & Garini, Y. (2015). Guidelines for the fitting of anomalous diffusion mean square displacement graphs from single particle tracking experiments. PLoS One, 10(2), e0117722.

Or how poorly the `MSD` actually performs at estimating this exponent:
Muñoz-Gil, G., Volpe, G., Garcia-March, M. A., Aghion, E., Argun, A., Hong, C. B., ... & Manzo, C. (2021). Objective comparison of methods to decode anomalous diffusion. Nature communications, 12(1), 6253.

In addition, the optimal number of points to use in the fitting is not valid when diffusion is anomalous. I would err on the side of caution here, and wait until we have a better picture of what to advise so that people don't get heavily biased estimates.

aafkevandenberg: ok, I'll leave out the detailed suggestion and just say that anomalous diffusion is typically determined by looking at the MSD curve.

docs/introduction/diffusion_intro/diffusion_intro.rst
```diff
@@ -0,0 +1,138 @@
...
+.. image:: figures/1D_diffusion.png
+
+
+
+The diffusion coefficient
+---------------------------
+
+A purely diffusive protein has the same probability to step to the right as to the left at any point, therefore the average displacement is zero. 
+Rather than through velocity, diffusion is therefore quantified through the *diffusion coefficient (D)*. The diffusion coefficient quantifies how fast a protein explores its environment; 
+the units of the diffusion coefficients are :math:`length^{2}/s`. For 1D diffusion along DNA, D is often expressed in :math:`µm^{2}/s` or :math:`kbp^{2}/s` (CITE).
```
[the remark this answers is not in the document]
aafkevandenberg: Nice suggestion, I do like how diffusion analysis is explained in this paper, will include it.

docs/introduction/diffusion_intro/diffusion_intro.rst
aafkevandenberg: Does it make sense to cite that first paper that you mention, Martin et al? Is that the one you are planning to implement? Do you know single molecule studies where this paper has been used?

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+.. image:: figures/1D_diffusion.png
+
+
+
+The diffusion coefficient
+---------------------------
+
+A purely diffusive protein has the same probability to step to the right as to the left from any location on the DNA, therefore the average displacement is zero. 
+Rather than through velocity, diffusion is therefore quantified through the *diffusion coefficient (D)*. The diffusion coefficient quantifies how fast a protein explores its environment;
```
JoepVanlier, line 30:
```suggestion
Since diffusion is distinct from directed motion, which would be characterized by a velocity, it is quantified with a different quantity named the *diffusion coefficient (D)*. The diffusion coefficient quantifies how fast a protein explores its environment; 
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+.. image:: figures/1D_diffusion.png
+
+
+
+The diffusion coefficient
+---------------------------
+
+A purely diffusive protein has the same probability to step to the right as to the left from any location on the DNA, therefore the average displacement is zero. 
+Rather than through velocity, diffusion is therefore quantified through the *diffusion coefficient (D)*. The diffusion coefficient quantifies how fast a protein explores its environment; 
+the units of the diffusion coefficients are :math:`length^{2}/s`. For 1D diffusion along DNA, D is often expressed in :math:`µm^{2}/s` or :math:`kbp^{2}/s` :cite:`Kochaniak2009`.
```
JoepVanlier, line 31: Should probably define what you mean by `kbp` in the text?

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Methods for obtaining the diffusion coefficient
+------------------------------------------------
```
JoepVanlier, line 34:
```suggestion
Methods for obtaining the diffusion coefficient
-----------------------------------------------
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+The first step for analyzing a diffusive protein from a kymograph, is to track each binding event. The resuling tracks give the positions of each bound protein over time.
```
JoepVanlier, line 36:
```suggestion
The first step to analyze a diffusive protein from a kymograph, is to track each binding event. The resulting tracks give the positions of each bound protein over time.
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+These tracks can then be analyzed further to determine the diffusion coefficient. Below we discuss two methods for diffusion analysis that are available in Pylake. 
+
+Mean Squared Displacement analysis
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+A common method for determining the diffusion coefficient, is to first compute the mean squared displacement (MSD) as a function of lag time for each track.
```
JoepVanlier, line 42:
```suggestion
A common method for determining the diffusion coefficient is to first compute the mean squared displacement (MSD) as a function of lag time for each track. 
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+from that time until that time + the lag time. So each point on the MSD curve, is an average of displacements for a specific lag time, see the :doc:`theory section on diffusive motion</theory/diffusion/diffusion>`
+for the equation for the MSD. 
+
+The largest possible lag time is the duration of a track and the shortest possible lag is the time resolution, the line time of the kymograph. 
+The image above does not show the complete MSD curve. The longest possible lag time would be 2500 seconds, but the MSD curve is only given up to a lag time of 5 seconds. 
+The full MSD curve looks as follows:
+
+.. image:: figures/MSD_full.png
+
+As you can see, the full MSD curve becomes less and less linear for larger lags times. The funky shape of the MSD curve for large lag times is caused by:
```
JoepVanlier, line 59: Funky? :D

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+A common method for determining the diffusion coefficient, is to first compute the mean squared displacement (MSD) as a function of lag time for each track. 
+For purely diffusive motion, the first part of the MSD curve is linear, with the slope proportional to the diffusion coefficient. For 1D diffusion the MSD is given by: :math:`MSD = 2D\Delta t`.
+The diffusion coefficient can be obtained by fitting a linear function to the MSD curve, as illustrated in the figure below (simulated data):
+
+.. image:: figures/MSD_analysis.png
+
+Note that for the MSD curve, the 'lag time' (aka 'time interval') is given on the x-axis, rather than 'time'. 
+To compute the MSD for a certain lag time, you scan over a track, one data point at a time and compute the displacement squared 
+from that time until that time + the lag time. So each point on the MSD curve, is an average of displacements for a specific lag time, see the :doc:`theory section on diffusive motion</theory/diffusion/diffusion>`
+for the equation for the MSD.
```
JoepVanlier, line 51: It might be instructive to mention how this works. That for a track without gaps, you have `N-1` intervals to average over for `lag 1`, `N-2` for `lag 2` etc. To really drive home that you are averaging over time.

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+.. image:: figures/MSD_full.png
+
+As you can see, the full MSD curve becomes less and less linear for larger lags times. The funky shape of the MSD curve for large lag times is caused by:
+
+#. The MSD at shorter time lags is averaged over more lags and therefore more accurate
+
+#. Consecutive data points in the MSD are highly correlated
+
+The MSD curve for larger lags is not suitable for a linear fit. An important step in MSD analysis, is therefore to determine which number of lags is optimal for a linear fit and obtaining the diffusion coefficient.
+Taking too many lags, may result in an MSD curve that is not fully linear, while taking too little lags may not give enough data points for a good estimate of the diffusion coefficient.
```
JoepVanlier, line 66:
```suggestion
Taking too many lags may result in an MSD curve that is not fully linear, while taking too few lags may not give enough data points for a good estimate of the diffusion coefficient.
```

JoepVanlier: Interestingly, if there is no observation noise and free diffusion, `2` lags is optimal.

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+Comparing CVE and MSD analysis
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+When tracks are long and diffusion clearly dominates over noise, CVE and MSD analysis perform equally well.
+MSD analysis can have a small bias when using very short tracks, or when the diffusion coefficient is very small (when a method is biased, the diffusion coefficients obtained via that method deviates 
+from the real diffusion constant). CVE on the other hand, is an unbiased method for determining the diffusion coefficient. 
+For a more detailed comparison between CVE and OLS, see the :doc:`theory section on diffusive motion</theory/diffusion/diffusion>`.
+
+Ensemble Based Estimate
+^^^^^^^^^^^^^^^^^^^^^^^^
```
JoepVanlier, line 84:
```suggestion
Ensemble Based Estimate
^^^^^^^^^^^^^^^^^^^^^^^
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+The estimate of the diffusion coefficient can further be improved by using an ensemble estimate.
```
JoepVanlier, line 86:
```suggestion
The estimate of the diffusion coefficient can be improved further by using an ensemble estimate. 
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+Refinement
+^^^^^^^^^^
+The tracking algorithm in the Pylake kymotracker does not always move from one pixel to the next, but sometimes skips a few pixels within a track. 
+In the figure below for example, the track (white) sometimes skips one or multiple pixels.
+The missing frames can be added through refinement. Further, refinement can improve the position estimate of the event. The orange line shows the update of the coordinates after refining the white track.
+
+*Note that refining missing frames is necessary for MSD analysis, but not for CVE.* MSD analysis works best when the time between all data points is the same. 
+
+.. image:: figures/refinement.png
```
JoepVanlier, line 101: I really like that you mention and show this.

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+Pylake has :ref:`two methods on refinement<localization_refinement>`:
+
+1) **Centroid refinement**
+
+Centroid refinement always refines missing frames. The disadvantage of this method, is that it does not perform well when tracks are close together or with high background. 
+At the moment, this method is the default in the kymotracker.
+
+2) **Gaussian refinement (recommended)**
+
+Gaussidan refinement fits a Gaussian function to improve the position estimate. This method is better at refining tracks that are close together.
```
JoepVanlier, line 112:
```suggestion
Gaussian refinement fits a Gaussian function to improve the position estimate. This method is better at refining tracks that are close together. 
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+1) **Centroid refinement**
+
+Centroid refinement always refines missing frames. The disadvantage of this method, is that it does not perform well when tracks are close together or with high background. 
+At the moment, this method is the default in the kymotracker.
+
+2) **Gaussian refinement (recommended)**
+
+Gaussidan refinement fits a Gaussian function to improve the position estimate. This method is better at refining tracks that are close together. 
+Using Gaussian refinement to improve the position estimate of tracks is always recommended. Gaussian refinement can also be used to refine missing frames, but this is not the default. When doing MSD analysis, 
+you have to activate the refinement of missing frames, by setting `refine_missing_frames = True` (see section on :ref:`gaussian refinement<kymotracker_gaussian_refinement>`).
```
JoepVanlier, line 114: Cross referencing is nice! 👍

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+
+FAQ on diffusion analysis
+--------------------------
+- **The diffusion coefficients for some of the tracks are negative, what does that mean?**
```
JoepVanlier, line 119:
```suggestion
FAQ on diffusion analysis
-------------------------

- **The diffusion coefficients for some of the tracks are negative, what does that mean?**
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+    Sometimes a diffusion coefficient of a single track is negative. For example, when your diffusion coefficient is low, you will have some diffusion coefficients above and some below zero. 
+    The negative values do not have a biological meaning by themselves, but the average of all the diffusion coefficients from different tracks should still give a good, positively values estimate of the diffusion
```
JoepVanlier, line 121:
```suggestion
    The negative values do not have a biological meaning by themselves, but the average of all the diffusion coefficients from different tracks should still give a good, positive-valued estimate of the diffusion 
```

docs/theory/diffusion/intro.rst
JoepVanlier: I really like the FAQ btw.

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+FAQ on diffusion analysis
+--------------------------
+- **The diffusion coefficients for some of the tracks are negative, what does that mean?**
+    Sometimes a diffusion coefficient of a single track is negative. For example, when your diffusion coefficient is low, you will have some diffusion coefficients above and some below zero. 
+    The negative values do not have a biological meaning by themselves, but the average of all the diffusion coefficients from different tracks should still give a good, positively values estimate of the diffusion 
+    coefficient (provided that you have enough tracks, and that the protein is diffusing freely). The negative data points *should not* be removed from the dataset, otherwise you get a bias in your estimate. 
+
+- **What happens to the diffusion coefficient if a protein bumps into another protein or against other obstacles?**
+    If there are many diffusive proteins binding to the DNA at once and they are not able to pass each other, the proteins will confine the motion, resulting in sub-diffusive or confined diffusive behavior.
```
JoepVanlier, line 125:
```suggestion
    If many diffusive proteins bind to the DNA at once and they are not able to pass each other, the proteins will confine the motion, resulting in sub-diffusive or confined diffusive behavior.
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+    If you are interested in free diffusion of a protein, it is best to keep the density of protein binding events low enough, such that the proteins don't meet often. 
+    When the DNA is held between two beads in an optical tweezer experiment, the beads can also confine the motion of a diffusive protein, if the protein diffuses far enough.
```
JoepVanlier, line 127:
```suggestion
    When the DNA is held between two beads in an optical tweezer experiment, the beads can also confine the motion of a diffusive protein if the protein diffuses far enough. 
```

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+    coefficient (provided that you have enough tracks, and that the protein is diffusing freely). The negative data points *should not* be removed from the dataset, otherwise you get a bias in your estimate. 
+
+- **What happens to the diffusion coefficient if a protein bumps into another protein or against other obstacles?**
+    If there are many diffusive proteins binding to the DNA at once and they are not able to pass each other, the proteins will confine the motion, resulting in sub-diffusive or confined diffusive behavior.
+    If you are interested in free diffusion of a protein, it is best to keep the density of protein binding events low enough, such that the proteins don't meet often. 
+    When the DNA is held between two beads in an optical tweezer experiment, the beads can also confine the motion of a diffusive protein, if the protein diffuses far enough. 
+    If you want to exclude the effect of the beads, you can consider cutting the track to exclude the part where the protein reaches the bead.
+    Another approach is to account for the presence of the beads when analyzing the data, but this option is not available in Pylake at the moment.
```
JoepVanlier, line 129: We should really add that feature.

docs/theory/diffusion/intro.rst
```diff
@@ -0,0 +1,140 @@
...
+
+- **How do I account for the motion of the DNA molecule itself?**
+    Though the DNA in an optical tweezer experiment is constrained, it still is fluctuating. The motion of the DNA sets a lower limit on what diffusion coefficient can be observed for a protein diffusing
```
JoepVanlier, line 132:
```suggestion
    Though the DNA in an optical tweezer experiment is constrained, it still fluctuates. The motion of the DNA sets a lower limit on what diffusion coefficient can be observed for a protein diffusing 
```

JoepVanlier (changes requested):
Awesome! I added a few small suggestions.

docs/introduction/diffusion_intro/diffusion_intro.rst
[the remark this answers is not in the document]
JoepVanlier: Yeah, I think it's better indeed. For these things, it would be really awesome if we had some sort of glossary functionality on ReadTheDocs. Where you hover over a word and see a small glossary item with an image or something.

Many of these things are so easy to show in a plot, but so hard to put into words.

docs/introduction/diffusion_intro/diffusion_intro.rst
JoepVanlier: Yeah, for now that's fine. We should really get some of this anomalous stuff into `pylake`.

JoepVanlier (approved):
Looks like a nice addition to the docs! Thank you!

PR #562, opened 2023-08-21
touchdown: use nonzero initial guess for decay

JoepVanlier:
Use non-zero initial guess for the decay parameter so it isn't exactly at the bound. `scipy=1.11.2` issues a `DivisionByZero` warning when the initial guess is exactly at the bound, but the solution is outside of the bounds.

Since in our case, it is better to provide a nonzero initial guess for the decay anyhow, this seems like the best solution.


tobiasjj (approved):
Looks like an adequate fix. 🚴

PR #563, opened 2023-08-22
merge `1.2` into main

JoepVanlier:
**Why this PR?**
I wanna merge the bug fixes so far into main since we have a test firing on CI that prevents the tests from running properly there.

I'd prefer to just merge back early rather than cherry-picking the fix since it makes it easier to keep track of what has and hasn't been merged back.


tobiasjj (approved):
LGTM! 🚴

PR #564, opened 2023-08-28
Reorganize the docs sidebar

rpauszek:
**Why this PR?**

Just some more spring cleaning for the docs. This one is in prep for merging #561, which is a nice beginners' intro to the diffusion theory. The idea moving forward would be that each topic gets this type of intro and builds up the theory to the most advanced stuff so we can address the needs of all levels of users

[Docs build here](https://lumicks-pylake.readthedocs.io/en/docs_reorg/index.html)

Pulled the theory link out of the User Guide section and made it its own section called Fundamentals (suggestions for a better name definitely welcome)

image

Also took the 2 theory pages we have currently and split them up into separate files. I think this is a bit nicer because everything is now split in more digestible chunks rather than being assaulted by a wall of text and equations.


aafkevandenberg:
Hi Ray, the reshuffling definitely makes it better, now the docs have a more logical structure with smaller pieces of information per section.

Some suggestions:

- The 'overview' section feels a bit redundant. It barely gives more information than when you click on 'Diffusion' and watch the section headings. Under 'User Guide' we also don't have an 'Overview' section. 
- An alternative name for 'Fundamentals' would be 'Learn more'

docs/theory/diffusion/index.rst
```diff
@@ -0,0 +1,25 @@
+Diffusion
+=========
+
+Pylake supports a number of methods for studying diffusive processes.
+These methods rely on quantifying the movement of the fluorophore.
```
aafkevandenberg, line 5: Without context (explaining about a kymograph, optical tweezers), this sentence can be a bit confusing. Currently, this context is given in the introduction to diffusion analysis. You could just remove the sentence from this paragraph.

docs/theory/force_calibration/active.rst
```diff
@@ -0,0 +1,82 @@
...
+
+Analogously to passive calibration, there is also a hydrodynamically correct theory for active calibration which should be used when inertial forces cannot be neglected.
+This involves fitting the thermal spectrum with the hydrodynamically correct power spectrum discussed earlier, but also requires using a hydrodynamically correct model for the peak:
+
+.. math::
+
+    P_\mathrm{hydro}^\mathrm{active}(f) = \frac{\left(A f_\mathrm{drive} \left|\gamma / \gamma_0\right|\right)^2 \delta \left(f - f_\mathrm{drive}\right)}{2 \left(\left(f_{c,0} + f \mathrm{Im}(\gamma/\gamma_0) - f^2/f_{m, 0}\right)^2 + \left(f \mathrm{Re}(\gamma / \gamma_0)\right)^2\right)} \tag{$\mathrm{m^2/Hz}$}
+
+We can also include a distance to the surface like before.
+This results in an expression for the drag coefficient :math:`\gamma` that depends on the distance to the surface which is given by the same equations as listed in the section on the hydrodynamically correct model.
```
JoepVanlier, line 82: Add a reference back to the hydro section?

docs/theory/diffusion/fitting.rst
```diff
@@ -0,0 +1,239 @@
+Fitting Methods
```
JoepVanlier, line 1: I find `Fitting Methods` a bit of an odd section header, since the stuff described in `Ensemble estimates` is also just a fitting method.

I feel that `Mean Squared Displacements`, `TAMSD`, `Ensemble MSD` and `Important Takeaways` are all pure background about MSDs and should probably be under a header `Mean Squared Displacements`.

Then the next three (OLS, GLS, CVE) are estimators that could have their own section.

JoepVanlier (changes requested):
Great! I really like splitting things up a bit more. These pages were definitely getting a bit overwhelming and doing this is definitely a step towards not getting that TL;DR feeling immediately.

I added a few small comments, but in addition to that, it seems that this change breaks the diffusion theory notebook.

I think it would be nice to merge the diffusion theory sections back into one notebook when the user downloads the page(s) as a notebook. For two reasons really, the first is that it is a pain when verifying whether the notebooks still work requires us to download all of them and run them. Some of them would also be kind of silly short. Second is that it's nice to not duplicate a bunch of it to make it executable.

Part of why the notebook was executable, is so that practitioners can try some numerical experiments and easily compare different approaches when using certain experimental conditions (certain track lengths, noise levels etc). I think having it as a single notebook is helpful for that.

I also think that now that it's split over sections, some more cross referencing would also be welcome.

docs/theory/diffusion/index.rst
```diff
@@ -0,0 +1,25 @@
...
+This section details some of the available methods and details some pitfalls with each of them.
+
+In this chapter, we will be making use of :doc:`simulated data<diffusion>`, so that we can study the quality of our
+estimates while knowing the ground truth. Note that the code to perform these simulation is included,
+so that you can play with them and produce similar graphs which are more tailored to your own experimental conditions.
+
+For more information on tracking the particles in a kymograph, refer to
+:doc:`Kymotracking </tutorial/kymotracking>` and specifically :ref:`the section on diffusion <kymotracking-diffusion>`
+regarding diffusion analysis on kymographs.
```
aafkevandenberg, line 14: I now also include these links in the introduction to diffusion.

PR #566, opened 2023-08-29
ci: small improvements to freezing test output

JoepVanlier:
**Why this PR?**
Just working with `pathlib.Path` provides cleaner looking code. In addition, by adding the zero-dimensional slice, we can easily serialize more complex objects to `npz` as well.


lumicks/pylake/conftest.py
```diff
@@ -182,24 +181,24 @@ def get_reference_data(reference_data, test_name=None, file_name=None):
...
-        reference_data_fn = os.path.join(reference_data_path, f"{ref_data_filename}.npz")
+        reference_data_path = calling_path / "reference_data" / calling_function
+        reference_file_path = reference_data_path / f"{ref_data_filename}.npz"
 
-        exists = os.path.exists(reference_data_fn)
-        if exists and not request.config.getoption("--update_reference_data"):
-            return np.load(reference_data_fn, allow_pickle=True)["arr_0"]
+        if reference_file_path.exists() and not request.config.getoption("--update_reference_data"):
+            with np.load(reference_file_path, allow_pickle=True) as npz_file:
+                return npz_file["arr_0"][()]
```
tobiasjj, line 191: 👍 context manager

lumicks/pylake/conftest.py
```diff
@@ -182,24 +181,24 @@ def get_reference_data(reference_data, test_name=None, file_name=None):
...
                 raise RuntimeError(
                     f"Test data for {calling_function}/{ref_data_filename} missing! Did you forget "
                     f"to check it in?"
                 )
 
-            os.makedirs(reference_data_path, exist_ok=True)
-            np.savez(reference_data_fn, reference_data)
-            print(f"\nWritten reference data {ref_data_filename} to {reference_data_fn}.")
+            reference_data_path.mkdir(parents=True, exist_ok=True)
+            np.savez(reference_file_path, reference_data)
```
tobiasjj, line 200: Why do you use `npz` as a file format when you only load the first array "arr_0" anyway? `np.save` would do the same without having to explicitly access `"arr_0" when loading.

tobiasjj (approved):
`pathlib.Path` produces indeed cleaner code. One minor comment about `npz` vs `npy`. LGTM! 🚴

lumicks/pylake/conftest.py
tobiasjj: An afterthought: You will probably want to specifically handle ragged nested sequences (e.g. [[1,2],[1,2,3]]) and explicitly convert them to object arrays, if you intend to freeze this kind of data. Otherwise, numpy will raise an `ValueError`.

JoepVanlier: > Why do you use `npz` as a file format when you only load the first array "arr_0" anyway? `np.save` would do the same without having to explicitly access `"arr_0" when loading.

My reasoning was that `npz` is a more general file format. If we ever were to switch to a compressed format or one where we do have multiple arrays, we could easily change, without requiring all old stuff to be updated. I didn't really see any downside other than the `arr_0` indexing.

> An afterthought: You will probably want to specifically handle ragged nested sequences (e.g. [[1,2],[1,2,3]]) and explicitly convert them to object arrays, if you intend to freeze this kind of data. Otherwise, numpy will raise an `ValueError`.

Yes, that is a good point. I'll add some handling to try to convert it to a `numpy` array, and if that fails, convert it to an object array.

PR #567, opened 2023-08-30
ci: convert convertible list of lists to `ndarray`

JoepVanlier:
**Why this PR?**
Non-ragged lists should just be converted to `ndarray`, since it facilitates convenient testing. If you don't do this, then you can't really use `np.testing.assert_allclose` conveniently, since you'll always explicitly have to do this step.


tobiasjj (approved):
As discussed. 🚴

PR #568, opened 2023-08-30
force calibration: new lower bound for f-diode

rjmoerland:
*Why*
During testing of robust fitting on axial force data I sometimes received a "division by zero" exception (or equivalent, can't fully remember). I traced it back to the lower bound of `f_diode`, which is zero. I think a still very reasonable lower bound is 1 Hz, one that is very unlikely to actually occur in practice but prevents 1/0.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -6,6 +6,7 @@
 
 * Fixed a bug that prevented resaving a `KymoTrackGroup` loaded from an older version of Pylake.
 * Fixed a bug that inadvertently made us rely on `cachetools>=5.x`. Older versions of `cachetools` did not pass the instance to the key function resulting in a `TypeError: key() missing 1 required positional argument: '_'` error when accessing cached properties or methods.
+* Fixed a bug that could cause a division by zero while fitting power spectra with `f_diode`. The lower bound of `f_diode` was changed from 0 to 1.0 Hz. This change should not impact results unless users were getting failed calibrations with this error.
 
 ## v1.2.0 | 2023-08-15
 
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -161,7 +161,7 @@ class DiodeModel(FilterBase):
                 description="Diode low-pass filtering roll-off frequency",
                 unit="Hz",
                 initial=14000,
-                lower_bound=0.0,
+                lower_bound=1.0,
                 upper_bound=_nyquist,
             ),
             Param(


JoepVanlier (approved):
Single character PR! :D

Nice catch! Seems reasonable to have a lower bound of 1 Hz.

I would consider this a bugfix; which means it should probably target the release branch (`release/1.2`). We typically merge bug-fixes there and merge those back to `main` when we do a patch release or when we release from `main`.

Just out of curiosity, why were you fitting axial force with a diode model? I thought those were fast sensors?

rjmoerland:
Ok, I changed the target to `release/1.2` :)

> Just out of curiosity, why were you fitting axial force with a diode model? I thought those were fast sensors?

Honestly I think I was just mucking about to get the fit to work. If I recall correctly, the data is for a larger bead, but we don't have the hydrodynamically correct model for Z, so I ended up abusing the `f_diode` for it. Does that make sense?

Shall I update the `changelog.md` as well?

JoepVanlier:
> Ok, I changed the target to `release/1.2` :)

Cool, thanks!

> > Just out of curiosity, why were you fitting axial force with a diode model? I thought those were fast sensors?
> 
> Honestly I think I was just mucking about to get the fit to work. If I recall correctly, the data is for a larger bead, but we don't have the hydrodynamically correct model for Z, so I ended up abusing the `f_diode` for it. Does that make sense?

It does to me. It's what they do in `Schäffer, E., Nørrelykke, S. F., & Howard, J. (2007). Surface forces and drag coefficients of microspheres near a plane surface measured with optical tweezers. Langmuir, 23(7), 3654-3665.` for cases where they don't have an accurate model as well (see section `Data Acquisition and Fitting`). Note that they use pretty small beads though. The alternative is to limit the upper frequency range (which is what we end up doing for axial currently, since phenomenological models were somewhat controversial).

For lateral calibration, we typically find that a phenomenological extra roll-off is better than not having the extra roll-off at all. I wouldn't be surprised if the same is true for axial (even though we can't really test it due to the absence of a frequency dependent model). If the actual fit is bad, the calibration is likely to be inaccurate, since it starts shifting the diode frequency to make up for the misfit.

> Shall I update the `changelog.md` as well?

Yes please. Just add a bug-fix entry. Make sure you mention that this shouldn't have an impact on results unless they were already seeing failed calibrations (so people don't worry too much about their existing calibrations changing or being bad).

agom (approved):
The elusive single character PR!

tobiasjj (approved):
🚴 🥳 Unbelievable!

rpauszek (approved):
exemplary work

PR #570, opened 2023-09-15
Add custom mapping of photon counter channels to RGB colors

agom:
This allows reconstruction of images from systems where the photon detector channels are not named with the standard "Red, "Green", and "Blue" names.

For example, this image is reconstructed from a file where the photon count channels are named `Detector 1`, `Detector 2`, `Detector 3` and `Detector 4`. 

output

Draft because needs additional tests.

### DIFF
--- a/lumicks/pylake/tests/test_file/test_file.py
+++ b/lumicks/pylake/tests/test_file/test_file.py
@@ -461,3 +461,13 @@ def test_h5_cropped_export_confocal(tmpdir_factory):
                 "PointScan1" not in f["Point Scan"]
                 or not f["Point Scan"]["PointScan1"].red_photon_count
             )
+
+
+def test_detector_mapping(h5_custom_detectors):
+    f = pylake.File.from_h5py(
+        h5_custom_detectors,
+        rgb_to_detectors={"Red": "Detector 1", "Green": "Detector 2", "Blue": "Detector 3"},
+    )
+    with pytest.raises(Exception):
+        f = pylake.File.from_h5py(h5_custom_detectors, match="Invalid RGB to detector mapping")
+    assert np.any(f.red_photon_count.data)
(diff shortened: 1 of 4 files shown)


JoepVanlier (changes requested):
I know it's still a draft, but I commented a few things I noticed.

I don't see the added value of using the enumerator `class` here. For public API, I feel this would only add unnecessary friction. Internally, we could use it, but then it would involve conversions on API endpoints such as `get_image`, and I'm not sure if it would add enough for that to be worth it?

I cannot comment on those particular lines, but note that `File.from_h5py` bypasses `__init__` and therefore needs a valid `rgb_to_detectors` to be defined.

I would also prefer to have some input validation when users provide these custom channels. Checking whether those datasets exist (if specified) and failing early seems better and more helpful than just silently returning empty channels and blank images later.

agom:
> I know it's still a draft, but I commented a few things I noticed.
> 
> I don't see the added value of using the enumerator `class` here. For public API, I feel this would only add unnecessary friction. Internally, we could use explicit enumeration, but then it would involve conversions on API endpoints such as `get_image`, and I'm not sure if it would add enough for that to be worth it? To me, it seems like just using a `str` -> `str` mapping is simpler.
> 
> I cannot comment on those particular lines, but note that `File.from_h5py` bypasses `__init__` and therefore needs a valid `rgb_to_detectors` to be defined.
> 
> I would also prefer to have some input validation when users provide these custom channels. Checking whether those datasets exist (if specified) and failing early seems better and more helpful than just silently returning empty channels and blank images later.

Thanks for the review! I was already on the fence with the `enum` so I'll just remove it, hopefully, I can come back to this in next few days.

lumicks/pylake/tests/test_file/test_file.py
```diff
@@ -461,3 +461,12 @@ def cropped_h5(name, crop=None):
                 "PointScan1" not in f["Point Scan"]
                 or not f["Point Scan"]["PointScan1"].red_photon_count
             )
+
+
+def test_detector_mapping(h5_custom_detectors):
+    f = pylake.File.from_h5py(
+        h5_custom_detectors,
+        rgb_to_detectors={"Red": "Detector 1", "Green": "Detector 2", "Blue": "Detector 3"},
+    )
```
JoepVanlier, line 470:
```suggestion
    )
    assert np.any(f.red_photon_count.data)
```
Maybe verify that we actually have the data present (I could still remove the mapping entirely without triggering an error).

JoepVanlier (approved):
Looks good! A few minor suggestions, but other than that :shipit:

PR #575, opened 2023-09-26
tracking widget: fix width of kymograph part of the widget layout

JoepVanlier:
**Why this PR?**
Workaround for https://github.com/jupyter-widgets/ipywidgets/issues/3822

Without this PR, the slider part of the widget ends up so narrow that it is invisible. Initially I thought `ipywidgets` was not installed correctly, but it turned out that the `output` widget was just eating up all the screen space.


mikhas (approved):

rpauszek (approved):
go fer it

PR #576, opened 2023-09-26
kymo: make single line scan return a valid `line_time_seconds`

JoepVanlier:
**Why this PR?**
Prior to this change, `Kymo`s consisting of a single or partial scan line provide no `line_time_seconds` which prevents plotting and other functionality. Now they provide the line time of what is actually shown in the plot (e.g. a full line, without deadtime if the line time is smaller or equal than a single line).

The second issue is that for a few Bluelake versions, Bluelake put some kymographs that were taken with a fixed line count in the `Scan` field. The last commit contains a fallback method so that they will not be loaded as corrupted `Scan` items, but end up in the `kymos` field.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,7 @@
 #### Improvements
 
 * Kymographs consisting of a single scan line now return a valid `line_time_seconds`. This allows certain downstream functionality, such as `Kymo.plot()`.
+* Issue a more descriptive error when attempting to compute a diffusion constant of a track with no points.
 
 ## v1.2.1 | 2023-10-17
 
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,10 @@
   * Shortcuts `"r"`, `"g"`, and `"b"` can now be used for plotting single color channels in addition to `"red"`, `"green"`, and `"blue"`.
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
 
+#### Improvements
+
+* Kymographs consisting of a single scan line now return a valid `line_time_seconds`. This allows certain downstream functionality, such as `Kymo.plot()`.
+
 ## v1.2.1 | 2023-10-17
 
 #### Bug fixes
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@
 
 * Kymographs consisting of a single scan line now return a valid `line_time_seconds`. This allows certain downstream functionality, such as `Kymo.plot()`.
 * Issue a more descriptive error when attempting to compute a diffusion constant of a track with no points.
+* Pylake can now handle kymographs that were erroneously stored in the `Scan` field. Kymographs with a pre-specified number of lines to record were incorrectly being marked on the timeline and exported as 'Scan' instead of 'Kymograph' in versions of Bluelake prior to `2.5.0`.
 
 ## v1.2.1 | 2023-10-17
 
--- a/lumicks/pylake/detail/image.py
+++ b/lumicks/pylake/detail/image.py
@@ -180,9 +180,11 @@ def first_pixel_sample_indices(infowave):
     if infowave.size == 0:
         return 0, 0
 
-    return np.argmax(infowave != InfowaveCode.discard), np.argmax(
-        infowave == InfowaveCode.pixel_boundary
-    )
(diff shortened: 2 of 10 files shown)


lumicks/pylake/detail/image.py
```diff
@@ -180,9 +180,11 @@ def first_pixel_sample_indices(infowave):
     if infowave.size == 0:
         return 0, 0
 
-    return np.argmax(infowave != InfowaveCode.discard), np.argmax(
-        infowave == InfowaveCode.pixel_boundary
-    )
+    pixel_boundary = np.argmax(infowave == InfowaveCode.pixel_boundary)
+    if infowave[pixel_boundary] != InfowaveCode.pixel_boundary:
+        raise RuntimeError("No completed pixel found in Scan")
```
rpauszek, line 185:
```suggestion
        raise RuntimeError("No completed pixel found in image")
```
maybe don't use "Scan" so as not to confuse with the class

lumicks/pylake/kymo.py
```diff
@@ -47,17 +38,17 @@ def _default_line_time_factory(self: "Kymo"):
 
             # Special case for no dead time. Is it really no deadtime, or a single line?
             if dead_time == 0 and not np.any(beyond_first_line):
-                raise RuntimeError(single_line_error)
+                return scan_time * infowave._src.dt * ns_to_sec
 
         except ValueError:  # ValueError occurs when beyond_first_line is empty
-            raise RuntimeError(single_line_error)
+            return scan_time * infowave._src.dt * ns_to_sec
```
rpauszek, line 44: I can't suggest on the whole block, but it seems that the `try ... except` was there mostly to re-throw a more descriptive error. Since we don't need to do that now, we could move the early out up and save some indentation

Personally, I find it cleaner, for lines 35-44:
```
beyond_first_line = infowave.data[start + scan_time :] != InfowaveCode.discard
if not np.any(beyond_first_line):
    return scan_time * infowave._src.dt * ns_to_sec

# Special case for no dead time. Is it really no deadtime, or a single line?
if (dead_time := np.argmax(beyond_first_line)) == 0:
    return scan_time * infowave._src.dt * ns_to_sec
```

I tried it and all of the tests pass

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -735,6 +735,12 @@ def determine_optimal_points(frame_idx, coordinate, max_iterations=100):
 
         num_slopes.add(num_slope)
 
+        if len(coordinate) <= 4:
+            raise RuntimeError(
+                "You need at least 5 time points to estimate the number of points to include in "
+                "the fit."
+            )
```
rpauszek, line 742: don't we have this check already somewhere?

rpauszek (approved):
good stuff, it's nice to handle our mistakes graciously 😅  just a left a couple minor comments for clarity

lumicks/pylake/kymo.py
JoepVanlier: Yeah, that's fair. I'd rather check for `len(beyond_first_line)` than using `np.any`, since that's the only case we need to be covering (to prevent `argmax` from getting an empty array).

changelog.md
```diff
@@ -7,6 +7,12 @@
 * Added more options for plotting color channels for images:
   * Shortcuts `"r"`, `"g"`, and `"b"` can now be used for plotting single color channels in addition to `"red"`, `"green"`, and `"blue"`.
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
+* Handle kymographs that were accidentally stored in the `Scan` field.
```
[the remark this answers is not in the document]
JoepVanlier: Good call on adding the version number!

And yeah that's fair, it's an improvement. I feel it's not necessarily a bugfix on pylake's part.

lumicks/pylake/kymotracker/detail/msd_estimation.py
JoepVanlier: We do have a similar check elsewhere, but you don't even get to that check if you run diffusion estimation on an empty track. Instead you are greeted by a `TypeError: expected non-empty vector for x` which is not helpful.

PR #579, opened 2023-10-09
Fix tracks csv export

rpauszek:
**Why this PR?**
Currently the minimum length field is formatted as `int` but we need `float`. Bug was discovered when importing from CSV and doing dwelltime analysis with new options results in an error (discretization time (line time) is larger than the minimum duration `0`)


JoepVanlier (approved):
Good catch! And what an embarrassing error.

One minor suggestion, namely that the error is caused on saving, rather than importing. It's important that people realize that they might have to verify files they saved in the meantime.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -213,8 +215,25 @@ def create_track(time, coord, min_length=None):
...
+        warnings.warn(
+            RuntimeWarning(
+                "This CSV file is from a version in which the minimum observable track duration "
+                "was incorrectly rounded to an integer. This may result in errors in certain "
+                "analyses. To remedy this error, use `lk.filter_tracks()` on the "
+                "tracks from the widget with the desired minimum track length and resave "
+                "using the `KymoTrackGroup.save()` method. Note that the minimum length should be "
+                "greater or equal to the minimum length used for the original tracking. For more "
+                "information refer to "
+                "https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html#migrating-old-track-files "
```
JoepVanlier, line 228: Feel free to ignore, but:
```suggestion
                "https://lumicks-pylake.readthedocs.io/en/latest/tutorial/nbwidgets.html#migrating-old-track-files"
```

JoepVanlier (approved):
Looks good.

PR #580, opened 2023-10-10
Update update instructions install.rst

aafkevandenberg:
I got the feedback that the installation instructions are super easy, but the update instructions are not so clear, so I made the steps more explicit


JoepVanlier (approved):
👍

PR #582, opened 2023-10-16
Fixup zenodo

rpauszek:
**Why this PR?**
Apparently Zenodo recently changed their REST API. These are corrections for as best I can tell what we should expect now


JoepVanlier (approved):
Looks good!

PR #585, opened 2023-10-25
force_calibration: coupling model for active calibration

JoepVanlier:
**Why this PR?**
When performing active calibration, we get a smaller velocity around the beads than expected when dealing with a dual trap configuration. This can lead to biases when not taken into account in the model that predicts the oscillation amplitude of the beads:

$R_{d, corrected} = c R_d$
$R_{f, corrected} = \frac{R_f}{c}$
$\kappa_{corrected} = \frac{\kappa}{c^2}$

correction_factor_whatsnew

Various models exist to calculate the correction factor `c`. In the context of optical tweezers, I have seen Oseen (acting force being a point function) and Rotne Prager used [1]. For us, this seems to correct well for large distances, but seems to fail at short distances*. [2] seems to fare better for both regimes, but only covers oscillations parallel to the bead axes. A fraction with a reciprocal polynomial for oscillations perpendicular to the bead axes can be found in [3].

Based on these two models, we can compute a combined model (see the documentation for more information on this). If we use this to correct data over the entire trap range, we see that this corrects the calibration factors pretty nicely. See graphs below.

Docs here: [whats new](https://lumicks-pylake.readthedocs.io/en/coupling/whatsnew/index.html), [theory](https://lumicks-pylake.readthedocs.io/en/coupling/theory/force_calibration/active.html#bead-bead-coupling), [tutorial](https://lumicks-pylake.readthedocs.io/en/coupling/tutorial/force_calibration.html#active-calibration-with-two-beads-far-away-from-the-surface) and [example](https://lumicks-pylake.readthedocs.io/en/coupling/examples/bead_coupling/coupling.html#).

Note: I have marked this functionality as alpha, since testing has been somewhat limited so far. That said, for conditions where they are valid, the results obtained with these models is much more reasonable than without.

Figure 22
_Calibrations are obtained by putting trap 2 in a static location and moving trap 1 in the x-direction. We are showing results for trap 2 here, since there is an unexplained additional dependence of the sensitivity on the x position for trap 1. Note that we have used the Stimson model for horizontal oscillations and the Wakiya approximation model as presented by Goldmann model for Y oscillation here._

X-oscillation-XY
_Active calibration with two 4.38 um beads (X). Here we moved a second bead around in the trap range of motion and calibrating the bead that was at a fixed position._
Y-oscillation-XY
_Active calibration with two 4.38 um beads (Y). Here we moved a second bead around in the trap range of motion and calibrating the bead that was at a fixed position._

[1] https://core.ac.uk/download/pdf/270293467.pdf
[2] Stimson, M., & Jeffery, G. B. (1926). The motion of two spheres in a viscous fluid. Proceedings of the Royal Society of London. Series A, Containing Papers of a Mathematical and Physical Character, 111(757), 110-116 (2007).
[3] Goldman, A. J., Cox, R. G., & Brenner, H. (1966). The slow motion of two identical arbitrarily oriented spheres through a viscous fluid. Chemical Engineering Science, 21(12), 1151-1170.

* Note that [2] contains a typographical error in equation 37, where it should read `4/3` and not `2/3`
* Note that the value reported in [2] for `alpha=0.5` should be `0.6596`.

### DIFF
--- a/docs/theory/force_calibration/active.rst
+++ b/docs/theory/force_calibration/active.rst
@@ -92,6 +92,8 @@ We can also include a distance to the surface like before. This results in an ex
 coefficient :math:`\gamma` that depends on the distance to the surface which is given by the same
 equations as listed in the section on the :doc:`hydrodynamically correct model<hyco>`.
 
+.. _bead_bead_theory:
+
 Bead-bead coupling
 ------------------
 
--- a/docs/theory/force_calibration/active.rst
+++ b/docs/theory/force_calibration/active.rst
@@ -91,3 +91,111 @@ hydrodynamically correct model for the peak:
 We can also include a distance to the surface like before. This results in an expression for the drag
 coefficient :math:`\gamma` that depends on the distance to the surface which is given by the same
 equations as listed in the section on the :doc:`hydrodynamically correct model<hyco>`.
+
+Bead-bead coupling
+------------------
+
+.. warning::
+
+    The implementation of the coupling correction models is still alpha functionality.
+    While usable, this has not yet been tested in a large number of different scenarios.
+
+The active calibration method presented in the previous sections relies on oscillating the nanostage with a known amplitude and frequency.
+The fluid in the flow-cell follows the stage motion.
+This in turn exerts a drag on the bead that leads to a sinusoidal displacement of the bead from the trap center.
+The amplitude of the detected displacement (measured in Volts) and the stage amplitude are then quantified.
+From the stage amplitude (measured in microns, since the stage position is calibrated) an expected bead displacement is calculated.
+
+When using two beads, the flow field around the beads is reduced (because the presence of the additional bead slows down the fluid).
+The magnitude of this effect depends on the bead diameter, distance between the beads and their orientation with respect to the fluid flow.
+Streamlines for some bead configurations are shown below (simulated using FEniCSx :cite:`the_fenics_project_developers_2023_10432590`).
+
+.. image:: figures/streamlines.png
+  :nbattach:
+
+As a result, the bead moves less than expected for a given stage motion.
+
+Since the displacement sensitivity (microns/V) is given by the ratio of the expected bead displacement (in microns) to detected displacement (in Volts) and we detected less displacement than expected (lower voltage amplitude), we obtain an artificially higher displacement sensitivity than expected.
+
+If we define a factor :math:`c` by which the velocity is reduced, we obtain the following relations for correcting for this reduced flow field:
+
+.. math::
+
+    \begin{align}
+    R_{d, corrected} & = c R_d\\
+    R_{f, corrected} & = \frac{1}{c} R_f\\
+    \kappa_{corrected} & = \frac{1}{c^2}\kappa
+    \end{align}
+
+Where :math:`R_d` is the displacement sensitivity, :math:`R_f` is the force sensitivity and :math:`\kappa` is the stiffness.
+As shown in the plot below, failing to account for this effect can result in substantial calibration error.
+
+.. image:: figures/errors.png
+  :nbattach:
+
+To calculate the desired correction factor :math:`c`, we need to determine what happens to the fluid around the beads.
+Considering the fluid velocity and viscosity, we can conclude that we typically operate in the regime where viscous effects are dominant (creeping flow).
+This can be checked by calculating the Reynolds number for the flow.
+Filling in the maximal velocity we expect during the oscillation, we find the following expression.
+
+.. math::
+
+    Re = \frac{\rho u L}{\eta} = 2 \pi f A d \frac{\rho}{\eta}
+
+Here :math:`\rho` refers to the fluid density, :math:`u` the characteristic velocity, :math:`L` the characteristic length scale, :math:`\eta` the viscosity, :math:`f` the oscillation frequency, :math:`A` the oscillation amplitude and :math:`d` the bead diameter.
+For microfluidic flow, this value is typically much smaller than `1`.
+
+In this limit, the Navier-Stokes equation reduces to the following expressions:
+
+.. math::
+
+    \begin{align}
+    \nabla^2 v & = \frac{1}{\eta} \nabla p \\
+    \nabla \cdot v & = 0
+    \end{align}
+
+Here :math:`\eta` is the viscosity, :math:`p` is the pressure and :math:`v` is the fluid velocity.
+Creeping flow is far removed from every day intuition as it equilibrates instantaneously.
+The advantage of this is that for sufficiently low frequencies, the correction factor can be based on the correction factor one would obtain for a steady state constant flow.
+
+For two beads aligned in the flow direction, we can use the analytical solution presented in :cite:`stimson1926motion`.
+This model uses symmetry considerations to solve the creeping flow problem for two solid spheres moving at a constant velocity parallel to their line of centers.
+We denote the correction factor obtained from this model as :math:`c_{\|}`.
+This correction factor is given by the ratio of the drag coefficient by the drag coefficient one would expected from a single bead in creeping flow (:math:`3 \pi \eta d v`).
+For beads aligned perpendicular to the flow direction, we use a model from :cite:`goldman1966slow`, which we denote as :math:`c_{\perp}`.
+
+From the derivations in these papers, it follows that the correction factors obtained depend on the bead diameter(s) :math:`d` and distance between the beads :math:`l`.
+For equally sized beads, this dependency is a function of the ratio of the distance between the beads over the bead diameter.
+
+Considering the linearity of the equations that describe creeping flow :cite:`goldman1966slow`, we can combine the two analytical solutions by decomposing the incoming velocity (in the direction :math:`\vec{e}_{osc}`) into a velocity perpendicular to the bead-to-bead axis :math:`\vec{e}_{\perp}` and a velocity component aligned with the bead-to-bead axis :math:`\vec{e}_{\|}`.
+
+.. math::
+
+    \begin{align}
+    v_{\|} & = (\vec{e}_{\|} \cdot\vec{e}_{osc}) c_{\|}\\
+    v_{\perp} & = (\vec{e}_{\perp} \cdot \vec{e}_{osc})  c_{\perp}
+    \end{align}
+
+This provides us with contributions for each of those axes, but we still need to project this back to the oscillation axis (since this is where we measure our amplitude).
+We can calculate our desired hydrodynamic correction factor as:
+
+.. math::
+
+    c_{total} = v_{\|} (\vec{e}_{\|} \cdot \vec{e}_{osc}) + v_{\perp} (\vec{e}_{\perp} \cdot \vec{e}_{osc})
+
+The response of this combined model for equally sized beads can be calculated as follows::
+
+    diameter = 1.0
+    l_d = np.arange(1.01, 8, 0.1) * diameter
+    zeros = np.zeros(l_d.shape)
+    plt.plot(l_d, lk.coupling_correction_2d(l_d, zeros, diameter, is_y_oscillation=False), label="horizontal alignment [Stimson et al]")
+    plt.plot(l_d, lk.coupling_correction_2d(zeros, l_d, diameter, is_y_oscillation=False), label="vertical alignment [Goldman et al]")
+    plt.plot(l_d, lk.coupling_correction_2d(l_d / np.sqrt(2), l_d / np.sqrt(2), diameter, is_y_oscillation=False), label="diagonal alignment")
+    plt.ylabel('Correction factor [-]')
+    plt.xlabel("l/d [-]")
+    plt.legend()
+
+.. image:: figures/correction_factor.png
+
+Here, when providing only a horizontal distance recovers the Stimson model :cite:`stimson1926motion`, while a vertical displacement recovers the Goldman model :cite:`goldman1966slow`.
+To find out more about how to use these correction factors, please refer to the :ref:`tutorial<bead_bead_tutorial>`.
--- a/docs/theory/force_calibration/active.rst
+++ b/docs/theory/force_calibration/active.rst
@@ -95,7 +95,7 @@ equations as listed in the section on the :doc:`hydrodynamically correct model<h
 .. _bead_bead_theory:
 
 Bead-bead coupling
-------------------
+^^^^^^^^^^^^^^^^^^
 
 .. warning::
 
@@ -147,7 +147,7 @@ Filling in the maximal velocity we expect during the oscillation, we find the fo
 Here :math:`\rho` refers to the fluid density, :math:`u` the characteristic velocity, :math:`L` the characteristic length scale, :math:`\eta` the viscosity, :math:`f` the oscillation frequency, :math:`A` the oscillation amplitude and :math:`d` the bead diameter.
 For microfluidic flow, this value is typically much smaller than `1`.
 
-In this limit, the Navier-Stokes equation reduces to the following expressions:
+In this limit, the Navier-Stokes equation describing fluid flow reduces to the following expressions:
 
 .. math::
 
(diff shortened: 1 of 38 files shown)


docs/tutorial/force_calibration.rst
```diff
@@ -396,6 +396,44 @@ However, if we do not provide the height above the surface, we can see that the
...
+    Rf_corrected = calibration["Rf"].value / factor
+    stiffness_corrected = calibration["kappa"].value / factor**2
+
+To correct a force trace, simply divide it by the correction factor::
+
+    corrected_force1x = f.force1x / factor
+
+.. note::
+
+    This coupling model is only valid far away from the surface.
```
RoelandvanWijk, line 435: @JoepVanlier 50 um up or only in the middle?

RoelandvanWijk (approved):
Looks ok to me. Checked a bunch of formulas in the papers and they match, as indicated it is indeed experimental. Maybe give an indication on how far from the surface the model can be applied so users have something to go by. Now it said "very far" from the surface (check the comment I left).

docs/tutorial/force_calibration.rst
JoepVanlier: I don't expect there to be a sharp cutoff point. My guess it will probably depend on the ratio of bead radius to distance like Faxen's law does.

My guess is that there is some regime where you might get away with using Faxen's law and this model together, but I don't know how far you have to be for this to be a reasonable model.

Maybe we should rephrase it as, `This coupling model neglects effects from the surface.` to basically state what it does. Right now we don't have enough information to give more user guidance than that.

JoepVanlier: I have rephrased it as: `This coupling model neglects effects from the surface. It is intended for measurements performed at the center of the flowcell.`

If I have some time, I will look at this a bit more with FEM.

JoepVanlier: I compared it a bit with some FEM simulations near the surface. While it certainly isn't perfect, it seems that a superposition of Faxen and this model is still a lot better than doing nothing. That said, to stay on the safe side for now, I would only recommend it for bulk, and recommend people not to use it yet near the surface. I am not sure how many people would be doing two-bead assays close to the surface, so I'm not sure whether there is a use-case for it.

image

rjmoerland:
It is my intention to look at this more closely, but for now a question: have you tried to correct the experimental active calibration with correction factors in Y obtained through FEM, and if so, did it get better than what you have now? Don't get me wrong, the results are highly convincing, but you mentioned the presence of an apparent offset for the Y calibration. I'm wondering if that offset disappears. I'm also wondering how much memory and time it took to do a single one of those calculations - depending on that , one might even consider doing a full FEM calculation for the exact geometry that a user has (sounds unlikely, but won't hurt to ask ;) )

JoepVanlier:
> It is my intention to look at this more closely, but for now a question: have you tried to correct the experimental active calibration with correction factors in Y obtained through FEM, and if so, did it get better than what you have now? Don't get me wrong, the results are highly convincing, but you mentioned the presence of an apparent offset for the Y calibration. I'm wondering if that offset disappears. I'm also wondering how much memory and time it took to do a single one of those calculations - depending on that , one might even consider doing a full FEM calculation for the exact geometry that a user has (sounds unlikely, but won't hurt to ask ;) )

I will try that next time I have some time to work on this stuff. I don't think implementing a FEM approach for calculating these factors is unreasonable (especially if just creeping Stokes flow suffices), but it brings a lot of difficult questions regarding how to package it and make sure there aren't many unexpected failure modes (in which it suddenly takes much longer to calculate or doesn't converge for whatever reason).

The one I used was too slow, but I also made no real effort to try and optimize it for performance with better solvers and/or use of MP.

rjmoerland:
In addition to the above (thank you for the extensive reply), did you investigate how a difference in bead height affects the calibration? Based on my [calculations](https://lumicks.atlassian.net/wiki/spaces/OCC/pages/2364441913/Trap+stiffnesses+versus+bead+size+for+PS+beads), two beads of nominally 4.4 um may differ in height by about 1 um, depending on the exact diameter. I'm wondering what that would do 🤔

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -175,6 +180,67 @@ def viscosity_of_water(temperature, molarity_nacl=None, pressure=None):
...
+    perpendicular and aligned with the bead-to-bead axis. For the aligned component, we use the
+    model presented in [1]_ to determine a correction factor, while for the perpendicular axis
+    we use the model presented in [2]_, [3]_ and [4]_.
+
+    Parameters
+    ----------
+    dx, dy: array_like or float
+        X and Y distances from bead 1 to bead 2 [um].
+    bead_diameter : float
+        Bead radius [um].
```
facutuesca, line 196:
```suggestion
        Bead diameter [um].
```

rjmoerland (approved):
I think it looks very good, and I love the look and feel of the images for the documentation. I think it's good that it's stressed that this is alpha functionality, but having this merged also allows us to test the functionality more easily. If not there, I'd still would recommend to warn users that differences of bead height (for whatever reason) are not taken into account because the assumption is that everything happens in the same plane.

docs/theory/force_calibration/active.rst
```diff
@@ -91,3 +91,94 @@ hydrodynamically correct model for the peak:
...
+Considering the fluid velocity and viscosity, we can conclude that we typically operate in the regime where viscous effects are dominant (creeping flow).
+This can be checked by calculating the Reynolds number for the flow.
+Filling in the maximal velocity we expect during the oscillation, we find the following expression.
+
+.. math::
+
+    Re = \frac{\rho u L}{\mu} = 2 \pi f A d \frac{\rho}{\mu}
+
+Here :math:`\rho` refers to the fluid density, :math:`u` the characteristic velocity, :math:`L` the characteristic length scale and :math:`\mu` the viscosity.
+For microfluidic flow, this value is typically much smaller than `1`.
```
aafkevandenberg, line 131: And f, A and d are....

docs/theory/force_calibration/active.rst
```diff
@@ -91,3 +91,94 @@ hydrodynamically correct model for the peak:
...
+    l_d = np.arange(1.01, 8, 0.1) * diameter
+    zeros = np.zeros(l_d.shape)
+    plt.plot(l_d, lk.coupling_correction_2d(l_d, zeros, diameter, is_y_oscillation=False), label="horizontal alignment [Stimson et al]")
+    plt.plot(l_d, lk.coupling_correction_2d(zeros, l_d, diameter, is_y_oscillation=False), label="vertical alignment [Goldman et al]")
+    plt.plot(l_d, lk.coupling_correction_2d(l_d / np.sqrt(2), l_d / np.sqrt(2), diameter, is_y_oscillation=False), label="diagonal alignment")
+    plt.ylabel('Correction factor [-]')
+    plt.xlabel("l/D [-]")
+    plt.legend()
+
+.. image:: figures/correction_factor.png
```
aafkevandenberg, line 181: The diameter, D is not defined in the text. And it is also not clear from the text how the correction factor is a function if I/D

docs/theory/force_calibration/active.rst
```diff
@@ -91,3 +91,94 @@ hydrodynamically correct model for the peak:
...
+From the stage amplitude (measured in microns, since the stage position is calibrated) an expected bead displacement is calculated.
+
+When using two beads, the flow field around the beads is reduced (because the presence of the additional bead slows down the fluid).
+The magnitude of this effect depends on the bead diameter, distance between the beads and their orientation with respect to the fluid flow.
+Streamlines for some bead configurations are shown below (simulated using FEniCSx :cite:`the_fenics_project_developers_2023_10432590`).
+
+.. image:: figures/streamlines.png
+  :nbattach:
+
+As a result, the bead moves less than expected for a given stage motion.
```
aafkevandenberg, line 118: Do you have an idea of how large the error can be when the bead-bead coupling is not accounted for? Is this discussed in one of the papers you are citing?

aafkevandenberg (commented):
Nice, I learned something new.
Added a few comments and suggestions.

docs/theory/force_calibration/active.rst
```diff
@@ -91,3 +91,94 @@ hydrodynamically correct model for the peak:
...
+In this limit, the Navier-Stokes equation describing fluid flow reduces to the following expressions:
+
+.. math::
+
+    \begin{align}
+    -\nabla \cdot ( \nabla v + pI ) & = 0\\
+    \nabla \cdot v & = 0
+    \end{align}
+
+Here :math:`p` is the pressure and :math:`v` is the fluid velocity.
```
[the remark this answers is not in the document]
JoepVanlier: `I` is the identity matrix. I can rewrite it a little bit to avoid having that in there.

docs/theory/force_calibration/active.rst
JoepVanlier: These were all defined above. But yeah, it's probably not a bad idea to not assume people having read the preceding section. Makes it nicer to read only this section.

docs/theory/force_calibration/active.rst
JoepVanlier: It was defined in the first section, but it doesn't hurt to redefine it per section, so I'll add it again here. I do note that for consistency, it is probably better to use small case `d` (since we use capital `D` for diffusion and that may end up being unnecessarily confusing).

docs/theory/force_calibration/active.rst
JoepVanlier: The error depends strongly on the bead-bead distance to diameter ratio. I have added a small section on the expected error if you don't take it into account.

errors

JoepVanlier:
> I think it looks very good, and I love the look and feel of the images for the documentation. I think it's good that it's stressed that this is alpha functionality, but having this merged also allows us to test the functionality more easily. If not there, I'd still would recommend to warn users that differences of bead height (for whatever reason) are not taken into account because the assumption is that everything happens in the same plane.

I will add a warning to the tutorial section about it, but I wouldn't mind doing more if you have suggestions.

```rst
.. warning::

    The model implemented here only supports beads that are aligned in the same plane.
    It does not take a mismatch in the `z`-position of the beads into account.
    In reality, the height above the surface depends on the bead radius and may be different for the two beads.
    At short bead-to-bead distances, such a mismatch could make the coupling slightly less pronounced than the model predicts.
```

I think it would be great if we had a source we could cite for this (could we cite py-optics for this?). I also think the graph you have made for the `z-position` as a function of bead diameter is very insightful as it shows that the magnitude of how much the z-position changes with bead size (the oscillations w.r.t. size).

I did some simulations (again only Stokes eq) with this (and quickly also did a prototype `coupling_correction_3d`, but based on these results I would expect the effect to be quite small in most relevant cases.

image

docs/tutorial/force_calibration.rst
```diff
@@ -400,6 +400,57 @@ However, if we do not provide the height above the surface, we can see that the
...
+.. note::
+
+    This coupling model neglects effects from the surface. It is intended for measurements performed at the center of the flowcell.
+
+.. note::
+
+    The model implemented here only supports beads that are aligned in the same plane.
+    It does not take a mismatch in the `z`-position of the beads into account.
+    In reality, the height above the surface depends on the bead radius and may be different for the two beads.
+    At short bead-to-bead distances, such a mismatch would make the coupling less pronounced than the model predicts.
```
rjmoerland, line 452:
```suggestion
    The model implemented here only supports beads that are aligned in the same plane.
    It does not take a mismatch in the `z`-position of the beads into account.
    In reality, the position in the focus depends on the bead radius and may be different for the two beads if they slightly differ in size [ref].
    At short bead-to-bead distances, such a mismatch would make the coupling less pronounced than the model predicts.
```
With [ref]: https://doi.org/10.1364/JOSAB.35.002654 (Fig. 3).

rjmoerland:
> I think it would be great if we had a source we could cite for this (could we cite py-optics for this?). I also think the graph you have made for the `z-position` as a function of bead diameter is very insightful as it shows that the magnitude of how much the z-position changes with bead size (the oscillations w.r.t. size).

I've added a comment with a reference where this phenomenon is mentioned in passing. It would be cool to cite py-optics, and I did intend to get it on Zenodo, but then got lost in the mayhem of daily affairs. It's still possible to do, and I could even create a specific example (notebook) that highlights the effect.
 
> I did some simulations (again only Stokes eq) with this (and quickly also did a prototype `coupling_correction_3d`, but based on these results I would expect the effect to be quite small in most relevant cases.
> image

According to my calculations, for polystyrene beads in water and a trapping wavelength of 1064 nm, the beads can jump around by about 1 µm for nominal bead sizes of 4-5 µm. That would mean about 0.4-0.5 times the bead radius. If I interpret your graphs correctly, then I'd agree that it barely affects the calibration. I'm not sure a 3D correction is warranted (unless we look at z forces? 😅 )

docs/tutorial/force_calibration.rst
JoepVanlier: Wonderful! Thanks. I don't think `sphinx-bibtex` supports citing figures directly in the same way latex does, but hopefully just citing it as `[...] (Fig. 3)` works well enough.

JoepVanlier:
> > I think it would be great if we had a source we could cite for this (could we cite py-optics for this?). I also think the graph you have made for the `z-position` as a function of bead diameter is very insightful as it shows that the magnitude of how much the z-position changes with bead size (the oscillations w.r.t. size).
> 
> I've added a comment with a reference where this phenomenon is mentioned in passing. It would be cool to cite py-optics, and I did intend to get it on Zenodo, but then got lost in the mayhem of daily affairs. It's still possible to do, and I could even create a specific example (notebook) that highlights the effect.

I think that reference will do for now. When you get round to adding `py-optics` and/or the notebook used to demonstrate this effect to Zenodo, I'd love to add it.

> > I did some simulations (again only Stokes eq) with this (and quickly also did a prototype `coupling_correction_3d`, but based on these results I would expect the effect to be quite small in most relevant cases.
> > image
> 
> According to my calculations, for polystyrene beads in water and a trapping wavelength of 1064 nm, the beads can jump around by about 1 µm for nominal bead sizes of 4-5 µm. That would mean about 0.4-0.5 times the bead radius. If I interpret your graphs correctly, then I'd agree that it barely affects the calibration. I'm not sure a 3D correction is warranted (unless we look at z forces? 😅 )

Yeah, I also wasn't suggesting pulling the 3D model in here. It is not really that much more complicated, but it will probably lead to people worrying about what to fill in for `z`, while the effect seems to be rather small. If we ever find a case where it matters, I think we can always add it later.

docs/tutorial/force_calibration.rst
rjmoerland: I think it will :)

rjmoerland:
> I think that reference will do for now. When you get round to adding `py-optics` and/or the notebook used to demonstrate this effect to Zenodo, I'd love to add it.

👍 

> Yeah, I also wasn't suggesting pulling the 3D model in here. It is not really that much more complicated (I basically do the same trick, where I interpret the problem as fluid moving under an angle and then project on to the relevant axes. Except, now I have one more perpendicular axis to project onto), but it will probably lead to people worrying about what to fill in for `z`, while the effect seems to be rather small. If we ever find a case where it matters, I think we can always add it later.

Totally agree, just meant to save you some work 🙂

docs/tutorial/force_calibration.rst
JoepVanlier: Sorry, I had somehow missed this comment. Yeah. For `Re << 1`, Navier Stokes reduces to just the Stokes equations and they are linear w.r.t. pressure and velocity. So yes, I would indeed expect this. But I feel the same way. We have not validated coupling behavior near the surface and I do not feel comfortable exposing that at this time.

rjmoerland: That's all right, I had forgotten about it myself 😆 
Goes on the TO DO list then :)

docs/theory/force_calibration/active.rst
aafkevandenberg: ah yes, I see they are defined in the previous section. Yes, good to repeat them here.

facutuesca:
giphy

PR #586, opened 2023-10-26
Fix fetching files from zenodo

agom:
The API changed the record lookup from using `record_id` to `id`

### DIFF
--- a/lumicks/pylake/file_download.py
+++ b/lumicks/pylake/file_download.py
@@ -120,10 +120,7 @@ def download_from_doi(doi, target_path="", force_download=False, show_progress=T
     file_names = []
     for file in record_metadata["files"]:
         file_name = file["filename"]
-        url = (
-            f"https://zenodo.org/api/records/{record_metadata['record_id']}"
-            f"/files/{file_name}/content"
-        )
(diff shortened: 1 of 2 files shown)


lumicks/pylake/file_download.py
```diff
@@ -121,8 +121,7 @@ def download_from_doi(doi, target_path="", force_download=False, show_progress=T
     for file in record_metadata["files"]:
         file_name = file["filename"]
         url = (
-            f"https://zenodo.org/api/records/{record_metadata['record_id']}"
-            f"/files/{file_name}/content"
+            f"https://zenodo.org/api/records/{record_metadata['id']}" f"/files/{file_name}/content"
         )
```
JoepVanlier, line 125: It fits on a single line now, so you could just do:
```python
url = f"https://zenodo.org/api/records/{record_metadata['id']}/files/{file_name}/content"
```

JoepVanlier (changes requested):
Looks good and passes all tests, but should target `release/1.2`.

lumicks/pylake/file_download.py
agom: `black` disagrees 🙂

JoepVanlier (approved):
:shipit:

PR #587, opened 2023-10-30
tracking: add `min_duration` argument to `lk.filter_tracks`

JoepVanlier:
**Why this PR?**
It makes more sense for users to filter by a specific track duration, rather than a number of points in a track.

*What to look out for*
It is important to set the minimum time appropriately. If we filter by length `min_duration`, then the first length we can actually observe is the next available duration `ceil(min_duration / line_time) * line_time`.

Below are the results of running dwell time analysis on the results from two simulated kymos (line time 0.25 and line time 0.1) with 1000 lines each.

If one computes the minimally observable duration correctly, the estimates are as expected:

Correctly taken into account
_Rounded to next step (ceil)_

If not, then well, we get biased results:

Incorrectly taken into account (round)
_Rounded to nearest value (round)_

Incorrectly taken into account (naive)
_Not rounded_

Note that the track durations were chosen (short compared to the line time) such that this effect would be quite prominent.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -8,6 +8,7 @@
   * Shortcuts `"r"`, `"g"`, and `"b"` can now be used for plotting single color channels in addition to `"red"`, `"green"`, and `"blue"`.
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
 * Added `duration` property to `KymoTrack` which returns the duration (in seconds) that the track was observed.
+* Added option to filter tracks by duration in seconds using `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)`.
 
 #### Improvements
 
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * Added more options for plotting color channels for images:
   * Shortcuts `"r"`, `"g"`, and `"b"` can now be used for plotting single color channels in addition to `"red"`, `"green"`, and `"blue"`.
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
+* Added `duration` property to `KymoTrack` which returns the duration (in seconds) that the track was observed.
 
 #### Improvements
 
--- a/changelog.md
+++ b/changelog.md
@@ -9,6 +9,7 @@
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
 * Added `duration` property to `KymoTrack` which returns the duration (in seconds) that the track was observed.
 * Added option to filter tracks by duration in seconds using `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)`.
+* Added `KymoTrackGroup.filter()` to filter tracks in-place. `tracks.filter(minimum_duration=2)` is equivalent to `tracks = lk.filter_tracks(tracks, minimum_duration=2)`.
 
 #### Improvements
 
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -438,6 +438,11 @@ class KymoTrack:
         return self._localization.position
 
     @property
+    def duration(self):
+        """Track duration in seconds"""
+        return self.seconds[-1] - self.seconds[0]
+
+    @property
     def _line_time_seconds(self):
         """Source kymograph line time in seconds."""
         return self._kymo.line_time_seconds
@@ -1504,7 +1509,7 @@ class KymoTrackGroup:
                 if exclude_ambiguous_dwells
                 else group
             )
-            dwelltimes_sec = np.array([track.seconds[-1] - track.seconds[0] for track in tracks])
+            dwelltimes_sec = np.array([track.duration for track in tracks])
             nonzero_dwelltimes_sec = dwelltimes_sec[dwelltimes_sec > 0]
             removed_zeros = removed_zeros or len(nonzero_dwelltimes_sec) != len(dwelltimes_sec)
 
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -1277,6 +1277,22 @@ class KymoTrackGroup:
 
         self._src = [track for track in self._src if not track.in_rect(rect, all_points)]
 
+    def filter(self, *, minimum_length=1, minimum_duration=0):
+        """Remove tracks shorter than specified criteria from the list.
+
+        Parameters
+        ----------
+        minimum_length : int, optional
+            Minimum number of tracked points for the track to be accepted (default: 1).
+        minimum_duration : seconds, optional
+            Minimum duration in seconds for a track to be accepted (default: 0).
+        """
+        from .kymotracker import filter_tracks  # local import to avoid circular import
+
+        self._src = filter_tracks(
+            self, minimum_length=minimum_length, minimum_duration=minimum_duration
+        )._src
+
     def __repr__(self):
         return f"{self.__class__.__name__}(N={len(self)})"
 
--- a/lumicks/pylake/kymotracker/kymotracker.py
+++ b/lumicks/pylake/kymotracker/kymotracker.py
@@ -367,8 +367,8 @@ def track_lines(
     return refine_tracks_centroid(kymotrack_group, track_width=line_width, bias_correction=True)
 
 
-def filter_tracks(tracks, minimum_length):
-    """Remove tracks shorter than a minimum number of time points from the list.
+def filter_tracks(tracks, minimum_length=1, *, minimum_duration=0):
+    """Remove tracks shorter than specified criteria from the list.
 
     This can be used to enforce a minimum number of frames a spot has to be detected in order
     to be considered a valid track.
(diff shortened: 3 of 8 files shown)


lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -367,7 +367,7 @@ def track_lines(
     return refine_tracks_centroid(kymotrack_group, track_width=line_width, bias_correction=True)
 
 
-def filter_tracks(tracks, minimum_length):
+def filter_tracks(tracks, minimum_length, *, minimum_duration=0):
```
rpauszek, line 370: this seems a bit odd to me to enforce a minimum number of frames on top of a minimum duration. Filtering by duration seems more natural and probably what you most often want, but how it's set up here you necessarily also still need to think about the number of tracked frames.

I could see multiple options to remedy this and give enough flexibility for users:
- add a default for `minimum_length=1`. This way you can do `filter_tracks(minimum_duration=x)` and it would work as you expect. You could also then choose to filter by both by doing `filter_tracks(n, minimum_duration=x)`
- deprecate this function and have 2 separate functions `filter_tracks_by_frames()` and `filter_tracks_by_duration` that you could then chain together if you want to do both

what do you think?

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -47,6 +48,16 @@ def test_kymo_track(blank_kymo, blank_kymo_track_args):
     assert str(k1) == "KymoTrack(N=3)"
 
 
+@pytest.mark.parametrize(
+    "time_idx, ref_value", [[[0, 1, 2], 2], [[0, 2], 2], [[1, 3], 2], [[1, 4, 5], 4]]
+)
+def test_kymotrack_seconds(blank_kymo, time_idx, ref_value):
```
rpauszek, line 54:
```suggestion
def test_kymotrack_duration(blank_kymo, time_idx, ref_value):
```

lumicks/pylake/kymotracker/tests/test_kymotrack.py
```diff
@@ -1505,3 +1516,28 @@ def test_photon_counts_api(blank_kymo):
         ).photon_counts,
         [1, 2, 5],
     )
+
+
+@pytest.mark.slow
+def test_integration_test_binding_times_with_minimum_time():
```
rpauszek, line 1522: name seems redundant. I'd suggest either
- `test_integration_binding_times_with_minimum_time`
- `binding_times_with_minimum_time_integration_test`

rpauszek (changes requested):
overall the code looks good, I just think there should be the option to filter by `min_frames` *or* `min_duration`. I think this can be accomplished pretty straightforward, see the comments for suggestions

lumicks/pylake/kymotracker/tests/test_kymotrack.py
JoepVanlier: It has to start with `test_` to be picked up by `pytest`, so I'd pick the former. 👍

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Yeah, that's a good point.

I agree that having to always think about both is not nice. I'm just slightly concerned about polluting the namespace with 3 very similarly named functions (2 in the future then).

Personally, I would have liked for both arguments to just be keyworded, so you can see immediately from the calling code what is being used. But unfortunately, that'd be a breaking change.

We could just default `minimum_length` and warn that it will be keyworded in the future.

lumicks/pylake/kymotracker/tests/test_kymotrack.py
rpauszek: ah it's the file that can end with `test`, not the functions. that's what I was thinking about

lumicks/pylake/kymotracker/kymotracker.py
rpauszek: yeah, I think I prefer the option to default both arguments. I think it's fine to warn that `minimum_length` will be keyword only in the future ... might be kinda annoying, but I guess that's the point 😉

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -367,7 +367,7 @@ def track_lines(
     return refine_tracks_centroid(kymotrack_group, track_width=line_width, bias_correction=True)
 
 
-def filter_tracks(tracks, minimum_length):
+def filter_tracks(tracks, *args, minimum_duration=0, **kwargs):
```
rpauszek, line 370: hmm, doing it this way I think is not the best option (although I get why you did, so we can throw the `FutureWarning`). It hit me right away because `minimum_length` isn't in the signature, but it is in the docstring. This leads to a very strange API stub:

image

I thought there might be a better way using `inspect` but unfortunately it doesn't tell you if an argument was called as positional or keyworded 😑 So I think we're left with a pretty ugly option, but user facing I think is better; something like

```
def filter_tracks(tracks, *args, minimum_length=None, minimum_duration=0):
    if len(args) > 1:
        raise ValueError("Passed too many positional arguments")
    elif len(args):
        if minimum_length is not None:
            raise ValueError("Passed too many positional arguments")
        else:
            minimum_length = args[0]
    minimum_length = 1 if minimum_length is None else minimum_length
```
It's still not optimal because you end up with the random `*args` but at least `minimum_length` is in the signature and the rendered API stub will look better.


An alternative that just came to mind (especially since going this route is going to be ugly no matter what) is to just completely deprecate `lk.filter_tracks()` and add a method `KymoTrackGroup.filter(self, *, minimum_length=1, minimum_duration=0)` which would return a new filtered group instance. In a way, I even think this option more closely matches our general API (we often have mutation methods that return a new instance rather than free functions) and `lk.filter_tracks()` already expects a group instance first anyways.

What do you think?

rpauszek (changes requested):

lumicks/pylake/kymotracker/kymotracker.py
JoepVanlier: Yeah, that was indeed the reason. The first option doesn't let us get rid of the positional option over time. Then again, I suppose we never really want to add additional positional arguments to this one, so that may be OK.

I don't know if I'm a huge fan of removing it and adding it to `KymoTrackGroup`. I think quite a few (if not most) of the methods on groups behave like the methods you'd expect on a `list` (which is mutate the object rather than return a new instance). If you look at `remove_tracks_in_rect`, `extend`, `split`, `merge`, they all mutate.

rpauszek: yeah, I was referring to the API outside `KymoTrackGroup` we generally don't mutate in place. But since we have that behavior anyway we could still do `KymoTrackGroup.filter()` and have it change in place. I guess any option is fine; I'm not particularly a fan of the random `*args` but 🤷 not the worst thing in the world. 

I guess the only other thing with the first option is it doesn't really follow our normal deprecation method, but again not a huge deal. It would be nice to somehow mark in the docstring prominently that the positional argument will be removed in the future. Not sure if there's a nice way to do that. And of course, we need to not forget to actually remove this at some point

JoepVanlier: No, I meant just leaving the argument positional, giving it a default and dropping the `*args`. Essentially dropping all the funky input logic we have now for the free function and accepting that that one argument can just be passed in a positional manner. Maybe that is the best option. We can still always provide it with a named argument in the docs to at least guide users into that habit.

I can add a separate `.filter` on the group and do `kwargs` from the get go there, but I think the free function has value of its own (since copying prior to mutation is annoying). It's kind of like having `np.reshape` vs `array.reshape`.

changelog.md
```diff
@@ -7,6 +7,9 @@
 * Added more options for plotting color channels for images:
   * Shortcuts `"r"`, `"g"`, and `"b"` can now be used for plotting single color channels in addition to `"red"`, `"green"`, and `"blue"`.
   * Two-channel visualizations can be plotted using `"rg"`, `"gb"`, or `"rb"`.
+* Added `duration` property to `KymoTrack` which returns the duration (in seconds) that the track was observed.
+* Added option to filter tracks with `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)` by track duration (in seconds).
+* Added `KymoTrackGroup.filter()` to filter tracks by duration or length (number of identified track points) in-place.
```
rpauszek, line 12:
```suggestion
* Added `duration` property to `KymoTrack` which returns the track duration in seconds.
* Added option to filter tracks by duration in seconds using `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)`.
* Added `KymoTrackGroup.filter()` to filter tracks in-place. `tracks.filter(minimum_duration=2)` is equivalent to `lk.filter_tracks(tracks, minimum_duration=2)`.
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -437,6 +437,10 @@ def position(self):
         """The tracked positional coordinates. The units depend on the spatial units of the tracked kymo."""
         return self._localization.position
 
+    @property
+    def duration(self):
```
rpauszek, line 441: let's add a simple docstring here so it shows up in the API stub

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -1272,6 +1276,22 @@ def remove_tracks_in_rect(self, rect, all_points=False):
...
+    def filter(self, *, minimum_length=1, minimum_duration=0):
+        """Remove tracks shorter than a minimum number of time points from the list.
+
+        Parameters
+        ----------
+        minimum_length : int, optional
+            Minimum length for the track to be accepted (default: 1).
+        minimum_duration : seconds, optional
+            Minimum duration for a track to be accepted (default: 0).
+        """
```
rpauszek, line 1288:
```suggestion
        """Remove tracks shorter than specified criteria from the list.

        Parameters
        ----------
        minimum_length : int, optional
            Minimum of tracked points for the track to be accepted (default: 1).
        minimum_duration : seconds, optional
            Minimum duration in seconds for a track to be accepted (default: 0).
        """
```

lumicks/pylake/kymotracker/kymotracker.py
```diff
@@ -377,10 +377,21 @@ def filter_tracks(tracks, minimum_length):
     ----------
     tracks : List[KymoTrack]
         Detected tracks on a kymograph.
-    minimum_length : int
-        Minimum length for the track to be accepted.
+    minimum_length : int, optional
+        Minimum length for the track to be accepted (default: 1).
+    minimum_duration : seconds, optional
+        Minimum duration for a track to be accepted (default: 0).
```
rpauszek, line 383:
```suggestion
    minimum_length : int, optional
        Minimum of tracked points for the track to be accepted (default: 1).
    minimum_duration : seconds, optional
        Minimum duration in seconds for a track to be accepted (default: 0).
```

rpauszek (approved):
Nice, I think this is good compromise. Just left some minor suggestions for wording

PR #588, opened 2023-10-31
force_calibration: split low-level fit from preparation and packaging

JoepVanlier:
**Why this PR?**
Currently, you can only use the spectral density fitting as part of one single function that prepares the initial guess, fits the spectrum and then packages the result.

This prohibits coming up with more advanced fitting schemes such as combining multiple datasets into a single fit or using priors. This PR splits the actual fitting procedure from the public API and allows more general models to be fitted by the internal function.


rpauszek (approved):
lgtm

PR #589, opened 2023-10-31
Fix scan aspect ratio

rpauszek:
**Why this PR?**
We currently calculate a default aspect ratio which results in square pixels. For scans acquired with non-square pixels, this results in distortion of the image

Current:
Screenshot 2023-10-31 at 15 05 56

Fixed:
Screenshot 2023-10-31 at 15 06 38

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -5,6 +5,7 @@
 #### Bug fixes
 
 * Fixed `lk.download_from_doi()` to align with new Zenodo REST API.
+* Fixed a bug in `Scan.plot()` in which the default aspect ratio was calculated such that pixels always appeared square. For scans with non-square pixel sizes, this would result in distortion of the image.
 
 ## v1.2.1 | 2023-10-17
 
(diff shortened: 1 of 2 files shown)


JoepVanlier (approved):
Looks good, but maybe add an example to the docstring showing the use of `aspect="equal"` for people who want the square pixels back?

wide

rpauszek:
> Looks good, but maybe add an example to the docstring showing the use of `aspect="equal"` for people who want the square pixels back?

I thought about this, but I truly can't think of a use case where you'd actually want to do this. Even in the case where users have started analysis and want consistency, I truly think everyone would rather just re-plot and get the correct image than force a messed up one, no?

JoepVanlier:
> > Looks good, but maybe add an example to the docstring showing the use of `aspect="equal"` for people who want the square pixels back?
> 
> I thought about this, but I truly can't think of a use case where you'd actually want to do this. Even in the case where users have started analysis and want consistency, I truly think everyone would rather just re-plot and get the correct image than force a messed up one, no?

👍

PR #590, opened 2023-11-02
calibration: add private api for motion blur model

JoepVanlier:
**Why this PR?**
This model can be used for camera-based calibrations with longer exposure times. Public API is pending.

See [Wong, W. P., & Halvorsen, K. (2006). The effect of integration time on fluctuation measurements: calibrating an optical trap in the presence of motion blur. Optics express, 14(25), 12517-12531](https://opg.optica.org/oe/fulltext.cfm?uri=oe-14-25-12517) for more information.

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -9,6 +9,8 @@ from .detail.drag_models import faxen_factor, brenner_axial
 from .detail.power_models import (
     g_diode,
     alias_spectrum,
+    motion_blur_peak,
+    motion_blur_spectrum,
     sphere_friction_coefficient,
     passive_power_spectrum_model,
     theoretical_driving_power_lorentzian,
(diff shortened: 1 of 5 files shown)


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -768,6 +785,24 @@ def _theoretical_driving_power(self, f_corner):
...
+
+        Parameters
+        ----------
+        acquisition_time : float
+            Acquisition time in seconds
+        """
+        new_model = copy(self)
+        new_model._calculate_power_spectral_density = motion_blur_spectrum(
+            self._calculate_power_spectral_density, acquisition_time
+        )
```
rpauszek, line 799: since this is subclassing the passive model and the first half is the same would it maybe be good to use `super()` and then just extend here?

rpauszek (approved):
looks good, just a suggestion to extend the superclass method rather than duplicating

lumicks/pylake/force_calibration/calibration_models.py
JoepVanlier: Good point. Fixing.

PR #592, opened 2023-11-07
Migrate test point scan

rpauszek:
**Why this PR?**
keep on keepin on...

### DIFF
--- a/lumicks/pylake/tests/test_imaging_confocal/conftest.py
+++ b/lumicks/pylake/tests/test_imaging_confocal/conftest.py
@@ -1,8 +1,10 @@
 import numpy as np
 import pytest
 
+from lumicks.pylake.point_scan import PointScan
+
 from ..data.mock_file import MockDataFile_v2
-from ..data.mock_confocal import generate_scan_json, generate_kymo_with_ref
+from ..data.mock_confocal import MockConfocalFile, generate_scan_json, generate_kymo_with_ref
 
 start = np.int64(20e9)
 dt = np.int64(62.5e6)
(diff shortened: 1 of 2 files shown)


lumicks/pylake/tests/test_imaging_confocal/conftest.py
```diff
@@ -198,3 +200,31 @@ def kymo_h5_file(tmpdir_factory, test_kymo):
...
+        infowave=np.zeros(data["red"].shape),
+        red_photon_counts=data["red"],
+        green_photon_counts=data["green"],
+        blue_photon_counts=data["blue"],
+    )
+    point_scan = PointScan("PointScan1", mock_file, start, stop, metadata)
+
+    reference = {
+        "data": data,
+        "timestamps": (np.arange(n_samples, dtype=np.int64) * dt) + start,
```
JoepVanlier, line 225: Bracket is not needed.

```suggestion
        "timestamps": np.arange(n_samples, dtype=np.int64) * dt + start,
```

lumicks/pylake/tests/test_imaging_confocal/test_point_scan.py
```diff
@@ -5,19 +5,21 @@
...
 
-    assert ps_red.data.shape == (90,)
+    for color, ref_data in ref["data"].items():
+        count = getattr(ps, f"{color}_photon_count")
 
-    np.testing.assert_allclose(ps_red.timestamps, reference_timestamps)
-    np.testing.assert_allclose(ps_red.data, reference_counts)
+        assert count.data.shape == ref_data.shape
 
+        np.testing.assert_allclose(count.timestamps, ref["timestamps"])
```
JoepVanlier, line 16:
```suggestion
        np.testing.assert_equal(count.timestamps, ref["timestamps"])
```

These are integers. You should really check for equality, not closeness.

You can see that this test actually doesn't test very well right now by changing the timestamps by a small number in `conftest.py`. This will not fire.

JoepVanlier (changes requested):
One of these tests don't work as intended right now. For big integers, you should really verify that there is an exact match.

JoepVanlier (approved):
👍

PR #596, opened 2023-11-22
Hidden Markov models

rpauszek:
**Why this PR?**
Adds models for analyzing time series data exhibiting transitions between discrete states. This is useful for constant-extension experiments or photon count time traces (especially FRET)

This PR lays down all of the internal algorithms for model training and state labeling. A second PR will focus on building up the rest of the public API with convenience functions for plots, histograms, etc...

Notes:
- Similar to `lk.GaussianMixtureModel` that wraps `sklearn`'s GMM, I defined an internal model class to just handle the parameters and initialization/update; we can then expose what we want from the model through the public API that wraps this
- There is a lot of internal code that uses notation seemingly from nowhere, however it is pretty standard in the HMM field and derives largely from Rabiner's seminal "A tutorial on HMMs and selected applications in speech recognition". Another very useful resource is the [HMM project](https://web.ece.ucsb.edu/Faculty/Rabiner/ece259/speech%20recognition%20course/term%20projects/HMM%20Project.pdf) from his speech recognition course, although not exactly sure if we should or how we should cite this
- Mostly I did this to avoid using things like `state_transition_matrix` sprinkled in throughout the algorithm, instead using the standard `A`. Open for discussion if there are suggestions for finding a middle ground here.
- For the public properties I use more descriptive attribute names (`means` instead of `mu`)
- Much of the code is pulled from my `smtirf` repo, written during my postdoc and refactored to fit `pylake`'s API style and vectorized where possible (since we're not yet using `numba`)
- The only intended public API is the `HiddenMarkovModel` class, although I haven't made it public yet. My idea is to match the interface with that of `lk.GaussianMixtureModel`. 

A typical workflow (after completing the public API) would be something like:
```
model = lk.HiddenMarkovModel.from_channel(force1x, n_states=3)
model.plot(force1x)
```

image

### DIFF
--- a/docs/refs.bib
+++ b/docs/refs.bib
@@ -445,3 +445,14 @@ publisher = {Biophysical Society}
   year={2020},
   publisher={AIP Publishing LLC}
 }
+
+@article{Rabiner1989HMM,
+  title={A Tutorial on Hidden Markov Models and Selected Applications in Speech Rrecognition},
+  author={Rabiner, Lawrence R.},
+  journal={Proceedings of the IEEE},
+  volume={77},
+  number={2},
+  pages={257-289},
+  year={1989},
+  publisher={IEEE}
+}
--- a/docs/tutorial/population_dynamics.rst
+++ b/docs/tutorial/population_dynamics.rst
@@ -93,13 +93,6 @@ We can also plot the time trace with each data point labeled with its most likel
 
 .. image:: figures/population_dynamics/gmm_labeled_trace.png
 
-We can extract a list of dwell times (how long the system stays in one state before transitioning to another) using the
-:meth:`~lumicks.pylake.GaussianMixtureModel.extract_dwell_times` method::
-
-    dwell_times = gmm.extract_dwell_times(force)
-
-In the next section we'll fit these data to obtain lifetimes for the model states.
-
 .. _downsampling_artifacts:
 
 Downsampling and data artifacts
@@ -133,6 +126,55 @@ two different states and do not arise from any (bio)physically relevant mechanis
 Furthermore, states with very short lifetimes can be averaged out of the data if the downsampling factor is too high. Therefore,
 in order to ensure robust results, it may be advisable to carry out the analysis at a few different downsampled rates.
 
+Hidden Markov Models
+--------------------
+
+The GMM is useful for identifying discrete states in the data, however, it does not use any of the
+temporal information available in channel measurements. To take advantage of this extra information,
+we can use a Hidden Markov Model (HMM).
+
+As suggested by the name, HMMs are used to describe a Markovian process in which the underlying
+physical state of the system is hidden or unobservable.
+
+- A Markov process (or Markov chain) is a stochastic process in which the state of the system at some
+  time point :math:`t` only depends on the state of the system at the previous time point :math:`t-1`;
+  in other words, the system is *memory-less*. The probability of transitioning from one state to
+  another is given by a transition probability matrix.
+- The (hidden) physical state that we are interested in here is the conformation of the DNA hairpin;
+  (folded or unfolded). However we cannot directly observe this state in the measurements. Instead
+  we must infer the conformational state from some observed signal; ie., the force measurements.
+  The observed signal for a particular state is described by a Normal (Gaussian) distribution.
+
+This description is the most basic implementation of an HMM. For detailed theoretical and practical
+explanations on the use and training of HMMs see the seminal work by Rabiner :cite:`Rabiner1989HMM`.
+
+We can fit an HMM with a single line of code::
+
+    hmm = lk.HiddenMarkovModel(force["0s":"2s"], n_states=2)
+
+We can get some information about the training process from the :attr:`~lumicks.pylake.HiddenMarkovModel.fit_info`
+property::
+
+    >>> print(hmm.fit_info)
+    HmmFitInfo(converged=True, n_iter=5, bic=1589.2881620885353, aic=1550.067858853102, log_likelihood=-768.033929426551)
+
+Just as with the GMM we trained earlier, we can visually inspect the quality of the fitted
+model. First, let's look at the probability distribution of the observed data::
+
+    plt.figure()
+    hmm.hist(force["0s":"2s"])
+    plt.show()
+
+.. image:: figures/population_dynamics/hmm_hist.png
+
+We can also inspect the temporal evolution of the fit::
+
+    plt.figure()
+    hmm.plot(force["0s":"1s"])
+    plt.show()
+
+.. image:: figures/population_dynamics/hmm_labeled_trace.png
+
 Dwell time analysis
 -------------------
 
@@ -150,8 +192,18 @@ to another state. The distribution can alternatively be parameterized by a rate 
 
     \mathrm{Exp}\left(t | a, k \right) = \sum_i^M a_i k_i \exp{\left( -k_i t \right)}
 
-The :class:`~lumicks.pylake.DwelltimeModel` class can be used to optimize the model parameters for an array of determined dwell times.
-Here we'll use the dwell times determined above for the high force state::
+The :class:`~lumicks.pylake.DwelltimeModel` class can be used to optimize the model parameters for an array
+of determined dwell times (how long the system stays in one state before transitioning to another).
+
+We can extract arrays of dwell times for all states in a GMM or HMM with
+:meth:`lk.GaussianMixtureModel.extract_dwell_times()<lumicks.pylake.GaussianMixtureModel.extract_dwell_times>`
+or :meth:`lk.HiddenMarkovModel.extract_dwell_times()<lumicks.pylake.HiddenMarkovModel.extract_dwell_times>`,
+respectively::
+
+    dwell_times = gmm.extract_dwell_times(force)
+
+This returns a dictionary with state indices as keys and lists of the observed dwell times as the values.
+We can fit a model by passing the array of dwell times to :class:`~lumicks.pylake.DwelltimeModel`::
 
     dwell_1 = lk.DwelltimeModel(dwell_times[1], n_components=1)
 
--- a/lumicks/pylake/population/detail/hmm.py
+++ b/lumicks/pylake/population/detail/hmm.py
@@ -1,0 +1,313 @@
+"""
+Parameters for internal API here conform to notation in Rabiner
+
+References
+----------
+.. [1] Rabiner, L. A Tutorial on Hidden Markov Models and Selected Applications in Speech
+        Recognition. Proceedings of IEEE. 77, 257-286 (1989).
+"""
+
+import warnings
+from dataclasses import dataclass
+
+import numpy as np
+
+from ..mixture import GaussianMixtureModel
+from .fit_info import HmmFitInfo
+from ...channel import Slice, Continuous
+from .validators import col, row
+
+
+def normalize_rows(matrix):
+    return matrix / col(np.sum(matrix, axis=1))
+
+
+@dataclass(frozen=True)
+class ClassicHmm:
+    """Model parameters for classic Hidden Markov Model.
+
+    Parameters
+    ----------
+    K : int
+        number of states
+    pi : np.ndarray
+        initial state probabilities, shape [K, ]
+    A : np.ndarray
+        state transition probability matrix, shape [K, K]
+    mu : np.ndarray
+        state means, shape [K, ]
+    tau : np.ndarray
+        state precision (1 / variance), shape [K, ]
+    """
+
+    K: int
+    pi: np.ndarray
+    A: np.ndarray
+    mu: np.ndarray
+    tau: np.ndarray
+
+    @classmethod
+    def guess(cls, data, n_states, gmm=None):
+        """Calculate an initial guess for the model from a GMM.
+
+        Parameters
+        ----------
+        data : np.ndarray
+            Training data
+        n_states : int
+            Number of hidden states
+        gmm : GaussianMixtureModel
+            Pre-trained GMM
+        """
+        data = Slice(Continuous(data, 0, 1))
+        if gmm is None:
+            gmm = GaussianMixtureModel.from_channel(data, n_states)
+        statepath = gmm.label(data)
+        A = normalize_rows(
+            np.vstack(
+                [
+                    [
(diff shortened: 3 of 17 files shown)


lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,361 @@
...
+        """Update model parameters from the temporary variables of the Baum Welch algorithm.
+
+        Parameters
+        ----------
+        data : np.ndarray
+            observed data, shape [T, ]
+        gamma : np.ndarray
+            probability of being in state i at time t, shape [T, K]
+        xi : np.ndarray
+            probaility of being in state i at time t and state j at time t + 1, shape [T, K, K]
```
JoepVanlier, line 159:
```suggestion
            probability of being in state i at time t and state j at time t + 1, shape [T, K, K]
```

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,361 @@
...
+        The number of hidden states in the model.
+    tol : float
+        The tolerance for training convergence.
+    max_iter : int
+        The maximum number of iterations to perform.
+    """
+
+    def __init__(self, data, n_states, tol, max_iter):
+        self.n_states = n_states
+        self._model, self._exit_flag = baum_welch(
```
JoepVanlier, line 39: I like the approach of separating the model that uses all the shorthand notation from the user API. I would probably move that model to the `detail` folder (similarly to how we've done for the FEC models) if it's not meant to be used by end users.

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,361 @@
...
+def col(x):
+    """Return x as column vector."""
+    return np.atleast_2d(x).T
+
+
+@dataclass(frozen=True)
+class ClassicHmm:
+    """Model parameters for classic Hidden Markov Model.
+
+    Parameters for this internal class conform to notation in Rabiner [ref]
```
JoepVanlier, line 95: Add ref.

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,361 @@
...
+
+    def __init__(self, data, n_states, tol, max_iter):
+        self.n_states = n_states
+        self._model, self._exit_flag = baum_welch(
+            data, ClassicHmm.guess(data, n_states), tol=tol, max_iter=max_iter
+        )
+
+    @property
+    def exit_flag(self):
+        return self._exit_flag
```
JoepVanlier, line 45: Personally, I find the use of `exit_flags` to indicate convergence failure fairly unpythonic.
I find this awkward with the GMM implementation we have as well.

Exit flags like this easily go unnoticed and long term, it would be nice to actually go to emit warnings for these.

Also, why not make these attributes full properties rather than sticking them in a dictionary?

JoepVanlier (changes requested):
1 of N:
Code-wise it looks nice. I like the separation of the internal model from the API. I also like that the model essentially is frozen, and how you set up the guess function to return one with an initial guess. The design I like.

What I'm really missing is tests though and this makes it a lot harder to review. Tests are usually quite helpful to see how you typically set up your models and simulations.

I've started reading the Rabiner paper, but I haven't had time yet to delve into the details. I have so far only commented a few concerns that popped out after an initial look over the code.

rpauszek:
Update Dec 8 2023:
@JoepVanlier  I've implemented your requested changes and added further convenience API to match that of our GMM (plotting, histograms, etc). There are a lot of small commits but I thought that might make it easier to review. If not I can squash down (I think this whole thing will end up as 1 commit in the end)

A few notes
- there's the fixup with the issue we ran into with the monkeypatching. I left that as a separate commit right at the beginning
- I forgot to move stuff to detail before making changes .... so I just did that in one go, literal copy/paste from one commit to the next
- rather than adding a `from_channel` class method like for the GMM I found it nicer to just deal with it in the constructor. I'm actually re-thinking if we even want to allow using these models with numpy arrays because a lot of the downstream stuff relies on time information. so to make it actually useful we'd need to either
   - allow users to optionally supply  the time step if they're using regular arrays
   - disallow this altogether
- I think I want to get rid of the `label` method for this and GMM and replace it either with `.path()` like I have here. Or might be more useful to have two separate methods `.state_path()` and `.emission_path()` But I'll do this in the next PR so this one doesn't go too crazy with adding functionality and changing existing API
- docs seem a bit incomplete, but I want to do another PR to make changes to the GMM API that will I think be nicer user experience in the end and avoids writing long/complex example code for certain visualizations I want to show for comparing between GMM and HMM

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+    data : numpy.ndarray | Slice
+        Data array used for model training.
+    n_states : int
+        The number of hidden states in the model.
+    tol : float
+        The tolerance for training convergence.
+    max_iter : int
+        The maximum number of iterations to perform.
+    gmm_guess : GaussianMixtureModel
+        Initial guess for the observation model parameters.
```
JoepVanlier, line 33: Maybe not for this PR (it's big enough 😅) but it would be nice if we could pass in an hmm as well. If a previous fit didn't converge and it took a while, we could continue the run?

JoepVanlier (changes requested):
Excellent work. A few more comments. I'm not fully done with reviewing it yet, but it's getting there. 👍 

Personally, rather than wrap changes to the HMM on the next PR, I'd prefer to just have the changes that you would do to the HMM in this one already so that I can see what the API will look like.

docs/refs.bib
```diff
@@ -445,3 +445,14 @@ @article{bullerjahn2020optimal
...
+@article{Rabiner1989HMM,
+  title={A Tutorial on Hidden Markov Models and Selected Applications in Speech Rrecognition},
+  author={Rabiner, Lawrence R.},
+  journal={Proceedings of the IEEE},
+  volume={77},
+  number={2},
+  pages={257-289},
+  year={1989},
+  publisher={IEEE}
+}
```
JoepVanlier, line 458: eolrip :)

docs/tutorial/population_dynamics.rst
```diff
@@ -133,6 +126,50 @@ two different states and do not arise from any (bio)physically relevant mechanis
...
+  time point :math:`t` only depends on the state of the system at the previous time point :math:`t-1`;
+  in other words, the system is *memory-less*. The probability of transitioning from one state to
+  another is given by a transition probability matrix.
+- The (hidden) physical state that we are interested in here is the conformation of the DNA hairpin;
+  (folded or unfolded). However we cannot directly observe this state in the measurements. Instead
+  we must infer the conformational state from some observed signal; ie., the force measurements.
+  The observed signal is described by a GMM.
+
+This description is the most basic implementation of an HMM. For detailed theoretical and practical
+explanations on the use and training of HMMs see the seminal work by Rabiner :cite:`Rabiner1989HMM`.
```
JoepVanlier, line 149: This is a nice way of dealing with not having a theory section on this yet.

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        if isinstance(data, Slice):
+            data = data.data
+
+        self.n_states = n_states
+        self._model, self._exit_flag = baum_welch(
+            data, ClassicHmm.guess(data, n_states, gmm=gmm_guess), tol=tol, max_iter=max_iter
+        )
+
+    @property
+    def exit_flag(self):
```
JoepVanlier, line 51: Same as above. You put a lot more info in here than just a flag.

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        """Label channel data as states.
+
+        Parameters
+        ----------
+        trace : Slice
+            Channel data to label.
+        """
+        return viterbi(trace.data, self._model)
+
+    def path(self, trace):
```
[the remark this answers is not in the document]
rpauszek: yeah, I kept going back and forth on the idea which is how I ended up here. but in the end I think it's best to have two separate methods

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        Parameters
+        ----------
+        trace : Slice
+            Data object to histogram.
+        trace_kwargs : Optional[dict]
+            Plotting keyword arguments passed to the data line plot.
+        path_kwargs : Optional[dict]
+            Plotting keyword arguments passed to the state path line plot.
+        show_path : bool
+            Whether or not to show state path.
```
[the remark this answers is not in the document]
rpauszek: at one point it did 😯

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        x : numpy.ndarray
+            Array of independent variable values at which to calculate the PDF.
+
+        Returns
+        -------
+        numpy.ndarray:
+            PDF array split into components for each state with shape (n_states, x.size).
+            The full normalized PDF can be calculated by summing across rows.
+        """
+        return np.exp(self._model.state_log_likelihood(x))
```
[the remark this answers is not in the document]
rpauszek: yeah, I wasn't quite happy with this because technically the HMM doesn't have a PDF in the traditional sense I guess. Considering right now we only use it for the histogram and it's a single line of code, I'm gonna get rid of the method altogether. We can always add it back (private or otherwise) if we find we need it in other places in the future

lumicks/pylake/population/detail/hmm.py
```diff
@@ -0,0 +1,296 @@
...
+            model = model.update(data, gamma, xi)
+
+            # E-step
+            gamma, xi, current_log_likelihood = calculate_temporary_variables(
+                model, *forward_backward(data, model)
+            )
+
+            # check for convergence
+            if (delta := current_log_likelihood - previous_log_likelihood) < 0:
+                warnings.warn(f"log likelihood decreasing by {delta:0.4e}")
```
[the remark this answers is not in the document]
rpauszek: Yeah I'm not sure. I've seen this in other EM implementations but I think it's just a numerical wiggle as the EM algorithm guarantees that the likelihood increases with each iteration. No idea where this comes from, but in all the data I've tried it's always small, on the order of 1e-14 or so. Maybe best to just get rid of the warning altogether?

lumicks/pylake/population/detail/hmm.py
```diff
@@ -0,0 +1,296 @@
...
+    statepath : np.ndarray
+        Hidden state path, shape [T, ]
+    """
+    T = data.size
+    B = model.state_log_likelihood(data).T
+    psi = np.zeros(B.shape, dtype=int)
+    statepath = np.zeros(T, dtype=int)
+
+    # initialization
+    delta = col(np.log(model.pi) + B[0])
```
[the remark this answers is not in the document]
rpauszek: yes, for very well separated states the initial state probability collapses to something like `[1, 0, 0]`. This is a peculiarity of training on a single trace since you only have one initial observation so you can't calculate a proper distribution. If you train on an ensemble of traces you would get an actual distribution. 

If you use data where the states are closer (or observations are noiser) you get something along the lines of `[1.28634832e-201 1.99176214e-049 1.00000000e+000]` which doesn't throw this warning. So I suggest to just ignore this warning using `np.errstate`

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        """
+        state_path = self.label(trace)
+        emission_path = self.means[state_path]
+
+        state_path = trace._with_data_source(trace._src._with_data(state_path))
+        emission_path = trace._with_data_source(trace._src._with_data(emission_path))
+
+        return state_path, emission_path
+
+    def extract_dwell_times(self, trace, *, exclude_ambiguous_dwells=True):
```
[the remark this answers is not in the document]
rpauszek: yes, but now getting rid of `label()` I use `state_path()` here but `label()` still in the GMM. I started moving stuff to a mixin class, but that'll now involve changes to the GMM and it got messy quickly. So I suggest leaving this (and `hist`) as it is right now and addressing that in the 2nd PR? that ok?

lumicks/pylake/population/detail/hmm.py
```diff
@@ -0,0 +1,298 @@
...
+    statepath = np.zeros(T, dtype=int)
+
+    # initialization
+    # for models trained on a single trace you can end up with zeros
+    # if states are well separated or observations have very low noise
+    with np.errstate(divide="ignore"):
+        delta = col(np.log(model.pi) + B[0])
+    # recursion
+    for t in range(1, T):
+        R = delta + np.log(model.A)
```
JoepVanlier, line 288: Not so likely for real data, but if a transition is never observed (and this element becomes zero), you could get a runtime warning taking the log here as well.

JoepVanlier (changes requested):
Looking great! Aside from validating that the HMM or GMM initial guess is passed actually contains the correct number of states, it's mostly minor things.

lumicks/pylake/population/detail/hmm.py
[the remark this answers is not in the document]
JoepVanlier: Yeah, I think that is appropriate, since `exp(-huge)` will just result in zero.

lumicks/pylake/population/hmm.py
```diff
@@ -0,0 +1,210 @@
...
+        """Model state means."""
+        return self._model.mu
+
+    @property
+    def variances(self):
+        """Model state variances."""
+        return 1 / self._model.tau
+
+    @property
+    def std(self) -> np.ndarray:
```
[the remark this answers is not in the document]
rpauszek: copy/paste from GMM. I think Tobias added it there but nothing else has annotations. I'll leave it out for now, maybe worthwhile at some point doing a specific look through for everything to annotate where appropriate. Still not quite sure how I feel about them as I think it makes the code much less readable

lumicks/pylake/population/detail/hmm.py
rpauszek: actually it's a real case when two states aren't connected to each other, you get a `0` for that element. added handling for this and a test

JoepVanlier: Excellent

JoepVanlier (approved):
:shipit:

PR #597, opened 2023-11-24
Small test improvements and bugfix on `Slice`.

JoepVanlier:
Fixes some issues that came up during mutation testing.

- Removes some unnecessary code from line time reconstruction.
- Improves tests by including more specific corner cases.
- Fixes a bug in `downsampled_like` that produced an unusable `Slice`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -11,6 +11,10 @@
 * Added option to filter tracks by duration in seconds using `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)`.
 * Added `KymoTrackGroup.filter()` to filter tracks in-place. `tracks.filter(minimum_duration=2)` is equivalent to `tracks = lk.filter_tracks(tracks, minimum_duration=2)`.
 
+#### Bug fixes
+
+* Fixed a bug in `Slice.downsampled_like` that would fail to raise an error due to a lack of overlap between the low frequency and high frequency channel when the high frequency channel starts within the last sample of the low frequency channel.
+
 #### Improvements
 
 * Kymographs consisting of a single scan line now return a valid `line_time_seconds`. This allows certain downstream functionality, such as `Kymo.plot()`.
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -38,10 +38,6 @@ def _default_line_time_factory(self: "Kymo"):
 
         dead_time = np.argmax(beyond_first_line)
 
-        # Special case for no dead time. Is it really no deadtime, or a single line?
-        if dead_time == 0 and not np.any(beyond_first_line):
-            return scan_time * infowave._src.dt * ns_to_sec
-
         return (scan_time + dead_time) * infowave._src.dt * ns_to_sec
 
     elif self.timestamps.shape[1] > 1:
--- a/lumicks/pylake/tests/test_channels/test_channels.py
+++ b/lumicks/pylake/tests/test_channels/test_channels.py
@@ -748,10 +748,18 @@ def test_downsampling_like():
     ):
         s.downsampled_like(s)
 
-    for offset in (-4 * 4, t_downsampled[-1] + 1):
+    timestep = 4
+    for offset in (with_offset(-4 * timestep), t_downsampled[-1] - timestep + 1):
         with pytest.raises(RuntimeError, match="No overlap between slices"):
-            s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), with_offset(offset), 4))
+            s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), offset, timestep))
             s.downsampled_like(reference)
+
+    for offset, ref in [
+        (with_offset(-4 * timestep + 1), 4),
+        (t_downsampled[-1] - timestep, 1),
+    ]:
+        s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), offset, timestep))
+        np.testing.assert_equal(s.downsampled_like(reference)[0].data, ref)
 
 
 def test_channel_plot():
--- a/lumicks/pylake/tests/test_channels/test_channels.py
+++ b/lumicks/pylake/tests/test_channels/test_channels.py
@@ -711,6 +711,20 @@ def test_downsampling_over_subset():
     )
 
 
+def test_downsampling_out_of_range():
+    num_points = 4
+    s = channel.Slice(channel.Continuous(np.arange(num_points), start=with_offset(100), dt=10))
+
+    for intervals in [(20, 100), (100 + 10 * num_points, 10000)]:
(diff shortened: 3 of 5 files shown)


changelog.md
```diff
@@ -11,6 +11,10 @@
 * Added option to filter tracks by duration in seconds using `lk.filter_tracks(tracks, minimum_duration=duration_in_seconds)`.
 * Added `KymoTrackGroup.filter()` to filter tracks in-place. `tracks.filter(minimum_duration=2)` is equivalent to `tracks = lk.filter_tracks(tracks, minimum_duration=2)`.
 
+#### Bug fixes
+
+* Fixed a bug in `downsampled_like` that would fail to indicate a lack of overlap between the low frequency and high frequency channel when sampling within the last sample of the low frequency channel.
```
rpauszek, line 16:
```suggestion
* Fixed a bug in `Slice.downsampled_like` that would fail to raise an error due to a lack of overlap between the low frequency and high frequency channel when the high frequency channel starts within the last sample of the low frequency channel.
```

lumicks/pylake/kymo.py
```diff
@@ -38,10 +38,6 @@ def _default_line_time_factory(self: "Kymo"):
 
         dead_time = np.argmax(beyond_first_line)
 
-        # Special case for no dead time. Is it really no deadtime, or a single line?
-        if dead_time == 0 and not np.any(beyond_first_line):
```
rpauszek, line 42: 😶

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -711,6 +711,20 @@ def test_downsampling_over_subset():
     )
 
 
+def test_downsampling_out_of_range():
+    num_points = 4
+    s = channel.Slice(channel.Continuous(np.arange(num_points), with_offset(100), 10))
+
+    for intervals in [(20, 100), (100 + 10 * num_points, 10000)]:
+        with pytest.raises(RuntimeError, match="No overlap"):
```
rpauszek, line 719: I think matching to partial messages was also something that caught us in the mutation checking, no?
```suggestion
        with pytest.raises(RuntimeError, match="No overlap between slices."):
```

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -711,6 +711,20 @@ def test_downsampling_over_subset():
     )
 
 
+def test_downsampling_out_of_range():
+    num_points = 4
+    s = channel.Slice(channel.Continuous(np.arange(num_points), with_offset(100), 10))
```
rpauszek, line 716: I think it might be nice to be explicit with the args here, it's just a bit more but then we don't have to remember the positional arguments if you're looking through things quickly

```suggestion
    s = channel.Slice(channel.Continuous(np.arange(num_points), start=with_offset(100), dt=10))
```

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -734,11 +748,16 @@ def test_downsampling_like():
...
+    timestep = 4
+    for offset in (with_offset(-4 * timestep), t_downsampled[-1] - timestep + 1):
         with pytest.raises(RuntimeError, match="No overlap between slices"):
-            s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), with_offset(offset), 4))
+            s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), offset, timestep))
             s.downsampled_like(reference)
 
+    for offset in (with_offset(-4 * timestep + 1), t_downsampled[-1] - timestep):
+        s = channel.Slice(channel.Continuous(np.array([1, 2, 3, 4]), offset, timestep))
+        s.downsampled_like(reference)
```
rpauszek, line 759: should we actually test that the downsampled result is valid and not just that it doesn't throw? or is that tested enough elsewhere?

lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -780,6 +799,19 @@ def test_with_data(data, new_data):
     np.testing.assert_allclose(data._with_data(new_data).timestamps, old_timestamps)
 
 
+def test_empty_timeseries():
+    with pytest.raises(IndexError, match="End of empty time series is undefined"):
+        _ = channel.TimeSeries([], []).stop
```
rpauszek, line 804: is it considered better to use the `_ = ` rather than just nothing at the front?

lumicks/pylake/tests/test_imaging_camera/test_image_stack.py
```diff
@@ -121,6 +122,25 @@ def test_image_stack(shape):
...
+    stack = ImageStack.from_dataset(mock_stack, name="hello", start_idx=1, stop_idx=2, step=3)
+    assert stack.name == "hello"
+    assert stack._start_idx == 1
+    assert stack._stop_idx == 2
+    assert stack._step == 3
+
+
+def test_stack_name(rgb_tiff_file):
+    cs = ImageStack(str(rgb_tiff_file), align=True)
+    cs._src.close()
```
rpauszek, line 141: why do you need to close the source? or alternatively, why don't you need to in the previous test?

lumicks/pylake/tests/test_imaging_camera/test_image_stack.py
```diff
@@ -990,10 +1010,11 @@ def test_alignment_multistack(rgb_alignment_image_data, gray_alignment_image_dat
     assert full_stack.num_frames == len(ref_ts)
 
 
-def test_frame_timestamp_ranges_include_true():
+@pytest.mark.parametrize("steps", [2, 4])
```
rpauszek, line 1013: why test additionally with `2`? is this just to avoid using hardcoded values in the actual test?

rpauszek (approved):
Always nice to have more robust tests, looks pretty good. In addition to some minor wording suggestions I left a couple questions, but nothing so serious to request changes; feel free to ignore

lumicks/pylake/tests/test_imaging_camera/test_image_stack.py
JoepVanlier: Otherwise the file stays open. This test loads a file on disk and the previous doesn't (it uses a mock).

lumicks/pylake/tests/test_channels/test_channels.py
JoepVanlier: It did yes, but I added ignores for those. That said, I don't mind adding it, while we're modifying this code anyhow. It should be `No overlap between range and selected channel.` though :)

lumicks/pylake/tests/test_channels/test_channels.py
JoepVanlier: It is, but not in this particular context, so I've added a quick assert with the result here now.

PR #598, opened 2023-11-24
powerspectrum: add bin width number of samples used to compute the power spectrum.

JoepVanlier:
**Why this PR?**
I find myself frequently adding this same calculation to notebooks.

When using `window_seconds`, it is not immediately apparent how many of the raw data points actually get used and what frequency bin width that leads to. You could compute the bin width from `ps.frequency[1] - ps.frequency[0]` but that comes with the caveat that someone could have chopped parts of the spectrum out giving you an incorrect value.

We know what the native bin width will be, so let's add it as a property.


rpauszek (approved):
🚀

PR #599, opened 2023-11-28
export_movie: force showing the figure

JoepVanlier:
**Why this PR?**
It seems that on jupyter notebook, when you export a movie, keeping a handle to `FuncAnimation` results in `matplotlib` attempting to disconnect a callback that no longer exists when you call `plt.close()`. This change mitigates that by not storing the variable.


rpauszek (approved):
:shipit:

PR #601, opened 2023-12-01
calibration: add model for salt water

JoepVanlier:
**Why this PR?**
Sometimes one may use a high sodium buffer. In that case, you need a different viscosity and density.

The first two commits introduce the functionality. The last commit I would say is optional, but I feel it's an improvement. It basically merges the functionality of `viscosity_of_salty_water` into `viscosity_of_water` by just giving it an extra argument. I feel this is better for discoverability of the API. Plan would be to just squash it before merge.

Paper here: https://srd.nist.gov/JPCRD/jpcrd176.pdf

Docs build here: https://lumicks-pylake.readthedocs.io/en/salty_branch/_api/lumicks.pylake.viscosity_of_water.html

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_models.py
+++ b/lumicks/pylake/force_calibration/calibration_models.py
@@ -9,7 +9,9 @@ from .detail.drag_models import faxen_factor, brenner_axial
 from .detail.salty_water import (
     _poly,
     pressure_factor,
+    molarity_to_molality,
     zero_pressure_viscosity,
+    _density_of_salt_solution,
     _check_salt_model_validity,
 )
 from .detail.power_models import (
@@ -68,7 +70,7 @@ def diode_params_from_voltage(
     )
 
 
-def density_of_water(temperature, molality, pressure=0.101325):
+def density_of_water(temperature, molarity, pressure=0.101325):
     """Determine the density of water with NaCl.
 
     This model is based on [1]_.
@@ -77,8 +79,8 @@ def density_of_water(temperature, molality, pressure=0.101325):
     ----------
     temperature : array_like
         Temperature in C
-    molality : float
-        Molality NaCl [mol/kg]
+    molarity : float
+        Molarity NaCl [mol/L]
     pressure : float, optional
         Pressure (default: 0.101325) [MPa].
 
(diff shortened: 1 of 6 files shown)


rpauszek (changes requested):
looks good, but the public functions need to accept molarity instead of molality (no one uses this in the bio area). We can convert internally and leave the private API accepting molality

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -60,15 +66,91 @@ def diode_params_from_voltage(
...
+    _check_salt_model_validity("Density", temperature, pressure, molality)
+
+    temperature_k = scipy.constants.convert_temperature(temperature, "C", "K")
+    weight = 58.4428 / 1000  # kg/mol for NaCl
+    mass_fraction = (
+        molality * weight / (1.0 + molality * weight)
+    )  # mol NaCl/kg water -> kg NaCl/kg total
+
+    i_ab = np.arange(-2, 3, 1)
+    i_other = np.arange(0, 3, 1)
```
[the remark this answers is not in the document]
JoepVanlier: I'll just call them `powers_ab` and `powers_other`. The former are powers for the a and b polynomials while the other one is shared between all other polynomials. The reason I had called them `i` is because that is what they were called in the paper. See page 73, equation 12, 13, 14.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -60,15 +66,91 @@ def diode_params_from_voltage(
...
+    """
+    import scipy.constants
+
+    _check_salt_model_validity("Density", temperature, pressure, molality)
+
+    temperature_k = scipy.constants.convert_temperature(temperature, "C", "K")
+    weight = 58.4428 / 1000  # kg/mol for NaCl
+    mass_fraction = (
+        molality * weight / (1.0 + molality * weight)
+    )  # mol NaCl/kg water -> kg NaCl/kg total
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -60,15 +66,91 @@ def diode_params_from_voltage(
...
+    """Determine the density of water with NaCl.
+
+    This model is based on [1]_.
+
+    Parameters
+    ----------
+    temperature : array_like
+        Temperature in C
+    molality : float
+        Molality NaCl [mol/kg]
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, it makes sense. I'd prefer to not use 1000 for the density of water (since we have an explicit model of that too), but we can convert exactly using some basic rootfinding. I'll add it in a separate commit.

lumicks/pylake/force_calibration/calibration_models.py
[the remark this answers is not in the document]
rpauszek: yeah I realized going through the rest of the internal api that there's probably just no good names for most of these variables. this is fine though

lumicks/pylake/force_calibration/detail/salty_water.py
```diff
@@ -0,0 +1,222 @@
...
+        (mass_fraction**2) * poly(powers_other, e_coeff),
+        -mass_fraction * poly(powers_other, f_coeff) * pressure,
+        -(mass_fraction**2) * poly(powers_other, g_coeff) * pressure,
+        -0.5 * poly(powers_other, h_coeff) * pressure**2,
+    ]
+
+    return 1.0 / np.sum(terms, axis=0)
+
+
+def water_molarity_to_molality(molarity, temperature, pressure, mol_weight=58.4428):
```
rpauszek, line 158: A couple points here, all kind of related
- I find this name confusing because you're not calculating the concentration of water. I would leave it generic, jus `molarity_to_molality`. 
- Because it's generic I think it might also be a good idea to not give a default value to `mol_weight` but rather pass in the MW of NaCl in the public `density_of_water` function (because that one is specifically for correcting for NaCl concentration)
- If you don't mind, I would change `mol_weight` to `molecular_weight` or at least `molec_weight`. This might be personal, but the chemist in me sees "mol" and immediately equates this to "moles". Since it's only used once, I think it's ok to have the longer name

lumicks/pylake/force_calibration/detail/salty_water.py
```diff
@@ -0,0 +1,222 @@
...
+            f"outside the valid range of the solution density model ("
+            f"{water_molality_to_molarity(0.0, temperature, pressure)}, "
+            f"{water_molality_to_molarity(6.0, temperature, pressure)})."
+        )
+
+    # Bounds are based on the valid bounds for the density model
+    return brentq(implicit_equation, 0.0, 6.0)
+
+
+def water_molality_to_molarity(molality, temperature, pressure, mol_weight=58.4428):
```
rpauszek, line 201: same as the points above

rpauszek (changes requested):
looks really good, just a very small request that I think will improve clarity

lumicks/pylake/force_calibration/detail/salty_water.py
JoepVanlier: - I find this name confusing because you're not calculating the concentration of water. I would leave it generic, jus molarity_to_molality.
Good point. I'll change it.

- Because it's generic I think it might also be a good idea to not give a default value to `mol_weight` but rather pass in the MW of NaCl in the public `density_of_water` function (because that one is specifically for correcting for NaCl concentration)
I've adjusted it. But technically they're not that generic. Since they internally depend on `_density_of_salt_solution`, they are specific to this solution. That said, changing it will make it easier in the future to possibly pass in a custom density function.

- If you don't mind, I would change mol_weight to molecular_weight or at least molec_weight. This might be personal, but the chemist in me sees "mol" and immediately equates this to "moles". Since it's only used once, I think it's ok to have the longer name
Yeah, makes sense. I'll update it.

lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -62,15 +70,53 @@ def diode_params_from_voltage(
...
+        Temperature (20°C <= T < 150°C), pressure <= 35 MPa and molality <= 6 mol/kg.
+
+    References
+    ----------
+    .. [1] Kestin, J., Khalifa, H. E., & Correia, R. J. (1981). Tables of the dynamic and
+       kinematic viscosity of aqueous NaCl solutions in the temperature range 20–150 C and
+       the pressure range 0.1–35 MPa. Journal of physical and chemical reference data, 10(1),
+       71-88.
+    """
+    molality = molarity_to_molality(molarity, temperature=temperature, pressure=pressure)
```
rpauszek, line 100: how can you call this without the molecular weight argument?

JoepVanlier: Good catch. Turns out it fell outside the test coverage when I reshuffled to use molarity. Fixed now.

rpauszek (approved):
looks good to go 👍

PR #608, opened 2023-12-12
Fixup GMM API

rpauszek:
**Why this PR?**
Some observations from working on the HMMs. We want to keep the two models' API in sync, but didn't want to mix changes to existing code with adding new features

Some notes:
- Decided to deprecate the `from_channel` class method as it only requires 2 lines of code to just do the check in the constructor. This is what I ended up doing for the HMM
- Deprecated `label()` in favor of `path()` that gives you `Slices` for both the state and emission paths in one go. Open to suggestions for a better name
- Switched up the exit flag type. Added key lookup to the base class so as not to break functionality (other than specifically testing against `dict` type), but maybe this is such a small change it's worth it to also break this

While working on these models again, I'm no questioning if we truly want to allow construction with raw numpy arrays. In general, these models are used for time series data and so we probably should either:
- only accept `Slice` and give users a way to make them from scratch
- have an additional `timestep_seconds` kwarg that would be required if constructing from numpy array. However, this wouldn't give the slice back to the user so maybe isn't even useful.

These points need to be explored more, so I'll leave it to a future PR


lumicks/pylake/population/detail/fit_info.py
```diff
@@ -10,8 +10,42 @@ class FitInfo:
...
+    aic: float
+        Akaike Information Criterion on fitted data
+
+        .. math::
+            AIC = 2 k - 2 \ln{(L)}
+
+        Where :math:`k` is the number of fitted parameters, :math:`n` is the number of observations
+        (data points) and :math:`L` is the maximized value of the likelihood function
+    lower_bound: float
+        Lower bound value on the log likelihood
```
JoepVanlier, line 40: Would be nice to add a bit more description on how to interpret this `lower_bound`. Right now, just saying `lower bound value on the log likelihood` doesn't tell users much. How are they supposed to interpret this value?

lumicks/pylake/population/mixture.py
```diff
@@ -78,27 +81,18 @@ def __init__(self, data, n_states, init_method, n_init, tol, max_iter):
         )
 
     @classmethod
+    @deprecated(
+        reason=(
+            "This method has been deprecated and will be removed in a future version. You can now "
+            "use `Slice` instances to construct this class directly."
+        ),
+        action="always",
```
JoepVanlier, line 89: Please update the docs to actually reflect this. The method `from_channel` is still used in `population.rst`.

JoepVanlier (changes requested):
Looks good.

I think there are still a few things that are odd with the API.

When plotting an HMM, you get the data, plus a state path. When plotting a GMM, you get a classified trace.

image

I also noticed that this means that the current docstring for the HMM is incorrect.

In the HMM we decided to defer that to this PR, but it would be good to deduplicate the shared functions (`extrtact_dwell_times`, `plot`, parts of `hist`) and make sure functions that have the same name behave the same for both models before we release this feature.

lumicks/pylake/population/mixture.py
rpauszek: oops, I noticed this but forgot to push the fixup commit

lumicks/pylake/population/detail/fit_info.py
rpauszek: this is the extent of the description in the `sklearn` docs for the model that we wrap, so I don't really want to go into any more detail. Could add a link to their docs for more info?

lumicks/pylake/population/mixture.py
```diff
@@ -137,60 +161,38 @@ def label(self, trace):
...
-        """
-        return self._bic
+        r"""Calculates the Bayesian Information Criterion:"""
+        return self.fit_info.bic
 
     @property
+    @deprecated(
+        reason=(
+            "This method has been deprecated and will be removed in a future version. Use "
+            "`GaussianMixtureModel.exit_flag.aic` instead."
```
JoepVanlier, line 188:
```suggestion
            "`GaussianMixtureModel.fit_info.aic` instead."
```

lumicks/pylake/population/detail/fit_info.py
rpauszek: from the `sklearn` docs I think it's because they have an option for a variational model where instead of optimizing on the log likelihood you optimize on the lower bound for the evidence; so this name seems to be a catch-all. I wonder if we should just use `log_likelihood` instead and then we don't need 2 separate dataclasses to differentiate between gmm and hmm

JoepVanlier (approved):
Beautiful!

I found two small errors in the deprecation warnings, but other than that, I think it's great and good to go! This and the previous PR in this series are awesome work.

PR #609, opened 2023-12-12
plot_correlated: fix off by one error

JoepVanlier:
**Why this PR?**
It fixes a bug that leads to an off-by-one error in the correlated plot.

- We downsample over the frame.
- The downsampled time axis uses the minimum timestamp of each frame as its timestamp.
- We subtract the first downsampled timestamp such that the origin starts at zero.
- We want to hold this first sample over the entire frame, as it corresponds to that frame. In other words, up to the next frame/time point. Therefore we should be using "post", not "pre" when setting up the step plot.
- To make sure we get an additional stick for the last frame, we must also hold the last value for the frame time.

Note:
I will defer the regression test (for now) as I plan to add an additional test that will cover this once we correctly handle the exposure time for `ImageStack`. Basically this means making sure that we integrate the channel data over the exposure time rather than the frame time when that option is selected.

With the bug:
bug_real_data_bead
bug_real_data_trap
Real data with the bug, trap and bead position

bug_simulated
Simulated data. There should be four frames, with the noise shown when the signal value is high.

Without the bug:
fixed_real_data_bead
fixed_real_data_trap

fixed_simulated_data
Simulated data. There should be four frames, with the noise shown when the signal value is high.

If you want to simulate the test data yourself:
```
from lumicks.pylake.tests.data.mock_widefield import make_frame_times, MockTiffFile
from lumicks.pylake.image_stack import TiffStack, ImageStack
from lumicks.pylake import channel
import numpy as np

a1 = np.random.normal(size=(4, 4))
a2 = np.zeros((4, 4))
make_frame_times(3)

fake_tiff = TiffStack(
    [
        MockTiffFile(
            data=[a2, a1, a2, a2],
            times=make_frame_times(4),
            description="stack",
            bit_depth=16,
        )
    ],
    align_requested=False,
)

stack = ImageStack.from_dataset(fake_tiff)
d = [0] * 5 + [2] * 5 + [0] * 10

s = channel.Slice(channel.Continuous(d, stack.start, 2), labels={"y": "test", "title": "y"})
stack.plot_correlated(s, frame=1)
```

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.3.2 | t.b.d.
+
+#### Bug fixes
+
+* Fixed a bug where the time indicator was off by one frame in [`ImageStack.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.plot_correlated) and [`ImageStack.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.export_video).
+
 ## v1.3.1 | 2023-12-07
 
 #### Bug fixes
--- a/lumicks/pylake/nb_widgets/correlated_plot.py
+++ b/lumicks/pylake/nb_widgets/correlated_plot.py
@@ -70,7 +70,21 @@ def plot_correlated(
 
     t0 = downsampled.timestamps[0]
     t, y = downsampled.seconds, downsampled.data
-    ax_channel.step(t, y, where="pre")
+
+    # We explicitly append the last frame time to make sure that it still shows up
+    last_dt = np.diff(
+        [
+            frame_range
+            for frame_range in frame_timestamps
+            if frame_range[0] >= channel_slice.start and frame_range[1] <= channel_slice.stop
+        ][-1]
+    )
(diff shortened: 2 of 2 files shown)


agom:
Shouldn't this go into the `release/1.3` branch?

JoepVanlier:
> Shouldn't this go into the `release/1.3` branch?

Yes, it should. Note that I made a small amendment since yesterday (We should take the last frame that's in range of the channel data for the final step, not the last frame of the movie). This doesn't affect anything other than the last "step" duration though (the value is the same as the previous).

changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.3.2 | 2023-12-12
```
agom, line 3:
```suggestion
## v1.3.2 | 2023-12-13
```

agom (approved):
Looks good. Just a question, in the last plot of the PR description, should the red overlay span the entire frame duration? Seems to be short, is that expected?

JoepVanlier:
> Looks good. Just a question, in the last plot of the PR description, should the red overlay span the entire frame duration? Seems to be short, is that expected?

It covers the exposure time which in my simulated example was a partial frame.

changelog.md
JoepVanlier: Oh good catch, I think pycharm autofilled that for me. It should actually be `t.b.d.` since we don't know when we'll release this patch yet.

PR #610, opened 2023-12-18
docs: fix import issue and add `id` field to notebook export

JoepVanlier:
**Why this PR?**
Our docs stopped building.

One is an import issue (they removed an import from `nbexport` [here](https://github.com/jupyter/nbconvert/commit/a7df9afbf165e397c31d6b79c66cff8288e4a42e)). Second was just a warning (but one that will become an error in the future) which is that an `id` field is missing from our exported notebooks.

### DIFF
--- a/docs/_ext/nbexport.py
+++ b/docs/_ext/nbexport.py
@@ -1,4 +1,5 @@
 import os
+import uuid
 import itertools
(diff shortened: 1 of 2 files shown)


rpauszek (approved):
👍

PR #613, opened 2024-01-16
force_calibration: fix small mistake in the documentation

JoepVanlier:
**Why this PR?**
Normalization constant is missing a factor `pi` in the theory code.

Note: Had to cherrypick the docs fixup from `main` to get the docs to build.


rpauszek (approved):
the sphinx commit should get dropped when you rebase on `main` now

PR #614, opened 2024-01-23
Move duplicated GMM/HMM properties to mixin

rpauszek:
**Why this PR?**
After adding HMMs and re-aligning the GMM API, I realized that we could de-duplicate a bit more code using the mixin if we implement a simple data structure for the GMM rather than just storing the `sklearn` model. This has the added benefit in that we can get rid of the `as_sorted` decorator and the private `_map` property.

All changes are to private API

Two tiny extras that I had to do, but I think it's worth it:
- store the `lower_bound` as a private attribute, but this can be removed when we remove `exit_flag`
- change the implementation of the deprecated `label()` method, but I realized we can do it even easier with other stuff we already had

### DIFF
--- a/lumicks/pylake/population/detail/mixin.py
+++ b/lumicks/pylake/population/detail/mixin.py
@@ -1,10 +1,33 @@
+from dataclasses import dataclass
+
 import numpy as np
 
+from .fit_info import PopulationFitInfo
 from ...channel import Slice
 from ..dwelltime import _dwellcounts_from_statepath
 
 
 class TimeSeriesMixin:
+    @property
+    def fit_info(self) -> PopulationFitInfo:
+        """Information about the model training exit conditions."""
+        return self._fit_info
+
+    @property
+    def means(self) -> np.ndarray:
+        """Model state means."""
+        return self._model.mu
+
+    @property
+    def variances(self) -> np.ndarray:
+        """Model state variances."""
+        return 1 / self._model.tau
+
+    @property
+    def std(self) -> np.ndarray:
+        """Model state standard deviations."""
+        return np.sqrt(self.variances)
+
     def extract_dwell_times(self, trace, *, exclude_ambiguous_dwells=True):
         """Calculate lists of dwelltimes for each state in a time-ordered state path array.
 
@@ -151,3 +174,10 @@ class TimeSeriesMixin:
         path_kwargs = {"c": "tab:blue", "lw": 2, **(path_kwargs or {})}
         emission_path = self.emission_path(trace)
         emission_path.plot(**path_kwargs)
+
+
+@dataclass(frozen=True)
+class LatentVariableModel:
+    K: int
+    mu: np.ndarray
+    tau: np.ndarray
--- a/lumicks/pylake/population/mixture.py
+++ b/lumicks/pylake/population/mixture.py
@@ -1,26 +1,31 @@
+from dataclasses import dataclass
+
 import numpy as np
 import scipy
 from deprecated.sphinx import deprecated
 
 from ..channel import Slice
-from .detail.mixin import TimeSeriesMixin
+from .detail.mixin import TimeSeriesMixin, LatentVariableModel
 from .detail.fit_info import PopulationFitInfo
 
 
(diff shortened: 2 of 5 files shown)


lumicks/pylake/population/mixture.py
```diff
@@ -1,26 +1,31 @@
...
-from .detail.mixin import TimeSeriesMixin
+from .detail.mixin import BaseModel, TimeSeriesMixin
 from .detail.fit_info import PopulationFitInfo
 
 
-def as_sorted(fcn):
-    """Decorator to return results sorted according to mapping array.
+@dataclass(frozen=True)
+class ClassicGmm(BaseModel):
+    """Model parameters for classic Hidden Markov Model.
```
JoepVanlier, line 14:
```suggestion
    """Model parameters for classic Gaussian Mixture Model.
```

lumicks/pylake/population/tests/test_hmm_algos.py
```diff
@@ -1,7 +1,7 @@
 import numpy as np
 import pytest
 
-from lumicks.pylake import HiddenMarkovModel, GaussianMixtureModel
```
JoepVanlier, line 4: Hmm. I'm surprised this dependency that is not needed wasn't picked up by flake earlier.

lumicks/pylake/population/mixture.py
```diff
@@ -62,22 +67,33 @@ def __init__(self, data, n_states, init_method="kmeans", n_init=1, tol=1e-3, max
...
-            self._model.converged_,
-            self._model.n_iter_,
-            self._model.bic(data),
-            self._model.aic(data),
-            np.sum(self._model.score_samples(data)),
+            _model.converged_,
+            _model.n_iter_,
+            _model.bic(data),
+            _model.aic(data),
+            np.sum(_model.score_samples(data)),
```
JoepVanlier, line 96: Very minor, but might be nice to assign these with keyword args, since they're spread out over lines anyhow. Or maybe just the last one at least. Then you immediately see that it is the log likelihood.

lumicks/pylake/population/mixture.py
```diff
@@ -62,22 +67,33 @@ def __init__(self, data, n_states, init_method="kmeans", n_init=1, tol=1e-3, max
...
             init_params=init_method,
             n_init=n_init,
             tol=tol,
             max_iter=max_iter,
         )
         data = np.reshape(data, (-1, 1))
-        self._model.fit(data)
+        _model.fit(data)
+
+        # todo: remove when exit_flag is removed
```
JoepVanlier, line 80: Nice reminder.

lumicks/pylake/population/detail/mixin.py
```diff
@@ -151,3 +174,10 @@ def plot_path(self, trace, *, trace_kwargs=None, path_kwargs=None):
         path_kwargs = {"c": "tab:blue", "lw": 2, **(path_kwargs or {})}
         emission_path = self.emission_path(trace)
         emission_path.plot(**path_kwargs)
+
+
+@dataclass(frozen=True)
+class BaseModel:
```
JoepVanlier, line 180: I feel the name `BaseModel` is a bit too general/nondescript. I'm afraid that might bite us in the future. Thinking what GMMs and HMMs have in common, maybe `LatentVariableModel` or `TimeSeriesModel`?

What do you think?

JoepVanlier (approved):
Great! It's really nice to get rid of the sorting in addition to the deduplication!

lumicks/pylake/population/tests/test_hmm_algos.py
rpauszek: maybe we don't enforce `flake8` on the tests? it was only recently we enforced `black` there

JoepVanlier: We should be enforcing it though. If I look at the config, we only exclude that warning for `__init.py__`. I'll look into it. Not part of this PR anyhow.
```
    # Ignore imported but not used for __init__.py files.
    __init__.py:F401
```

PR #615, opened 2024-01-23
ImageStack: allow closing `ImageStack` file handles

JoepVanlier:
**Why this PR?**
It would be nice to have `ImageStack` close the file handle when it is garbage collected. People have run into not being able to delete files on windows because the handle was still open.

In addition, a `.close()` is added for cases where it should be closed immediately.

I considered two alternatives before going for the wrapper class:
- Using `weakref.finalizer`, but this cannot pass the instance it is closing as argument (and is therefore not a solution for this).
- Adding the handling on `TiffStack`, but this is unfortunately also not an option, since the handle is actually associated with the `TiffFile` and `TiffStack` can return new `TiffStack` instances that reference the same (list of) `TiffFile`s.


rpauszek (approved):
awesome! super nice to have this handled automatically

PR #616, opened 2024-01-25
docs: fix logo on pypi

JoepVanlier:
***Why this PR?***
Unfortunately, supporting a dark and light-mode logo broke our logo on pypi. It's a known issue, but there is [no solution yet](https://github.com/pypi/warehouse/issues/11251).

Looking at other projects, none of the supported dark/light mode options seem to work. They either would show nothing on pypi or both logos (like the option below).

```
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="https://media.githubusercontent.com/media/lumicks/pylake/main/docs/logo_dark.png">
  <source media="(prefers-color-scheme: light)" srcset="https://media.githubusercontent.com/media/lumicks/pylake/main/docs/logo_light.png">
  <img alt="pylake logo" src="https://media.githubusercontent.com/media/lumicks/pylake/main/docs/logo_light.png" width="auto" height="120">
</picture>
```

Personally, I think until there is proper support, it'd probably be best to just use the light mode logo.
I would also be open to having a customized logo here and two logo's on pypi.


rpauszek (approved):
seems reasonable until there's a proper fix, I don't think we have to go with a more complicated solution for this

PR #617, opened 2024-01-30
ci: fix tests for `pytest>=8.0.0`

JoepVanlier:
Pytest changed its warning/exception order starting from `8.0.0` (see https://github.com/pytest-dev/pytest/issues/9036).

This updates our tests to pass under the new and old version of `pytest`.

- For kymographs, this meant explicitly passing a motion blur parameter such that we don't issue that error instead of what we are looking for.
- For the dwelltime analysis, we have to make sure we don't get any of the deprecation warnings so that we can see the warning we are testing for.
- For touchdown, we have to explicitly silence the covariance matrix is undefined error (which actually already tells us that the problem will be non-identifiable i.e. too little data, which is exactly what the test tests for).

After this gets merged, I will merge `1.3` back into `main`.


rpauszek (approved):
:shipit:

PR #618, opened 2024-01-30
imagestack: fix adjustment single channel images

JoepVanlier:
**Why this PR?**
We introduced a regression when it comes to color adjustments for single channel WIT stacks in `1.3.0` [here](https://github.com/lumicks/pylake/pull/571).

If you try to do a color adjustment on a stack and then plot without specifying a color channel or specifying `"rgb"` no adjustment is applied.

This PR fixes that and adds a test to cover the single channel case. I will squash to a single commit when approved, but left it separate so you can verify that the test fires without the fix.

Note: Tests are not passing because of https://github.com/lumicks/pylake/pull/617 .

Example:
```
plt.subplot(1, 2, 1)
stack.plot(frame=300, adjustment=lk.ColorAdjustment(0, 97, "percentile"))
plt.title("")

plt.subplot(1, 2, 2)
stack.plot("r", frame=300, adjustment=lk.ColorAdjustment(0, 97, "percentile"))
plt.title("")
```

Expected:
image

Result:
image


changelog.md
```diff
@@ -11,6 +11,7 @@
...
 * Fixed a bug where the time between frames was incorrectly not excluded when calling [`ImageStack.frame_timestamp_ranges()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.frame_timestamp_ranges) with `include_dead_time=False`. Note that `Scan` and `Kymo` are not affected.
 * Fixed a bug where [`ImageStack.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/v1.3.0/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.plot_correlated) was not excluding the dead time between frames.
+* Fixed a bug where color adjustments on a single channel [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack) would not be applied for single channel stacks unless a single channel was provided as argument. This bug was introduced in `v1.3.0`.
```
rpauszek, line 14:
```suggestion
* Fixed a bug where color adjustments on a single channel [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack) would not be applied unless a channel was provided as argument. This bug was introduced in `v1.3.0`.
```

rpauszek (approved):
👍

changelog.md
JoepVanlier: ah yeah, I guess the fact that it is about stacks is already implied by the API reference

PR #623, opened 2024-02-13
kymo: crop beads out of kymograph

JoepVanlier:
**Why this PR?**
Selecting ROIs by hand can be cumbersome. This function provides approximate bead edges that can be used for setting the kymotracking ROI (or directly crop the kymograph).

Might be a good initial approach until we find something better. Marked as beta because while tested on a fair number of kymographs, we might still find a better algorithm in the future.

Screenshot 2024-02-13 at 10 33 17

Screenshot 2024-02-13 at 10 34 21

Screenshot 2024-02-13 at 10 35 31

API docs [here](https://lumicks-pylake.readthedocs.io/en/find_beads/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.bead_roi) and [here](https://lumicks-pylake.readthedocs.io/en/find_beads/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.crop_beads)

### DIFF
--- a/lumicks/pylake/detail/bead_cropping.py
+++ b/lumicks/pylake/detail/bead_cropping.py
@@ -1,0 +1,101 @@
+import numpy as np
+
+
+def find_beads(
+    summed_kymo,
+    bead_diameter_pixels,
+    plot=False,
+    threshold_percentile=70,
+    allow_negative_beads=True,
+):
+    """Find bead edges in kymograph image sum
+
+    Attempts to search for the bead edges by checking where the fluorescence drops below a
+    threshold. Only intended for stationary beads.
+
+    Parameters
+    ----------
+    summed_kymo : np.ndarray
+        Kymograph image data (single channel summed along the time axis).
+    bead_diameter_pixels : float
+        Estimate for the bead size in pixels.
+    plot : bool
+        Plot result
+    threshold_percentile : int
+        Percentile below which to drop.
+    allow_negative_beads : bool
+        Tries to estimate a background from the dominant mode of the photon counts. This estimate
+        is subtracted from the raw counts prior to analysis.
+
+    Returns
+    -------
+    list of int
+       List of the two edge positions in pixels.
+
+    Raises
+    ------
+    RuntimeError
+        When the algorithm fails to locate two edges during the bead finding stage.
+    """
+    import scipy.ndimage
+    import matplotlib.pyplot as plt
+
+    data = summed_kymo.astype(float)
+
+    # Makes sure we can handle dark beads
+    if allow_negative_beads:
+        # Try to guess the background by finding the most prominent mode
+        kde_estimate = scipy.stats.gaussian_kde(data, 0.02)
+        interpolated_kde = np.arange(min(data), np.max(data))
+        baseline = interpolated_kde[np.argmax(kde_estimate.pdf(interpolated_kde))]
+        data = np.abs(data - baseline)
+    else:
+        baseline = 0
+
+    # Get rid of small fluorescence peaks.
+    only_beads = scipy.ndimage.grey_opening(data, int(bead_diameter_pixels / 4) * 2 + 1)
+
+    # Get rough estimate of where the beads are.
+    blurred, blurred_derivative = (
+        scipy.ndimage.gaussian_filter(
+            only_beads, bead_diameter_pixels, mode="constant", cval=0, order=order
+        )
+        for order in (0, 1)
+    )
+
+    # Find negative zero crossings.
+    bead_edges = np.where(np.diff(np.sign(blurred_derivative)) == -2)[0]
+
+    if plot:
+        plt.plot(summed_kymo, label="input data")
+        plt.axhline(baseline, label="baseline")
+        plt.plot(data, label="transformed data")
+        plt.plot(only_beads, label="only beads")
+        plt.plot(blurred, label="Gaussian filtered")
+
+        for be in bead_edges:
+            plt.axvline(be, label="center guess", color="k", linestyle="--", alpha=0.5)
+
+        plt.legend()  # In case we don't make it past the exception
+
+    if len(bead_edges) != 2:
+        raise RuntimeError("Did not find two beads")
+
+    # We move inwards until the opened trace drops below the Gaussian blurred trace. This will be
+    # roughly halfway on the slope.
+    bead_edges[0] += np.where(only_beads[bead_edges[0] :] < blurred[bead_edges[0] :])[0][0]
+    bead_edges[1] -= np.where(
+        np.flip(only_beads[: bead_edges[1]]) < np.flip(blurred[: bead_edges[1]])
+    )[0][0]
+
+    # We then keep moving until we've hit a low enough signal level that we are confident we
+    # have reached the tether.
+    final_cutoff = np.percentile(data[bead_edges[0] : bead_edges[1]], threshold_percentile)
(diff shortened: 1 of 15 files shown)


lumicks/pylake/detail/bead_cropping.py
```diff
@@ -0,0 +1,92 @@
...
+
+    Parameters
+    ----------
+    data : np.ndarray
+        Kymograph image data (single channel summed along the time axis).
+    bead_diameter_pixels : float
+        Estimate for the bead size in pixels.
+    plot : bool
+        Plot result
+    percentile : int
```
rpauszek, line 20: maybe something like `threshold_percentile` would be more descriptive?

lumicks/pylake/detail/bead_cropping.py
```diff
@@ -0,0 +1,92 @@
...
+        Percentile below which to drop.
+    allow_negative_beads : bool
+        Tries to estimate a background from the dominant mode of the photon counts. This estimate
+        is subtracted from the raw counts prior to analysis.
+
+    Raises
+    ------
+    RuntimeError
+        When the algorithm fails to locate two edges during the bead finding stage.
+    """
```
rpauszek, line 30: it might be nice to add a super short `Returns` block. I think it'd be helpful for our future selves to quickly be able to tell that you expect a list of 2 edge positions in pixels (I'm assuming just from reading the code)

lumicks/pylake/kymo.py
```diff
@@ -4,6 +4,8 @@
 
 import numpy as np
 
+from lumicks.pylake.detail.bead_cropping import find_beads
+
```
rpauszek, line 8: We use relative imports for `pylake` stuff (absolute imports in the tests). I think it'll also get rid of the extra space

```suggestion
from .detail.bead_cropping import find_beads
```

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
         ax_hist.set_ylabel("counts")
         ax_hist.set_title(self.name)
 
+    def bead_roi(self, bead_diameter, *, channel="green", plot=False, percentile=70, **kwargs):
```
rpauszek, line 575: I think `bead_edges` sounds better. Additionally, we try to use verbs for method names so I would go with `estimate_bead_edges()`. This also makes it clear that it's an estimation and not like 100% bullet proof

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
...
+        """Determine approximate positional coordinates of the bead edges.
+
+        Searches for the bead edges by checking where the fluorescence drops below a threshold.
+        Only intended for non-moving beads.
```
rpauszek, line 579:
```suggestion
        Only intended for stationary beads.
```

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
...
+            values will make the bounds go closer to the bead edge but risk the algorithm failing
+            or having high background.
+        **kwargs
+            Forwarded to bead finding function.
+
+        Raises
+        ------
+        RuntimeError
+            When the algorithm fails to locate two edges during the bead finding stage.
+        """
```
rpauszek, line 607: a `Returns` section would be helpful here I think, especially making it clear that the values are in microns

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
...
+                bead_diameter_pixels=bead_diameter / self.pixelsize_um[0],
+                plot=plot,
+                percentile=percentile,
+                **kwargs,
+            )
+            * self.pixelsize_um[0]
+        )
+
+    def crop_beads(self, bead_diameter, channel="green", **kwargs):
+        """Crop beads out of kymograph
```
rpauszek, line 621: maybe something like:

```suggestion
        """Estimates the edges of stationary beads and returns a copy of the kymograph cropped to these limits.
```

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
...
+            a large number of different scenarios. The API can still be subject to change without
+            any prior deprecation notice! If you use this functionality keep a close eye on the
+            changelog for any changes that may affect your analysis.
+
+        Parameters
+        ----------
+        bead_diameter : float
+            Rough estimate for the bead size (microns).
+        channel : 'red', 'green', 'blue', optional
+            Channel to use for bead detection.
```
rpauszek, line 593: Not for this PR, but just curious did you try this by summing across channels also? Often the beads show up in all channels and using all the signal would significantly boost the beads features even more above anything on the tether

lumicks/pylake/kymo.py
```diff
@@ -570,6 +572,74 @@ def plot_with_time_histogram(
...
+        **kwargs
+            Forwarded to :meth:`Kymo.bead_roi()`.
+        """
+        to_current_units = self.pixelsize[0] / self.pixelsize_um[0]
+        crop_locations = np.asarray(self.bead_roi(bead_diameter, channel=channel, **kwargs))
```
rpauszek, line 640: In a way it's nice that you take care to handle any units (although the docstring specifically mentions microns which could be confusing since it needs to be in whatever the current units are). I wonder though if we should just `raise` for this and the previous method if the kymo isn't in microns? It doesn't make sense really to calibrate and then crop, and there's less chance of some odd corner case popping up that we haven't taken into account (which is just a nice way of saying it's less for us to worry about 😉 ) What do you think?

rpauszek (approved):
Awesome, I think this will be super useful functionality that users will appreciate. I approved, but take a look at my comments on the method name and disallowing for kymos not in microns. Neither are necessary, but I think they might be nice...up to you

lumicks/pylake/detail/bead_cropping.py
JoepVanlier: Yeah, good point.

lumicks/pylake/detail/bead_cropping.py
JoepVanlier: Works for me!

lumicks/pylake/kymo.py
JoepVanlier: `estimate_bead_edges` it is!

lumicks/pylake/kymo.py
JoepVanlier: Yeah, probably a good idea to explicitly exclude this.

lumicks/pylake/kymo.py
JoepVanlier: Great idea! I did just give this a go, but for the kymos that I have it doesn't seem to make a big difference. If you have one where you see a big difference in SNR that way, it would be good to include it in the test set.

I still have a few failure mode kymos. Basically, if the center gets really crowded or there are mobile proteins, the small peak rejection fails. Usually this can be recovered by going for a different channel or taking a subset of the kymo.

rpauszek: I'm not so worried about failure modes because the beads can be really variable so I'm not expecting 100% accuracy for this estimate. I just wonder, if using the signal summed across channels isn't any worse off than using a single channel maybe we just do this by default and not have the channel argument at all?

PR #624, opened 2024-02-23
kymo: finding bead edges part 2

JoepVanlier:
**Why this PR?**
The algorithm added in https://github.com/lumicks/pylake/pull/623 works fairly robustly, but can fail occasionally for narrow beads. The algorithm in this PR is a bit more general and can be used as fallback.

It makes use of cross correlation to determine features that recur in each frame. Basically, for each window, it attempts to use a series of all possible templates of a given size, which it then cross correlates with the next frame (taking the maximum found cross correlation). This is repeated for each frame. After this, we take the average along the positional axis which gives you a score for how "templatey" that bit is. 

There is a parameter (which is not explicitly documented in the function on kymo), which relaxes the requirement of the template to stay in the same place. I have seen this help when considering experiments where people are pulling on beads; though in most cases, it does result in some more failures (hence it is not the default).

I think initially, on the public side of the API, we should expose as little as possible. I would like to see how users get on with this functionality. If we find out that they do need the customizability that the arguments offer, then we can expose them when needed, but I would initially like to try it with the defaults first.

API doc [here](https://lumicks-pylake.readthedocs.io/en/option_2/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.estimate_bead_edges).

### DIFF
--- a/lumicks/pylake/detail/bead_cropping.py
+++ b/lumicks/pylake/detail/bead_cropping.py
@@ -21,7 +21,7 @@ def _move_inward_percentile(data, bead_edges, threshold_percentile):
 
 
 def find_beads_brightness(
-    summed_kymo,
+    kymograph_image,
     bead_diameter_pixels,
     plot=False,
     threshold_percentile=70,
@@ -34,8 +34,8 @@ def find_beads_brightness(
 
     Parameters
     ----------
-    summed_kymo : np.ndarray
-        Kymograph image data (single channel summed along the time axis).
+    kymograph_image : np.ndarray
+        2D kymograph image
     bead_diameter_pixels : float
         Estimate for the bead size in pixels.
     plot : bool
@@ -59,6 +59,7 @@ def find_beads_brightness(
     import scipy.ndimage
     import matplotlib.pyplot as plt
 
+    summed_kymo = kymograph_image.sum(axis=1)
     data = summed_kymo.astype(float)
 
     # Makes sure we can handle dark beads
--- a/lumicks/pylake/detail/bead_cropping.py
+++ b/lumicks/pylake/detail/bead_cropping.py
@@ -1,7 +1,26 @@
 import numpy as np
 
 
-def find_beads(
+def _guess_background(data):
+    """Determine background level by determining most prominent mode"""
+    import scipy.stats
+
+    kde_estimate = scipy.stats.gaussian_kde(data, 0.02)
+    interpolated_kde = np.arange(min(data), np.max(data))
+    return interpolated_kde[np.argmax(kde_estimate.pdf(interpolated_kde))]
+
+
+def _move_inward_percentile(data, bead_edges, threshold_percentile):
+    """Move inward until a specific percentile is reached"""
+
+    final_cutoff = np.percentile(data[slice(*bead_edges)], threshold_percentile)
+    return [
+        bead_edges[0] + np.where(data[bead_edges[0] :] <= final_cutoff)[0][0],
+        bead_edges[1] - np.where(np.flip(data[: bead_edges[1]]) <= final_cutoff)[0][0],
+    ]
+
+
+def find_beads_brightness(
     summed_kymo,
     bead_diameter_pixels,
     plot=False,
@@ -29,7 +48,7 @@ def find_beads(
 
     Returns
     -------
-    list of int
+    bead_edges : list[int]
        List of the two edge positions in pixels.
 
     Raises
@@ -45,9 +64,7 @@ def find_beads(
     # Makes sure we can handle dark beads
     if allow_negative_beads:
         # Try to guess the background by finding the most prominent mode
-        kde_estimate = scipy.stats.gaussian_kde(data, 0.02)
-        interpolated_kde = np.arange(min(data), np.max(data))
-        baseline = interpolated_kde[np.argmax(kde_estimate.pdf(interpolated_kde))]
+        baseline = _guess_background(data)
         data = np.abs(data - baseline)
     else:
         baseline = 0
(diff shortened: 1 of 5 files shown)


lumicks/pylake/detail/bead_cropping.py
```diff
@@ -88,14 +105,178 @@ def find_beads(
...
+            normalization_factor = 1.0 + np.std(padded_img[:, t + 1]) * np.std(template)
+
+            if allow_movement:
+                correlation_image.append(
+                    np.max(scipy.signal.correlate(padded_img[:, t + 1], template, mode="same"))
+                    / normalization_factor
+                )
+            else:
+                im = padded_img[template_origin : template_origin + template_size_pixels, t + 1]
+                correlation_image.append(np.sum(im * template) / normalization_factor)
```
rpauszek: would it be worth it to try to pull this out in a separate function? the double `for` loop using `append`s could then become a pretty simple nested list comprehension I think. Or do you think this would be a big hit for performance because of significant function calls?

lumicks/pylake/detail/tests/test_bead_crop.py
```diff
@@ -5,7 +5,7 @@
 import pytest
 import matplotlib.pyplot as plt
 
-from lumicks.pylake.detail.bead_cropping import find_beads
+from lumicks.pylake.detail.bead_cropping import find_beads, find_beads_template
```
rpauszek, line 8: since we have 2, maybe we should have a more distinct name for `find_beads`? `find_beads_brightness`?

lumicks/pylake/kymo.py
```diff
@@ -586,18 +589,40 @@ def estimate_bead_edges(
...
+
+        2. Template
+
+        Downsamples the kymograph to a specific number of frames (specified with the optional
+        parameter `num_time_frames`). For each frame, we traverse the scan line pixel by pixel,
+        extracting a template at each pixel. This template is then correlated with the next frame.
+        This process is repeated until all frames have been processed and a 2D matrix is acquired.
+        This matrix is then summed over the temporal axis obtaining a similarity score over time.
+        This score can be interpreted as a measure for how long a template taken at that particular
+        location is present. The outer maxima of this curve provide the bead centers.
```
rpauszek, line 611: This looks odd to me when rendered. It's difficult to tell if the next paragraph applies to the "template" option still or to both
image

I would suggest something like this
image

```suggestion
        * *Brightness: brightness based bead edge detection.*
          Searches for bead edges by summing along the temporal direction and removing small features.
          The result is smoothed and peaks are detected. Bead edges are found by checking where the
          fluorescence drops below a threshold.

        * *Template: template correlation-based bead edge detection.*
          Downsamples the kymograph to a specific number of frames (specified with the optional
          parameter `num_time_frames`). For each frame, we traverse the scan line pixel by pixel,
          extracting a template at each pixel. This template is then correlated with the next frame.
          This process is repeated until all frames have been processed and a 2D matrix is acquired.
          This matrix is then summed over the temporal axis obtaining a similarity score over time.
          This score can be interpreted as a measure for how long a template taken at that particular
          location is present. The outer maxima of this curve provide the bead centers.
```

lumicks/pylake/kymo.py
```diff
@@ -586,18 +589,40 @@ def estimate_bead_edges(
...
+
+        Once the centers are found, we start searching from a bead_diameter from the center until
+        the fluorescence drops below a baseline level.
+
         Parameters
         ----------
         bead_diameter : float
             Rough estimate for the bead size (microns).
+        algorithm : str
+            Which algorithm to use. Valid options are "brightness", "template".
```
rpauszek, line 621:
```suggestion
        algorithm : 'brightness', 'template'
            Which algorithm to use. 
```

lumicks/pylake/kymo.py
```diff
@@ -586,18 +589,40 @@ def estimate_bead_edges(
...
         channel : 'red', 'green', 'blue', optional
             Channel to use for bead detection.
         plot : bool, optional
             Plot result
-        threshold_percentile : int
-            Percentile to drop down to before accepting that we have left the bead area. Higher
-            values will make the bounds go closer to the bead edge but risk the algorithm failing
-            or having high background.
         **kwargs
             Forwarded to bead finding function.
```
rpauszek, line 627: I wonder if we should even expose this in the public function at first? It would require users digging into the docstrings for the internal functions anyway. Only downside is it'd make things more difficult for us for testing, but we can just use the private functions directly. 

If we really want to expose it, I think I'd prefer to expose all of the individual parameters with defaults rather than `kwargs` and only the relevant ones get forwarded. What do you think?

lumicks/pylake/kymo.py
```diff
@@ -618,18 +643,24 @@ def estimate_bead_edges(
...
+
+        if algorithm == "brightness":
+            return (
+                np.array(find_beads(image.sum(axis=1), **shared_parameters)) * self.pixelsize_um[0]
+            )
+        elif algorithm == "template":
+            return np.array(find_beads_template(image, **shared_parameters)) * self.pixelsize_um[0]
+        else:
+            raise ValueError(
+                f'Unrecognized algorithm {algorithm} selected. Choose "brightness" or "template"'
```
rpauszek, line 660: If you go with my suggestion for both private functions to take the full kymo image you could do something like
```
if algorithm not in ("brightness", "template"):
    raise ValueError(...)
finder = find_beads if algorithm == "brightness" else find_beads_template
return np.array(finder(image, **shared_parameters)) * self.pixelsize_um[0]
```

rpauszek (commented):
Functionally everything looks really great and seems to work well on lots of different "kinds of beads". I've left some comments, nothing super important but I would like to get your thoughts on some of the suggestions before approving

lumicks/pylake/detail/bead_cropping.py
```diff
@@ -88,14 +105,178 @@ def find_beads(
...
+    ----------
+    kymograph_image : np.ndarray
+        2D kymograph image
+    bead_diameter_pixels : int
+        Template size in pixels (will be rounded up to an odd number). Note that
+        the template size should be equal or larger than the bead size. It should
+        capture as much of the bead (including the fluorescent side lobe if present)
+        as possible.
+    num_time_frames : int
+        Number of time frames to downsample to (must be larger than 3).
```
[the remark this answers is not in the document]
JoepVanlier: 👍

lumicks/pylake/detail/tests/test_bead_crop.py
JoepVanlier: yeah, good call

lumicks/pylake/detail/bead_cropping.py
```diff
@@ -88,14 +105,178 @@ def find_beads(
...
+):
+    """Determine bead edges using a cross correlation procedure
+
+    This function seeks a bead-sized template that appears in subsequent frames.
+    It then moves inward by half the template size to provide bead edges.
+
+    Parameters
+    ----------
+    kymograph_image : np.ndarray
+        2D kymograph image
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, I'll add a commit for making this consistent. I agree. I just held off on it because I wanted to constrain this PR to the new functionality, but it's good to just do it now.

lumicks/pylake/detail/bead_cropping.py
JoepVanlier: Nah, I'm not too worried about perf here. Most of the time is gonna be spent in `correlate`. For `allow_movement=False`, there's a faster option (see last commit), but you get two pretty distinct branches then because of the normalization required. Not sure if this is worth the harder to read code for a function that is pretty fast.

lumicks/pylake/kymo.py
JoepVanlier: I've added them to the docstring. The reason I held off on it was because I didn't want to have to do breaking changes quickly. But I think the alpha warning covers that this functionality may change quickly.

The defaults should be good for most cases.

rpauszek (approved):
LGTM

PR #625, opened 2024-02-27
some final tasks before release

JoepVanlier:
**Why this PR?**
Running through all our notebooks, I noticed that there were a few rough edges that needed fixing before release.
In addition, this PR aims to fix a few last remaining issues and incorporates a `slack` variable into the bead edge detector based on review comments.

Unfortunately, it seems that for the widget to be able to use the new functionality at all, we do need to add them to the list of parameters. Initially I thought of adding them to the widget as direct parameters, but this would require ughly additional workarounds to make it work properly. For now I have added them non-UI visible because I feel the guidance on how to best choose it is still being discussed. Right now the default is set to the legacy value, but we could easily change that in the future if we want to.

I also had to remove a test for the zenodo API, which tested an intentionally invalid link. It started returning a different HTTP error and in retrospect, I don't think that test makes a lot of sense and it is bound to be brittle.

I think the other changes are fairly uncontroversial and are mainly minor docs changes. I will squash all the smaller changes before merge, but left them separate for easy review.

Docs build here: https://lumicks-pylake.readthedocs.io/en/release_v1_4_0/whatsnew/1.4.0/1_4_0.html

### DIFF
--- a/docs/examples/cas9_kymotracking/kymowidget.png
+++ b/docs/examples/cas9_kymotracking/kymowidget.png
@@ -1,3 +1,3 @@
 version https://git-lfs.github.com/spec/v1
-oid sha256:fb0b0c9c5cfec5c2c21cd5c0511282485495198c86484bb1c6640b46d89458b3
-size 141272
+oid sha256:49589f03786a0925c21475debd380bb9c43bfa03bfb07af2f89e20d41c843c91
+size 335322
--- a/docs/tutorial/kymographs.rst
+++ b/docs/tutorial/kymographs.rst
@@ -98,6 +98,15 @@ crop the region from `9.5` micron to `26` microns using the following command::
 
 .. image:: figures/kymographs/kymo_cropped.png
 
+If we know the bead diameter, we can automatically crop the kymo to an estimate of the bead edges using :meth:`~lumicks.pylake.kymo.Kymo.crop_beads()`.
+This can be convenient when batch processing many kymographs::
+
+    plt.figure()
+    kymo.crop_beads(4.84, algorithm="brightness").plot("rgb", aspect="auto", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+    plt.show()
+
+.. image:: figures/kymographs/kymo_cropped_bead_edges.png
+
 .. note::
 
     Note, slicing in time is currently only supported for unprocessed kymographs. If you want to both crop and slice a kymo,
--- a/docs/whatsnew/1.4.0/1_4_0.rst
+++ b/docs/whatsnew/1.4.0/1_4_0.rst
@@ -3,7 +3,7 @@ Pylake 1.4.0
 
 .. only:: html
 
-Get a sneak peak at the new features coming in `v1.4.0`:
+Pylake `v1.4.0` has been released with new features and improvements to existing analyses. Here’s some of the highlights:
 
 Hidden Markov Models
 --------------------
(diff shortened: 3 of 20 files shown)


lumicks/pylake/kymo.py
```diff
@@ -577,6 +577,7 @@ def estimate_bead_edges(
         algorithm,
         *,
         channel="green",
+        slack=0,
```
rpauszek, line 580: I don't think anyone is going to understand the intention of this argument from the name. Maybe something like `padding` or `extra_cropping`, `additional_cropping`?

lumicks/pylake/kymo.py
```diff
@@ -663,21 +666,27 @@ def estimate_bead_edges(
...
+        if np.diff(bead_edges) <= self.pixelsize_um[0]:
+            raise RuntimeError(
+                f"Detected bead edges in combination with selected slack ({slack:.2f}) lead to "
+                f"empty kymograph. Cropping region: {bead_edges[0]:.2f}, {bead_edges[1]:.2f}."
             )
 
-    def crop_beads(self, bead_diameter, algorithm, *, channel="green", **kwargs):
+        return bead_edges
+
+    def crop_beads(self, bead_diameter, algorithm, *, channel="green", slack=0, **kwargs):
```
rpauszek, line 689: same with the `slack` parameter as above

docs/examples/cas9_kymotracking/kymowidget.png
rpauszek, line 1: it's nit-picky but could you remake this with the light theme? the dark background is pretty jarring when everything else in the docs is light theme

docs/tutorial/kymographs.rst
```diff
@@ -98,6 +98,15 @@ crop the region from `9.5` micron to `26` microns using the following command::
 
 .. image:: figures/kymographs/kymo_cropped.png
 
+If we know the bead diameter, we can also do this automatically using :meth:`~lumicks.pylake.kymo.Kymo.crop_beads()`.
```
rpauszek, line 101: I think it's good anytime talking about this functionality we reinforce that it's an estimate and not guaranteed to be 100% accurate in all cases
```suggestion
If we know the bead diameter, we can automatically crop the kymo to an estimate of the bead edges using :meth:`~lumicks.pylake.kymo.Kymo.crop_beads()`.
```

docs/whatsnew/1.4.0/1_4_0.rst
```diff
@@ -18,4 +18,34 @@ Check out the :ref:`hmm-section` tutorial and the
...
 .. figure:: hmm_hairpin.png
 
-   HMM analysis of a tethered DNA hairpin held at three different bead separations.
\ No newline at end of file
+   HMM analysis of a tethered DNA hairpin held at three different bead separations.
+
+Automatic bead cropping
+-----------------------
+
+Added :meth:`~lumicks.pylake.kymo.Kymo.estimate_bead_edges()` and :meth:`~lumicks.pylake.kymo.Kymo.crop_beads()` for roughly cropping the beads out of a kymograph. This can help when batch processing of kymographs.
```
rpauszek, line 26:
```suggestion
Added :meth:`~lumicks.pylake.kymo.Kymo.estimate_bead_edges()` and :meth:`~lumicks.pylake.kymo.Kymo.crop_beads()` for quickly crop the beads out of a kymograph using an estimate of the bead edges. This can help when batch processing of kymographs.
```

docs/whatsnew/1.4.0/1_4_0.rst
```diff
@@ -18,4 +18,34 @@ Check out the :ref:`hmm-section` tutorial and the
...
+To do this, we added two additional parameters:
+
+- `filter_width` allows customizing the filter applied prior to detection.
+- `adjacency_filter` applies a filter on the detected peaks, removing any fluorescent peaks that do not have a detected peak in an adjacent frame.
+
+This allows using lower thresholds, while keeping false detections in check.
+
+.. figure:: tracking_comparison.png
+
+.. figure:: tracking_comparison_threshold.png
```
rpauszek, line 43: could you remake the plots using something like `viridis` for the colormap? I don't know if it's the colorblindness but I can barely distinguish the tracks on the green background. The perceptually uniform colormaps are anyways best practice for data visualization (at least in a single-channel format)

lumicks/pylake/simulation/diffusion.py
```diff
@@ -34,9 +34,11 @@ def simulate_wiener(sigma, num_steps, time_step):
 def _get_blank_kymo(*, line_time_seconds):
     """Generate kymo with particular properties"""
 
-    return _kymo_from_array(
+    kymo = _kymo_from_array(
         np.zeros((0, 0)), "r", line_time_seconds=line_time_seconds, pixel_size_um=1.0
     )
+    kymo._motion_blur_constant = 0
```
rpauszek, line 40: we don't have to do it now, but I guess we should add this as a parameter for `_kymo_from_array`

rpauszek (changes requested):
Almost good to go, just a couple minor requests

docs/whatsnew/1.4.0/1_4_0.rst
JoepVanlier: Good point. Switched to lime on gray.

lumicks/pylake/kymo.py
JoepVanlier: Fair point. Let's go with `extra_cropping`. `padding` makes me think we go towards the outside and additional is just longer.

lumicks/pylake/simulation/diffusion.py
JoepVanlier: The reason I would hold off on that for now is that I'd rather compute it from the exposure rather than expose the blur constant directly. I would pick this up if/when we start supporting diffusive analysis on camera images.

rpauszek (approved):
great! 🚀

PR #627, opened 2024-03-04
tests: force close tiff files

JoepVanlier:
**Why this PR?**
Garbage collection isn't run after each test / parametrization, which means that in some cases, handles can still be open which `pytest` raises an exception for. Best to just force close the TIFFs.

Normally, I would have preferred to use a context manager, but I feel providing this API on `ImageStack` would give the wrong impression of what happens in the following case:

```python
with lk.ImageStack("tiff.tiff") as im:
  im2 = im[50:100, 50:100]
```

If we were to close `im` here, `im2` would be invalidated. This could be avoided by opening the file only on demand and closing it immediately or copying. For now I would propose just fixing the tests.


rpauszek (approved):
🚢

PR #628, opened 2024-03-04
Replace hardcoded names in `File.__str__`

rpauszek:
just something I noticed and was a quick fix


JoepVanlier (approved):
🤣

lumicks/pylake/file.py
```diff
@@ -186,6 +186,9 @@ def print_force(field_name):
             calibration = "  .calibration\n" if field.calibration else ""
             return f".{field_name}\n{calibration}" if field else ""
 
+        rng = range(4)
```
JoepVanlier, line 189: minor optional: could do `range(1, 5)`and leave out the `+1` below?

PR #629, opened 2024-03-08
docs: fix some small oversights in the documentation

JoepVanlier:
**Why this PR?**
- We have systems out in the field with `100` kHz sampling rate. We should ideally not hardcode the `78125` Hz in our docs examples, because people may end up copying these and getting invalid results.
- We forgot to provide the public function `time_idx` with a docstring which makes it not show up as an API endpoint in the documentation.

### DIFF
--- a/docs/tutorial/force_calibration.rst
+++ b/docs/tutorial/force_calibration.rst
@@ -326,7 +326,7 @@ Let's have a look to see if this peak indeed appears in our power spectrum.
 To see it clearly, we reduce the blocking amount and show the spectrum all the way up to a frequency of 10 Hz::
 
     show_peak = lk.calculate_power_spectrum(
-        volts.data, sample_rate=78125, num_points_per_block=5, fit_range=(10, 23000)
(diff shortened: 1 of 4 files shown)


docs/tutorial/force_calibration.rst
```diff
@@ -326,7 +326,7 @@ Let's have a look to see if this peak indeed appears in our power spectrum.
 To see it clearly, we reduce the blocking amount and show the spectrum all the way up to a frequency of 10 Hz::
 
     show_peak = lk.calculate_power_spectrum(
-        volts.data, sample_rate=78125, num_points_per_block=5, fit_range=(10, 23000)
+        volts.data, sample_rate=volts.sample_rate, num_points_per_block=5, fit_range=(10, 23000)
```
tobiasjj, line 329: 👍

tobiasjj (approved):
LGTM! I did not double check that all occurrences of `sample_rate=78125` were replaced, but I assume that they were: 🚴

PR #630, opened 2024-03-08
imagestack: support two-color tiff stacks

JoepVanlier:
**Why this PR?**
Some systems in the field have the ability to export two-color tiff stacks. Would be good to add support for them.

*Note*
I have made the judgment call to map the export function to a three color tiff stack with the correct mapping, since it makes it more convenient to load into other software. Not many software packages support two color tiffs out of the box, besides, if people want the original two-color tiff, they have it already.

### DIFF
--- a/lumicks/pylake/detail/widefield.py
+++ b/lumicks/pylake/detail/widefield.py
@@ -322,6 +322,14 @@ class TiffStack:
             file.close()
 
 
+def _parse_wavelength(wavelength_field):
+    """Parse excitation wavelength"""
+    try:
+        return float(wavelength_field.split("/")[0])  # Format is wavelength / width
+    except ValueError:
+        return
+
+
 class ImageDescription:
     """Wrapper around TIFF ImageDescription tag and other pylake specific metadata.
 
@@ -381,11 +389,24 @@ class ImageDescription:
 
         self.channel_order = (0, 1, 2)
         if len(excitation_colors) == 2:
-            self.channel_order = [
+            self.channel_order = tuple(
                 ix
                 for ix, color in enumerate(("Red", "Green", "Blue"))
                 if color in excitation_colors
-            ]
+            )
+
+        excitation_filter_wavelengths = [
+            _parse_wavelength(self.json[f"Channel {channel} detection wavelength (nm)"])
+            for channel in range(3)
+            if f"Channel {channel} detection wavelength (nm)" in self.json
+        ]
+
+        if excitation_filter_wavelengths and all(excitation_filter_wavelengths):
+            # Validate that the wavelengths are in ascending order
+            if not all(np.diff(excitation_filter_wavelengths) < 0):
+                raise RuntimeError(
+                    f"Wavelengths are not in descending order {excitation_filter_wavelengths}"
+                )
 
         self._alignment_matrices = {
             rgb_channel: TransformMatrix.from_alignment(
@@ -477,6 +498,14 @@ class ImageDescription:
             for j in range(3):
                 if f"Channel {j} alignment" in out:
                     out[f"Applied channel {j} alignment"] = out.pop(f"Channel {j} alignment")
+
+        # Add some Pylake specific metadata
+        out["Pylake"] = {
+            "Channel mapping": {
+                ("red", "green", "blue")[channel_idx]: idx
+                for idx, channel_idx in enumerate(self.channel_order)
+            }
+        }
         return out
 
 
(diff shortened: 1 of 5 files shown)


agom (approved):
🟥 🟩 🟦

lumicks/pylake/detail/widefield.py
```diff
@@ -364,13 +375,23 @@ def __init__(self, tiff_file, align_requested):
...
-                    self._raw_alignment_matrix(channel), *self.offsets
-                ).invert()
-                for channel in range(3)
-            }
+        excitation_colors = [
+            key.split(" ")[0] for key in self.json.keys() if "Excitation Laser wavelength" in key
+        ]
+
+        # For two-color tiffs, the channel order may be different. Green is always channel 1
+        # and used as reference. The other channel will be stored in channel zero.
```
tobiasjj, line 383: I guess this is dictated by BL and there is no other way to get the channel order information directly from BL!?

tobiasjj (approved):
LGTM. It would be great if the assumption of the ordering of 2 channel images could be removed (see comment). 🚴

lumicks/pylake/detail/widefield.py
JoepVanlier: Yeah, actually, I want to verify this on a system this week.

The reference image is always the one in `channel[1]` (second channel). That makes the transformation matrices always in the order of the raw tiff channels.

In terms of relevant information, we have the following information available:
```
{'Alignment region of interest (x, y, width, height)': [0, 0, 2048, 682],
 'Blue Excitation Laser power level (%)': 2.0,
 'Blue Excitation Laser power on': 'true',
 'Blue Excitation Laser wavelength (nm)': 488,
 'Camera': 'WT',
 'Channel 0 alignment': [1.001670002937317,
  0.01721999980509281,
  -4.593649864196777,
  -0.017550000920891762,
  1.0012500286102295,
  33.25389099121094],
 'Channel 0 detection wavelength (nm)': '600/50',
 'Channel 1 alignment': [1.0, 0.0, 0.0, 0.0, 1.0, 0.0],
 'Channel 1 detection wavelength (nm)': '525/40',
 'Green Excitation Laser power level (%)': 2.0,
 'Green Excitation Laser power on': 'true',
 'Green Excitation Laser wavelength (nm)': 561,
```

Initially, I was under the impression that green would always have the unit transformation (but given the metadata I'm not so sure that is correct anymore).

If I look at the metadata that comes with the channels, I see that `Blue` is the one with the unit transform. I will add an extra commit that pulls the detection wavelengths from the BL metadata and uses those. I'd rather not hardcode the values, so I will order it by wavelength.

lumicks/pylake/detail/widefield.py
```diff
@@ -364,13 +383,39 @@ def __init__(self, tiff_file, align_requested):
...
+
+        self.channel_order = (0, 1, 2)
+        if len(excitation_colors) == 2:
+            self.channel_order = tuple(
+                [
+                    ix
+                    for ix, color in enumerate(("Red", "Green", "Blue"))
+                    if color in excitation_colors
+                ]
+            )
```
tobiasjj, line 398:
```suggestion
            self.channel_order = tuple(
                ix
                for ix, color in enumerate(("Red", "Green", "Blue"))
                if color in excitation_colors
            )
```

lumicks/pylake/detail/widefield.py
```diff
@@ -364,13 +383,39 @@ def __init__(self, tiff_file, align_requested):
...
+        excitation_filter_wavelengths = [
+            _parse_wavelength(self.json[f"Channel {channel} detection wavelength (nm)"])
+            for channel in range(3)
+            if f"Channel {channel} detection wavelength (nm)" in self.json
+        ]
+
+        if excitation_filter_wavelengths and all(excitation_filter_wavelengths):
+            # Sort by wavelength
+            self.channel_order = tuple(
+                self.channel_order[ix] for ix in np.argsort(excitation_filter_wavelengths)[::-1]
```
tobiasjj, line 409:
```suggestion
                self.channel_order[ix] for ix in np.argsort(excitation_filter_wavelengths[::-1])
```
I would reverse the order of the `excitation_filter_wavelengths` list to indicate that you want the order of colors with decreasing wavelengths, i.e. "red", "green" and "blue".

JoepVanlier: I don't think it's the same?

Let's say I have wavelengths `[2, 3, 1]`,  then `np.argsort(c)[::-1]` gives `[1, 0, 2]`, while `np.argsort(c[::-1])` gives `[0, 2, 1]`, but that's a moot point now since .

tobiasjj: Oh right, that was a quick shot :-(

PR #634, opened 2024-03-28
plot_correlated: add option to not downsample the channel data

JoepVanlier:
**Why this PR?**
Sometimes, you want to see things that happen within a frame. Currently, we downsample to the frame time. This PR makes this downsampling optional. This means you can just provide a slice that is downsampled the way you want it, and it'll just work.

This is especially important for `Scan` items, because they have long frame times.

This PR proposes adding a new argument for this. Which means you'd call `plot_correlated` or `export_video` with `downsample_channel=False` to get the behaviour where it doesn't downsample extra. The default is `True` for backwards compatibility; but if the old behavior is so bad that one might consider it faulty, we could consider making the default `False`.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * Added support for loading two-color `TIF` files with [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack).
 * Made ([`CalibrationResults`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html#calibrationresults)) callable to evaluate the fitted model power spectral density at the specified frequencies.
 * Added `fitted_params` field to ([`CalibrationResults`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html#calibrationresults)) for convenience.
+* Added option to disable downsampling channel data to frame rates with correlated plots ([`ImageStack.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.plot_correlated), [`Scan.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.Scan.html#lumicks.pylake.Scan.plot_correlated)) and exported videos ([`ImageStack.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.export_video), [`Scan.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.Scan.html#lumicks.pylake.Scan.export_video)) using `downsample_channel=False`.
 
 ## v1.4.0 | 2024-02-28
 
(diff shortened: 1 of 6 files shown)


changelog.md
```diff
@@ -2,7 +2,10 @@
 
 ## v1.5.0 | t.b.d.
 
+#### New features
+
 * Added support for loading two-color `TIF` files with [`ImageStack`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack).
+* Allow making correlated plots ([`ImageStack.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.plot_correlated)) and videos [`ImageStack.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.export_video) of correlated plots without downsampling the time channel data to the frame timestamps by passing the argument `downsample_channel=False`.
```
rpauszek, line 8:
```suggestion
* Added option to disable downsampling channel data to frame rates with correlated plots ([`ImageStack.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.plot_correlated), [`Scan.plot_correlated()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.Scan.html#lumicks.pylake.Scan.plot_correlated)) and exported videos ([`ImageStack.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.export_video), [`Scan.export_video()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.Scan.html#lumicks.pylake.Scan.export_video)) using `downsample_channel=False`.
```

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -148,6 +149,8 @@ def export_video(
             When specified, we export a video correlated to channel data
         vertical : bool, optional
             Render with the plots vertically aligned (default: True).
+        downsample_channel : bool, optional
+            Downsample the channel data over the frame (default: True).
```
rpauszek, line 153: perhaps:
```suggestion
            Downsample the channel data over frame timestamp ranges (default: True).
```
(would be for the other docstrings too)

lumicks/pylake/nb_widgets/correlated_plot.py
```diff
@@ -46,10 +47,16 @@ def plot_correlated(
...
+        Downsample the channel data over the frame (default: True).
     """
     import matplotlib.pyplot as plt
 
-    downsampled = channel_slice.downsampled_over(frame_timestamps, where="left", reduce=reduce)
+    downsampled = (
+        channel_slice.downsampled_over(frame_timestamps, where="left", reduce=reduce)
+        if downsample_channel
+        else channel_slice[frame_timestamps[0][0] : frame_timestamps[-1][-1]]
+    )
```
rpauszek, line 59: nitpick: I wonder if it'd be clearer to overwrite `channel_slice` here since `downsampled` doesn't necessarily make sense anymore?

rpauszek (approved):
LGTM, just some minor wording suggestions if you like

lumicks/pylake/nb_widgets/correlated_plot.py
JoepVanlier: I also doubted about this. I'm not a huge fan of the word downsample anymore, but we can't replace it since for the downsampled case, we need the original channel once below for the last frame.

How about `processed_channel`? I could also go for `downsampled_or_sliced_channel` but it feels painfully verbose.

PR #636, opened 2024-03-29
calibration: expose an estimate of the local drag

JoepVanlier:
**Why this PR?**
It may be confusing that we calculate back our drag coefficient to bulk. The reason we do this is so that experiments can be compared and drag coefficients can be carried over from one calibration axis to the other.

However, it may not always be immediately clear how to calculate this value back to the surface-relevant value.

Since one may require the local drag coefficient for other analyses, it is a good idea to also expose a local drag coefficient.

When not using hydro, the drag coefficient we calculate _is_ the local drag.
When hydro _is_ active, then the drag coefficient we obtain with the model is the drag we would have obtained in bulk (all the height dependence is in the model). That means that we have to explicitly convert that drag coefficient to a "local" drag. For consistency, we do this with the same hydrodynamic model.

### DIFF
--- a/lumicks/pylake/force_calibration/convenience.py
+++ b/lumicks/pylake/force_calibration/convenience.py
@@ -96,7 +96,10 @@ def calibrate_force(
     excluded_ranges : list of tuple of float, optional
         List of ranges to exclude specified as a list of (frequency_min, frequency_max).
     drag : float, optional
-        Overrides the drag coefficient to this particular value.
+        Overrides the drag coefficient to this particular value. Note that you want to use the
+        bulk drag coefficient for this (obtained from the field `gamma_ex`). This can be used to
+        carry over an estimate of the drag coefficient obtained using an active calibration
+        procedure.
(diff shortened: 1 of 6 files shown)


lumicks/pylake/force_calibration/convenience.py
```diff
@@ -96,7 +96,10 @@ def calibrate_force(
     excluded_ranges : list of tuple of float, optional
         List of ranges to exclude specified as a list of (frequency_min, frequency_max).
     drag : float, optional
-        Overrides the drag coefficient to this particular value.
+        Overrides the drag coefficient to this particular value. Note that you want to use the
+        bulk drag coefficient for this (obtained from the field gamma_ex). This can be used to
```
rpauszek, line 100:
```suggestion
        bulk drag coefficient for this (obtained from the field `gamma_ex`). This can be used to
```

rpauszek (approved):
🚀

PR #637, opened 2024-04-10
plotting: rename to `downsample_to_frames`

JoepVanlier:
**Why this PR?**
This parameter name is clearer for end users. Given that it is unreleased functionality it's best to make this breaking change now.


alessiamarcolini (approved):

aafkevandenberg (approved):
Great, this is much more clear!

PR #638, opened 2024-04-10
kymo: add `plot_with_channels`

JoepVanlier:
**Why this PR?**
It would be nice to just plot any channel data alongside kymos and to have control over how these get downsampled.

With this new call (`kymo.plot_with_channels()`), you can make plots like the following with a single call:

image

and

image

Note that since not all our channels have y-label metadata (most do not), the y-labels are customizable (by passing a list of labels).

Docs [here](https://lumicks-pylake.readthedocs.io/en/plot_with_channel/tutorial/kymographs.html#correlating-with-channel-data) and [here](https://lumicks-pylake.readthedocs.io/en/plot_with_channel/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_with_channels).

I figured that since we are generalizing this behavior and making it more composable, users may also want to be able to plot more than one correlated channel at once. To reduce friction, the function accepts both a single slice or multiple.

### DIFF
--- a/docs/tutorial/kymographs.rst
+++ b/docs/tutorial/kymographs.rst
@@ -278,6 +278,24 @@ We achieved this by passing `np.sum` to the `reduce` parameter of :func:`~lumick
 This results in summing the photon counts rather than taking their average.
 The argument `title_vertical=True` places the channel names along the y-axis instead of the axis title allowing a slightly more compact plot.
 
+Note that the plot can be further customized by specifying custom `labels`, `colors` and a `scale_bar`::
+
+    kymo.plot_with_channels(
+        [
+            file.force1x.downsampled_by(100),
+            file["Photon count"]["Green"].downsampled_over(kymo.line_timestamp_ranges(), reduce=np.sum),
+        ],
+        "rgb",
+        adjustment=lk.ColorAdjustment(5, 98, "percentile"),
+        aspect_ratio=0.2,
+        title_vertical=True,
+        scale_bar=lk.ScaleBar(10.0, 5.0),
+        colors=[[1.0, 0.2, 0.5], "green"],
+        labels=["My force", "My photons"],
+    )
+
+.. image:: ./figures/kymographs/kymo_plot_with_channels_customized.png
+
 There is also a convenience function :meth:`~lumicks.pylake.kymo.Kymo.plot_with_force` to plot a kymograph along with a
 downsampled force trace::
 
--- a/docs/tutorial/kymographs.rst
+++ b/docs/tutorial/kymographs.rst
@@ -235,15 +235,15 @@ and therefore each pixel no longer has an identifiable time. For example, we can
 Additionally, a downsampled kymograph cannot be sliced (same as cropped kymographs mentioned above). Therefore you should
 first slice the kymograph and then downsample.
 
-Correlating with force
-----------------------
+Correlating with channel data
+-----------------------------
 
 We can downsample channel data according to the lines in a kymo. We can use
 :func:`~lumicks.pylake.kymo.Kymo.line_timestamp_ranges()` for this::
 
     line_timestamp_ranges = kymo.line_timestamp_ranges()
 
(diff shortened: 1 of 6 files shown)


tobiasjj (approved):
Great new feature that allows users to quickly visualize and correlate kymograph images with arbitrary channel data. Documentation is clear and concise. Code is well tested. 🚴

docs/tutorial/kymographs.rst
```diff
@@ -267,3 +267,25 @@ The function can also take a dictionary of extra arguments to customize the kymo
 These parameter values get forwarded to :func:`matplotlib.pyplot.imshow`.
 
 .. image:: ./figures/kymographs/kymo_correlated.png
+
```
aafkevandenberg, line 270: I'd introduce `kymo.plot_with_channels` first. And after that introduce `kymo.plot_with_force` as a convenience function for a specific user case.

lumicks/pylake/kymo.py
```diff
@@ -413,37 +414,60 @@ def plot(
...
-        color_channel,
+        channels,
+        color_channel="rgb",
+        *,
         aspect_ratio=0.25,
-        reduce=np.mean,
         kymo_args=None,
         adjustment=no_adjustment,
+        title_vertical=False,
+        labels=None,
```
aafkevandenberg, line 426: I have no idea how hard this is, but is it possible to allow for other kinds of plot arguments, such as the color of the plot line? Not essential, but would be nice. The most common things users may want to change the color of the line and maybe to add a legend.

aafkevandenberg (commented):
Looks great! Added a few small comments.

lumicks/pylake/kymo.py
JoepVanlier: Any remaining keyworded arguments get forwarded to `Slice.plot()`.

So you could already do:
```
kymo.plot_with_channels(
    [
        h5_file.force1x.downsampled_by(100),
        h5_file["Photon count"]["Green"].downsampled_over(kymo.line_timestamp_ranges(), reduce=np.sum),
    ],
    "rgb",
    adjustment=lk.ColorAdjustment(5, 98, "percentile"),
    aspect_ratio=0.2,
    title_vertical=False,
    color="blue",
)
```

resulting in

image

for example.

What do you mean by a legend? Every graph has its own subplot so I don't see where one would add a legend. Or do you mean a scalebar?

That is possible since you can also pass kymograph arguments. Unfortunately, that is a little more complicated (because either the kymo customization or the line plots can be forwarded easily, but not both).

```
kymo.plot_with_channels(
    [
        h5_file.force1x.downsampled_by(100),
        h5_file["Photon count"]["Green"].downsampled_over(kymo.line_timestamp_ranges(), reduce=np.sum),
    ],
    "rgb",
    adjustment=lk.ColorAdjustment(5, 98, "percentile"),
    aspect_ratio=0.2,
    title_vertical=False,
    kymo_args={"scale_bar": lk.ScaleBar(10.0, 5.0)},
    color="blue",
)
```

returns:

image

aafkevandenberg: I actually meant individual colors. (eg, plot the green photon counts in green, the force in black). And I meant a legend, but I agree that that does not add much value.

JoepVanlier: Maybe I should add it as example in the docs.

JoepVanlier: Alright. I've added a `colors` and `scale_bar` keyword arg to allow for easier customization. I've also added a section that demonstrates these to the docs.

JoepVanlier: Alright, `colors` and `labels` are now exposed as well.
image

tobiasjj (approved):
The extra customization options requested are a great addition and well explained in the docs. Slicing the kymo channel to its own limits before plotting seems like a useful addition. 🚴

PR #640, opened 2024-04-12
ci: add option to serialize test dictionaries

JoepVanlier:
**Why this PR?**
We have a convenience fixture for storing reference values for tests called `reference_data`. These functions store reference results in `npz` files. That's nice for arrays and such.

Sometimes however, there is merit to comparing a set of simple values, and to also quickly be able to see their changes in a diff when a PR comes up. For this I would propose adding a simple convenience function that basically performs a similar function but provides a human readable diff.


lumicks/pylake/conftest.py
```diff
@@ -184,27 +218,68 @@ def get_reference_data(reference_data, test_name=None, file_name=None):
...
-
-            np.savez(reference_file_path, reference_data)
             print(f"\nWritten reference data {ref_data_filename} to {reference_file_path}.")
             return reference_data
 
     return get_reference_data
+
+
+@pytest.fixture
+def compare_to_reference_dict(request, reference_data):
```
tobiasjj, line 242: Is the `request` fixture needed here?

lumicks/pylake/conftest.py
```diff
@@ -184,27 +218,68 @@ def get_reference_data(reference_data, test_name=None, file_name=None):
...
+    Intended to store and/or compare to a reference dictionary stored in a human-readable json
+    file. Only intended for single numeric values. See `reference_data` for usage information.
+
+    Examples
+    --------
+    ::
+
+        @pytest.mark.parametrize()
+        def test_freezing(compare_to_reference_dict):
+            compare_to_reference_dict({"a": 5, "b": 1e-12, rtol=1e-6})
```
tobiasjj, line 254: Did you intend for `rtol=1e-6` to be a part of the data?
```suggestion
            compare_to_reference_dict({"a": 5, "b": 1e-12}, rtol=1e-6)
```

tobiasjj (approved):
LGTM! I would probably add another unit test for `compare_to_reference_dict` where the comparison fails, see also my other two comments. 🚴

lumicks/pylake/conftest.py
```diff
@@ -184,27 +218,68 @@ def get_reference_data(reference_data, test_name=None, file_name=None):
...
+        all_keys = set(ref_data.keys()) | set(data.keys())
+
+        difference_strings = []
+        same = True
+        for key in all_keys:
+            info, match = check_similarity(key, data, ref_data)
+            difference_strings.append(info)
+            same = same and match
+
+        if not same:
```
tobiasjj, line 280:
```suggestion
        difference_strings, match = zip(*[check_similarity(key, data, ref_data] for key in all_keys)

        if not all(match):
```

lumicks/pylake/conftest.py
JoepVanlier: No! 😅

lumicks/pylake/conftest.py
JoepVanlier: Good catch! It was required in a previous implementation, but isn't anymore.

PR #641, opened 2024-04-12
calibration: provide error estimates for kappa and Rd

JoepVanlier:
**Why this PR?**
Currently, we provide error estimates for the parameters fitted when fitting the passive calibration spectrum.

Of interest are the estimation errors on `kappa` and `Rd` however. This PR does some basic error propagation to provide an error estimate for those quantities.

Note that these error estimates propagate errors that arise from fitting which neglects errors in temperature, height above the surface and viscosity (which likely dominate the experimental error, but for which we need pieces of information that are experiment-dependent (e.g. bead variability, how much the laser warms up the fluid, z-drift etc)). Note that they also assume sufficient blocking and averaging has taken place.


lumicks/pylake/force_calibration/calibration_models.py
```diff
@@ -828,18 +838,19 @@ def __init__(
         )
         self.driving_frequency_guess = driving_frequency_guess
         self.sample_rate = sample_rate
-        self.num_windows = num_windows
+        self.num_windows = num_windows  # caveat: misnamed parameter
```
rpauszek, line 841: perhaps mark this with `# todo: ` instead of `caveat`? I'll sometimes grep for `todo` to see if there's things to pick up

rpauszek (approved):
looks good, but double check in the docs I think there's some places we print out the results structure where these new ones should be included now

JoepVanlier:
> looks good, but double check in the docs I think there's some places we print out the results structure where these new ones should be included now

Good point. We had two in the force calibration tutorial. I replaced them now.

PR #642, opened 2024-04-15
fdfit: better error handling `parameter_trace()` 

JoepVanlier:
**Why this PR?**
A few tiny fixes to reduce API friction based on a use case where we do not fit a model first.

`parameter_trace()` can be used to fit a single parameter on a per datapoint basis while keeping all others fixed. Typically, one would fit a model to _different_ data, and then use that model to compute such a parameter trace. However, there are use cases for when one just wants to do it with a set of parameters obtained earlier (or via other means).

Currently, users might attempt to pass a regular dictionary with values to `parameter_trace()` analogously to what we allow for `model.__call__()`. The error you get when you do that is not helpful, nor does the docstring indicate any way to do this.

One thing I doubted about is whether we should just fall back to the default bounds specified in the model for the parameter to be profiled if bounds are unavailable, but I decided against it, because one could start hitting those limits and users would not really understand where the limits might be coming from. I think a better approach is to point them to the correct way to obtain bounded parameters instead.

I also noticed that we don't warn when we notice that the values start hitting the bounds. I added a warning for this case, as users probably want to change their fitting limits or check their data when this happens. Unfortunately, enforcing some limits is still preferable to none, to avoid non-physical behaviour (negative contour lengths and such).

Docs build here: https://lumicks-pylake.readthedocs.io/en/fdfit_model/_api/lumicks.pylake.parameter_trace.html

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -14,6 +14,7 @@
 #### Improvements
 
 * Added error message when parameters are passed to [`lk.parameter_trace()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.parameter_trace.html) that do not have the required attributes. 
+* Warn when parameter estimates are hitting the fitting bounds when using [`lk.parameter_trace()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.parameter_trace.html).
 
 ## v1.4.1 | t.b.d.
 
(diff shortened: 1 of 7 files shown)


aafkevandenberg (commented):
Thank you for the additions! I had to use this function without fit yesterday, because I did not have a regular FD curve available. It is useful if you want to check whether the data make sense, independent of the FD fit. For the final analysis, it is of course recommended to do the FD fit first.
Can you explain more about the bounds that you mention in your description? I assumed that all parameters are completely fixed, except for the one that is being refitted. And that one is completely free. But I guess that assumption is wrong.

changelog.md
```diff
@@ -9,6 +9,12 @@
...
+* Allow accessing FdFit model defaults through key access notation (i.e. `model["m/Lc"]`).
+
+#### Improvements
+
+* Added error message when parameters are passed to [`lk.parameter_trace()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.parameter_trace.html) that do not have the required attributes. 
+* Warn when estimates are hitting the parameter bounds when using [`lk.parameter_trace()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.parameter_trace.html).
```
rpauszek, line 17:
```suggestion
* Warn when parameter estimates are hitting the fitting bounds when using [`lk.parameter_trace()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.parameter_trace.html).
```

lumicks/pylake/fitting/parameter_trace.py
```diff
@@ -9,6 +11,20 @@ def parameter_trace(model, params, inverted_parameter, independent, dependent, *
...
+        parameter estimate does not hit these bounds. Note that `Pylake` will issue a warning
+        whenever estimates are at the bounds.
+
+        For example, when creating a worm-like chain model the parameter `contour length` is bounded
+        between a very small value (corresponding to the length of one base pair) and infinity
+        (since zero and negative values are non-physical) by default. If the contour length is then
+        estimated to be exactly at the lower bound by this function, a warning is issued. Note
+        that these bounds can be modified by changing them through the `lower_bound` and
+        `upper_bound` attributes of a `Parameter`.
+
```
rpauszek, line 27:
```suggestion
    .. note::

        Inverting the model with respect to a particular parameter enforces any bounds
        defined for that parameter in `params`. Results can only be trusted for values where the
        parameter estimate does not hit these bounds. Note that `Pylake` will issue a warning
        whenever this condition is met.

        For example, when creating a worm-like chain model the parameter `contour length` has a lower bound
        of the length of one base pair by default,
        since zero and negative values are non-physical. If the contour length is then
        estimated to be exactly at the lower bound by this function, a warning is issued. Note
        that these bounds can be modified by changing them through the `lower_bound` and
        `upper_bound` attributes of a `Parameter`.

```

rpauszek (approved):
looks good, just some minor wording suggestions

PR #644, opened 2024-05-03
tests: hash parameterized tests for shorter paths

JoepVanlier:
**Why this PR?**
We typically serialize some tests for more convenient reference data handling.

Using the full parametrization in the filename is ill-advised as that can lead to very long filenames which can become a problem on windows (with the 256 character path limit).

Because this risks making the human readability of the `json`-based tests less helpful, I have decided to also modify the json serialization such that we also store the parametrization in the json, such that we don't use the ability to quickly evaluate in the diff what changed. I think it's actually a lot nicer than what we had before because encoding it in the filename didn't always result in something useful (as they get clipped in previews).

Seeing something like:

image

gives you a very quick overview of what changed in the test results for a particular parametrization.

I have also changed the directory name from `reference_data` to `ref_data` to reduce the character count a little bit more.


lumicks/pylake/conftest.py
```diff
@@ -107,14 +108,37 @@ def configure_mpl():
...
+    """Hash a filename for a test
+
+    Parameters
+    ----------
+    string_to_hash : str
+        String to return the hash of.
+
+    We use a base36 hash to ensure that we can safely store these hashes as filenames on all
+    operating systems. Base64 contains /, and has issues with case sensitivity on windows. Base36
+    is still shorter than a hexadecimal hash.
```
dean0x7d, line 121: You could use `urlsafe_b64encode()` which doesn't produce `/`: https://docs.python.org/3/library/base64.html#base64.urlsafe_b64encode

Also, take a look at this https://stackoverflow.com/questions/2510716/short-python-alphanumeric-hash-with-minimal-collisions. Something like this might work: 
```py
hash = hashlib.md5(string_to_hash.encode())
b64 = base64.urlsafe_b64encode(hash.digest())
return b65[:10]
```

lumicks/pylake/force_calibration/tests/ref_data/good_fit_integration_test/1VTOTCZXAH24I3XX2E0PKZN7N.json
dean0x7d, line 1: It's a bit funny that this test is called `test_..._test`. Maybe drop the trailing `_test`?

dean0x7d (approved):
Nice work! Having readable parameters inside the file is a great upside of this! The code looks very good. Just see my suggestion about `urlsafe_b64encode()`. One more question: Have you considered putting the test name inside the reference file instead of the folder? Or would that be going too far?

lumicks/pylake/conftest.py
JoepVanlier: I suspect base64 would still lead to a lot of collisions since unlike URLs, I don't think the windows file system is case sensitive?

JoepVanlier:
> One more question: Have you considered putting the test name inside the reference file instead of the folder? Or would that be going too far?

You mean a flat folder then? That's not a bad idea. But I think it would only work for the human readable formats like json. But then again, those are the only ones where you'd really tend to do visual inspection.

I'll try it tomorrow and play around a bit with it to see how I end up liking it.

lumicks/pylake/conftest.py
dean0x7d: Right, the case. Good thinking. Yeah, it would be good to add it to the comment.

rpauszek (approved):
:shipit:

PR #646, opened 2024-05-08
Support python 3.12 / notebook>=7 and drop support for python 3.9.

JoepVanlier:
**Why this PR?**
It's good to stay relatively up to date with Python versions. This PR drops support for Python `3.9` and adds support for `3.12`.

We could consider keeping `3.9` for one more release and immediately drop it on the next (to ease any transition issues) or just rip the band-aid off now. I have a mild preference for the latter, since `3.9` is old, and all the dependency changes we _are_ already doing for `jupyter notebook` might require a new environment for some anyway.

*Jupyter*
With `3.12` comes a necessary change, which is the switch to jupyter `notebook>=7`. Reason being that `notebook<7` pins us to `pyzmq<25`. For a while, they [added](https://github.com/jupyter/notebook/pull/6749) this pin officially, but they have since dropped [this](https://github.com/jupyter/notebook/pull/7305) since binaries for `pyzmq<25` are not being created for python 3.12 and this effectively locks you into old versions of python. That said, I did do a bunch of testing, and for us, without the pin, I still get `Uncaught exception in ZMQStream callback` after a while on windows, after which the interactive widgets stop working responding entirely (this is the issue that was the original reason for the pin).

We should update. I have tested the latest installation instructions with `notebook>7` on `conda/windows`, `pip/windows` and `pip/mac`and I think we should just do it. If there are issues, we will resolve them as they come. Staying with something that is essentially deprecated is just kicking the can down the road.

*Interactive widget*
Since we do want interactive widgets, I have added `ipympl` as an extra dependency.
Note that this means using the magic `%matplotlib widget` rather than `%matplotlib notebook`.

*conda*
I have also dropped some of the special conda instructions in the installation guide, as it seems they are also no longer needed.

Docs build [here](https://lumicks-pylake.readthedocs.io/en/py312/install.html). Most relevant is probably the [FAQ](https://lumicks-pylake.readthedocs.io/en/py312/install.html#frequently-asked-questions).

Note that the tests aren't green because the `3.9` action isn't being run any more. I will flip the switch on that once I get a greenlight on this PR.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -418,18 +418,6 @@ To enable interactive plots, you have to invoke the correct `magic commands <htt
 in the notebook. When using Jupyter notebook, the following command will switch the `matplotlib` backend from the inline
 one (which renders images) to the interactive backend::
 
-    %matplotlib notebook
-
-You can also choose to install `ipympl`, which can perform better in some cases. You can install it with `pip`::
-
-    pip install ipympl
-
-or `conda`::
-
-    conda install -c conda-forge ipympl
-
-The `ipympl` backend can be activated by invoking the following magic command in a notebook::
-
     %matplotlib widget
 
 *Note that switching backends typically requires you to restart the Jupyter kernel*. When using JupyterLab, `ipympl` is
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -1,7 +1,7 @@
 Installation
 ============
 
-Pylake can be installed on Windows, Linux and Mac and requires Python 3.9 or newer.
+Pylake can be installed on Windows, Linux and Mac and requires Python 3.10 or newer.
 
 .. note::
 
@@ -19,7 +19,7 @@ The easiest way to install Python and SciPy is with `Anaconda`_, a free scientif
 
 .. rubric:: Windows
 
-#. Go to the `Anaconda`_ website and download the Python 3.9 (or newer) installer.
+#. Go to the `Anaconda`_ website and download the Python 3.10 (or newer) installer.
 
 #. Run it and accept the default options during the installation.
 
@@ -69,7 +69,7 @@ This concludes the Pylake installation procedure. Check out the :doc:`Tutorial <
 
 .. rubric:: Linux
 
-#. Go to the `Anaconda`_ website and download the Python 3.9 (or newer) installer.
+#. Go to the `Anaconda`_ website and download the Python 3.10 (or newer) installer.
 
 #. Open a terminal window and run::
 
@@ -112,7 +112,7 @@ This concludes the Pylake installation procedure. Check out the :doc:`Tutorial <
 
 .. rubric:: macOS
 
-#. Go to the `Anaconda`_ website and download the Python 3.9 (or newer) installer.
+#. Go to the `Anaconda`_ website and download the Python 3.10 (or newer) installer.
 
 #. Run it and accept the default options during the installation.
 
@@ -147,7 +147,7 @@ This concludes the Pylake installation procedure. Check out the :doc:`Tutorial <
 Installation using pip
 ----------------------
 
-If you're already familiar with Python and have Python >= 3.9 installed, installing Pylake on Windows, Linux or Mac can be done using `pip`, Python's usual package manager::
+If you're already familiar with Python and have Python >= 3.10 installed, installing Pylake on Windows, Linux or Mac can be done using `pip`, Python's usual package manager::
 
     pip install lumicks.pylake
 
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -320,6 +320,33 @@ Important to note is that packages on `conda` and `pip` are typically *not* comp
 Frequently asked questions
 --------------------------
 
+**Why are the plots in my notebook not interactive?**
+
+Jupyter notebooks support the use of interactive figures.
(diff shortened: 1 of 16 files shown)


setup.py
```diff
@@ -66,11 +66,11 @@ def read(filename):
...
     extras_require={
         "notebook": [
-            # Notebook upper limit is a workaround for issues with IPython not being defined.
-            "notebook>=4.4.1,<7",
+            "notebook>=7",
             "ipywidgets>=7.0.0",
-            "jupyter_client<8",  # https://github.com/jupyter/notebook/issues/6748
-            "pyzmq<25",  # https://github.com/jupyter/notebook/issues/6748
+            "jupyter_client>=8",
+            "ipympl",
```
dean0x7d, line 72: It would be nice to have some minimum defined here.

setup.py
```diff
@@ -66,11 +66,11 @@ def read(filename):
...
+            "pyzmq",
```
dean0x7d, line 73: Is `pyzmq` even needed here anymore? It's not a direct dependency and we don't pin it. So probably doesn't even need to be in the list.

dean0x7d (approved):
The updates to the docs look great! Nice to be moving forward.

docs/install.rst
```diff
@@ -326,6 +320,35 @@ Important to note is that packages on `conda` and `pip` are typically *not* comp
...
 
+**Why are the plots in my notebook not interactive?**
+
+Jupyter notebooks support the use of interactive figures.
+To enable interactive plots, you have to invoke the correct `magic commands <https://ipython.readthedocs.io/en/stable/interactive/magics.html>`_ in the notebook.
+When using Jupyter notebook, the following command will switch the `matplotlib` backend from the inline one (which renders images) to the interactive backend::
+
+    %matplotlib widget
+
+When using JupyterLab, `ipympl` is the only backend that provides interactive plots with Pylake.
```
aafkevandenberg, line 331: What do you want to tell the user with this comment? That for Jupyter Lab you also need `%matplotlib widget`?

setup.py
JoepVanlier: Nope. Removing...

docs/install.rst
```diff
@@ -326,6 +320,35 @@ Important to note is that packages on `conda` and `pip` are typically *not* comp
...
+
+.. note::
+
+    Switching plotting backends typically requires you to restart the Jupyter kernel.
+
+**I tried using interactive plots using Jupyter notebook, but get the error `Javascript Error: IPython is not defined`**
+
```
aafkevandenberg, line 338: I would start with an instruction how to solve the issue, an then add the explanation. So, something like:

Starting from Pylake `1.4.0`, Pylake uses a more recent version of Jupyter Notebook (`notebook >=7)`. This newer version requires a different backend: `%matpltolib notebook` has to be replaced by `%matplotlib widget`.

Note that switching between interactive backend requires a kernel restart (select kernel -> restart from the menu at the top of the Jupyter Notebook)

The reason for the update is that the backend which handled interactive figures in `notebook<7` is no longer supported by Jupyter Notebooks. See also the matplotlib documentation ...

aafkevandenberg (approved):
Looking good. It would be good to test the update on several laptops internally before release.

docs/install.rst
JoepVanlier: Yeah, good idea to start with the solution.

I'd rephrase it a little bit, because Pylake didn't really switch anything internally though. You could in principle install with `pip` without `[notebook]`, install `notebook<7` yourself and get the old situation. It would still work.

All we are doing is updating what we install by default on `conda` (what we ship with essentially).

```
Starting from Pylake `1.4.0`, Pylake ships with a more recent version of Jupyter Notebook (`notebook >=7`).
This newer version requires a different backend when using interactive figures: `%matplotlib notebook` has to be replaced by `%matplotlib widget`.
Note that switching between interactive backends requires a kernel restart (select kernel -> restart from the menu at the top of the Jupyter Notebook).

The reason for this change is that the backend which handled the interactive figures in `notebook<7` is no longer supported by Jupyter notebooks.
Moving forward, `ipympl` will be used for interactive figures.
See also the `matplotlib documentation <https://matplotlib.org/stable/users/explain/figure/interactive.html#jupyter-notebooks-jupyterlab>`_ on this.
```

docs/install.rst
JoepVanlier: I dropped the second sentence. The fact that it uses `ipympl` is probably less interesting to users.

JoepVanlier:
> Looking good. It would be good to test the update on several laptops internally, MacOS and Windows, before release.

Is there a particular configuration or installation procedure that you are worried about?

I have tested this on three machines. My desktop at home (Windows, w/pip), my DELL at work (Windows, both pip and conda) and my work mac (MacOS with pip inside a conda environment and with conda itself). I also tried installing the dependencies for the last version of Pylake on Windows and then updating to the latest and that also worked.

You're free to try it on additional machines, but to me it seems like it would lead to diminishing returns.

Especially since that to install latest, the installation method is always going to be slightly different than what we would truly have on a release.

aafkevandenberg (commented):
Having it tested on 3 devices with 2 different operating systems should be sufficient. I will then test it right after release as well.

setup.py
JoepVanlier: Yeah, good catch on this front actually, since we need the latest `ipympl` for the version of `matplotlib` that we are using.

rpauszek (approved):
fantastic! 🚀🚀🚀🚀

PR #600, opened 2023-11-30
export_movie: add the option to correlate a channel slice when exporting to a movie

JoepVanlier:
**Why this PR?**
Highly requested feature. This feature allows you to export channel data alongside a movie. Users want these kind of movies in talks and whatnot.

I've manually tested all permutations I could think of (cropping by time, pixels, including color adjustments for both scans and image stacks). I tested on jupyter notebook (with `notebook` backend), jupyter lab (with `widget` backend) and on the non-interactive backend.

test_scan
with_channel

Docs [here](https://lumicks-pylake.readthedocs.io/en/correlated_movie/whatsnew/1.3.0/1_3_0.html), [here](https://lumicks-pylake.readthedocs.io/en/correlated_movie/tutorial/imagestack.html#stack-plotting) and [here](https://lumicks-pylake.readthedocs.io/en/correlated_movie/tutorial/scans.html#confocal-plotting). API docs [here](https://lumicks-pylake.readthedocs.io/en/correlated_movie/_api/lumicks.pylake.scan.Scan.html#lumicks.pylake.scan.Scan.export_video) and [here](https://lumicks-pylake.readthedocs.io/en/correlated_movie/_api/lumicks.pylake.ImageStack.html#lumicks.pylake.ImageStack.export_video).

PR #647, opened 2024-05-14
video export: fix two bugs in video export

JoepVanlier:
**Why this PR?**
Working on the what's new section, I noticed two bugs in the video export functionality.

The first is that the final time step when not downsampling is too big (it takes the frame time step rather than the channel time step we should actually be using when not downsampling to frames).

Without it, you get this:
image

This feature was never released (therefore I did not include a changelog entry).

The second bug is that the `start_frame` argument is ignored. This bug was introduced in Pylake 1.3.0 in this [PR](https://github.com/lumicks/pylake/pull/600).

Note that I am targeting `main`, since we want to release `1.5.0` soon and have no plans for `1.4.1`, but we could easily backport the `start_frame` fix if there is a need.

In both cases, I've hardened the tests to cover these specific errors.


rpauszek (approved):
🚢

PR #651, opened 2024-05-28
tests: fix test for tracking `ImageStack`-based `Kymo`

JoepVanlier:
*Why this PR?*
Unfortunately, one of our tests is broken.

The test in question tests whether a `Kymo` coming from an `ImageStack` can be tracked successfully.

`track_greedy` raises when the threshold is equal or smaller than the minimum pixel value.

The minimum pixel value of the `Kymo` exported from the `ImageStack` here is `3`. That means this test can randomly fail during tracking, depending on whether the floating point pixel value (after filtering) is slightly above or below this critical value.

It seems that numerically, we typically ended up with `2.9999999999999996` for the minimum pixel value. Meaning the threshold would be above.

### DIFF
--- a/lumicks/pylake/tests/test_imaging_camera/test_image_stack.py
+++ b/lumicks/pylake/tests/test_imaging_camera/test_image_stack.py
@@ -1205,7 +1205,7 @@ def test_integration_test_to_kymo(
         (tether_start * pixelsize, 10 * pixelsize), (26 * pixelsize, 10 * pixelsize)
     )
     kymo = with_tether.to_kymo(half_window=half_window, reduce=np.sum)
-    lines = track_greedy(kymo, "red", pixel_threshold=3)
+    lines = track_greedy(kymo, "red", pixel_threshold=4)
     np.testing.assert_allclose(
         lines[0].position, [(position - tether_start) * pixelsize] * num_images
     )


changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.5.1 | t.b.d.
+
+#### Bug fixes
+
+* Fixed brittle test which was testing tracking a `Kymograph` generated from an artificial `ImageStack`. This test erroneously attempted tracking a line with the threshold set to the minimum pixel intensity of the image. A threshold set to the minimum intensity is typically rejected by `track_greedy`, but in the case of this test, floating point errors ended up making the minimum pixel value end up slightly lower or higher than the threshold. The former was more common making the test fail only sporadically.
```
rpauszek, line 7: I don't think we usually put this kind of internal stuff in the changelog, do we? At any rate, is it really a bug fix?

rpauszek (approved):
👍

changelog.md
JoepVanlier: Yeah, you are right. When I initially opened this PR, I thought we may still have to do a hotfix release to get it to pass on `conda-forge`, in which case a re-release without any notes would have been pretty strange.

Rerunning the action ended up making the error disappear though (different runner/env maybe), so I'll drop the entry.

PR #652, opened 2024-05-30
Update piezotracking.rst

aafkevandenberg:
Correct me if I'm wrong, but normally one should use a 1D polynomial, right? I always choose a region that is linear (and not too far from the region of interest for recording the fd curve).


aafkevandenberg:
Using a 1D polynomial made the piezo distance deviate more from the camera tracking distance, so for this data, using a 2D polynomial is better.

PR #653, opened 2024-05-30
file: fix detector mapping bug

JoepVanlier:
**Why this PR?**
The custom detector mappings in `1.4.0` introduced a bug that disallows opening files without an explicit detector mapping when only a subset of the photon channels is available. This PR fixes that and adds some more extensive testing to make a regression in this functionality in the future less likely.

I have also demoted the exception to a warning, as it seems too restrictive to disallow creation entirely in retrospect (the user will just get blank channels if they get the name wrong). The warning still notifies them, to hopefully resolve the confusion they may feel when they end up with blank channels in a `Kymo` or `Scan` without making it a hard requirement to read the file at all.


lumicks/pylake/file.py
```diff
@@ -26,8 +26,10 @@ class File(Group, Force, DownsampledFD, BaselineCorrectedForce, PhotonCounts, Ph
     filename : str | os.PathLike
         The HDF5 file to open in read-only mode
 
-    rgb_to_detectors : Dict[Color, str]
-        Dictionary that maps RGB colors to a photon detector channel (either photon counts, or photon time tags)
+    rgb_to_detectors : Optional[Dict[Color, str]]
```
agom, line 29: The color ended up being just a `str`
```suggestion
    rgb_to_detectors : Optional[Dict[str, str]]
```

lumicks/pylake/file.py
```diff
@@ -80,21 +78,43 @@ def _check_file_format(self):
...
+                detectors = set(self.h5["Photon time tags"])
+            elif "Photon count" in self.h5:
+                detectors = set(self.h5["Photon count"])
+
+            # Only check if detector data was exported
+            if not detectors:
+                return
+
+            # Off indicates that the user doesn't want to plot data for that particular channel.
+            if not_found := (set(rgb_to_detectors.values()) - {"Off"}) - detectors:
```
agom, line 104: Maybe "None" is more intuitive than "Off" since the is no "On" option.
```suggestion
            if not_found := (set(rgb_to_detectors.values()) - {"None"}) - detectors:
```

agom (approved):
Looks good! Just left a couple of suggestions.

lumicks/pylake/file.py
JoepVanlier: I initially had `"None"` and changed it for `"Off"` because confusing it with `None` (not a string) would cause errors. That said, you have a good point about there not being an `"On"`. I'll change it to `"None"`.

lumicks/pylake/file.py

I added one extra check to verify that the keys are correct. Otherwise you could accidentally use an incorrect capitalization like "red" and wonder why it doesn't work.

rpauszek (approved):
👍

PR #657, opened 2024-06-04
Update piezotracking.rst

aafkevandenberg:
Setting the sign is an important part of the analysis workflow, as it does sometimes happen that the signs are swapped.

### DIFF
--- a/docs/tutorial/piezotracking.rst
+++ b/docs/tutorial/piezotracking.rst
@@ -142,8 +142,9 @@ distance without the tether. :math:`F_{1x}` and :math:`F_{2x}` are the forces me
 
 To do this in Pylake, we set up the piezo distance calibration as follows::
 
-    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
+    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x, signs=(1, -1))
 
+where `signs` specifies the sign of force 1x and force 2x respectively. You can determine the signs by viewing the respective force channels; the channel that becomes more negative as force increases requires a `-1` sign.
 We now have all the calibrations we need to do piezo tracking on our experimental data.
 
 Calculating the Fd Curve


docs/tutorial/piezotracking.rst
```diff
@@ -142,9 +142,9 @@ distance without the tether. :math:`F_{1x}` and :math:`F_{2x}` are the forces me
 
 To do this in Pylake, we set up the piezo distance calibration as follows::
 
-    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
+    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x, signs=(1, - 1))
```
JoepVanlier, line 145: We usually follow `black` formatting.

```suggestion
    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x, signs=(1, -1))
```

docs/tutorial/piezotracking.rst
```diff
@@ -142,9 +142,9 @@ distance without the tether. :math:`F_{1x}` and :math:`F_{2x}` are the forces me
...
 
-We now have all the calibrations we need to do piezo tracking on our experimental data.
+where `signs` contains the sign of force 1x and force 2x respectively. We now have all the calibrations we need to do piezo tracking on our experimental data.
```
JoepVanlier, line 147:
```suggestion
where `signs` specifies the sign of force 1x and force 2x respectively. We now have all the calibrations we need to do piezo tracking on our experimental data.
```

Are the signs often different from the default? Should we tell people in the tutorial how to determine them?

JoepVanlier (changes requested):
Good point. If the default doesn't always apply, it's good to mention it in the tutorial.

docs/tutorial/piezotracking.rst
aafkevandenberg: I updated the text and squashed the commits, let me know your thoughts.

docs/tutorial/piezotracking.rst
```diff
@@ -142,8 +142,9 @@ distance without the tether. :math:`F_{1x}` and :math:`F_{2x}` are the forces me
 
 To do this in Pylake, we set up the piezo distance calibration as follows::
 
-    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x)
+    piezo_calibration = lk.PiezoForceDistance(distance_calibration, baseline_1x, baseline_2x, signs=(1, -1))
 
+where `signs` specifies the sign of force 1x and force 2x respectively. You can determine the signs by viewing the respective force channels; the channel that becomes more negative as force increases requires a '-1' sign.
```
JoepVanlier, line 147:
```suggestion
where `signs` specifies the sign of force 1x and force 2x respectively. You can determine the signs by viewing the respective force channels; the channel that becomes more negative as force increases requires a `-1` sign.
```
Backticks if you intend for this to show up highlighted.

JoepVanlier (approved):
This is very nitpicky, but we usually try to keep the commit messages in the same-ish format. Maybe something like: `docs: document sign piezotracking tutorial`. The prefix makes it easy to see which commits affect code and which don't. It's convenient when you're looking through past history for when a bug or issue may have been introduced for instance.

Once you squashed it down to a single commit again remember to select the `rebase and merge` option to merge it to main.

PR #659, opened 2024-06-04
file: populate calibration field when accessing force via square brackets

JoepVanlier:
**Why this PR?**
It's a small thing, but it is often awkward to explain the difference between

`f.force1x`, `f.downsampled_force1x` and `f["Force HF"]["Force 1x"]` and `f["Force LF"]["Force 1x"]`. The former nets you metadata, while the latter does not, even though most other channel access happens with square brackets.

This PR adds that the force metadata is available when accessing with square brackets as well.


rpauszek (approved):
👍

PR #660, opened 2024-06-05
zenodo: add `allow_overwrite` as parameter

JoepVanlier:
**Why this PR?**
Currently, when there is a hash mismatch between the file on Zenodo and the one on disk, your only recourse is to either use `force_download=True` which re-downloads all the files or delete the file manually (which can be tricky) and then rerun the function. This one adds an argument `allow_overwrite` to allow overwriting those files where there is a hash mismatch only.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.6.0 | t.b.d.
+
+#### New features
+
+* Added parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
+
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.6.0 | t.b.d.
+
+#### New features
+
+* Add parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
```
```suggestion
* Added parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
```

rpauszek (approved):

PR #662, opened 2024-06-14
force: provide better API for recalibrating offline

JoepVanlier:
**Why this PR?**
Currently, it is difficult for users to see exactly which parameter is what when looking at a Bluelake calibration item. What doesn't help is that the parameters often have slightly different names on the Pylake and Bluelake side.

This PR is part of a planned two PRs. In this PR I mostly set up a helper function, that allows users to extract a dictionary of calibration parameters (that can be fed directly to `lk.calibrate_force`) to reproduce a Bluelake calibration. I also have added a `start` and `stop` property to the calibration item, which ensures that you can slice with these items as well.

The idea of these API additions are that you can very quickly make a small-ish change to the calibration, without having to piece together what goes where every time (it is very easy to forget the exclusion ranges or fixed diode parameters for instance).

So, let's say you forgot to apply the hydrodynamically correct model, now you could do this for instance:

```python
previous_item = f.force1x.calibration[0]
item = f.force1x.calibration[1]
calibration_params = item.calibration_params()

with_hydro = calibration_params | {"hydrodynamically_correct": True}  # override an option
force_slice = f.force1x[item] / previous_item.force_sensitivity  # decalibrate

lk.calibrate_force(force_slice, driving_data=nano, **with_hydro)
```

In the second PR, I want to add a table-like view, with human readable descriptions of most parameters like we already have for the calibration items coming from Pylake so that these items become less opaque for people looking at this data.


lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
+import re
+from collections import UserDict
+
+
+def _read_from_bl_dict(bl_dict):
+    mapping = {
```
rpauszek, line 6: simple docstring might be nice, just so you don't have to think about all the different `dict`s being passed around

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+        self._verify_full()
+
+        return {
+            "num_points_per_block": self.num_points_per_block,
+            "sample_rate": self.sample_rate,
+            "excluded_ranges": self.excluded_ranges,
+            "fit_range": self.fit_range,
+        }
+
+    def _model_params(self):
```
rpauszek, line 64: could also be a property

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+        # TODO: Model needs to support fixed_diode and fixed_alpha, keep API private until finalized
+        self._verify_full()
+        model_params = _read_from_bl_dict(self.data)
+        model_params["sample_rate"] = self.sample_rate
+        model_params["fast_sensor"] = self.fast_sensor
+        model_params["axial"] = bool(self.data.get("Axial calibration"))
+        model_params["hydrodynamically_correct"] = bool(
+            self.data.get("Hydrodynamic correction enabled")
+        )
+        return model_params
```
rpauszek, line 76: just a thought, but using a `dict` union could be a nice pattern to use here and elsewhere

```suggestion
        return _read_from_bl_dict(self.data) | {
            "sample_rate": self.sample_rate,
            "fast_sensor": self.fast_sensor,
            "axial": bool(self.data.get("Axial calibration")),
            "hydrodynamically_correct": bool(self.data.get("Hydrodynamic correction enabled")),
        }
```

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+
+    def calibration_params(self):
```
rpauszek, line 78: should be a property

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+        return self.data.get("driving_frequency (Hz)") is not None
+
+    @property
+    def fast_sensor(self):
+        """Returns whether it was a fast sensor or not"""
+        self._verify_full()
+
+        # If it is not a fixed diode or free diode, it's a fast sensor
+        diode_fields = ("Diode frequency (Hz)", "f_diode (Hz)")
+        return not any(f in self.data.keys() for f in diode_fields)
```
rpauszek, line 166: I would just put the tuple directly in the generator

```suggestion
        return not any(f in self.data.keys() for f in ("Diode frequency (Hz)", "f_diode (Hz)"))
```

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+    return {key_param: bl_dict.get(key_bl) for key_param, key_bl in mapping.items()}
+
+
+class ForceCalibrationItem(UserDict):
+    def _verify_full(self):
+        if (kind := self.data.get("Kind", "Unknown")) != "Full calibration":
+            raise ValueError(
+                "These parameters are only available for a full calibration. Instead, this is a "
+                f"calibration item where the following operation was applied: {kind}."
+            )
```
rpauszek, line 30: I think you can easily make this a decorator, which would make the syntax below nicer

```suggestion
    def _verify_full(method):
        def wrapper(self, *args):
            if (kind := self.data.get("Kind", "Unknown")) != "Full calibration":
                raise ValueError(
                    "These parameters are only available for a full calibration. Instead, this is a "
                    f"calibration item where the following operation was applied: {kind}."
                )
           return method(self, *args)
      return wrapper
```

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+            new_calibration = lk.calibrate_force(calib_slice.data, **calibration_params)
+            new_calibration.plot()
+
+            # Make a new calibration, but change the amount of blocking
+            less_blocking_params = calibration_params | {"num_points_per_block": 200}
+            less_blocking = lk.calibrate_force(calib_slice.data, **less_blocking_params)
+            less_blocking.plot()
+        """
+        model_params = self._model_params()
+        model_params["active_calibration"] = self.active_calibration
```
rpauszek, line 107:
```suggestion
        model_params = self._model_params() | {"active_calibration": self.active_calibration}
```

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+
+        def check_defined(pair):
+            return pair[1] is not None
+
+        return dict(filter(check_defined, (self.power_spectrum_params() | model_params).items()))
```
rpauszek, line 112: interestingly, I think the `black` formatting for a longer line using a `lambda` might actually make this more readable

```suggestion
        return dict(
            filter(
                lambda pair: pair[1] is not None,
                (self.power_spectrum_params() | model_params).items(),
            )
        )
```

lumicks/pylake/conftest.py
```diff
@@ -305,12 +305,20 @@ def check_similarity(key, test_dict, ref_dict):
                 return f"{key}: missing vs {ref_dict[key]} (reference only)", False
             elif key not in ref_dict:
                 return f"{key}: {test_dict[key]} vs missing (test only)", False
-            elif not np.allclose(
-                (test_value := test_dict[key]), (ref_value := ref_dict[key]), rtol=rtol, atol=atol
-            ):
-                return f"{key}: {test_value} vs {ref_value} (difference)", False
             else:
-                return f"{key}: {test_value} vs {ref_value} (match)", True
+                match (test_value := test_dict[key], ref_value := ref_dict[key]):
```
rpauszek, line 309: our first `match` huzzah!

rpauszek (changes requested):
very nice first step on unifying all the various ways we store calibration information. Most of my comments are fairly nit-picky but I'm curious what you think

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+
+class ForceCalibrationItem(UserDict):
+    def _verify_full(self):
+        if (kind := self.data.get("Kind", "Unknown")) != "Full calibration":
+            raise ValueError(
+                "These parameters are only available for a full calibration. Instead, this is a "
+                f"calibration item where the following operation was applied: {kind}."
+            )
+
+    def power_spectrum_params(self):
```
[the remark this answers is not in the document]
JoepVanlier: I was doubting about that to be honest.

The reason I left it as a non-property is because I was thinking about leaving the option open to allow overriding arguments directly in this call in the future already. In case people end up finding the merge operation with dictionaries (`params | other params`) confusing.

lumicks/pylake/calibration.py
JoepVanlier: Yeah, I prefer that actually.

lumicks/pylake/calibration.py
JoepVanlier: Same as before, I'm not confident it will never have parameters. I could imagine one might want a `full=True` (to get all parameters specified for example, rather than just the one BL happened to export which were relevant); or a list of overrides.

lumicks/pylake/calibration.py
JoepVanlier: Yeah makes sense.

lumicks/pylake/calibration.py
JoepVanlier: Yeah, same as before. I think the parameter ones may have optional customization in the future.

lumicks/pylake/calibration.py
JoepVanlier: 👍

lumicks/pylake/conftest.py
JoepVanlier: Second, check the first commit ;)

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,218 @@
...
+        "driving_frequency_guess": "Driving data frequency (Hz)",
+        "distance_to_surface": "Bead center height (um)",
+        # Fixed diode parameters (only available when diode was fixed)
+        "fixed_alpha": "Diode alpha",
+        "fixed_diode": "Diode frequency (Hz)",
+        # Only available for axial with active
+        "drag": "gamma_ex_lateral (kg/s)",
+    }
+
+    return {key_param: bl_dict.get(key_bl) for key_param, key_bl in mapping.items()}
```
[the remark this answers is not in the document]
JoepVanlier: Yes, because they may not exist. `get` does not raise but returns `None` instead. Only a subset of these keys will be available. Initially, I considered having the option to always return the full list of possible arguments (populating it with defaults), but decided against it to see if anyone asks for it first.

For now, we could just omit the ones that aren't present at the door, then we also don't need that filter at the end.
```
return {
        key_param: bl_dict[key_bl] for key_param, key_bl in mapping.items() if key_bl in bl_dict
    }
```

lumicks/pylake/conftest.py
JoepVanlier: Yep haha. These walruses are a match :D

lumicks/pylake/calibration.py
JoepVanlier: Yeah, I agree. But considering the change to `_read_from_bl_dict`, filtering at the end is no longer needed.

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
rpauszek: ah yeah, I forgot about the default `None` return. either is fine

lumicks/pylake/calibration.py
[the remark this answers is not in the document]
rpauszek: yeah ok, it makes sense to leave it like this then

lumicks/pylake/calibration.py
```diff
@@ -1,3 +1,224 @@
...
+    def wrapper(self, *args, **kwargs):
+        if (kind := self.data.get("Kind", "Unknown")) != "Full calibration":
+            raise ValueError(
+                "These parameters are only available for a full calibration. Instead, this is "
+                f"a calibration item where the following operation was applied: {kind}."
+            )
+
+        return method(self, *args, **kwargs)
+
+    return wrapper
```
rpauszek, line 45: you can even keep this scoped locally in the class, which I think might be better because it relies on that specific class (for now at least)

rpauszek (approved):
looks great, just one extra little comment

lumicks/pylake/calibration.py
JoepVanlier: Sounds good. Should also still include `wraps` to forward the docstring.

PR #664, opened 2024-06-28
calibration: force calibration API pretty printing

JoepVanlier:
**Why this PR?**
Given that we now have actual properties defining the physical properties of each force calibration result, it's time to actually present these to the user.

In the updated API, the html and string representation of a calibration item give you a short summary of what's in there. Note the last column, which indicates whether the `Slice` you got this calibration item from actually contains the data you would need to recalibrate. I hope this will make it more clear when a sufficient amount of data was exported.

image

Access to the items is still by indexing them. And this now lists a table of properties.

image

**What to look out for?**
Technically, this PR is breaking, because we used to return a raw `list` of force calirations from `.calibration`. A regular list has some more operations defined than the one added here, but I would argue that many of those didn't really make sense for what we were going for. 

In the new setup, the calibrations are always packaged with sufficient metadata to construct the visual table. Also, in the new setup, we don't allow adding items (so we can be confident we are only dealing with force calibration items).

You can however still cast it down to a regular `list()` if you really want to. We could add a new property (`calibrations`?) and deprecate `.calibration` to not be breaking, but I wonder if the removal of things that weren't really intended as valid use cases warrants this additional friction of changing the property name for users.

**Note**
If we decide to go with this approach, I will open a last one in this series that updates the documentation, but I'm holding off on that since this will involve making many screenshots, and it does not seem like a good use of time to finalize all that before deciding on the final API.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * Added parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
 * Added force calibration information to channels accessed directly via the square bracket notation (e.g. `file["Force HF"]["Force 1x"].calibration`).
 * Calibration results and parameters are now accessible via properties which are listed when items are printed. See [calibration results](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html) and [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) API documentation for more information.
+* Added `applied_at` property to a [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) obtained from a force slice. This property returns the timestamp in nanoseconds at which the force calibration was applied.
 * Added improved printing of calibration items under `channel.calibration` providing a more convenient overview of the items associated with a `Slice`.
 * Added parameter `titles` to customize title of each subplot in `[`Kymo.plot_with_channels()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_with_channels).
 
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * Added parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
 * Added force calibration information to channels accessed directly via the square bracket notation (e.g. `file["Force HF"]["Force 1x"].calibration`).
 * Calibration results and parameters are now accessible via properties which are listed when items are printed. See [calibration results](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html) and [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) API documentation for more information.
+* Added improved printing of calibration items under `channel.calibration` providing a more convenient overview of the items associated with a `Slice`.
 * Added parameter `titles` to customize title of each subplot in `[`Kymo.plot_with_channels()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_with_channels).
 
 ## v1.5.1 | 2024-06-03
--- a/changelog.md
+++ b/changelog.md
@@ -9,7 +9,8 @@
 * Calibration results and parameters are now accessible via properties which are listed when items are printed. See [calibration results](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html) and [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) API documentation for more information.
 * Added `applied_at` property to a [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) obtained from a force slice. This property returns the timestamp in nanoseconds at which the force calibration was applied.
 * Added improved printing of calibration items under `channel.calibration` providing a more convenient overview of the items associated with a `Slice`.
-* Added parameter `titles` to customize title of each subplot in `[`Kymo.plot_with_channels()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_with_channels).
+* Added improved printing of calibrations performed with `Pylake`.
+* Added parameter `titles` to customize title of each subplot in [`Kymo.plot_with_channels()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.kymo.Kymo.html#lumicks.pylake.kymo.Kymo.plot_with_channels).
 
 ## v1.5.1 | 2024-06-03
 
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -28,6 +28,11 @@ class ForceCalibrationItem(UserDict, CalibrationPropertiesMixin):
             return self[bluelake_key]
 
     @property
+    def applied_at(self):
+        """Time the calibration was applied in nanoseconds since epoch"""
+        return self.data.get("Timestamp (ns)")
+
+    @property
     def _fitted_diode(self):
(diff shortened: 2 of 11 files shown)


lumicks/pylake/calibration.py
```diff
@@ -46,7 +54,7 @@ def _sensor_type(self):
     def kind(self):
         kind = self.data.get("Kind", "Unknown")
         if kind == "Full calibration":
-            return "Active calibration" if self.active_calibration else "Passive calibration"
+            return "Active" if self.active_calibration else "Passive"
         else:
             return kind
```
rpauszek, line 59: I noticed we have a mismatch between this type and `CalibrationResults`. I think the shorter version here is good; but `CalibrationResults.kind` returns `"active calibration" | "passive calibration"`. This is the kind of stuff that's gonna bite us because we have to maintain two different classes.

For this one in particular, I wonder if it's better to move it to the mixin and just test the type there. Not super great as far as coding practice, but at least it avoids this kind of mistake

lumicks/pylake/calibration.py
```diff
@@ -310,12 +360,57 @@ def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibr
...
         return ForceCalibration(time_field=time_field, items=items)
 
+    def print_summary(self, tablefmt):
+        return tabulate(
+            (
+                (
+                    idx,
+                    datetime.datetime.fromtimestamp(int(item.applied_at / 1e9)).strftime("%x %X")
+                    if item.applied_at
+                    else "-",
```
rpauszek, line 378: nitpick: curious, if you wrap this in `()` does it stay on one line? would be nice, but if not it's fine like this

lumicks/pylake/calibration.py
```diff
@@ -310,12 +360,57 @@ def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibr
...
+                        and (item.start >= self._slice_start)
+                        and self._slice_stop
+                        and (item.stop <= self._slice_stop)
+                    ),
+                )
+                for idx, item in enumerate(self._src)
+            ),
+            tablefmt=tablefmt,
+            headers=(
+                "Idx",
```
rpauszek, line 396: `idx` is very programmer-y. I'd use the full `"Index"`

lumicks/pylake/calibration.py
```diff
@@ -310,12 +360,57 @@ def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibr
...
-                attrs = calibration_item[force_channel].attrs
+                attrs = dict(calibration_item[force_channel].attrs)
                 if time_field in attrs.keys():
+                    # Copy the timestamp at which the calibration was applied into the item
+                    attrs["Timestamp (ns)"] = calibration_item.attrs.get("Timestamp (ns)")
                     items.append(ForceCalibrationItem(attrs))
 
         return ForceCalibration(time_field=time_field, items=items)
 
+    def print_summary(self, tablefmt):
```
rpauszek, line 371: how does this differ from the `_print_properties` output from the mixin? I'm super confused at this point

rpauszek (changes requested):
Overall I like the result of this PR, the pretty printing is super nice. I have serious concerns about how complex the calibration stuff is getting (see my comment in the changelog) but I don't think that's something we can fix immediately as we'll very likely need to introduce breaking changes. We might be getting to the point where that is warranted though. Let's chat offline

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -33,21 +33,143 @@
...
+            headers=("Property", "Description", "Value"),
+        )
+
+    @property
+    def _repr_properties(self):
+        """Returns properties in a dataclass-style formatting"""
+        properties = ", ".join(
+            (f"{prop}={str(getattr(self, prop))}" for prop in self._results + self._parameters)
+        )
+        return f"{self.__class__.__name__}({properties})"
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, you're right, good catch. I think initially, I thought I might be using it in more places, but even then, we could just invoke the repr.

I don't think the str is needed. Removing.

lumicks/pylake/calibration.py
JoepVanlier: This is unrelated to the mixin since this isn't the calibration item. This is the list of calibration items. Maybe renaming it would help?

I think part of what makes this iffy is some of the naming. I feel calling this `ForceCalibration` was a mistake (but that's a name that wasn't introduced in this PR).

I think a lot of the mess could already be cleared up with three changes:
- Change the name of this thing to `ForceCalibrationList`.
- Moving the mixin to `detail`.
- Moving the ForceCalibrationItem to its own file.

I feel a combination of questionable naming (legacy) and having too much in one file is making this seem messier than it needs to be already.

In reality, the specialization classes are actually relatively small.

changelog.md
```diff
@@ -7,6 +7,8 @@
 * Added parameter `allow_overwrite` to [`lk.download_from_doi()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.download_from_doi.html#lumicks.pylake.download_from_doi) to allow re-downloading only those files where the checksum does not match.
 * Added force calibration information to channels accessed directly via the square bracket notation (e.g. `file["Force HF"]["Force 1x"].calibration`).
 * Calibration results and parameters are now accessible via properties which are listed when items are printed. See [calibration results](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html) and [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) API documentation for more information.
+* Added `applied_at` property to [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) which returns the timestamp in nanoseconds at which the force calibration was applied.
```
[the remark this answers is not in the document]
JoepVanlier: I don't think two user facing specializations is that bad, since they serve different purposes and one cannot be substituted for the other. They are always going to have a few properties the other one doesn't have.

For calibration items, they're gonna be timestamps. For calibration results they're going to be the ability to graph and validate the results.

I do think the naming right now is kind of terrible though (see previous comment). But yeah, let's discuss tomorrow. I can quickly set up a branch in the morning where some things are shuffled around to hopefully make things a bit more apparent. I didn't want to move stuff yet in the middle of relatively big changes (since it complicated review), but maybe I should have.

lumicks/pylake/calibration.py
JoepVanlier: Yeah, let's move this one to the mixin.

lumicks/pylake/calibration.py
JoepVanlier:
```
                    (
                        datetime.datetime.fromtimestamp(int(item.applied_at / 1e9)).strftime(
                            "%x %X"
                        )
                        if item.applied_at
                        else "-",
                        item.kind,
                    ),
```
😬 

Actually, let me put that thing in a tiny inline function. If I just define:
```python
        def format_timestamp(timestamp):
            return datetime.datetime.fromtimestamp(int(timestamp)).strftime("%x %X")
```

right before, then that blurb becomes far more readable.

lumicks/pylake/force_calibration/power_spectrum_calibration.py
```diff
@@ -33,21 +33,143 @@
...
+    def _print_properties(self, tablefmt="text"):
+        def parse_prop(v):
+            # Some of the fields are string-typed. As a result, the automatic floating point
+            # formatting in tabulate fails which results in excessively long values.
+            if np.isscalar(v):
+                if isinstance(v, bool):
+                    return str(v)
+
+                value = f"{v:.5g}"
+                return " " * (5 - (value + ".").find(".")) + value
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, that line is probably my least favorite in the entire PR (if you know of a better way of doing it, please tell me :D). I have added your comment and fixed the space thing.

lumicks/pylake/calibration.py
JoepVanlier: I was trying to lose characters, since the line was getting overly wide on text mode. How about "#"?

lumicks/pylake/calibration.py
```diff
@@ -310,12 +331,58 @@ def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibr
...
+        def format_timestamp(timestamp):
+            return datetime.datetime.fromtimestamp(int(timestamp)).strftime("%x %X")
```
rpauszek, line 344: I think this should probably be private

changelog.md
```diff
@@ -7,6 +7,8 @@
...
 * Added force calibration information to channels accessed directly via the square bracket notation (e.g. `file["Force HF"]["Force 1x"].calibration`).
 * Calibration results and parameters are now accessible via properties which are listed when items are printed. See [calibration results](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html) and [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) API documentation for more information.
+* Added `applied_at` property to a [calibration item](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.calibration.ForceCalibrationItem.html) obtained from a force slice. This property returns the timestamp in nanoseconds at which the force calibration was applied.
+* Added improved printing of calibration items under `channel.calibration` providing a more convenient overview of the items associated with a `Slice`.
```
rpauszek, line 11: there's also improved printing for the pylake calibration class too right?

rpauszek (approved):
Nice, I think this looks good with the improvements and with the following PRs looks like it'll be cleaned up more addressing most of my concerns about the complexity 👍

PR #665, opened 2024-07-01
calibration: private api for exclusion/fit ranges

JoepVanlier:
**Why this PR?**
In an effort to make Pylake calibration results more similar to Bluelake calibration results, this adds both fit range and exclusion ranges to the `CalibrationResults`.

Note that it is private in this PR, because the next PR will add a layer on top that unifies the properties for both classes (`CalibrationResults` and `ForceCalibrationItem`).


rpauszek (approved):
LGTM

PR #666, opened 2024-07-01
calibration: expose all relevant calibration parameters as properties

JoepVanlier:
**Why this PR?**
It's about time we defined a unified calibration API that exposes (but also documents) all the relevant properties.

This PR adds almost all of the plumbing.

Previously, we always exposed the raw dictionary data for calibration items. This has lead to subtle differences between calibration items coming from Bluelake and ones performed by Pylake. In addition, many of the names of these dictionaries can be sort of cryptic if you are not intimately familiar with force calibration.

This commit adds an actual calibration API with properties that have human readable names for all the important parameters and results. In addition to that, there are extensive doc-strings now that describe what each parameter means. To make sure the items do not go out of sync again (and keep docstrings on both sides the same), the properties are defined in a mixin.

The dictionary key access is kept around for people that just want to directly interact with the raw data, but we could consider deprecating direct access once we hop to a next major version (although I don't really see a need to yet).

You can find the docs build [here](https://lumicks-pylake.readthedocs.io/en/all_the_properties/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html).

### DIFF
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -2,6 +2,8 @@ import re
 from functools import wraps
 from collections import UserDict
 
+from lumicks.pylake.force_calibration.power_spectrum_calibration import CalibrationPropertiesMixin
+
 
 def _read_from_bl_dict(bl_dict):
     """Returns a dictionary of calibration parameters from a calibration dictionary exported from
@@ -33,7 +35,7 @@ def _read_from_bl_dict(bl_dict):
     }
 
 
-class ForceCalibrationItem(UserDict):
+class ForceCalibrationItem(UserDict, CalibrationPropertiesMixin):
     @staticmethod
     def _verify_full(method):
         @wraps(method)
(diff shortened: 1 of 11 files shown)


lumicks/pylake/calibration.py
```diff
@@ -79,18 +81,39 @@ def power_spectrum_params(self):
...
+            "drag": self.transferred_lateral_drag_coefficient,
+            # Other properties
             "sample_rate": self.sample_rate,
             "fast_sensor": self.fast_sensor,
             "axial": bool(self.data.get("Axial calibration")),
-            "hydrodynamically_correct": bool(self.data.get("Hydrodynamic correction enabled")),
+            "hydrodynamically_correct": self.hydrodynamically_correct,
         }
 
+        return dict(filter(lambda item: item[1] is not None, params.items()))
```
rpauszek, line 115: any reason to use `filter` instead of a comprehension?
```
{key: value for key, value in params.items() if value is not None}
```

rpauszek (approved):
very nice, just left a few rather nit-picky comments.

lumicks/pylake/calibration.py
```diff
@@ -48,6 +20,36 @@ def wrapper(self, *args, **kwargs):
...
+        return "f_diode (Hz)" in self or "alpha" in self
+
+    @property
+    def _sensor_type(self):
+        if self.fast_sensor:
+            return "fast_sensor"
+
+        if "f_diode (Hz)" in self.data:
+            return "slow sensor (fitted)"
+        elif "Diode frequency (Hz)" in self.data:
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, that's a fair point. I'll change it to standard.

lumicks/pylake/calibration.py
JoepVanlier: no reason, I'll swap it for a comprehension

PR #667, opened 2024-07-03
Bugfix: handle tuple of excluded ranges passed in

rpauszek:
Came across this as I was trying to get a handle on calibration stuff for reviewing the other PRs. In the notebook, we pass in a `tuple`, which you can't use to extend a `list` and so you get an error

```
power_spectrum = lk.calculate_power_spectrum(
    volts.data,
    sample_rate=volts.sample_rate,
    fit_range=(1e2, 23e3),
    num_points_per_block=2000,
    excluded_ranges=([19348, 19668], [24308, 24548])
)
```


JoepVanlier (approved):
Great catch! Thanks!

PR #669, opened 2024-07-04
tracker: allow widget backend

JoepVanlier:
**Why this PR?**
The widget won't open on versions of `matplotlib` later than 3.9, since the magic `%matplotlib widget` now returns the widget name `widget` rather than `ipympl`. In addition, it seems that 3.9 [lowercases](https://github.com/matplotlib/matplotlib/pull/27948) all the backend names, but older ones didn't, so I decided to do that defensively in our check too.

I've added a second option for a magic, since in this case, if we had had this in the error message, the user would have likely found `%matplotlib ipympl` to be functional (it does work already in 3.9.0).


rpauszek (approved):
👍

PR #670, opened 2024-07-05
testing: add small script to test notebooks we ship

JoepVanlier:
**Why this PR?**
This is a little Friday afternoon project. It's happened more than once that notebooks we ship broke or used deprecated API. Considering how our collection of notebooks is growing (and expected to grow further), I think we need some sort of semi-automated solution to catch these problems. I did have a quick look at [galata](https://github.com/jupyterlab/jupyterlab/tree/main/galata), but I feel that for what we are doing here it is overkill for now.

Running this little script before release should prevent that.

Basically, you just run it from the command line _after_ building the docs:

`python docs/run_notebooks.py --reset`

It caches which notebooks already passed, so rerunning it without `reset` only reruns the needed ones.

`python docs/run_notebooks.py`

Finally, if you do find issues, you can copy the current versions of the notebooks in the built docs folder such that you can debug them where the files are:

`python docs/run_notebooks.py --copy`

Do note that it fetches the notebooks from the built docs folder and expects them to be built in the usual folder (`build/html`).

**Why did you not makes this part of CI testing?**
It is much too slow to put on `ci` (and pulls in far too much data). I think it makes much more sense to just run these integration tests once manually prior to each release we do.

Considering it downloads a whole bunch of files from Zenodo (and the total size is bulky), I would recommend just keeping the folder it makes around. It is on the ignore list, so it shouldn't be a problem.

Note:  While doing this, I noticed some of the notebooks did not ship with the data, so I will be opening a [second PR](https://github.com/lumicks/pylake/pull/671) where I fix them up.

### DIFF
--- a/docs/run_notebooks.py
+++ b/docs/run_notebooks.py
@@ -1,11 +1,12 @@
 import io
 import os
-import sys
+import json
+import time
 import pathlib
+import argparse
 import warnings
 import traceback
-from json import load
-from contextlib import redirect_stderr, redirect_stdout, contextmanager
+from contextlib import contextmanager, redirect_stderr, redirect_stdout
 
 import matplotlib as mpl
 import matplotlib.pyplot as plt
@@ -25,7 +26,7 @@ def execute_notebook(nb_test_dir, filename):
     os.environ["JUPYTER_PLATFORM_DIRS"] = "1"
 
     with open(filename) as fp:
-        nb = load(fp)
+        nb = json.load(fp)
 
     mpl.use("agg")  # Use a non-interactive backend
 
@@ -39,7 +40,60 @@ def execute_notebook(nb_test_dir, filename):
                     plt.close("all")
 
 
-def run_notebooks(include_list):
+def read_report(filename):
+    """Reads a dictionary from a file"""
+    if not filename.is_file():
+        return {}
+
+    try:
+        with open(filename, "r") as f:
+            return json.load(f)
+    except json.decoder.JSONDecodeError:
+        return {}
+
+
+def process_notebook(notebook, report, nb_test_dir):
+    """Process a single notebook and return a report on it
+
+    Parameters
+    ----------
+    notebook : pathlib.Path
+        Filename of a notebook
+    report : dict
+        Dictionary with notebook results
+    nb_test_dir : pathlib.Path
+        Path to the testing directory.
+    """
+    finished_success = report.get("result", "") == "success"
+
+    if finished_success:
+        print(f"Skipping notebook: {notebook} (already successful)")
+    else:
+        print(f"Testing notebook: {notebook}")
+
+        try:
+            tic = time.time()
+            execute_notebook(nb_test_dir, notebook)
+            report["time"] = f"{time.time() - tic:.2f}"
+        except Exception as e:
+            print("\nAn exception was raised:\n")
+            print(f"   {e}\n")
+            print(traceback.format_exc())
+
+            report["result"] = str(e)
+        else:
+            report["result"] = "success"
+
+    return report
+
+
+def run_notebooks(include_list, reset_cache):
+    """Run all notebooks
+
+    Parameters
+    ----------
+    include_list : List[str]
+    """
     exclude_list = [
         "nbwidgets",  # Exclude the notebook widgets since those require interaction
         "cas9_kymotracking",
@@ -48,6 +102,9 @@ def run_notebooks(include_list):
     base_dir = pathlib.Path(__file__).parent.parent.resolve()
     nb_test_dir = base_dir / "nb_test"
     os.makedirs(nb_test_dir, exist_ok=True)
+
+    report_file = nb_test_dir / "test_report.txt"
+
     print(f"Output folder: {nb_test_dir}")
     notebook_folder = base_dir / "build" / "html" / "nbexport"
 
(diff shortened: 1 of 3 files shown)


docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+        Wipe the cache of which notebooks have already been run.
+    only_copy : bool
+        Only copy the notebooks to the notebook testing folder, but don't run them.
+    """
+    exclude_list = [
+        "nbwidgets",  # Exclude the notebook widgets since those require interaction
+        "cas9_kymotracking",
+    ]
+    base_dir = pathlib.Path(__file__).parent.parent.resolve()
+    nb_test_dir = base_dir / pathlib.Path("nb_test")
```
rpauszek, line 97: do you need to cast this specifically to `pathlib.Path`?

JoepVanlier: No, I don't think so. Removed it now.

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+import warnings
+import traceback
+from shutil import copyfile
+from contextlib import redirect_stderr, redirect_stdout
+
+import matplotlib as mpl
+import matplotlib.pyplot as plt
+
+
+def execute_notebook(nb_test_dir, filename):
```
alessiamarcolini, line 16: would it be easier to run it via jupyter with `jupyter execute notebook.ipynb`?

https://docs.jupyter.org/en/latest/running.html#using-a-command-line-interface

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+    os.environ["JUPYTER_PLATFORM_DIRS"] = "1"
+
+    with open(filename) as fp:
+        nb = json.load(fp)
+
+    mpl.use("agg")  # Use a non-interactive backend
+    os.chdir(nb_test_dir)
```
alessiamarcolini, line 23: changing the current working directory in scripts it's usually discouraged, but I understand the need here
as a best practice, let's reset the previous CWD at the end of the function

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+    """
+    exclude_list = [
+        "nbwidgets",  # Exclude the notebook widgets since those require interaction
+        "cas9_kymotracking",
+    ]
+    base_dir = pathlib.Path(__file__).parent.parent.resolve()
+    nb_test_dir = base_dir / "nb_test"
+    os.makedirs(nb_test_dir, exist_ok=True)
+
+    report_file = nb_test_dir / pathlib.Path("test_report.txt")
```
alessiamarcolini, line 100:
```suggestion
    report_file = nb_test_dir / "test_report.txt"
```

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+
+    print(f"Output folder: {nb_test_dir}")
+    notebook_folder = base_dir / "build/html/nbexport"
```
alessiamarcolini, line 103: this may or may not work on Windows (outside of WSL)

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+
+    if not notebook_folder.exists():
+        raise FileNotFoundError(
+            "Could not find notebook folder. Are you on the root pylake folder and did you compile"
```
alessiamarcolini, line 107:
```suggestion
            f"Could not find notebook folder at {str(notebook_folder)}. Are you on the root pylake folder and did you compile"
```

docs/run_notebooks.py
```diff
@@ -0,0 +1,196 @@
...
+
+    try:
+        for root, _, files in os.walk(notebook_folder):
+            root = pathlib.Path(root)
+            notebooks = [
+                f for f in files if pathlib.Path(f).suffix == ".ipynb" and "checkpoint" not in f
+            ]
+
+            for nb_file in notebooks:
+                notebook = root / pathlib.Path(nb_file)
```
alessiamarcolini, line 138:
```suggestion
                notebook = root / nb_file
```

docs/run_notebooks.py
JoepVanlier: Good point! I'll add a little context manager to do this. Then it also doesn't clutter up the code too much.

docs/run_notebooks.py
JoepVanlier: Referring to the slashes in the string? I initially used the `/` operator. AFAIK, `/` is a valid path separator on all the platforms where we release, but I wouldn't mind changing it back just to be on the safe side.

docs/run_notebooks.py
JoepVanlier: Always a good idea to show the path it is looking for 👍 

Explicit `str` is not needed though since it's the default, right?

docs/run_notebooks.py
JoepVanlier: I considered that, but I figured it'd be easier to silence specific (expected) warnings and promote (and capture) others to exceptions when evaluating it this route.

docs/run_notebooks.py
alessiamarcolini: yeah not needed, I got slightly confused due to pathlib

alessiamarcolini (approved):

docs/run_notebooks.py
rpauszek: I think using the `/` operator with a `Path` object handles this regardless of the os but I'm not 100% sure

alessiamarcolini: I believe that concatenating a `Path` with a string containing slashes does not perform sanitization, so on Windows you could get a result like `a\b\c/d/e` 🤔 so always better to concatenate with the pathlib `/` operator..
but I don't have means to try

JoepVanlier: I quickly tested on my old laptop (`win10`) and it does seem like it changes them to `/` even in the string part; but I am also not sure whether that is always guaranteed on all platforms, so maybe we should just err on the side of safety.

PR #672, opened 2024-07-11
calibration: move force calibration items into more appropriate locations

JoepVanlier:
**Why this PR?**
The handling of force calibration items is getting a bit unwieldy. Part of the reason is some poor naming. Part is that files have not been split up in a while.

This renames `ForceCalibration` to `ForceCalibrationList` to reflect that it is a list of calibration markers and moves some of the force calibration items, to split them up into sensible units a bit more.

I will squash this down before merge, but kept it separate so that it's easier to see the new item suggestions, without all the downstream update noise that results from it.


lumicks/pylake/calibration_item.py
```diff
@@ -0,0 +1,227 @@
+import re
```
rpauszek, line 1: I don't like this module at the top level personally? It feels more like an implementation detail. could we move it into `details` maybe?

rpauszek (approved):
I like this a lot better, I think the organization is a lot cleaner. Only thing is, I would move `calibration_item.py` to `details`. Just a personal preference, seems less cluttered that way

lumicks/pylake/calibration_item.py
JoepVanlier: Yeah, good point to not clutter up the main folder, but as we discussed, probably move it to `force_calibration` rather than `details` (considering it is user facing).

PR #673, opened 2024-07-15
Emulate numeric types for parameter classes

rpauszek:
**Why this PR?**

We have some parameter classes (`CalibrationParameter` and Fd-fitting `Parameter`) which hold metadata about the parameter (description, unit, etc...) along with a value. Currently, to use these types in any sort of arithmetic you have to specifically request the `value` attribute

```
result = array * parameter.value
```

Using this mixin allows for a simpler syntax while maintaining the useful metadata
```
result = array * parameter
```

This PR focuses on implementing the necessary dunder methods as described in [Emulating Numeric Types](https://docs.python.org/3/reference/datamodel.html#emulating-numeric-types) and tests (there are a lot of possibilities here, so suggestions on extra tests or things I've missed are very welcome). Following PRs will apply this mixin to the appropriate parameter classes.

### DIFF
--- a/lumicks/pylake/detail/tests/test_value.py
+++ b/lumicks/pylake/detail/tests/test_value.py
@@ -1,0 +1,210 @@
+import math
+import operator
+
+import numpy as np
+import pytest
+
+from lumicks.pylake.detail.value import ValueMixin
+
+
+class Parameter(ValueMixin):
+    def __init__(self, value, description):
+        super().__init__(value)
+        self.description = description
+
+
+class BadParameter(ValueMixin):
+    def __init__(self, description):
+        self.description = description
+
+
+operators = {
+    operator.add,
+    operator.sub,
+    operator.mul,
+    operator.truediv,
+    operator.floordiv,
+    operator.mod,
+    divmod,
+    operator.pow,
+    operator.and_,
+    operator.or_,
+    operator.xor,
+}
+
+
+def test_mixin():
+    a = Parameter(2, "two")
+    a + 2
+
+    b = BadParameter("two")
+    with pytest.raises(AttributeError, match="'BadParameter' object has no attribute 'value'"):
+        b + 2
+
+
+def test_operators():
+    a = Parameter(a0 := 5, "five")
+    b = Parameter(b0 := 8, "eight")
+
+    for op in operators:
+        assert op(a0, 2) == op(a, 2)
+        assert op(2, a0) == op(2, a)
+        assert op(a0, b0) == op(a, b)
+        assert op(b0, a0) == op(b, a)
+
+
+def test_comparisons():
+    a = Parameter(5, "five")
+
+    assert a == 5
+    assert 5 == a
+
+    assert a != 4
+    assert 4 != a
+
+    assert a < 10
+    assert 10 > a
+
+    assert a > 2
+    assert 2 < a
+
+    assert a >= 5
+    assert 5 >= a
+    assert a <= 5
+    assert 5 <= a
+
+
+def test_operators_str():
+    a = Parameter(a0 := "a", "alpha")
+    b = Parameter(b0 := "b", "beta")
+
+    assert a0 + "c" == a + "c"
+    assert "c" + a0 == "c" + a
+    assert a0 + b0 == a + b
+    assert a0 * 5 == a * 5
+    assert a0 * 5 == a * Parameter(5, "five")
+    assert a0 * Parameter(5, "five") == a * 5
+
+    assert a == "a"
+
+    with pytest.raises(TypeError, match="not all arguments converted during string formatting"):
+        a % 1
+
+    assert "Hello, venus" == Parameter("Hello, %s", "fmt") % "venus"
+
+    for c in (1, "c", True):
+        for op in operators - {operator.add, operator.mul, operator.mod}:
+            with pytest.raises(TypeError):
+                op(a, c)
+
+
+def test_operators_bool():
+    a = Parameter(a0 := True, "true")
+
+    for op in operators:
+        assert op(a0, True) == op(a, True)
+        assert op(False, a0) == op(False, a)
+
+
+def test_not_implemented_operators():
+    a = Parameter(a0 := 2, "two")
+
+    for op in (operator.lshift, operator.rshift):
+        with pytest.raises(TypeError):
+            op(a, 2)
+        with pytest.raises(TypeError):
+            op(2, a)
+
+    # __matmul__ has not been specifically implemented, but it still works
+    A0 = np.array([[1, a0], [3, 4]])
+    A = np.array([[1, a], [3, 4]])
+    B0 = np.array([[11, 12], [13, 14]])
+    np.testing.assert_equal(A @ B0, A0 @ B0)
+
+    with pytest.raises(ValueError, match="matmul: Input operand 1 does not have enough dimensions"):
+        A @ 2
+
+    with pytest.raises(ValueError, match="matmul: Input operand 0 does not have enough dimensions"):
+        2 @ A
+
+    # not directly implemented but default fallback means it still works
+    assert isinstance(a, Parameter)
+    a0 += 4
+    a += 4
+    assert a0 == a
+    assert isinstance(a, int)
+
+    a = Parameter(a0 := 2, "two")
+    b = a
+    a += 4
+    assert a0 + 4 == a
+    assert a0 == b
+
+
+def test_casting():
+    a = Parameter(a0 := 2, "two")
+    assert float(a0) == float(a)
+
+    b = Parameter(b0 := 3.14, "pi")
+    assert int(b0) == int(b)
+
+    assert complex(a0) == complex(a)
+    assert complex(b0) == complex(b)
+    assert complex(a0, 2.1) == complex(a, 2.1)
+    assert complex(b0, 2.1) == complex(b, 2.1)
+
+
+def test_none():
+    a = Parameter(None, "none")
+    assert not a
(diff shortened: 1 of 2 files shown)


lumicks/pylake/detail/value.py
```diff
@@ -0,0 +1,162 @@
...
+        return other & self.value
+
+    @extract_value
+    def __rxor__(self, other):
+        return other ^ self.value
+
+    @extract_value
+    def __ror__(self, other):
+        return other | self.value
+
```
JoepVanlier, line 116: I think you're missing one that's pretty important for control flow here.

```suggestion
    def __bool__(self):
        return bool(self.value)
```

lumicks/pylake/detail/tests/test_value.py
```diff
@@ -0,0 +1,183 @@
...
+
+    assert "Hello, venus" == Parameter("Hello, %s", "fmt") % "venus"
+
+    for c in (1, "c", True):
+        for op in operators - {operator.add, operator.mul, operator.mod}:
+            with pytest.raises(TypeError):
+                op(a, c)
+
+
+def test_operators_bool():
```
JoepVanlier, line 101: I think you're missing the cast to bool, which is pretty important I think as it is required for a lot of control flow.

The following example fails right now for instance:
```python
    a = Parameter(None, "five")
    assert not a
```
And that seems a particularly risky one, since it might be checked to see whether something is defined or not.

There's a second gotcha, where I don't think you can get around as easily which is the check:

```python
assert a is None
```
I know that one is used frequently to see if something is set at all, and that may be one where the abstraction may be dicey since it can never evaluate to true, since it isn't the same entity.

I have to think about this one.

lumicks/pylake/detail/tests/test_value.py
```diff
@@ -0,0 +1,183 @@
...
+    b = Parameter(b0 := -2, "two")
+    assert +b0 == +b
+
+    assert abs(b0) == abs(b)
+
+    p = Parameter(p0 := 3.141592, "pi")
+    assert round(p0) == round(p)
+    assert round(p0, 3) == round(p, 3)
+
+    assert math.trunc(p0) == math.trunc(p)
```
JoepVanlier, line 164: Would be nice to add a test for `max(a, b)` and `min(a, b)`

JoepVanlier (changes requested):
Great! I think for the parameter class in the `FdFitter` and probably in the `CalibrationResults` dictionary, this one mimicks the behaviour of a regular numeric type pretty well.

There's three places where it currently fringes a bit. Around conditionals, `None` and around specific interactions with `numpy` (see the comments). The first and some of the third point can easily be mitigated by adding the bool dunder; but the issues with `None` are trickier.

lumicks/pylake/detail/tests/test_value.py
```diff
@@ -0,0 +1,183 @@
...
+
+    with pytest.raises(TypeError):
+        divmod(arr, a)
+
+    for op in operators - {divmod}:
+        np.testing.assert_equal(op(arr, a), op(arr, a0))
+        np.testing.assert_equal(op(a, arr), op(a0, arr))
+
+    arr = np.array([1, 2, a])
+    assert isinstance(arr[-1], Parameter)
```
JoepVanlier, line 181: Is this type preservation inside a numpy array something you specifically want?

Because if not, it might be worth just implementing `__array__` as:
```python
    def __array__(self):
        return np.asarray(self.value)
```
Which would make things like `np.testing.assert_allclose()` and various other numpy methods work.

lumicks/pylake/detail/tests/test_value.py
```diff
@@ -0,0 +1,183 @@
...
+    with pytest.raises(ValueError, match="matmul: Input operand 1 does not have enough dimensions"):
+        A @ 2
+
+    with pytest.raises(ValueError, match="matmul: Input operand 0 does not have enough dimensions"):
+        2 @ A
+
+    # not directly implemented but default fallback means it still works
+    assert isinstance(a, Parameter)
+    a0 += 4
+    a += 4
```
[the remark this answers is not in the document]
rpauszek: not exactly, with primitives when you do `b == a` it's by value not reference apparently. So the following passes
```
a = 2
b = a                 # a=2, b=2
a += 4                # a=6, b=2
assert a0 + 4 == a
assert a0 == b
```

but indeed, good call to explicitly test this

lumicks/pylake/detail/tests/test_value.py
```diff
@@ -0,0 +1,183 @@
...
+
+def test_numpy():
+    a = Parameter(a0 := 5, "five")
+    arr = np.arange(3) + 1
+
+    with pytest.raises(TypeError):
+        divmod(arr, a)
+
+    for op in operators - {divmod}:
+        np.testing.assert_equal(op(arr, a), op(arr, a0))
```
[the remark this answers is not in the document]
rpauszek: well right now we have to have `.value` anyways so nothing lost. but implementing `__array__` should address this

lumicks/pylake/detail/tests/test_value.py
rpauszek: yeah this is kinda tricky, but on the otherhand the intent is to be able to use a class with a numeric `value` as if it were a scalar. So I'm not too worried about emulating all possible functionality for any possible type in `value`.

That being said, I've added explicit tests
```
Parameter(None) == None  # passes
Parameter(None) is None  # fails

File: lumicks/pylake/detail/tests/test_value.py [hunk 7]
```diff
@@ -0,0 +1,183 @@
...
+    p = Parameter(p0 := 3.141592, "pi")
+    assert round(p0) == round(p)
+    assert round(p0, 3) == round(p, 3)
+
+    assert math.trunc(p0) == math.trunc(p)
+    assert math.floor(p0) == math.floor(p)
+    assert math.ceil(p0) == math.ceil(p)
+
+
+def test_numpy():
```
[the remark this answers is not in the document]
rpauszek: nice, didn't know about (or had forgotten about) the `__array__()` dunder.

lumicks/pylake/detail/tests/test_value.py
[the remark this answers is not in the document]
JoepVanlier: Well, the concern would be more for when you change dictionary entries that are currently true numeric types.

lumicks/pylake/detail/tests/test_value.py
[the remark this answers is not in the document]
JoepVanlier: yeah, that's as expected :)

lumicks/pylake/detail/tests/test_value.py
JoepVanlier: This particular one concerns me because the "false positive" (`None` claiming it is not `None`) could lead to subtle hard to spot bugs.

We use `None` as a signaling value for unspecified values for optionals and many other packages do too. Because just doing `if (var)` would also be `False` for zero, which is often not what you intend.

We could enforce to be explicit on our end (using the equality operator), but other packages are not under our control and as far as I know `a is None` is considered idiomatic and less ambiguous, as it doesn't invoke the equality operator which could be overloaded.

JoepVanlier (approved):
That's all I can think of, but I would hold off on replacing the dictionary outputs for the calibration routines until the next major release so that we can clearly signal the breaking change.

`FdFitter` and new calibration properties are fine though but please wait until the fixes and moves are merged for the latter.

Before merging, I think it would be worth asking Dean to have a quick peek at this one to see if we missed anything since we intend to use this in many places.

lumicks/pylake/detail/value.py
```diff
@@ -0,0 +1,170 @@
...
+    def __array__(self):
+        return np.array(self.value)
```
JoepVanlier, line 120: I just realized something. This should take more arguments.

Array can take dtype and other keys.
```suggestion
    def __array__(self, *args, **kwargs):
        return np.array(self.value, *args, **kwargs)
```

JoepVanlier (changes requested):
Argh, I just realized something. `__array__` should really forward any arguments it gets, otherwise some functions that will try to cast it will fail.

Like this for instance:
image

lumicks/pylake/detail/value.py
rpauszek: ah good catch. fixed and added a test

JoepVanlier (approved):
thanks 👍

lumicks/pylake/detail/value.py
```diff
@@ -0,0 +1,170 @@
+import math
+
+import numpy as np
+
+
+class ValueMixin:
```
dean0x7d, line 6: I think it would be good to give this class an `__init__` method. It makes it clearer that derived classes should initialize it:
```py
class ValueMixin:
	def __init__(self, value):
	    self.value = value

class Parameter(ValueMixin):
    def __init__(self, value, description):
        super().__init__(value)
        self.description = description
```

You could also consider narrowing down the type a bit here. Maybe:
```py
from numbers import Number

class ValueMixin:
	def __init__(self, value: Number | bool | str):
	    self.value = value
```

dean0x7d (commented):
Looks pretty good to me! From an implementation point of view, I can only point out adding `__init__` unless there's a good reason not to (see my comment below) and maybe some type hints.

From a more meta point of view: This seems like pretty useful functionality. I'm surprised there wasn't something already out there. Did you consider any third-party libraries with similar features (physical units)? The implementation here looks solid so I'm not suggesting to swap anything out. It's just always good to mention possible alternatives.

PR #674, opened 2024-07-18
Examples droplet fusion

aafkevandenberg:
An example of how to analyze a droplet fusion event

Build can be found here: https://lumicks-pylake.readthedocs.io/en/examples_droplet_fusion/examples/droplet_fusion/droplet_fusion.html

### DIFF
--- a/docs/examples/droplet_fusion/droplet_fusion.rst
+++ b/docs/examples/droplet_fusion/droplet_fusion.rst
@@ -1,0 +1,255 @@
+.. warning::
+    Disclaimer: This analysis has not yet been tested in a large number of different scenarios.
+
+Droplet fusion
+==============
+
+.. only:: html
+
+    :nbexport:`Download this page as a Jupyter notebook <self>`
+
+.. _droplet_fusion:
+
+Analyzing a droplet fusion event
+--------------------------------
+
+The data in this notebook were acquired by moving two fluorescently labeled RNA droplets together at a constant speed.
+The droplets were held by optical tweezers and the right tweezers is moving while the left one is static.
+As the droplets come close together, they fuse to form one, larger droplet.
+The relaxation time of the fusion event, 𝜏, together with the radii of the droplets can reveal something about the material properties of the droplets;
+when plotting 𝜏 vs the average droplet radius for many droplets, the slope is given by 𝜂/𝛾, viscosity (Pa*s) /surface tension (N/m), assuming a Newtonian fluid [1]_. 
+The ratio 𝜂/𝛾 is also known as the inverse capillary velocity.
+
+Both the PSD signal (usually used to determine force exerted by the optical tweezers) and a scan were recorded during the experiment.
+The PSD signal has a much higher time resolution than the images, therefore it is best to use PSD signal to determine the relaxation time of the fusion event. 
+In this Notebook, we will first obtain the relaxation time from the PSD signal and then estimate the size of the droplets from the scan.
+
+Download the droplet fusion data
+--------------------------------
+
+The droplet fusion data are stored on zenodo.org.
+We can download the data directly from Zenodo using the function :func:`~lumicks.pylake.download_from_doi`.
+The data will be stored in the folder called `"test_data"`::
+
+    filenames = lk.download_from_doi("10.5281/zenodo.12772709", "test_data")
+
+Relaxation time of fusion event
+-------------------------------
+
+First, plot the PSD signal using Pylake. 
+Since the assumptions underlying force calibration are not met during the fusion event, the absolute value of the force is not reliable, and we label the y-axis as 'PSD signal'::
+
+    f = lk.File("test_data/Droplet_fusion_data.h5")
+    plt.figure()
+    f["Force HF"]["Force 2x"].plot()  
+    plt.ylabel("PSD signal (a.u.)")
+    plt.show()
+
+.. image:: force_signal.png
+
+The jump in the signal after 5 seconds shows the typical exponential relaxation for a droplet fusion event.
+
+Select data for fit
+^^^^^^^^^^^^^^^^^^^
+
+Below we are selecting the force and trap data at the fusion event. When fitting the fusion relaxation time, 
+it is important that the traps holding the droplets are either both static, or one of the traps is moving at a constant speed. 
+We plot the trap position over time to check which of these conditions is met::
+
+    start = "5.414s"
+    stop = "5.9s"
+
+    force_selection = f.force2x[start:stop]
+    trap_selection = f["Trap position"]["1X"][start:stop]
+
+    plt.figure()
+    plt.subplot(2, 1, 1)
+    force_selection.plot()
+    plt.ylabel("PSD signal (a.u.)")
+    plt.subplot(2, 1, 2)
+    trap_selection.plot()
+    plt.ylabel(r"x-coordinate ($\mu$m)")
+    plt.tight_layout()
+    plt.show()
+
+.. image:: selected_data.png
+
+Model for fusion
+^^^^^^^^^^^^^^^^
+
+The force data during the fusion event is fitted with the following equation: :math:`f(t) = ae^{-t/\tau}+bt+c`
+
+The term :math:`bt` accounts for the movement of the trap, assuming a constant trap speed ([2]_, [3]_). 
+(When both traps are static, the term :math:`bt` should be removed from the model.)
+The parameter of interest is :math:`𝜏`, the relaxation time scale of the fusion event::
+
+    from scipy.optimize import curve_fit
+
+    def relaxation_model(t, tau, a, b, c):
+        return a * np.exp(-t / tau) + b * t + c
+
+Fit the data and plot the result::
+
+    time = force_selection.seconds
+    force = force_selection.data
+
+    popt, pcov = curve_fit(relaxation_model, time, force, [0.1, force[0], 0, 0])
+    plt.figure()
+    plt.plot(time, force)
+    plt.plot(time, relaxation_model(time,*popt), label=fr"$\tau$ = {popt[0]:0.4f}s")
+    plt.ylabel(r"PSD signal (a.u.)")
+    plt.xlabel("Time (s)")
+    plt.legend()
+    plt.show()
+
+.. image:: fit.png
+
+The array :math:`popt` contains all the fitted parameters::
+
+    >>> print(popt)
+
+    [ 0.0557409   3.78890848 -4.19208305 -6.46343689]
+
+The first parameter in :math:`popt` is :math:`𝜏` and the other 3 parameters are :math:`a`, :math:`b` and :math:`c` respectively, as defined in the model above.
+The matrix :math:`pcov` is the covariance matrix and the standard deviation errors in the fitted parameters can be obtained as::
+
+    >>> np.sqrt(np.diag(pcov))
+
+    [0.00035864, 0.01059128, 0.02548404, 0.00926486]
+
+The relaxation time obtained from the fit is 0.0557 +- 0.0004 seconds. 
+
+In practice, the obtained relaxation time also depends on the data selection. 
+It is recommended to repeat the fit for multiple time intervals, and determine the uncertainty in the relaxation time accordingly.
+
+Now, we will proceed to determine the size of the droplets before the fusion event.
+
+Droplet size
+------------
+
+First load the scan and print the relevant metadata::
+
+    >>> for name, scan in f.scans.items():
+    ...        print(f"num frames: {scan.num_frames}")
+    ...        frame_duration = (scan.frame_timestamp_ranges()[0][1]-scan.frame_timestamp_ranges()[0][0])/1e9
+    ...        print(f"frame duration: {frame_duration} s")
+
+    num frames: 27
+    frame duration: 0.4977664 s
+
+Plot a frame before the fusion event::
+
+    framenr = 2
+    plt.figure()
+    scan.plot(channel="green", frame=framenr)
+    plt.show()
+
+.. image:: frame2.png
+
+Plot a frame after the fusion event::
+
+    framenr = 6
+    plt.figure()
+    scan.plot(channel="green", frame=framenr)
+    plt.show()
+
+.. image:: frame6.png
+
+If the droplets are in focus, the size of the droplet can be estimated from the 2D scan. 
+The estimate has limited precision because the sphere edges in the scanned images are not very sharp.
+For experimental data such as the one used in this notebook, we would expect an error on the order of ~10%.
+
+The first step, is to use image segmentation to identify the two droplets in the image. 
+The threshold may need to be optimized for your data::
+
+    from skimage.measure import label, regionprops
+
+    framenr = 2  # Choose a frame before the fusion event on which you want to identify and measure droplets
+
+    image = scan.get_image(channel="red")[framenr]
+    image = image / np.max(image)
+    threshold = 0.5
+    blobs = image > threshold
+    label_img = label(blobs)
+
+    plt.figure()
+    plt.subplot(2, 1, 1)
+    plt.title("original, normalized image")
+    plt.imshow(image)
+    plt.subplot(2, 1, 2)
+    plt.title("Identified objects")
+    plt.imshow(label_img)
+    plt.tight_layout()
+    plt.show()
+
+    if (blobs := len(np.unique(label_img))) != 3:
+        raise RuntimeError(f"Expected 2 blobs, found {blobs - 1} instead! Maybe adjust the threshold?")
+
+.. image:: image_segmentation.png
+
+For this scan, the fast axis is along the horizontal coordinate (you can check the direction of the fast axis by typing :attr:`scan.fast_axis <lumicks.pylake.scan.Scan.fast_axis>`). 
+Therefore, we estimate the size of the droplets by looking at the width of the identified objects:::
+
+    def get_center_and_width(scan, mask, axis):
+        """Grabs the center and width along the fast scanning axis"""
+        widths = np.sum(mask, axis=axis)
+        max_width = np.max(widths)
+
+        # Grab the position
+        coordinate_weighted_mask = np.indices(mask.shape)[axis] * mask
+        centers = np.sum(coordinate_weighted_mask, axis=axis) / np.clip(np.sum(mask, axis=axis), 1, np.inf)
+        
+        # Since some scanlines can have the same width, we'd want the vertical position to be the average of these
+        max_scanline = int(np.mean(np.nonzero(max_width == widths)[0]))
+
+        if axis:
+            center = (centers[max_scanline], max_scanline)
+        else:
+            center = (max_scanline, centers[max_scanline])
+        
+        return center, max_width
+
+
+    def plot_width(scan, center, width, axis):
+        plt.plot(center[0], center[1], "ko")
+        if axis:
+            plt.plot([center[0] - 0.5 * width, center[0] + 0.5 * width], [center[1], center[1]])
+        else:
+            plt.plot([center[0], center[0]], [center[1] - 0.5 * width, center[1] + 0.5 * width])
+
+    droplet_radii = np.array([]) 
+
+    fig, ax = plt.subplots()
+    ax.imshow(image, cmap=plt.cm.gray)
+    plt.xlabel("x (pixels)")
+    plt.ylabel("y (pixels)")
+    axis = 1 if scan.fast_axis == "X" else 0
+    center, width = get_center_and_width(scan, label_img == 1, axis)
+    droplet_radii = np.append(droplet_radii, 0.5 * width)
+    plot_width(scan, center, width, axis)
+    center, width = get_center_and_width(scan, label_img == 2, axis)
+    droplet_radii = np.append(droplet_radii, 0.5 * width)
+    plot_width(scan, center, width, axis)
+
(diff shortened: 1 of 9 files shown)


docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    plt.show()
+
+.. image:: selected_data.png
+
+Model for fusion
+^^^^^^^^^^^^^^^^
+
+We will be fitting the force data during the fusion event with the following equation: $f(t) = ae^{-t/\tau}+bt+c$                 
+
+The term $bt$ accounts for the movement of the trap, assuming a constant trap speed [1]_. The parameter of interest is $\tau$, the relaxation time schale of the fusion event::
```
JoepVanlier, line 69:
```suggestion
The term :math:`bt` accounts for the movement of the trap, assuming a constant trap speed [1]_. The parameter of interest is $\tau$, the relaxation time schale of the fusion event::
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    def relaxation_model(t, tau, a,b,c):
+        return a*np.exp(-t / tau)+b*t+c
+
+Fit the data and plot the result::
+
+    time = force_selection.seconds
+    force = force_selection.data
+
+    popt, pcov = curve_fit(relaxation_model, time, force, [0.1,force[0],0,0])
+    print(popt)
```
JoepVanlier, line 82: Minor, but it might be nice to format this a little bit.

Something like (I didn't test this):
```suggestion
    tau, amplitude, movement_term, offset = popt
    print(f"Relaxation time: {tau:.2f}, amplitude: {amplitude:.2f}, movement_term: {movement_term:.2f}, offset: {offset:.2f}")
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    plt.figure()
+    plt.plot(time,force)
+    plt.plot(time,relaxation_model(time,*popt),label=f'$\\tau$ = {popt[0]:0.2f}s')
```
JoepVanlier, line 85:
```suggestion
    plt.plot(time, relaxation_model(time, *popt), label=fr"$\tau$" = {tau:0.2f}s")
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    filenames = lk.download_from_doi("10.5281/zenodo.12772709", "test_data")
+
+Relaxation time of fusion event
+-------------------------------
+
+First, plot the force signal using Pylake. The force has a characteristic jump during a fusion event. 
+Since the force calibration breaks down during the fusion event, the absolute value of the force is not reliable, and we label the y-axis as 'laser signal'::
+
+    plt.figure()
+    f["Force HF"]["Force 2x"].plot()
```
JoepVanlier, line 34: `f` is never defined?

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    force_selection = f.force2x[start:stop]
+    trap_selection = f["Trap position"]["1X"][start:stop]
+
+    plt.figure()
+    plt.subplot(2, 1, 1)
+    force_selection.plot()
+    plt.ylabel("Laser signal (a.u.)")
+    plt.subplot(2, 1, 2)
+    trap_selection.plot()
+    plt.ylabel("x-coordinate ($\mu$m)")
```
JoepVanlier, line 58:
```suggestion
    plt.ylabel(r"x-coordinate ($\mu$m)")
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+
+Droplet size
+------------
+
+First load the scan and print the relevant metadata::
+
+    >>> for name, scan in f.scans.items():
+            print(f"num frames: {scan.num_frames}")
+            frame_duration = (scan.frame_timestamp_ranges()[0][1]-scan.frame_timestamp_ranges()[0][0])/1e9
+            print(f"frame duration: {frame_duration} s")
```
JoepVanlier, line 103: This will not compile to a jupyter notebook correctly. If you want a multi-line command with output, you have to extend the input lines with `...` (e.g.)
```suggestion
    >>> for name, scan in f.scans.items():
    ...     print(f"num frames: {scan.num_frames}")
    ...     frame_duration = (scan.frame_timestamp_ranges()[0][1]-scan.frame_timestamp_ranges()[0][0])/1e9
    ...     print(f"frame duration: {frame_duration} s")
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+
+    num frames: 27
+    frame duration: 0.4977664 s
+
+Plot a frame before the fusion event::
+
+    framenr = 2
+    plt.figure()
+    scan.plot(channel='green',frame=framenr)
```
JoepVanlier, line 112:
```suggestion
    scan.plot(channel="green", frame=framenr)
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+    plt.show()
+
+.. image:: frame2.png
+
+Plot a frame after the fusion event::
+
+    framenr = 6
+    plt.figure()
+    scan.plot(channel='green',frame=framenr)
```
JoepVanlier, line 121:
```suggestion
    scan.plot(channel="green", frame=framenr)
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+First, we use image segmentation to identify the two droplets in the image::
+
+    import math
+    from skimage.measure import label, regionprops
+
+    framenr = 2  # Choose a frame before the fusion event on which you want to identify and measure droplets
+
+    image = scan.get_image(channel='red')[framenr]
+    image = image / np.max(image)
+    threshold = 0.5
```
JoepVanlier, line 136: Interesting. I wonder to what extent the determined size depends on this threshold.

One thing I wonder about is the normalization on a per image basis. How does this maximum change over time?

How would it affect size estimates if you were to normalize the whole stack at once and just use one threshold for all the images?

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+
+    average_droplet_radii_um = average_droplet_radii * scan.pixelsize_um[0]
+
+The average radii for the droplets in micrometers are::
+
+    >>> print(average_droplet_radii)
+
+    [11.05117733 10.73977514]
+
+We now determined the relaxation time as well as the droplet sizes. The next step is to measure these two quantities for many different fusion events and plotting 𝜏 vs average radius.
```
JoepVanlier, line 196:
```suggestion
We now determined the relaxation time as well as the droplet radii. The next step is to measure these two quantities for many different fusion events and plotting 𝜏 vs average radius.
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,200 @@
...
+The slope of this plot is given by 𝜂/𝛾 (viscosity/surface tension) for a Newtonian fluid [2]_.
```
JoepVanlier, line 197: Preference thing, but:
```suggestion
For a Newtonian fluid, the slope of this plot is given by 𝜂/𝛾 (viscosity/surface tension) [2]_.
```

Is this dynamic or kinematic viscosity? Might be nice to add units here.

JoepVanlier (changes requested):
Did a quick first pass and made a few notes on things I noticed.

docs/examples/droplet_fusion/droplet_fusion.rst
aafkevandenberg: That is a good point. First of all, I don't expect the size determination to be perfect. Scan images are always a bit 'blurry' and I expect the error in the size to have a ~10% error. There might be better choices for the threshold, that are closer to the real size. I found 50% a good 'null' estimate. In the end, the size has to be determined for many separate scans from different fusion events. The final method has to be insensitive to intensity fluctuations between different scans. I expect the variation between scans to be much larger than within a scan. I could try different values for the threshold and normalization, and see whether the final estimated is affected a lot.

aafkevandenberg: I tried different values for the threshold and normalization, and the current choice actually works quite well. (some extreme values eg result in detecting more than two, or just 1 object). I added a remark that the error is ~10%, which accounts for the range of values I got for the radii.

aafkevandenberg: And I realized that the drops are not spherical, because they move while the scan is being recorded... This adds another layer of uncertainty for the droplet size...

JoepVanlier: Your last point is a good one.

Well, considering that in the dataset, the fast axis is horizontal, wouldn't it make sense to specifically look at the largest horizontal diameter then (least amount of error) rather than find an ellipse?

Then you also don't run the risk of comparing different orientations with eachother.

If you ensure that the width you compute is fast axis-aligned, you could even correct for the motion blur exactly! My guess is the precision of such an estimate would be slightly lower (since you're not averaging as much), but the accuracy would be higher (lower bias).

Maybe something like this would be an option?
```python
fig, ax = plt.subplots()
ax.imshow(image, cmap=plt.cm.gray)
average_droplet_radii = np.array([])  # List of arrays to collect all droplet radii

# For each label, find the maximal extent in the direction of the fast axis
# This results in the least error due to movement artefacts
def get_center_and_width(scan, mask, axis):
    """Grabs the center and width along the fast scanning axis"""
    widths = np.sum(mask, axis=axis)
    max_width = np.max(widths)

    # Grab the position
    coordinate_weighted_mask = np.indices(mask.shape)[axis] * mask
    centers = np.sum(coordinate_weighted_mask, axis=axis) / np.clip(np.sum(mask, axis=axis), 1, np.inf)
    
    # Since some scanlines can have the same width, we'd want the vertical position to be the average of these
    max_scanline = int(np.mean(np.nonzero(max_width == widths)[0]))

    if axis:
        center = (centers[max_scanline], max_scanline)
    else:
        center = (max_scanline, centers[max_scanline])
    
    return center, max_width


def plot_width(scan, center, width, axis):
    plt.plot(center[0], center[1], "ko")
    if axis:
        plt.plot([center[0] - 0.5 * width, center[0] + 0.5 * width], [center[1], center[1]])
    else:
        plt.plot([center[0], center[0]], [center[1] - 0.5 * width, center[1] + 0.5 * width])


ax.imshow(image, cmap=plt.cm.gray)
axis = 1 if scan.fast_axis == "X" else 0
center, width = get_center_and_width(scan, label_img == 1, axis)
plot_width(scan, center, width, axis)
center, width = get_center_and_width(scan, label_img == 2, axis)
plot_width(scan, center, width, axis)

plt.title("Fast scanning axis aligned width")
plt.show()
```

image

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,208 @@
...
+    plt.title(f"Identified objects")
+    plt.imshow(label_img)
+    plt.tight_layout()
+    plt.show()
+
+.. image:: image_segmentation.png
+
+Estimate the size of the droplet by measuring the length of the longest axis and the shortest axis and taking the average::
+
+    regions = regionprops(label_img)
```
JoepVanlier, line 162: Since you can't be sure that the threshold works well for them, consider explicitly asserting that there are only two. Something like:
```python
if (blobs := len(np.unique(label_img))) != 3:
    raise RuntimeError(f"Expected 2 blobs, found {blobs - 1} instead! Maybe adjust the threshold?")
```

docs/examples/droplet_fusion/droplet_fusion.rst
aafkevandenberg: I like that suggestion, will add it!

JoepVanlier: Awesome. I did do a quick estimate of the kind of error one might expect from the movement, but when aligned on the fast axis like this, it's at the subpixel level and I think it's really completely negligible compared to the potential error we'd be making already with thresholding the image anyhow. I get numbers on the order of 0.01 pixels.

JoepVanlier (commented):
Looking great! I only have a last few tiny requests.

docs/examples/droplet_fusion/droplet_fusion.rst
aafkevandenberg: Great, thanks. I adjusted the comment and now I write that the uncertainty of the sizes is caused by the blurryness of the image (and not anymore by the movement of one of the droplets)

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+.. image:: frame6.png
+
+If the droplets are in focus, the size of the droplet can be estimated from the 2D scan. 
+The estimate will not be extremely precise, because the boarders of the scanned images are not sharp.
+The estimate in the size will therefore have an error of ~10%.
+
+The first step, is to use image segmentation to identify the two droplets in the image. 
+The threshold may need to be optimized for your data::
+
+    import math
```
JoepVanlier, line 161: Do you actually use this import? If not, should probably remove it.

JoepVanlier (changes requested):
Looking good. A few small suggestions. Don't forget to remove the old link, because the docs won't build if you keep it around (see comments).

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+
+
+    def plot_width(scan, center, width, axis):
+        plt.plot(center[0], center[1], "ko")
+        if axis:
+            plt.plot([center[0] - 0.5 * width, center[0] + 0.5 * width], [center[1], center[1]])
+        else:
+            plt.plot([center[0], center[0]], [center[1] - 0.5 * width, center[1] + 0.5 * width])
+
+    average_droplet_radii = np.array([])
```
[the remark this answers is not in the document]
aafkevandenberg: yep, removed `average`

docs/examples/droplet_fusion/droplet_fusion.rst
aafkevandenberg: it is not used anymore in the current workflow, removed it

docs/examples/index.rst
```diff
@@ -20,4 +20,6 @@ For all of the examples, it is assumed that the following lines precede any othe
     twlc_fitting/twlc_fitting
     reca_fitting/reca_fitting
     cas9_kymotracking/cas9_kymotracking
+    hairpin_fitting/hairping_unfolding
```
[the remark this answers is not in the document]
JoepVanlier: Yes, that is what I meant.

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    >>>> print(popt)
+
+    [ 0.04754597  5.39347958 -4.5188036  -6.2716303 ]
+
+The first parameter in :math:`popt` is :math:`𝜏` and the other 3 parameters are :math:`a`, :math:`b` and :math:`c` respectively, as defined in the model above.
+The matrix :math:`pcov` is the covariance matrix and the standard deviation errors in the fitted parameters can be obtained as:: 
+
+    >>>> np.sqrt(np.diag(pcov))
+
+    [0.00021354, 0.01104559, 0.0218276 , 0.00779505]
```
JoepVanlier, line 117:
```suggestion
The matrix :math:`pcov` is the covariance matrix and the standard deviation errors in the fitted parameters can be obtained as::

    >>> np.sqrt(np.diag(pcov))

    [0.00021354, 0.01104559, 0.0218276 , 0.00779505]
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    droplet_radii = np.array([]) 
+
+    fig, ax = plt.subplots()
+    ax.imshow(image, cmap=plt.cm.gray)
+    plt.xlabel("x (pixels)")
+    plt.ylabel("y (pixels)")
+    axis = 1 if scan.fast_axis == "X" else 0
+    center, width = get_center_and_width(scan, label_img == 1, axis)
+    droplet_radii = np.append(
+            droplet_radii, 0.5* width)
```
JoepVanlier, line 225: Not sure why the line break is here.
```suggestion
    droplet_radii = np.append(droplet_radii, 0.5 * width)
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    plot_width(scan, center, width, axis)
+    center, width = get_center_and_width(scan, label_img == 2, axis)
+    droplet_radii = np.append(
+            droplet_radii, 0.5* width)
```
JoepVanlier, line 229:
```suggestion
    droplet_radii = np.append(droplet_radii, 0.5 * width)
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+
+Droplet size
+------------
+
+First load the scan and print the relevant metadata::
+
+    >>> for name, scan in f.scans.items():
+    >>>        print(f"num frames: {scan.num_frames}")
+    >>>        frame_duration = (scan.frame_timestamp_ranges()[0][1]-scan.frame_timestamp_ranges()[0][0])/1e9
+    >>>        print(f"frame duration: {frame_duration} s")
```
JoepVanlier, line 131:
```suggestion
First load the scan and print the relevant metadata::

    >>> for name, scan in f.scans.items():
    ...        print(f"num frames: {scan.num_frames}")
    ...        frame_duration = (scan.frame_timestamp_ranges()[0][1]-scan.frame_timestamp_ranges()[0][0])/1e9
    ...        print(f"frame duration: {frame_duration} s")
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    plt.ylabel(r"x-coordinate ($\mu$m)")
+    plt.xlabel("Time (s)")
+    plt.legend()
+    plt.show()
+
+.. image:: fit.png
+
+The array :math:`popt` contains all the fitted parameters::
+
+    >>>> print(popt)
```
JoepVanlier, line 108:
```suggestion
    >>> print(popt)
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    plt.ylabel("PSD signal (a.u.)")
+    plt.show()
+
+.. image:: force_signal.png
+
+The jump in the signal after 5 seconds shows the typical exponential relaxation for a droplet fusion event.
+
+Select data for fit
+^^^^^^^^^^^^^^^^^^^
+Below we are selecting the force and trap data at the fusion event. When fitting the fusion relaxation time,
```
JoepVanlier, line 54:
```suggestion
Select data for fit
^^^^^^^^^^^^^^^^^^^

Below we are selecting the force and trap data at the fusion event. When fitting the fusion relaxation time, 
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+For experimental data such as the one used in this notebook, we would expect an error on the order of ~10%.
+
+The first step, is to use image segmentation to identify the two droplets in the image. 
+The threshold may need to be optimized for your data::
+
+    from skimage.measure import label, regionprops
+
+    framenr = 2  # Choose a frame before the fusion event on which you want to identify and measure droplets
+
+    image = scan.get_image(channel='red')[framenr]
```
JoepVanlier, line 165:
```suggestion
    image = scan.get_image(channel="red")[framenr]
```

docs/examples/droplet_fusion/droplet_fusion.rst
```diff
@@ -0,0 +1,252 @@
...
+    threshold = 0.5
+    blobs = image > threshold
+    label_img = label(blobs)
+
+    plt.figure()
+    plt.subplot(2, 1, 1)
+    plt.title("original, normalized image")
+    plt.imshow(image)
+    plt.subplot(2, 1, 2)
+    plt.title(f"Identified objects")
```
JoepVanlier, line 176: Doesn't need to be an f-string.
```suggestion
    plt.title("Identified objects")
```

JoepVanlier (changes requested):
A few small last changes. One thing that's curious is that the code seems to generate a different y-label for a figure than the figure you uploaded.

JoepVanlier (approved):
Looks great! :shipit:

PR #675, opened 2024-07-19
ci: some ci maintenance work

JoepVanlier:
**Why this PR?**
The `pyproject.toml` file is decribed in [PEP 621](https://peps.python.org/pep-0621/). It is a more modern way of packaging python.

In addition, this PR sets up commit hooks. The last commit is just updates from updating `black`.


pyproject.toml
```diff
@@ -1,3 +1,58 @@
+[build-system]
+requires = ["setuptools>=62.3.0", "wheel"]
```
alessiamarcolini, line 2: let's be careful with the setuptools version... was it already pinned like this? 
latest setuptools versions changed some things

some context:

> setuptools in turn has recently switched to adhering to PEP 625, which enforces that hyphens in package names be normalized to underscores to eliminate a parsing ambiguity in the sdist file format.

this impacts the dot that it's in the package name 

https://peps.python.org/pep-0625/
https://packaging.python.org/en/latest/specifications/binary-distribution-format/#escaping-and-unicode
https://github.com/pypa/setuptools/issues/3593
https://github.com/pypa/twine/pull/1096

.pre-commit-config.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - id: requirements-txt-fixer
+  - id: trailing-whitespace
+    exclude: ^.*\.png$ ^.*\.gif$
+-   repo: https://github.com/pycqa/flake8
+    rev: 7.1.0
+    hooks:
+    - id: flake8
+      language_version: python3
+      additional_dependencies:
+        - flake8-bugbear==24.4.26
```
alessiamarcolini, line 24: this has one indentation too much

pyproject.toml
JoepVanlier: Interesting. I wasn't aware of the new filename normalization. Thanks for the heads up. I don't think setuptools was pinned at all, but if I look in the last deploy that happened, I see `setuptools` `65.5.0` being installed.

I imagine going to a newer `setuptools` might need some adaptation in the script we use on the feedstock `https://github.com/conda-forge/lumicks.pylake-feedstock/blob/main/recipe/meta.yaml`.

What would you suggest? Upper bound it to `69.3`for now or cut the cord?

It should only impact generated file names though, right?

alessiamarcolini: within this PR, I'd suggest pinning an upper bound, and then investigate the potential impact of upgrading setuptools in a separate PR

alessiamarcolini: We noticed this behavior in other projects, where we have CI workflows for testing the packaging and what goes inside the distribution, so might be interested to look into that as well

alessiamarcolini:
maybe out of scope here but since we are in maintenance mode:

now you could also put the flake8 configuration in `pyproject.toml`

pyproject.toml
```diff
@@ -1,3 +1,58 @@
...
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.10",
+    "Programming Language :: Python :: 3.11",
+    "Programming Language :: Python :: 3.12",
+    "Programming Language :: Python :: Implementation :: CPython",
+]
+dependencies = [
+    "pytest>=3.5",
+    "h5py>=3.4, <4",
+    "numpy>=1.24, <2",  # 1.24 is needed for dtype in vstack/hstack (Dec 18th, 2022)
```
dean0x7d, line 30: Not for this PR, but it reminded me: think about lifting this upper limit soon. `numpy` v2.1.0 is already out. In general, v2.x brings some really nice performance improvements for Mx Macs and with a much smaller package size (thanks to using the Accelerate SDK).

pyproject.toml
```diff
@@ -20,6 +75,20 @@ exclude = '''
 [tool.isort]
 profile = "black"
 line_length = 100
-py_version=310
+py_version = 310
 length_sort_sections = ["stdlib", "thirdparty", "firstparty", "localfolder"]
 lines_between_sections = 1
+
+[tool.pytest.ini_options]
+minversion = "3.5"
```
dean0x7d, line 83: I'm fairly sure that you're using `pytest` features newer than v3.5. Maybe bump this up to v7 or v8?

lumicks/pylake/__about__.py
```diff
@@ -1,5 +1,7 @@
+from importlib.metadata import version
+
 __title__ = "lumicks.pylake"
-__version__ = "1.5.1"
+__version__ = version("lumicks.pylake")
```
dean0x7d, line 4: I would swap this to work the other way around with `pyproject.toml` dynamic metadata: https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata

That way, `pyproject.toml` reads from this file rather than the other way around. IMO this is nicer because users will be doing `import lumicks.pylake` a lot more often than `pip install lumicks.pylake` so it's good to avoid extra overhead on the more frequently used path.

You can share `__summary__`, `__url__` and other info similarly.

dean0x7d (commented):
Great improvements! Nice to see the modernization to `pyproject.toml`. It makes things so much easier (like checking the minimum Python version). But I feel like you should swap the where the `version` is stored. Check out my last comment.

lumicks/pylake/__about__.py
JoepVanlier: Yeah, I inverted the dependency since I didn't manage to make it work with the editable install (it complained about missing dependencies). But I only just realized that I should really just directly pull it from the `__about__` file instead. Must have been tired last Friday 😬.

This is actually nicer for a second reason on top of the overhead, since I would indeed want the package code to be leading in the case of an editable install, rather than the time I happened to install it.

I don't think `__summary__` and `__url__` can be shared through an attribute, and I would prefer not to put them in a separate file, since I feel that adding file reading to `__init__` or `__about__` is messier than having these properties we virtually never change in two places.

pyproject.toml
JoepVanlier: Yeah, I will. I'll play around with it soon, but indeed, I don't want to piggyback that now.

JoepVanlier:
> Great improvements! Nice to see the modernization to `pyproject.toml`. It makes things so much easier (like checking the minimum Python version). But I feel like you should swap the where the `version` is stored. Check out my last comment.

Did you read the small discussion about the filename normalization? For now I upper bounded to `69.3`, but do you know if any packaging we do (other than conda) relies on the actual filename of the generated package?

dean0x7d:
I think it's only Conda. I think you should just go ahead and unlock the upper bound and see. There really isn't any other way in the end.

pyproject.toml
```diff
@@ -20,6 +75,23 @@ exclude = '''
...
 lines_between_sections = 1
+
+[tool.pytest.ini_options]
+minversion = "7.4"
+addopts = "--doctest-modules --ignore=setup.py --ignore=docs/conf.py"
+norecursedirs = ".* *.egg build docs/*"
+filterwarnings = "error"
+
+[tool.setuptools.dynamic]
+version = {attr = "lumicks.pylake.__about__.__version__"}
```
dean0x7d, line 89: 👌

dean0x7d (approved):
Great work!

PR #677, opened 2024-07-22
scan: allow loading truncated scans

JoepVanlier:
**Why this PR?**
It seems that in some rare conditions, we see some truncated confocal kymos. While we should solve this upstream, the fact that these were generated means that we should likely provide a fallback option for opening them.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,10 @@
 
 ## v1.5.2 | t.b.d.
 
+#### Improvements
+
+* Added a fallback which allows loading scans or kymographs that have truncated photon count channels.
+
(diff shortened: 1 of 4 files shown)


changelog.md
```diff
@@ -2,6 +2,10 @@
 
 ## v1.5.2 | t.b.d.
 
+#### Improvements
+
+* Added a fallback which allows loading scans which have been truncated at the end.
```
```suggestion
* Added a fallback which allows loading scans or kymographs that have truncated photon count channels. 
```

rpauszek (approved):
LGTM

PR #680, opened 2024-07-25
tools: small fix to `run_notebooks`

JoepVanlier:
**Why this PR?**
I didn't realize this, but we use the same filenames quite frequently. This silently makes the checker skip these files. This commit fixes that by encoding the relevant part of the path in the name used in both the report and the target filename in case of a `--copy`.

I also added two quality of life features.
- If you run this script with `-n specific_notebook.ipynb`, it force runs it (despite it having passed before).
- It now checks a hash whether a notebook changed, and if it did, it will rerun it even if the last run was a success (this should have probably been the behavior initially).


rpauszek (approved):
👍

PR #681, opened 2024-07-25
fdfitter: fix inversion for the scalar case

JoepVanlier:
**Why this PR?**
Noticed something odd in external code using the call operator on an inverted model (wrapping the independent variable in a list only to immediately extract it again after the call).

Turns out, calling the model with a scalar returns an error instead of working properly. Not ideal, so let's fix it.

### DIFF
--- a/lumicks/pylake/fitting/detail/derivative_manipulation.py
+++ b/lumicks/pylake/fitting/detail/derivative_manipulation.py
@@ -42,7 +42,10 @@ def inversion_functions(model_function, f_min, f_max, derivative_function, tol):
         return single_estimate.x[0]
 
     def manual_inversion(distances, initial_guess):
-        """Invert the dependent and independent variable for a list"""
+        """Invert the dependent and independent variable"""
+        if distances.ndim == 0:
+            return fit_single(distances, initial_guess)
+
         return np.array([fit_single(distance, initial_guess) for distance in distances])
 
     return manual_inversion, fit_single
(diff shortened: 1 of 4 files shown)


lumicks/pylake/fitting/model.py
```diff
@@ -161,8 +161,24 @@ def __call__(self, independent, params):
...
         independent = np.asarray(independent, dtype=np.float64)
+
+        if independent.ndim > 1:
+            raise TypeError(
+                "Only 0- or 1-dimensional arrays are supported for the independent variable, "
+                f"got {independent.ndim}"
+            )
+
         missing_parameters = set(self.parameter_names) - set(params.keys())
         if missing_parameters:
```
tiagobonetti, line 183: Nitpick: I saw this and couldn't resist the suggestion. 😆 
```suggestion
        if missing_parameters := set(self.parameter_names) - set(params.keys()):
```

lumicks/pylake/fitting/detail/derivative_manipulation.py
```diff
@@ -43,6 +43,9 @@ def fit_single(single_distance, initial_guess):
 
     def manual_inversion(distances, initial_guess):
         """Invert the dependent and independent variable for a list"""
```
tiagobonetti, line 45: The "for a list" part of the docstring may be outdated now...

tiagobonetti (approved):
Looks good to me.
`honk`

lumicks/pylake/fitting/model.py
JoepVanlier: Nice one! Fixing... :D

rpauszek (approved):

PR #682, opened 2024-07-26
slice: add option to highlight time region

JoepVanlier:
**Why this PR?**
While working on the force calibration docs (not done yet), I found myself wanting to annotate some regions on a channel plot.

Instead of repeating the code in multiple notebooks / tutorials, I figured it might be a nice feature to just have in Pylake.

Docs build here: https://lumicks-pylake.readthedocs.io/en/highlight/tutorial/file.html#highlighting

```python
force.highlight_time_range(("50s", "85s"), annotation="region 2")
```
image

```python
force.highlight_time_range("50s", annotation="start 1", color="C0")
force.highlight_time_range("90s", annotation="start 2", color="C1")
```
image

```python
file.force1x.highlight_time_range(file.markers["Tether: 11"], annotation="marker")
```
image

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -6,6 +6,7 @@ from typing import Union
 import numpy as np
 import numpy.typing as npt
 
+from .detail.plotting import _annotate
 from .detail.timeindex import to_timestamp
 from .detail.utilities import downsample
 from .nb_widgets.range_selector import SliceRangeSelectorWidget
@@ -503,6 +504,129 @@ class Slice:
         plt.xlabel(self.labels.get("x", "Time") + " (s)")
         plt.ylabel(self.labels.get("y", "y"))
         plt.title(self.labels.get("title", "title"))
+
+    def highlight_time_range(
+        self,
+        item,
+        *,
+        color=None,
+        annotation=None,
+        annotation_direction=None,
+        label=None,
+        start=None,
+        alpha=0.15,
+        font_color=None,
+    ):
+        """Highlight some points or regions in a plot.
+
+        Parameters
+        ----------
+        item : Marker | tuple(int, int) | int
+            Time point or range to highlight. Can be an item with a start and stop property such
+            as a `Marker` or `CalibrationItem`, a tuple of timestamps, a single timestamp or a
+            string in time format (e.g. "5s" for 5 seconds).
+        color : str | None
+            Color to use for drawing elements. Should be a matplotlib compatible color.
+        annotation : str | None
+            Annotation text
+        annotation_direction : "horizontal" | "vertical" | None
+            Direction of annotation text
+        label : str | None
+            Label to use in legend entry if a legend is plotted
+        start : int | None
+            Plot with respect to a different "zero" timestamp.
+        alpha : float
+            Opacity of highlighted time range
+        font_color : str | None
+            Annotation text color. Defaults to `color` if available, otherwise black.
+
+        Examples
+        --------
+        ::
+
+            from lumicks import pylake
+
+            f = lk.File("data.h5")
+
+            plt.figure()
+            slc = ["Distance"]["Distance 1"]
+            slc.plot()
+
+            # Draws the time region spanned by marker 1
+            slc.highlight_time_range(f.markers["1"])
+
+            # Draws an annotation at the 1-second mark
+            slc.highlight_time_range("1s", annotation="first second")
+
+            # Draws a highlighted region from 1 to 3 seconds and marks it with the text "region"
+            slc.highlight_time_range(("1s", "3s"), annotation="region")
+
+            # Draws a highlighted region from 4 to 5 seconds, marks it, but forces text vertically
+            slc.highlight_time_range(("1s", "3s"), annotation="region", annotation_direction="vertical")
+
+
+            plt.figure()
+            f.force1x.plot()
+
+            item = f.force1x.calibration[1]
+
+            # Fills the region between calibration_item.start and calibration_item.stop with
+            # color "C2"
+            f.force1x.highlight_time_range(item, annotation="calibration 1", color="C2")
+
+            # Draws a line with annotation at the time the calibration was applied.
+            f.force1x.highlight_time_range(item.applied_at, annotation="calibration 1 applied")
+        """
+        import matplotlib.pyplot as plt
+
+        def to_seconds(timestamp):
+            start_ts = self.start if start is None else start
+
+            if isinstance(timestamp, str):
+                timestamp = to_timestamp(timestamp, start_ts, self.stop)
+
+            return (timestamp - start_ts) / 1e9
+
+        def get_time_range(time_item):
+            try:
+                return to_seconds(time_item.start), to_seconds(time_item.stop)  # Marker-like item?
+            except AttributeError:
+                pass
+
+            # Have to process single items separately, since otherwise strings might end up getting
+            # indexed in the last case. Note that single entry numpy arrays return false
+            # for np.isscalar, hence the check with ndim instead.
+            if np.ndim(time_item) == 0:
+                try:
+                    return to_seconds(time_item), None
+                except TypeError:
+                    raise TypeError(
+                        "Provided item must be either timestamp, two timestamps or an item with a "
+                        f"start and stop property, got {time_item} instead."
+                    )
+
+            try:
+                return to_seconds(time_item[0]), to_seconds(time_item[1])
+            except (TypeError, IndexError):
+                raise TypeError(
+                    "Provided item must be either timestamp, two timestamps or an item with a "
+                    f"start and stop property, got {time_item} instead."
+                )
+
+        start_time, stop_time = get_time_range(item)
+        if annotation:
+            _annotate(
+                start_time,
+                annotation,
+                annotation_direction,
+                stop_time,
+                color=font_color if font_color else color,
+            )
+
+        plt.axvline(start_time, label=label if not stop_time else "_", color=color)
+        if stop_time:
+            plt.axvline(stop_time, color=color)
+            plt.axvspan(start_time, stop_time, label="_", alpha=alpha, color=color)
 
     def range_selector(self, show=True, **kwargs) -> SliceRangeSelectorWidget:
         """Show a range selector widget
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -7,7 +7,7 @@ import numpy as np
 import numpy.typing as npt
 
 from .detail.plotting import _annotate
-from .detail.timeindex import to_timestamp
+from .detail.timeindex import to_seconds, to_timestamp
 from .detail.utilities import downsample
 from .nb_widgets.range_selector import SliceRangeSelectorWidget
 
@@ -579,17 +579,12 @@ class Slice:
         """
         import matplotlib.pyplot as plt
 
-        def to_seconds(timestamp):
-            start_ts = self.start if start is None else start
-
-            if isinstance(timestamp, str):
-                timestamp = to_timestamp(timestamp, start_ts, self.stop)
-
-            return (timestamp - start_ts) / 1e9
+        def to_sec(timestamp):
+            return to_seconds(timestamp, self.start if start is None else start, self.stop)
 
         def get_time_range(time_item):
             try:
-                return to_seconds(time_item.start), to_seconds(time_item.stop)  # Marker-like item?
+                return to_sec(time_item.start), to_sec(time_item.stop)  # Marker-like item?
             except AttributeError:
                 pass
 
@@ -598,7 +593,7 @@ class Slice:
             # for np.isscalar, hence the check with ndim instead.
             if np.ndim(time_item) == 0:
                 try:
-                    return to_seconds(time_item), None
+                    return to_sec(time_item), None
                 except TypeError:
                     raise TypeError(
                         "Provided item must be either timestamp, two timestamps or an item with a "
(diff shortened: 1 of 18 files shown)


lumicks/pylake/channel.py
```diff
@@ -504,6 +504,137 @@ def plot(self, start=None, **kwargs):
...
+        def to_seconds(timestamp):
+            start_ts = self.start if start is None else start
+
+            if isinstance(timestamp, str):
+                timestamp = to_timestamp(timestamp, start_ts, self.stop)
+
+            return (timestamp - start_ts) / 1e9
+
+        def get_time_range(time_item):
+            if hasattr(time_item, "start") and hasattr(time_item, "stop"):
```
mikhas, line 588: Why ask for permission when there is `try:`?

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,137 @@ def plot(self, start=None, **kwargs):
...
+                return to_seconds(time_item.start), to_seconds(time_item.stop)
+            elif np.isscalar(time_item) == 1:
+                return to_seconds(time_item), None
+            elif len(time_item) == 2:
+                return to_seconds(time_item[0]), to_seconds(time_item[1])
```
mikhas, line 593: Other things than list or tuples have a `len(...)`.

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,137 @@ def plot(self, start=None, **kwargs):
...
+            else:
+                raise ValueError(
+                    "Provided item must be either timestamp, two timestamps or an item with a "
+                    f"start and stop property, got {time_item} instead."
+                )
+
+        start_time, stop_time = get_time_range(item)
+        if annotation:
```
mikhas, line 601: I'd split out the annotation handling into a separate function, esp. because it's optional.

mikhas (commented):
Nice feature! Not sure if it will be used by others, but it's valuable even if it only helps with Pylake docs.

lumicks/pylake/channel.py
JoepVanlier: Yeah, that's fair. I'll split it to its own function. It does indeed help readability.

I do think it should be part of this function in terms of public API though.

Right now, I don't think splitting this over two API calls helps discoverability and the behavior is pretty linked to this range annotation (with all the logic regaring start and optional stop).

lumicks/pylake/channel.py
JoepVanlier: Personally I found LBYL easier to read than EAFP in this instance, but it's a fair point. I've switched it over and actually did find one more interesting corner case that is now fixed (pertaining to `isscalar`, see the new comment). I would have used pattern matching, but it gets clunky for the length check.

lumicks/pylake/channel.py
JoepVanlier: I think it's fixed in the new version, but I'm still curious what you were hinting at here specifically?

JoepVanlier:
> Nice feature! Not sure if it will be used by others, but it's valuable even if it only helps with Pylake docs.

I have at least one application in mind myself, for highlighting the regions where a force calibration was measured vs applied. It's something that often leads to confusion with users attempting to recalibrate their data.

lumicks/pylake/channel.py
mikhas: The function seems to be very accepting on terms of what `time_item` can be, and that's by design.
From the code it also seems that anything that the function _does not_ accept should land in the final exception handling branch. But here, there was the possibility that  a `time_item` could slip through and end in this case, which wouldn't handle it properly. Say for instance, a string.

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
         plt.ylabel(self.labels.get("y", "y"))
         plt.title(self.labels.get("title", "title"))
 
+    @staticmethod
+    def _annotate(start_time, annotation, annotation_direction, stop_time, color):
+        """Annotate a region or time point with some text
```
mikhas, line 509: Private methods get fully documented as well? Interesting.

JoepVanlier: Why not? It promotes reuse if applicable. They should stay hidden when rendering the sphinxdocs.

mikhas: I just wasn't aware.

JoepVanlier: Ah okay. Well, I don't think we are always consistent in it, but I try when I remember to do so.

lumicks/pylake/channel.py
JoepVanlier: Yeah, the string case should be handled now.

rpauszek:
> > Nice feature! Not sure if it will be used by others, but it's valuable even if it only helps with Pylake docs.
> 
> I have at least one application in mind myself, for highlighting the regions where a force calibration was measured vs applied. It's something that often leads to confusion with users attempting to recalibrate their data.

We actually have a "hand drawn" diagram of this in the tutorial, would be super nice to replace with some actual code with this

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
         plt.ylabel(self.labels.get("y", "y"))
         plt.title(self.labels.get("title", "title"))
 
+    @staticmethod
```
rpauszek, line 507: any reason to not just have this as a free function rather than `staticmethod`?

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
...
+        """
+        import matplotlib.pyplot as plt
+
+        def to_seconds(timestamp):
+            start_ts = self.start if start is None else start
+
+            if isinstance(timestamp, str):
+                timestamp = to_timestamp(timestamp, start_ts, self.stop)
+
+            return (timestamp - start_ts) / 1e9
```
rpauszek, line 636: I know it's not needed per se but considering we have the `to_timestamp` function in `detail.timeindex` could throw this one there too

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
...
+                    'Invalid value passed for annotation_direction. Expected "horizontal" or '
+                    f'"vertical", got {annotation_direction}.'
+                )
+
+    def highlight_time_range(
+        self,
+        item,
+        annotation=None,
+        annotation_direction=None,
+        color=None,
```
rpauszek, line 563: be careful with using the same arg name as `mpl`. If you do `channel.highlight_time_range(..., c="g")` you'll get an error
```
TypeError: Got both 'color' and 'c', which are aliases of one another
```

at some point I wrote a little decorator or something to handle this, but I don't think it got merged and I don't remember which feature it was for

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
...
+        item : Marker | tuple(int, int) | int | str
+            An item with a start and stop property such as a `Marker` or `CalibrationItem`, a
+            tuple of timestamps, a single timestamp or a string in time format (e.g. "5s" for
+            5 seconds).
+        annotation : str | None
+            String to annotate with
+        annotation_direction : "horizontal" | "vertical" | None
+            Which direction to show the text when annotating.
+        color : matplotlib.color | None
+            Matplotlib color
```
rpauszek, line 582: this is a bit vague as it doesn't say what the color is applied to. see my comment below too

lumicks/pylake/channel.py
```diff
@@ -504,6 +504,172 @@ def plot(self, start=None, **kwargs):
...
+                )
+
+        start_time, stop_time = get_time_range(item)
+        if annotation:
+            self._annotate(start_time, annotation, annotation_direction, stop_time, color)
+
+        plt.axvline(start_time, label=label if not stop_time else "_", color=color, **kwargs)
+        if stop_time:
+            plt.axvline(stop_time, **kwargs)
+            plt.axvspan(start_time, stop_time, label="_", alpha=alpha, color=color, **kwargs)
```
rpauszek, line 671: you're missing passing `color` in the 2nd `plt.axvline()`, if you pass in `color` for a time range it only gets applied to one of the lines
image

considering also the error I described above, I think it's better to not have `color` or `alpha` as explicit arguments here but just accept `**kwargs` and let the color cycler deal with it for the default case. We can still give a default `alpha` using:
```suggestion
        plt.axvline(start_time, label=label if not stop_time else "_", **kwargs)
        if stop_time:
            plt.axvline(stop_time, **kwargs)
            plt.axvspan(start_time, stop_time, label="_", **{"alpha": 0.5} | kwargs)
```

rpauszek (changes requested):
Super nice feature, I think it'll come in handy. 

Just be careful with forwarding `mpl` arguments and aliases; I've been bitten by this before. Especially now with `dict` unions, I think there's a nice way to handle this and avoid the pitfalls (see my last comment)

lumicks/pylake/channel.py
JoepVanlier: Good catch!

I think I was initially passing it over `**kwargs`, but felt that explicitly adding `color` as an input argument makes the api more discoverable (that's probably when the bug snuck in). I've swapped it out now, but it seems that `c` is not a valid shorthand for `color` for `axvspan`.

The color cycler does not apply to `axvline` or `axhline`. I did look into incrementing the color cycle manually while I was working on this. It's possible to create a custom color cycler from `plt.rcParams['axes.prop_cycle']`, but it would require us maintaining the color cycle state ourselves, which might have unexpected corner cases in how it plays with the rest of the `mpl` ecosystem.

lumicks/pylake/channel.py
JoepVanlier: I think `c` isn't valid for `plt.axvspan`. You end up with `AttributeError: Polygon.set() got an unexpected keyword argument 'c'`.

lumicks/pylake/channel.py
JoepVanlier: Yeah, that's a fair point. It's gone now. Ideally, I would still like there to be a "default" option that just gets applied to all (the convenience option). So I would propose forwarding the color to `annotation_kwargs` if it does not specify colors itself.

lumicks/pylake/channel.py
JoepVanlier: Fair point. I've split it out now and added a little test for it.

lumicks/pylake/channel.py
rpauszek: but it is valid for `plt.axvline`. I think it's still good practice not to duplicate/forward mpl args specifically like this if it can be avoided. I've been bitten by this in the past

JoepVanlier: Yeah. I've removed it now.

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        self,
+        item,
+        color=None,
+        annotation=None,
+        annotation_direction=None,
+        label=None,
+        start=None,
+        alpha=0.15,
+        font_color=None,
+    ):
```
rpauszek, line 518: let's make these keyword only to avoid same headaches we created in the past
```suggestion
    def highlight_time_range(
        self,
        item,
        *,
        color=None,
        annotation=None,
        annotation_direction=None,
        label=None,
        start=None,
        alpha=0.15,
        font_color=None,
    ):
```

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        """Highlight some points or regions in a plot.
+
+        Parameters
+        ----------
+        item : Marker | tuple(int, int) | int | str
+            An item with a start and stop property such as a `Marker` or `CalibrationItem`, a
+            tuple of timestamps, a single timestamp or a string in time format (e.g. "5s" for
+            5 seconds).
```
rpauszek, line 526:
```suggestion
        item : Marker | tuple(int, int) | int | str
            Time point or ange to highlight. Can be an item with a start and stop property such as a `Marker` or `CalibrationItem`, a
            tuple of timestamps, a single timestamp or a string in time format (e.g. "5s" for
            5 seconds).
```

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        color : str | None
+            Color to use for drawing elements. Should be a matplotlib compatible color.
+        annotation : str | None
+            String to annotate with
```
rpauszek, line 530:
```suggestion
            Annotation text
```

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        annotation_direction : "horizontal" | "vertical" | None
+            Which direction to show the text when annotating.
```
rpauszek, line 532:
```suggestion
            Direction of annotation text
```

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        label : str | None
+            Label to use in legend entry if a legend is plotted
+        start : int | None
+            Plot with respect to a different "zero" timestamp.
+        alpha : float
+            Transparency to use when plotting
```
rpauszek, line 538:
```suggestion
            Transparency of highlighted time range
```

lumicks/pylake/channel.py
```diff
@@ -504,6 +505,123 @@ def plot(self, start=None, **kwargs):
...
+        font_color : str | None
+            Color to use for displaying the font. When omitted, `color` is used.
```
rpauszek, line 540:
```suggestion
            Annotation text color. Defaults to `color` if available, otherwise black.
```

rpauszek (approved):
Great, I think this is the best way to go without the `**kwargs` makes it a lot cleaner.

I left some nitpick suggestions for wording, but I do really think we should make most of the parameter keyword only. Otherwise good to go!

PR #683, opened 2024-07-30
fdfit: add docstring to `Parameter`

JoepVanlier:
**Why this PR?**
Without these docstrings, the properties simply don't show up in the docs.

### DIFF
--- a/lumicks/pylake/fitting/parameters.py
+++ b/lumicks/pylake/fitting/parameters.py
@@ -5,6 +5,25 @@ from tabulate import tabulate
 
 
 class Parameter:
+    """A single model parameter, usually part of a :class:`Params` dictionary.
+
+    Examples
+    --------
+    ::
+
+        import lumicks.pylake as lk
+        fit = lk.FdFit(lk.ewlc_odijk_distance("my_model"))
+
+        print(fit.params)  # Prints the model parameters
+        parameter = fit["my_model/Lc"] = 5  # parameter is of the type `Parameter` here.
+
+        # You can extract and/or modify fitting bounds
+        lower_bound = parameter.lower_bound
+
+        # Or read out the fitting error after the model has been fitted.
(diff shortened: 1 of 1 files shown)


mikhas (commented):
Personally I find it more confusing that there are `Params` and `Parameters` in the table of contents about FD fitting. At least for `Params` is says that those are model parameters, but it does not explain in the doc what these parameters do (it might be too obvious, but I would still clarify _because_ of the existence of the other ones).

Also, is the user expected to mess with `profile` or `stderr` properties at all? If not, why even document that they exist?

JoepVanlier:
> Personally I find it more confusing that there are `Params` and `Parameters` in the table of contents about FD fitting. At least for `Params` is says that those are model parameters, but it does not explain in the doc what these parameters do (it might be too obvious, but I would still clarify _because_ of the existence of the other ones).

It's `Parameter`, not `Parameters`, but it's a good point that the class itself doesn't have a docstring, which indeed makes its listing confusingly blank. I have added a top level docstring for the class now. Do you think the distinction is clear now?

> Also, is the user expected to mess with `profile` or `stderr` properties at all? If not, why even document that they exist?

Yes. People would read them out. It's the only way to get them.

Profile is the result of a profile likelihood analysis and std err is the standard error of the parameter estimate. Both of these you would be interested in as they tell you how estimable your parameter estimate is.

mikhas:
> It's `Parameter`, not `Parameters`, but it's a good point that the class itself doesn't have a docstring, which indeed makes its listing confusingly blank. I have added a top level docstring for the class now. Do you think the distinction is clear now?
> 
> Here is a build for your convenience: https://lumicks-pylake.readthedocs.io/en/parameter_docstring/api.html#fd-fitting

Thanks, looks better now.

rpauszek (approved):

PR #685, opened 2024-08-02
kymotrack: add `sample_from_channel`

JoepVanlier:
**Why this PR?**
It can be pretty error prone for users to do this themselves. To do this properly, they have to index `line_timestamp_ranges()` themselves and have to make sure the kymograph they apply that on is the one they have tracks of (and that it is cropped in the same way).

image

Docs build [here](https://lumicks-pylake.readthedocs.io/en/downsampled_over_track/tutorial/kymotracking.html#averaging-channel-data-over-tracks) and [here](https://lumicks-pylake.readthedocs.io/en/downsampled_over_track/_api/lumicks.pylake.kymotracker.kymotrack.KymoTrack.html#lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_channel).

### DIFF
--- a/docs/tutorial/kymotracking.rst
+++ b/docs/tutorial/kymotracking.rst
@@ -397,6 +397,32 @@ Here `num_pixels` is the number of pixels to sum on either side of the track.
 .. note::
     For tracks obtained from tracking or :func:`~lumicks.pylake.refine_tracks_centroid`, the photon counts found in the attribute :attr:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.photon_counts` are computed by :func:`~lumicks.pylake.kymotracker.kymotrack.KymoTrack.sample_from_image` using `num_pixels=np.ceil(track_width / pixelsize) // 2` where `track_width` is the track width used for tracking or refinement.
 
+Averaging channel data over tracks
+----------------------------------
+
+It is also possible to average channel data over the track using :meth:`~lumicks.pylake.kymotracker.Kymotrack.sample_from_channel()`.
+For example, let's find out what the force was during a particular track::
+
+    force_slice = file.force1x
+    track_force = longest_track.sample_from_channel(force_slice, include_dead_time=True)
+
+When you call this function with a :class:`~lumicks.pylake.Slice`, it returns another :class:`~lumicks.pylake.Slice` with the downsampled channel data.
+For every point on the track, the corresponding kymograph scan line is looked up and the channel data is averaged over the entire duration of that scan line.
+The parameter `include_dead_time` specifies whether the dead time (the time it takes the mirror to return to its initial position after each scan line) should be included in this average.
+For tracks which have not been refined (see :ref:`localization_refinement`) the result from this function may skip some scan lines entirely.
+
+We can plot these slices just like any other.
+Plotting this slice, we can see that the protein detaches shortly after the force drops::
+
+    plt.figure()
+    force_slice.plot(label="force (whole file)")
+    track_force.plot(start=force_slice.start, marker=".", label="force (longest track)")
+    plt.legend(loc="upper left")
+
+.. image:: figures/kymotracking/sample_from_channel.png
+
+We plotted the track here putting the time zero at the start time of the entire force plot by passing its `start` time.
+
 Plotting binding histograms
 ---------------------------
 
--- a/lumicks/pylake/kymotracker/kymotrack.py
+++ b/lumicks/pylake/kymotracker/kymotrack.py
@@ -3,6 +3,7 @@ import pathlib
 import itertools
 from copy import copy
 
+from ..channel import Slice, empty_slice
 from ..__about__ import __version__
 from ..detail.utilities import replace_key_aliases
 from .detail.peakfinding import _sum_track_signal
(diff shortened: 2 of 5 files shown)


docs/tutorial/kymotracking.rst
```diff
@@ -397,6 +397,32 @@ Here `num_pixels` is the number of pixels to sum on either side of the track.
...
+----------------------------------
+
+It is also possible to average channel data over the track using :meth:`~lumicks.pylake.kymotracker.Kymotrack.sample_from_channel()`.
+For example, let's find out what the force was during a particular track::
+
+    force_slice = file.force1x
+    track_force = longest_track.sample_from_channel(force_slice, include_dead_time=True)
+
+When you call this function with a :class:`~lumicks.pylake.Slice`, it returns another :class:`~lumicks.pylake.Slice` with the downsampled channel data.
+What happens is that for every point on the track, the correct kymograph scan line is looked up and the channel data is averaged over the entire duration of that scan line.
```
mikhas, line 410: Would clarify. It's either the entire _length_ of the scan line or the entire _duration_ of the line _scanning_.

As in: the scan line is already an artificial construct where the time it took to record it has been factored in, so talking about duration here again is potentially confusing.

lumicks/pylake/kymotracker/tests/test_image_sampling.py
```diff
@@ -114,3 +116,41 @@ def test_origin_warning_sample_from_image():
...
+    kymo = _kymo_from_array(
+        img,
+        "r",
+        line_time_seconds=1.0,
+        exposure_time_seconds=0.6,
+        start=_FIRST_TIMESTAMP + int(1e9),
+    )
+
+    data = Slice(Continuous(np.arange(100), start=_FIRST_TIMESTAMP, dt=int(1e8)))  # 10 Hz
+    kymotrack = KymoTrack(time_idx, [0, 1, 2], kymo, "red", 0)
```
mikhas, line 140: Lacks the case where the queried time indices are larger than the given track points, e.g., `time_idx = [0, 1, 2, 3]` (or even `[0, 3]`, `[2, 3]`).  I am guessing that `sample_from_channel` would fail and return an empty slice but yield no warning to the user, correct?

Or is that the scenario you test for in the `test_sample_from_channel_no_overlap` test? Note: It's not clear to me what `no_overlap` refers to in that test.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -621,6 +637,34 @@ def _split(self, node):
...
+        channel_slice : Slice
+            Sample from this channel data.
+        include_dead_time : bool
+            Include the time that the mirror returns to its start position after each kymograph
+            line.
+        """
+        ts_ranges = self._kymo.line_timestamp_ranges(include_dead_time=include_dead_time)
+        try:
+            return channel_slice.downsampled_over(
+                [ts_ranges[time_idx] for time_idx in self.time_idx]
```
mikhas, line 663: How is it guaranteed that each available time index also exists in `ts_ranges`, especially if `include_dead_time` is `False`? Wouldn't `ts_ranges` then miss a couple of indices that exist in `self.time_idx`?

docs/tutorial/kymotracking.rst
JoepVanlier: Isn't the difference already clarified by the line immediately after?

`The parameter `include_dead_time` specifies whether the time it takes the mirror to return to its initial position after each scan line should be included in this average.`

What do you think it should read?

lumicks/pylake/kymotracker/tests/test_image_sampling.py
JoepVanlier: If a time index outside the `Kymo` is required, it will raise. I don't really see how this would be possible in a real scenario though, since the `KymoTrack` is derived from the `Kymo` being sampled and the `Kymo` is private and cannot be modified.

The no overlap case is the case where the channel data has _no_ overlap with the kymotrack (which would happen if you happen to evaluate it for the wrong slice for example). It is important to handle that case, since it is not unthinkable that a user might try to match channels from the wrong file with another kymo.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Why would `include_dead_time` matter? Include dead time impacts the duration of each timestamp range interval, not the number of ranges that are there.

mikhas: Don't you get a different set of indices, depending on that flag?

docs/tutorial/kymotracking.rst
rpauszek, line 410:
```suggestion
For every point on the track, the corresponding kymograph scan line is looked up and the channel data is averaged over the entire duration of that scan line.
```

docs/tutorial/kymotracking.rst
```diff
@@ -397,6 +397,32 @@ Here `num_pixels` is the number of pixels to sum on either side of the track.
...
+The parameter `include_dead_time` specifies whether the time it takes the mirror to return to its initial position after each scan line should be included in this average.
```
rpauszek, line 411:
```suggestion
The parameter `include_dead_time` specifies whether the dead time (the time it takes the mirror to return to its initial position after each scan line) should be included in this average.
```

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -621,6 +637,34 @@ def _split(self, node):
 
         return before, after
 
+    def sample_from_channel(self, channel_slice, include_dead_time=True) -> Slice:
```
rpauszek, line 640: since this is effectively another downsampling method, I think it would be good to include an optional `reduce` function with `np.mean` as the default like we do for `Slice.downsample_over()`

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -621,6 +637,34 @@ def _split(self, node):
...
+        """Sample channel data using the time points corresponding to this track
+
+        For each time point in the track, sample the channel data corresponding to that kymograph
+        scan line.
```
rpauszek, line 644:
```suggestion
        """Downsample channel data corresponding to the time points of this track

        For each time point in the track, downsample the channel data over the time limits of the corresponding kymograph scan line.
```

rpauszek (approved):
This is a really nice convenience feature. I left some suggestions for wording and I think it's a good idea to include an optional `reduce` argument like we do for other downsampling methods

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Yeah, I was doubting about that. I'll add it.

PR #686, opened 2024-08-02
api: add low level api for confocal objects / slice

JoepVanlier:
**Why this PR?**
Other projects would benefit from us providing a stable API that is a bit lower level than what we typically expose. By covering it with tests, we can make sure that we don't break things they depend on accidentally.

In particular, the ability to create confocal objects directly from bluelake metadata and slices from arrays would be useful. Considering that we may want to still the freedom to change some of the constructors of `Slice` and `Continuous` in the future, I would suggest exposing a free function to create them for now.


lumicks/pylake/tests/test_channels/test_channels.py
```diff
@@ -981,3 +982,23 @@ def start(self):
 
     with pytest.raises(TypeError, match=invalid_range):
         s[BadType(0)]
+
+
+def test_low_level_construction():
+    start = 1388534400000000000
+    with pytest.raises(ValueError, match="Starting timestamp must be larger than"):
+        _ = make_continuous_slice(np.array([1, 2, 3]), start - 1, int(1e9 / 78125))
```
tiagobonetti, line 990: Is the `_` necessary?
```suggestion
        make_continuous_slice(np.array([1, 2, 3]), start - 1, int(1e9 / 78125))
```

lumicks/pylake/tests/data/mock_confocal.py
```diff
@@ -121,28 +126,19 @@ def generate_photon_count_line(line):
 
 
 class MockConfocalFile:
```
tiagobonetti, line 128: Is this class necessary anymore?
How about just renaming the static functions:
`MockConfocalFile.from_image` -> `mock_confocal_from_image`
`MockConfocalFile.from_streams` -> `mock_confocal_`

lumicks/pylake/detail/confocal.py
```diff
@@ -167,14 +171,36 @@ def from_json(cls, json_string):
         return cls(axes, json_dict["scan volume"]["center point (um)"], json_dict["scan count"])
 
 
+class ConfocalFile:
```
tiagobonetti, line 174: This is not really a file so how about naming this something like `ConfocalFileProxy`?

lumicks/pylake/lowlevel.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+def create_confocal_object(
+    name,
+    infowave,
+    json_metadata,
+    red_channel=empty_slice,
+    green_channel=empty_slice,
+    blue_channel=empty_slice,
+) -> Kymo | Scan:
```
tiagobonetti, line 18: I wonder why the mock version of this function got fully typed arguments while this doesn't. Is this a conscious choice?

tiagobonetti (commented):
I am feeling way more comfortable with Python these days but still, I'm not sure if I should be the one approving it. But overall the code looks good!

I also think the commits should be squashed into a single one.

lumicks/pylake/detail/confocal.py
JoepVanlier: Yeah, good point. Fixing...

lumicks/pylake/tests/test_channels/test_channels.py
JoepVanlier: Not needed indeed.

lumicks/pylake/tests/data/mock_confocal.py
JoepVanlier: Not needed anymore indeed. It's a bad habit of mine to sometimes use classes as namespaces. I'll replace them with free functions. Doing this, I realized that this thing is actually not taking slices. It's taking raw numpy arrays, so I changed the second one to `mock_confocal_from_arrays`.

lumicks/pylake/detail/confocal.py
```diff
@@ -167,14 +171,36 @@ def from_json(cls, json_string):
...
+
 class BaseScan(PhotonCounts, ExcitationLaserPower):
     """Base class for confocal scans
 
     Parameters
     ----------
     name : str
         confocal scan name
-    file : lumicks.pylake.File
+    file : lumicks.pylake.File | lumicks.pylake.ConfocalFile
```
tiagobonetti, line 203:
```suggestion
    file : lumicks.pylake.File | lumicks.pylake.ConfocalFileProxy
```

lumicks/pylake/kymo.py
```diff
@@ -71,7 +71,7 @@ class Kymo(ConfocalImage):
     ----------
     name : str
         Kymograph name
-    file : lumicks.pylake.File
+    file : lumicks.pylake.File | lumicks.pylake.detail.confocal.ConfocalFile
```
tiagobonetti, line 74:
```suggestion
    file : lumicks.pylake.File | lumicks.pylake.detail.confocal.ConfocalFileProxy
```

lumicks/pylake/scan.py
```diff
@@ -18,7 +18,7 @@ class Scan(ConfocalImage, VideoExport, FrameIndex):
     ----------
     name : str
         Scan name
-    file : lumicks.pylake.File
+    file : lumicks.pylake.File | lumicks.pylake.detail.confocal.ConfocalFile
```
tiagobonetti, line 21:
```suggestion
    file : lumicks.pylake.File | lumicks.pylake.detail.confocal.ConfocalFileProxy
```

lumicks/pylake/lowlevel.py
JoepVanlier: This was intentional. We usually do it either in the docstring (for our (semi-)public stuff) or in the parameter list, but not both to not repeat ourselves.

For internal stuff like this, we usually keep the docstring very short as people working on the code will gauge what it is doing quickly, hence I figured the type hints would be preferable here. I'm not opposed to changing it though.

JoepVanlier:
> I also think the commits should be squashed into a single one.

Yeah I agree. I usually try to keep it into smaller commits than I intend to merge to make it easier to review as units.

lumicks/pylake/detail/confocal.py
JoepVanlier: Haha, that was fast :D

rpauszek (approved):

PR #687, opened 2024-08-07
confocal: fix bug with empty power channels

JoepVanlier:
**Why this PR?**
If the confocal diagnostics channel has no points before the item, these properties raise, while they should just return either an empty slice (if there is really no data) or a slice with the data that is still present.

### DIFF
--- a/lumicks/pylake/detail/mixin.py
+++ b/lumicks/pylake/detail/mixin.py
@@ -1,4 +1,5 @@
 """Mixin class which add properties for predefined channels"""
+
 from ..channel import Slice, empty_slice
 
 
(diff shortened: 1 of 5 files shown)


lumicks/pylake/detail/mixin.py
```diff
@@ -205,22 +206,28 @@ class ExcitationLaserPower:
 
     def _get_laser_power(self, name):
         power_data = self.file["Confocal diagnostics"][f"Excitation Laser {name}"]
-        # fetch the timestamp of the last datapoint before the beginning of the item
-        start_time = power_data[: self.start].timestamps[-1]
+
+        # fetch the timestamp of the last datapoint before the beginning of this item
+        before_item = power_data[: self.start]
+        start_time = before_item.timestamps[-1] if before_item else self.start
```
mikhas, line 212: Since `before_item` isn't used elsewhere, perhaps:
```suggestion
        start_time = before_item.timestamps[-1] if before_item := power_data[: self.start] else self.start
```

JoepVanlier: Just fyi, that's not valid python. You need a pair of brackets there.

```python
start_time = before_item.timestamps[-1] if (before_item := power_data[: self.start]) else self.start
```

Personally, I find the separate assignment more readable since it reads more sequentially but I don't have a strong preference, so I wouldn't mind changing it either.

mikhas: The extra parentheses make this pattern less visually attractive. I was surprised to learn that the Walrus operator has lower precedence than even `=`: https://peps.python.org/pep-0572/#relative-precedence-of  when intuitively, it should have had equal or higher precedence.

JoepVanlier: True. But it is what it is. I'm sure they had reasons for it. The low precedence leads to a few potential gotchas. Something like `if value := reduce(x) > 0.0` is a very easy mistake to make and one I've seen a few times.

Do you prefer the old variant now that it got uglier? 😅

mikhas: It's entirely your choice.

mikhas: > Something like `if value := reduce(x) > 0.0` is a very easy mistake to make and one I've seen a few times.

What I've learned from the PEP is that `:=` expressions need parentheses in all cases but the iconic
```python
if y := f(x):
    ...
```

JoepVanlier: Pretty much yeah. But that iconic case is probably mostly what it was introduced for.

rpauszek (approved):

PR #688, opened 2024-08-09
caching: centralize caching

JoepVanlier:
### 1. Allows us to monitor memory consumption of all the caches.

By centralizing the channel caching, we can monitor the memory consumption by the caches. In certain cases, because we now store caches uniquely by location (rather than independently in every single channel object), it can lead to less memory consumption than before.

### 2. It fixes subtle sources of bugs

We can also enforce that the slices are read only resulting. This prevents the following source of "bugs":

```python
ch1 = file["Force HF"]["Force 1x"]
ch2 = file["Force HF"]["Force 1x"]

ch1.data[0:100] = 0
print(ch2.data)  # Shows zeroes for the ones we overwrote in channel 1.
```
And instead prompts the user to explicitly make a copy if they want to change the data. We already do this for certain arrays coming from confocal objects.

```python
ch1 = file["Force HF"]["Force 1x"]
ch2 = file["Force HF"]["Force 1x"]
ch1.data[0:100] = 0  # ValueError: assignment destination is read-only

copy = ch1.data.copy()  # Explicit copy!
copy[0:100] = 0
print(ch2.data)  # Data in ch1 was not modified
```

Note that in-place operations on slices are still fine though (which is nice, because those are often used in de/recalibrating). For example::

```python
ch1 = file.force1x
ch2 = file.force1x

assert id(ch1.data) == id(ch2.data)
ch1 /= 5

assert id(ch1.data) != id(ch2.data)
np.testing.assert_allclose(ch1.data * 5, ch2.data)
```

### 3. It makes certain access patterns much more performant

Currently doing something like:
```python
file = lk.File("file.h5")
for t in tracked_lines:
    track_force = t.sample_from_channel(file.force1x, include_dead_time=True)
```
Is horrendously slow clocking in at `42 +- 5` seconds for a small `30` second kymograph with `50` tracks. This reduces to `1.6 +- 0.1` seconds with the new cache. The following would have also worked to fix the problem without the cache:

```python
file = lk.File("file.h5")
ch = file.force1x
for t in tracked_lines:
    track_force = t.sample_from_channel(ch, include_dead_time=True)
```

But would require additional user training and doesn't have the benefit of allowing us to monitor how much data is going into these caches going forward.

### 1. Small versus big slices

While doing this, I noticed that for `Continuous` slices, we always read the whole slice even if we take only a small segment. Note that this was also the case before the caching was introduced (bug?).

At first, this seemed wasteful, so I set up the cache to allow slicing first and then grab the data (including start index and stop index in the location), but when you take many small slices, the performance is frequently far worse than reading the big slice as a whole at once. As a result, the case discussed above with the tracks still took 23 seconds.

One possible improvement to the cache system could be to include a start and stop index in the location again for continuous channels. Then keep track of how frequently a slice of a continuous channel with a particular path is accessed and if that exceeds some threshold assume that the user will be taking more chunks, dump the small chunks and load the whole slice instead. Then in the cache grabber, we could always check if the parent location (the one without additional indexing) is already present, and slice from that if present.

Right now, from my testing so far, this additional complexity does not seem worth it though. Most `h5` files contain one or two items and generally, you look at the whole thing before you start interacting with sub-slices. The only iffy part would be if a single slice would exceed the memory capacity. But right now, you also have no way of loading those without going through the raw `h5` structure.

### 2. What is a good default cache size?

I currently hardcoded 1 GB, but I suspect this should depend on the user system somehow? I think making it proportional to a size in bytes makes more sense than number of items however. Especially considering how much the items can vary in size.

### 3. API

Do we want to expose an API for users to set cache size or disable it entirely?

### Some final notes

1. When reading data from `TimeTag` channels, we actually read the data 3 times already even when we don't want to access it at all 4 by the time we actually read it. Note that we can fix this without having to resort to a global cache however.

2. The default functools lru cache is great, but items can have pretty different sizes, so. I think it's better to go with one where you can set a custom `size_of`.

3. Considering that disallowing data modification is a breaking change, I would propose we iterate on this a bit now, and roll this out when `2.0` rolls around. Thoughts?

### DIFF
--- a/lumicks/pylake/channel.py
+++ b/lumicks/pylake/channel.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import numbers
 from typing import Union
+from functools import lru_cache
 
 import numpy as np
 import numpy.typing as npt
@@ -9,6 +10,40 @@ import numpy.typing as npt
 from .detail.timeindex import to_timestamp
 from .detail.utilities import downsample
 from .nb_widgets.range_selector import SliceRangeSelectorWidget
+
+
+@lru_cache(maxsize=100)
+def _get_array(cache_object):
+    return cache_object.read_array()
+
+
+class LazyCache:
+    def __init__(self, location, dset):
+        """A lazy globally cached wrapper around an object that is convertible to a numpy array"""
+        self._location = location
+        self._dset = dset
+
+    def __len__(self):
+        return len(self._dset)
+
+    def __hash__(self):
+        return hash(self._location)
+
+    @staticmethod
+    def from_h5py_dset(dset):
+        location = f"{dset.file.filename}{dset.name}"
+        return LazyCache(location, dset)
+
+    def read_array(self):
+        arr = np.asarray(self._dset)
+        arr.flags.writeable = False
+        return arr
+
+    def __eq__(self, other):
+        return self._location == other._location
+
+    def __array__(self):
+        return _get_array(self)
 
 
 class Slice:
@@ -550,7 +585,6 @@ class Continuous:
 
     def __init__(self, data, start, dt):
         self._src_data = data
-        self._cached_data = None
         self.start = start
         self.stop = start + len(data) * dt
         self.dt = dt  # ns
@@ -571,7 +605,7 @@ class Continuous:
         start = dset.attrs["Start time (ns)"]
         dt = int(1e9 / dset.attrs["Sample rate (Hz)"])  # ns
         return Slice(
-            Continuous(dset, start, dt),
+            Continuous(LazyCache.from_h5py_dset(dset), start, dt),
             labels={"title": dset.name.strip("/"), "y": y_label},
             calibration=calibration,
         )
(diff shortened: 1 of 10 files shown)


tiagobonetti:
How about starting by adding the caching as an opt-in global config? This would allow us to test the caching in Lakeview.

lumicks/pylake/channel.py
```diff
@@ -796,32 +801,16 @@ def _apply_mask(self, mask):
...
+        data = (
+            LazyCache.from_h5py_dset(dset, field="Value")
+            if caching.global_cache
+            else dset.fields("Value")
+        )
+        timestamps = (
+            LazyCache.from_h5py_dset(dset, field="Timestamp")
+            if caching.global_cache
+            else dset.fields("Timestamp")
+        )
```
tiagobonetti, line 813: I think we should move the cache checking into functions in the `caching` package.
Those would be concerned about what caching to use while this code is only about loading the channel data.

I added a basic example below, but I think we can add other ones that cover the other usages from loading `numpy` arrays and such.

```py
# in caching
def from_h5py(dset, field=None):
    global global_cache
    return (
        LazyCache.from_h5py_dset(dset, field=field)
        if global_cache
        else dset.fields(field)
    )

# here
    @staticmethod
    def from_dataset(dset, y_label="y", calibration=None) -> Slice:
        data = caching.from_h5py(dset, field="Value")
        timestamps = caching.from_h5py(dset, field="Timestamp")
```

JoepVanlier: Yeah, that's a good point. That keeps it all in one place. Easier to change.

I've added the method you suggested and put the retrieval in `caching` as well. Decided to go with a mixin here, since then we can keep the storage with the code that stores.

PR #689, opened 2024-08-19
ci: don't modify test parametrization during test

JoepVanlier:
**Why this PR?**
This fixes this particular test failing after the first time it is run. I changed the input to a tuple, since it also discourages a similar thing from happening again in the future.

### DIFF
--- a/lumicks/pylake/tests/test_scalebar.py
+++ b/lumicks/pylake/tests/test_scalebar.py
@@ -1,5 +1,3 @@
-import json
-
 import numpy as np
 import pytest
 import matplotlib.pyplot as plt
@@ -25,7 +23,7 @@ def _validate_elements(ref_elements, item):
             else:
                 ref_elements.pop(idx)
         else:
-            raise AssertionError("Missing element")
+            raise AssertionError(f"Missing element {item}")
 
 
 @pytest.mark.parametrize(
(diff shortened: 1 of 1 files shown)


mikhas (commented):
Neither commit message nor PR description explain
1. What caused the bug
2. How the changed code (enforcing a list -> generator related issue?) fixes the issue.

Probably needs a code comment even so someone doesn't refactor it back to the old version.

JoepVanlier:
> Neither commit message nor PR description explain
> 
>     1. What caused the bug
> 
>     2. How the changed code (enforcing a list -> generator related issue?) fixes the issue.
> 
> 
> Probably needs a code comment even so someone doesn't refactor it back to the old version.

It is in the commit message? We were modifying the test parametrization.

mikhas:
I didn't mean to sound passive-aggressive, apologies if I did.

> I've added the following to the message and added a comment to the code:

To me the code comment is useful, but that's also because I was unaware of the whole issue. Thanks!

> ```
> Previously, the test parametrization was using a mutable type, which got modified by the test itself. 
> This results in the test failing every run after the first. By changing it to an immutable type, we 
> are forced to first convert it into a modifiable list explicitly in the test (copying its contents) 
> making sure we don't modify the source data in the parametrization.
> ```

You probably didn't notice, but this is slightly different from what you wrote in the PR description :) Initially, I really thought this test would fail every second time it's run: `1. pass 2. fail 3. pass 4. fail` and so on. But it's `1. pass 2. fail 3.fail 4. fail` which makes more sense of course.

> This kind of bug is really common in `python`-land and the reason linters usually pick up on mutable types in input arguments.

I had no idea this is common.

JoepVanlier:
> I didn't mean to sound passive-aggressive, apologies if I did.

No worries.

> > I've added the following to the message and added a comment to the code:
> 
> To me the code comment is useful, but that's also because I was unaware of the whole issue. Thanks!

Yeah, I think it's also less obvious with these test parametrizations. I figured changing it to a tuple would've been sufficient to prevent the bug from coming back (since there is no reason to change the tuple back to a list), but the comment doesn't hurt.

> > ```
> > Previously, the test parametrization was using a mutable type, which got modified by the test itself. 
> > This results in the test failing every run after the first. By changing it to an immutable type, we 
> > are forced to first convert it into a modifiable list explicitly in the test (copying its contents) 
> > making sure we don't modify the source data in the parametrization.
> > ```
> 
> You probably didn't notice, but this is slightly different from what you wrote in the PR description :) Initially, I really thought this test would fail every second time it's run: `1. pass 2. fail 3. pass 4. fail` and so on. But it's `1. pass 2. fail 3.fail 4. fail` which makes more sense of course.

Ah, yes! Let me edit that. It is indeed confusing.

> > This kind of bug is really common in `python`-land and the reason linters usually pick up on mutable types in input arguments.
> 
> I had no idea this is common.

Yeah, parametrization in this case is just passed into the test directly, so we end up with references to the actual data in the test parametrization. It is a bit of a gotcha.

Actually, on second thought, I do think the situation linters warn about is different as that pertains to defaults. In Python, functions are just objects and the defaults are evaluated and stored on evaluating the `def` statement that defines them (as in: not when the function is actually called). So if you mutate the mutable in the function, then call the function again, it persists.

It feels a bit similar in that it is a mutable gotcha, but the mechanics are different.

rpauszek (approved):

PR #690, opened 2024-08-19
dwelltime: add standard errors based on hessian approximation

JoepVanlier:
**Why this PR?**
For some purposes, we need a quantity that calculates an approximate standard error rapidly.
Asymptotics are useful especially when the problem is well constrained (lots of data and a well-suited model).

image

_2-component fit based on 100 dwells. Profile likelihood compared to asymptotic intervals. Confidence intervals would be given by the locations where the curves meet the threshold (dashed horizontal line)._

image

_2-component fit based on 1000 dwells. Profile likelihood compared to asymptotic intervals. Confidence intervals would be given by the locations where the curves meet the threshold (dashed horizontal line). Note how for lots of data, these two approaches produce almost identical results._

I have added the ability to plot them alongside the profiles, as it can be instructive to see how these type of errors compare. I think in the future, it would also be useful to explore this on the FdFitting side as well.

Small note on the implementation:

It is important to consider the constraint on the amplitudes when doing this uncertainty analysis. Without it, you get huge confidence intervals for alpha (since the problem is underdertermined). There are two ways of doing this. One you add entries to the Hessian corresponding to the constraint or you transform the derivatives into a subspace that fulfills the constraint, invert there and convert back. I chose the latter, since otherwise, we would have to choose how heavily we set this constraint constant. In the plot below you can see what varying the constraint does. You can basically see it converge to the solution we have now.

image

_Comparing the current approach (dash-dot) with an approach where we take into account the effect of the constraint by adding values to components of the Hessian manually. Note how the approaches agree for large values of C._

The risk with the constant based one is that if you choose the constant too large, it blows up (see figure below), whereas if you choose it too small you don't take the constraint into account sufficiently.

image

_Comparing the current approach (dash-dot) with an approach where we take into account the effect of the constraint by adding values to components of the Hessian manually. Note how excessively large values result in numerical problems._

Considering @rpauszek  is working on the dwell time documentation at the moment, I have deferred writing documentation for this specifically at this time.

### DIFF
--- a/lumicks/pylake/population/dwelltime.py
+++ b/lumicks/pylake/population/dwelltime.py
@@ -106,8 +106,10 @@ class DwelltimeProfiles:
         )
         parameters = Params(
             **{
-                key: Parameter(param, lower_bound=lb, upper_bound=ub)
-                for key, param, (lb, ub) in zip(keys, dwelltime_model._parameters, bounds)
+                key: Parameter(param, lower_bound=lb, upper_bound=ub, stderr=std_err)
+                for key, param, (lb, ub), std_err in zip(
+                    keys, dwelltime_model._parameters, bounds, dwelltime_model._std_errs
+                )
             }
         )
 
@@ -204,7 +206,7 @@ class DwelltimeProfiles:
         """Number of components in the model."""
         return self.model.n_components
 
-    def plot(self, alpha=None):
+    def plot(self, alpha=None, **kwargs):
         """Plot the profile likelihoods for the parameters of a model.
 
         Confidence interval is indicated by the region where the profile crosses the chi squared
@@ -219,13 +221,23 @@ class DwelltimeProfiles:
         """
         import matplotlib.pyplot as plt
 
+        with_stderr = kwargs.pop("with_stderr") if "with_stderr" in kwargs else False
+
+        std_errs = self.model._std_errs[~np.isnan(self.model._std_errs)]
         if self.n_components == 1:
-            next(iter(self.profiles.values())).plot(significance_level=alpha)
+            next(iter(self.profiles.values())).plot(
+                significance_level=alpha,
+                std_err=std_errs[0] if with_stderr else None,
+            )
         else:
             plot_idx = np.reshape(np.arange(1, len(self.profiles) + 1), (-1, 2)).T.flatten()
-            for idx, profile in zip(plot_idx, self.profiles.values()):
+            for par_idx, (idx, profile) in enumerate(zip(plot_idx, self.profiles.values())):
                 plt.subplot(self.n_components, 2, idx)
-                profile.plot(significance_level=alpha)
+                profile.plot(
+                    significance_level=alpha,
+                    std_err=std_errs[par_idx] if with_stderr else None,
+                    **kwargs,
+                )
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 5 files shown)


changelog.md
```diff
@@ -17,6 +17,7 @@
...
 * Show the ranges that were excluded from a power spectrum or calibration fit by passing `show_excluded=True` to [`PowerSpectrum.plot()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum.html#lumicks.pylake.force_calibration.power_spectrum.PowerSpectrum.plot) or [`CalibrationResults.plot()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html#lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.plot).
 * Plot the active calibration peak for a calibration result using `show_active_peak=True` with [`CalibrationResults.plot()`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.html#lumicks.pylake.force_calibration.power_spectrum_calibration.CalibrationResults.plot).
+* Added `err_amplitudes` and `err_lifetimes` to [`DwelltimeModel`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.DwelltimeModel.html). These return asymptotic standard errors and should only be used for well identified models.
```
rpauszek, line 20: this should be removed as the functionality is private for now

lumicks/pylake/population/dwelltime.py
```diff
@@ -1328,7 +1360,16 @@ def jac_fun(params):
...
     current_params[fitted_param_mask] = result.x
-    return current_params, -result.fun
+
+    if use_jacobian:
+        std_errs = np.full(current_params.shape, np.nan)
+        std_errs[fitted_param_mask] = _calculate_std_errs(
+            jac_fun, constraints, num_free_amps, current_params, fitted_param_mask
+        )
+    else:
+        std_errs = np.full(current_params.shape, np.nan)
```
rpauszek, line 1370: I think you can get away with just this

```suggestion
    std_errs = np.full(current_params.shape, np.nan)
    if use_jacobian:
        std_errs[fitted_param_mask] = _calculate_std_errs(
            jac_fun, constraints, num_free_amps, current_params, fitted_param_mask
        )        
```

rpauszek (approved):
cool, I think this will be super useful to have merged and it'll make reliability testing easier

lumicks/pylake/population/dwelltime.py
```diff
@@ -632,34 +632,6 @@ def lifetimes(self):
...
-        Returns an asymptotic standard error on the amplitude parameters. These error estimates
-        are only reliable for estimates where a lot of data is available and the model does
-        not suffer from identifiability issues. To verify that these conditions are met, please
-        use either :meth:`DwelltimeModel.profile_likelihood()` method or
-        :meth:`DwelltimeModel.calculate_bootstrap()`.
-
-        Note that `np.nan` will be returned in case the parameter was either not estimated or no
-        error could be obtained."""
-        return self._std_errs[self.n_components :]
-
```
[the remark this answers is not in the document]
JoepVanlier: Nah, I think that's maybe a better option actually. I maybe jumped the gun removing them entirely.

PR #694, opened 2024-09-26
calibration: expose diode calibration

JoepVanlier:
**Why this PR?**
- This gives users an option to interrogate the parameter values and results of the diode calibration.
- After discussing this with hrishi, I have decided to return a small dataclass rather than expose direct access to the free function to ensure that we can change this model in the future if need be.

Docs build here: https://lumicks-pylake.readthedocs.io/en/expose_diode_model/_api/lumicks.pylake.calibration.ForceCalibrationItem.html#lumicks.pylake.calibration.ForceCalibrationItem.diode_calibration

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -41,6 +41,7 @@ Force calibration
     ActiveCalibrationModel
     force_calibration.power_spectrum.PowerSpectrum
     force_calibration.power_spectrum_calibration.CalibrationResults
+    force_calibration.calibration_models.DiodeCalibrationModel
 
     :template: function.rst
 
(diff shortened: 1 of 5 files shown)


docs/api.rst
```diff
@@ -157,4 +158,4 @@ Simulation
     :toctree: _api
     :template: function.rst
 
-    simulation.simulate_diffusive_tracks
\ No newline at end of file
+    simulation.simulate_diffusive_tracks
```
tiagobonetti, line 161: Is GitHub tripping?

tiagobonetti (commented):
The code looks good.
Maybe a more scientific review should be added still.

docs/api.rst
JoepVanlier: I added a newline at the end of the file 😅

rpauszek (approved):
LGTM

PR #695, opened 2024-09-27
PowerSpectrum: improve diagnostic value power spectrum api

JoepVanlier:
### Why this PR?
- A simpler constructor will allow easier construction of a `PowerSpectrum`. The benefits of this may not be evident in this PR, but a followup one will have pathological spectra constructed directly rather than ifft'ing the spectrum only to immediately fft it again.
- Keeping track of the uncut data will allow us to plot what we excluded.
- Keeping track of the high frequency data and performing the downsampling and exclusion/fit ranges lazily will allow us to do multiple analyses without the requirement to always recompute the expensive part (the FFT) or keep two separate power spectra around. It also means that we will be getting the full spectrum in the calibration routine, which will allow us to apply some improvements going forward.
- Plotting the driving peak for an active calibration result gives a feel for how far above the noise floor it sits.

image

#### Increased size
Keeping the high frequency spectrum around does come at a cost, namely the increased memory footprint. For a typical calibration (100 kHz sample rate) this is about 4 MB worth of data per axis for passive, 8 for active.

#### Downsampling and frequency exclusion ranges
There is one annoying issue with showing the frequency exclusion ranges. Exclusion ranges are applied _prior_ to downsampling, which means that downsampling can subsequently pull a point outside the exclusion ranges into it. This is not really a bug, as this allows us to make cuts smaller than the blocking size, but it can look weird.

image

### DIFF
--- a/lumicks/pylake/force_calibration/calibration_results.py
+++ b/lumicks/pylake/force_calibration/calibration_results.py
@@ -91,12 +91,36 @@ class CalibrationResults(CalibrationPropertiesMixin):
         """Number of fitted samples (-)."""
         return self.ps_data.total_sampled_used
 
-    def plot(self):
-        """Plot the fitted spectrum"""
+    def plot(self, show_excluded=False, show_active_peak=False):
+        """Plot the fitted spectrum
+
+        Parameters
+        ----------
+        show_excluded : bool
+            Show fitting regions excluded from the fit
+        show_active_peak : bool
+            Show active calibration peak when available
+
+        Raises
+        ------
+        ValueError
+            If specifying `show_active_peak=True` for a passive calibration.
+        """
         import matplotlib.pyplot as plt
 
-        self.ps_data.plot(label="Data")
+        if show_active_peak and not hasattr(self.model, "output_power"):
+            raise ValueError(
+                "Requested to plot an active calibration peak while this is not an active "
+                "calibration. Please specify `show_active_peak=False` when plotting passive"
+                "calibration results."
+            )
+
+        self.ps_data.plot(label="Data", show_excluded=show_excluded)
         self.ps_model.plot(label="Model")
+
+        if show_active_peak:
+            self.model.output_power.ps.plot(label="Active peak")
+
         plt.legend()
 
     def plot_spectrum_residual(self):
--- a/lumicks/pylake/force_calibration/power_spectrum.py
+++ b/lumicks/pylake/force_calibration/power_spectrum.py
@@ -22,7 +22,60 @@ class PowerSpectrum:
         The total duration of the original data. [seconds]
     """
 
-    def __init__(self, data, sample_rate, unit="V", window_seconds=None):
+    def __init__(
+        self,
+        frequency,
+        power,
+        sample_rate,
+        total_duration,
+        unit="V",
+        *,
+        num_points_per_block=1,
+        total_samples_used=None,
+        variance=None,
+    ):
+        """Power spectrum
+
+        frequency : array_like
+            Frequency axis
+        power : array_like
+            Power spectral values
+        sample_rate : int
+            Sample rate
+        total_duration : float
+            Total measurement duration
+        unit : str
+            Unit of the spectrum
+        num_points_per_block : int
+            Number of points per block of the source data
+        total_samples_used : int
+            Total samples used to compute FFT
+        variance : array_like
+            Variance of each power spectral density point
+        """
+        self._frequency = np.asarray(frequency)
+        self._power = np.asarray(power)
+        self.sample_rate = sample_rate
+        self.total_duration = total_duration
+        self.unit = unit
+
+        self.total_samples_used = total_samples_used
+        self._raw_variance = variance
+        self._src_num_points_per_block = num_points_per_block
+
+        self._downsampling_factor = 1
+        self._fit_range = (
+            np.nextafter(frequency.min(), -np.inf),
+            np.nextafter(frequency.max(), np.inf),
+        )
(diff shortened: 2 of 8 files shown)


lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -22,7 +46,60 @@ class PowerSpectrum:
...
+        """
+        self._frequency = np.asarray(frequency)
+        self._power = np.asarray(power)
+        self.sample_rate = sample_rate
+        self.total_duration = total_duration
+        self.unit = unit
+
+        self.total_samples_used = total_samples_used
+        self._raw_variance = variance
+        self._src_num_points_per_block = num_points_per_block
```
rpauszek, line 88: since this is private anyway, what's the `src` prefix for/mean?

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -72,37 +146,109 @@ def squared_fft(d):
...
+        return self.total_samples_used
 
     @property
-    def frequency_bin_width(self):
+    def frequency_bin_width(self) -> float:
         """Returns the frequency bin width of the spectrum"""
-        return self.sample_rate / self.total_sampled_used * self.num_points_per_block
+        return self.sample_rate / self.total_samples_used * self.num_points_per_block
+
+    def _apply_transforms(self, data, with_exclusions=True, downsample_data=True) -> np.ndarray:
```
rpauszek, line 176: I think a little docstring would be helpful here as it's not apparent to those of us that haven't delved deep into the intricacies of this math what transform needs to be applied or what it means

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -276,24 +408,50 @@ def with_spectrum(self, power, num_points_per_block=1, variance=None) -> "PowerS
...
+            self.sample_rate,
+            self.total_duration,
+            self.unit,
+            num_points_per_block=num_points_per_block,
+            total_samples_used=self.total_samples_used,
+            variance=variance,
+        )
 
-    def plot(self, **kwargs):
+    def plot(self, *, show_excluded=False, **kwargs):
```
rpauszek, line 422: 👍

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -276,24 +408,50 @@ def with_spectrum(self, power, num_points_per_block=1, variance=None) -> "PowerS
...
+            Show ranges that were excluded from fitting.
         **kwargs
             Forwarded to :func:`matplotlib.pyplot.plot`.
         """
         import matplotlib.pyplot as plt
 
-        plt.plot(self.frequency, self.power, **kwargs)
+        if show_excluded:
+            if self._downsampling_factor > 1:
+                plt.plot(self.unfiltered_frequency, self.unfiltered_power, label="_")
```
rpauszek, line 436: I had no idea that you could exclude labels from the legend like this 🤯

rpauszek (approved):
Looks good, although I'm not gonna pretend that I got to deep into the intricacies of the math that was going on but looks mostly like rearranging and as always thoroughly tested. Just some minor nitpicks, mostly some extra comments would be good for those of us that don't know force calibration like the back of our hand

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: Fair point! I'll add a docstring.

lumicks/pylake/force_calibration/power_spectrum.py
```diff
@@ -7,6 +7,30 @@
...
+    breaks = [e[1] for e in excluded_ranges]
+    break_idx = np.searchsorted(frequency, breaks, "left")
+
+    nan_insertion_points = np.hstack((break_idx, break_idx - 1))
+    return (
+        np.insert(frequency, nan_insertion_points, np.nan),
+        np.insert(power, nan_insertion_points, np.nan),
+        frequency[break_idx - 1],
+        power[break_idx - 1],
+    )
```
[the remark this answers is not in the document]
JoepVanlier: Yep! Added.

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: Yeah. It's pretty nice. :)

lumicks/pylake/force_calibration/calibration_results.py
```diff
@@ -91,12 +91,27 @@ def number_of_samples(self):
...
+        show_excluded : bool
+            Show fitting regions excluded from the fit
+        show_active_peak : bool
+            Show active calibration peak when available
+        """
         import matplotlib.pyplot as plt
 
-        self.ps_data.plot(label="Data")
+        if show_active_peak and not hasattr(self.model, "output_power"):
+            raise ValueError("This is not an active calibration")
```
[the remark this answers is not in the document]
JoepVanlier: I was doubting about this too, but the error on the user side in this case is that they specified an incorrect value (they did not attempt to access this attribute themselves). That said, I will add a raises section, since it's generally a good idea to mention those.

lumicks/pylake/force_calibration/power_spectrum.py
JoepVanlier: It stands for source and was a reminder that that refers to the num points per block of the source material (pre blocking). That said, it is probably superfluous, since you can so clearly see it in the property already.

PR #696, opened 2024-10-02
download: allow non-url characters in filenames

JoepVanlier:
**Why this PR?**
We should not restrict what file names can be used in zenodo links as it will prohibit people from using this in their own notebooks.

The surface calibration data contains filenames with spaces which made me realize that we don't support non-url characters in file names.


tiagobonetti (approved):
Simple, tested an well documented.
Easy `honk`!

tobiasjj (approved):
Well tested. LGTM! 🚴

PR #697, opened 2024-10-02
docs: support surface calibration

JoepVanlier:
**Why this PR?**
Currently, users cannot reproduce the full surface calibration workflow.

This worked example will allow them to do so, and hopefully see the advantage active calibration brings in that situation.


aafkevandenberg (commented):
Hi Joep, this is super nice, this will be very useful.

I have to read this a couple of times to understand the details. But as a very general initial feedback I would say to splits this notebook in 2:

1) First just list the steps that a user has to take to start using the activate calibration, in exactly the right order. So, somebody who downloads the notebook just can load his/her own data and quickly go through the steps. (And, link to the page on the pylake docs that explains how to recalibrate the force using the obtained stiffness)
2) The second part of the docs can explain the background and the mechanisms that are taken into account, and how active and passive calibration are compared.

JoepVanlier:
> Hi Joep, this is super nice, this will be very useful.
> 
> I have to read this a couple of times to understand the details. But as a very general initial feedback I would say to splits this notebook in 2:
> 
>     1. First just list the steps that a user has to take to start using the activate calibration, in exactly the right order. So, somebody who downloads the notebook just can load his/her own data and quickly go through the steps. (And, link to the page on the pylake docs that explains how to recalibrate the force using the obtained stiffness)
> 
>     2. The second part of the docs can explain the background and the mechanisms that are taken into account, and how active and passive calibration are compared.

Ah yes, 1 is part of the force calibration rewrite. I spun this off from that branch, since the size of it was getting out of hand. This PR really only addresses point 2 for one more advanced use case (surface calibration).

PR #698, opened 2024-10-07
ci: make sure we used fixed seed in fixtures

JoepVanlier:
**Why this PR?**
I notice some of our tests fail in some rare cases. I verified that these were false positives and they were related to the rng involved in making the fixture. This PR makes sure that all the fixtures are using a fixed seed.

Note that in the first commit I added a folder to the gitignore list, this is since on `main` we ignore a big `nb_test` folder that is used for semi-automatically testing notebooks that I would not like to have to stash or drop when rebasing every time. I intend to squash this to one commit.

### DIFF
--- a/lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
+++ b/lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
@@ -8,7 +8,7 @@ from lumicks.pylake.kymotracker.kymotrack import *
 
 @pytest.fixture(scope="module")
 def kymos():
-    image = np.random.poisson(10, size=(10, 10, 3))
+    image = np.random.default_rng(1337).poisson(10, size=(10, 10, 3))
 
     def make_kymo(line_time_seconds, pixel_size_um):
         return _kymo_from_array(
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@ dist/
 venv/
 docs/changelog.md
 docs/_api/
+nb_test/
 
 __pycache__/
 *.py[cod]
--- a/lumicks/pylake/kymotracker/tests/conftest.py
+++ b/lumicks/pylake/kymotracker/tests/conftest.py
@@ -55,7 +55,8 @@ def kymo_integration_tracks(kymo_integration_test_data):
 @pytest.fixture
 def kymo_pixel_calibrations():
     image = raw_test_data()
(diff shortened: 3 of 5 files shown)


lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
```diff
@@ -8,7 +8,7 @@
 
 @pytest.fixture(scope="module")
 def kymos():
-    image = np.random.poisson(10, size=(10, 10, 3))
+    image = np.random.default_rng(1337).poisson(10, size=(10, 10, 3))
```
mikhas, line 11: Would turn `np.random.default_rng(number)` into an `rng` fixture itself, if possible since the pattern repeats so often.

mikhas (approved):

lumicks/pylake/kymotracker/tests/test_kymotrackgroup_sources.py
JoepVanlier: The scope of many of these fixtures is module (meaning they get constructed once for the whole module, not per test; which saves a lot of generation time). However, that means I can only use fixtures defined at the module scope inside of them (pytest limitation).

This in turn means that if I were to turn this generator into a fixture, all of the uses of that fixture would get the same generator (which only gets created once) which introduces a new source of randomness again (test order).

I could make it a free function, but it is already a free function, so I'm not sure that adds much. If it's merely about length, I could do a full import to remove the `np.random` part.

What do you think?

mikhas: Not worth it. Keep as is.

tobiasjj (approved):
I guess with a fixed seed you generate _predetermined_ random numbers and still get representative samples cheaply?!
Out of curiosity: How did the sampling of `rng` cause the false positives in the tests? Did it generate really unlikely distributed samples?
LGTM! 🚴

JoepVanlier:
> I guess with a fixed seed you generate _predetermined_ random numbers and still get representative samples cheaply?! Out of curiosity: How did the sampling of `rng` cause the false positives in the tests? Did it generate really unlikely distributed samples? LGTM! 🚴

Yes. The background level for the Kymo happened to bump the percentile that is used for determining the threshold in some tracking tests just over the next integer value thereby changing the results.

PR #704, opened 2024-10-30
channel: support recalibrating force data

JoepVanlier:
**Why this PR?**
It is good to be able to recalibrate your force data.

By providing an API to do this, you basically end up with a force slice that can be treated as any other force slice (with a calibration item attached that conforms to a common API).

For instance, a recalibrated force slice can now directly be fed to the piezo tracking API (which reads from the calibration item).

Prior to implementing the new functionality, I removed some old cruftiness from the API that didn't really gel well with the new API. I left this as a separate commit.

Usage example:
```python
  import lumicks.pylake as lk
  f = lk.File("passive_calibration.h5")
  calibration = f.force1x.calibration[1]  # Grab a calibration item for force 1x

  # Slice the data corresponding to the calibration we want to reproduce.
  calib_slice = f.force1x[calibration]

  # De-calibrate to volts using the calibration that was active before this slice.
  previous_calibration = calib_slice.calibration[0]
  calib_slice = calib_slice / previous_calibration.force_sensitivity
  calibration_params = previous_calibration.calibration_params()
  new_calibration = lk.calibrate_force(calib_slice.data, **calibration_params)
  new_calibration.plot()

  # Make a new calibration, but change the amount of blocking
  less_blocking_params = calibration_params | {"num_points_per_block": 200}
  less_blocking = lk.calibrate_force(calib_slice.data, **less_blocking_params)
  less_blocking.plot()

  # Recalibrate the force channels
  recalibrated_force1x = f.force1x.recalibrate_force(less_blocking)
```

### DIFF
--- a/docs/tutorial/force_calibration/calibration_items.rst
+++ b/docs/tutorial/force_calibration/calibration_items.rst
@@ -160,11 +160,9 @@ We can also see that the residual now should less systematic deviation::
 
 .. image:: figures/residual_better.png
 
-Now that we have our new calibration item, we can recalibrate a slice of force data.
-To do so, take the slice and multiply it by the ratio of the old and new calibration factors::
+Now that we have our new calibration item, we can recalibrate a slice of force data::
 
-    correction_factor = recalibrated_hyco.force_sensitivity / old_calibration.force_sensitivity
-    recalibrated_force1x = force1x_slice * correction_factor
+    recalibrated_force1x = force1x_slice.recalibrate_force(recalibrated_hyco)
 
     plt.figure()
     force1x_slice.plot()
--- a/lumicks/pylake/calibration.py
+++ b/lumicks/pylake/calibration.py
@@ -5,13 +5,13 @@ from tabulate import tabulate
 from lumicks.pylake.force_calibration.calibration_item import ForceCalibrationItem
 
 
-def _filter_calibration(time_field, items, start, stop):
+def _filter_calibration(items, start, stop):
     """filter calibration data based on time stamp range [ns]"""
     if len(items) == 0:
         return []
 
     def timestamp(x):
-        return x[time_field]
+        return x.stop if x.stop else x.applied_at  # Pylake items do not have a start and stop (yet)
 
     items = sorted(items, key=timestamp)
 
@@ -38,25 +38,22 @@ class ForceCalibrationList:
         calibration = f.force1x.calibration[1]  # Grab a calibration item for force 1x
     """
 
-    def __init__(self, time_field, items, slice_start=None, slice_stop=None):
+    def __init__(self, items, slice_start=None, slice_stop=None):
         """Calibration item
 
         Parameters
         ----------
-        time_field : string
-            name of the field used for time
         items : list[ForceCalibrationItem]
             list of force calibration items
         slice_start, slice_stop : int
             Start and stop index of the slice associated with these items
         """
-        self._time_field = time_field
         self._src = items
         self._slice_start = slice_start
         self._slice_stop = slice_stop
 
     def _with_src(self, _src):
-        return ForceCalibrationList(self._time_field, _src)
+        return ForceCalibrationList(_src)
 
     def __getitem__(self, item):
         if isinstance(item, slice):
@@ -86,14 +83,13 @@ class ForceCalibrationList:
         stop  : int
             time stamp at stop [ns]"""
         return ForceCalibrationList(
-            self._time_field,
-            _filter_calibration(self._time_field, self._src, start, stop),
+            _filter_calibration(self._src, start, stop),
             start,
             stop,
         )
 
     @staticmethod
-    def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibrationList":
+    def from_field(hdf5, force_channel) -> "ForceCalibrationList":
         """Fetch force calibration data from the HDF5 file
 
         Parameters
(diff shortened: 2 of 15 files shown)


aafkevandenberg (commented):
Very nice addition, this is much easier and cleaner way to recalibrate. 

The example you mention would fit well in the force calibration docs. I guess you do that in a separate PR?

JoepVanlier:
> Very nice addition, this is much easier and cleaner way to recalibrate.
> 
> The example you mention would fit well in the force calibration docs. I guess you do that in a separate PR?

Yeah. I was considering that too, but I figure, with the big shuffle, it doesn't make a lot of sense to add it now.

However, I could rebase this one on the docs rewrite and just add it there.

lumicks/pylake/channel.py
```diff
@@ -239,6 +239,65 @@ def calibration(self) -> list:
...
+                "Slice does not contain any calibration items. Is this an unprocessed force "
+                "channel?"
+            )
+
+        # Grab sensitivities. If for whatever reason, we don't know part of the calibration, we
+        # can't reasonably recalibrate, so we return NaNs for those time points.
+        timestamps = self.timestamps
+        force_sensitivities = np.full(len(timestamps), np.nan)
+        for c in self.calibration:
+            force_sensitivities[timestamps >= c.applied_at] = c.force_sensitivity
```
rpauszek, line 291: is it that expensive here to look up `timestamps` twice-ish that it should be a separate variable? I can't imagine a slice with like 50 calibrations where this loop would run signifcantly long. If not, I think I prefer to just use `self.timestamps` at each location

lumicks/pylake/piezo_tracking/tests/conftest.py
```diff
@@ -95,8 +95,10 @@ def piezo_tracking_test_data(poly_baseline_data, camera_calibration_data):
 
         def implicit_trap_position_equation(x):
             trap_trap_dist = x - trap2_ref
-            displacement = 2 * (force + baseline(x)) / stiffness_um
-            return (tether_dist + 2 * bead_radius + displacement - trap_trap_dist) ** 2
+            displacement = 2 * (force + baseline(x)) / stiffness_um  # noqa: B023
```
rpauszek, line 98: what's a noqa?

rpauszek (approved):
Very nice feature 🏆 Just left some nit-picky comments and thoughts, feel free to use or ignore

rpauszek:
one thing I realized after reviewing, if it's feasible it might be a good idea to add a check that the slice is actually force data. Not exactly sure of the best way to go about this, but just something to think about

docs/tutorial/force_calibration/calibration_items.rst
```diff
@@ -163,8 +163,7 @@ We can also see that the residual now should less systematic deviation::
 Now that we have our new calibration item, we can recalibrate a slice of force data.
 To do so, take the slice and multiply it by the ratio of the old and new calibration factors::
```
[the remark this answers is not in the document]
JoepVanlier: Good catch!

lumicks/pylake/calibration.py
```diff
@@ -102,23 +98,25 @@ def from_field(hdf5, force_channel, time_field="Stop time (ns)") -> "ForceCalibr
...
                 attrs = dict(calibration_item[force_channel].attrs)
-                if time_field in attrs.keys():
-                    # Copy the timestamp at which the calibration was applied into the item
-                    attrs["Timestamp (ns)"] = calibration_item.attrs.get("Timestamp (ns)")
-                    items.append(ForceCalibrationItem(attrs))
 
-        return ForceCalibrationList(time_field=time_field, items=items)
+                # Copy the timestamp at which the calibration was applied into the item
+                attrs["Timestamp (ns)"] = calibration_item.attrs.get("Timestamp (ns)")
+                items.append(ForceCalibrationItem(attrs))
```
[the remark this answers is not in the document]
JoepVanlier: Looks nice. Then we can inline it with the construction of the `ForceCalibrationList` as well.

lumicks/pylake/force_calibration/calibration_results.py
```diff
@@ -48,6 +50,17 @@ def __call__(self, frequency):
...
         return self.model(frequency, *self.fitted_params)
 
+    def _with_timestamp(self, applied_timestamp):
+        """Return a copy of this item with a timestamp of when it was applied"""
+        from lumicks.pylake.force_calibration.power_spectrum_calibration import CalibrationParameter
+
+        item = copy.copy(self)
+        item.params = copy.deepcopy(self.params)
+        item.params["Timestamp"] = CalibrationParameter(
+            "Timestamp when item was applied", applied_timestamp, "nanoseconds"
```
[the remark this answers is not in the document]
JoepVanlier: Currently it's not, item and results have their own different implementation for this one property. I think that is something we should tackle when we get named parameters on the other side as I'm not sure whether moving the string elsewhere will make it any safer.

lumicks/pylake/channel.py
JoepVanlier: Not expensive. I'll swap it out.

lumicks/pylake/piezo_tracking/tests/conftest.py
JoepVanlier: It silences the linter about B023.

In this case, the linter warns us about the fact that that function doesn't bind `tether_dist` and `force` from the enclosing loop. That might be an issue if you call the function _after_ the loop, but in this case the behavior you get is fine since the function is only ever used inside the loop.

JoepVanlier:
> one thing I realized after reviewing, if it's feasible it might be a good idea to add a check that the slice is actually force data. Not exactly sure of the best way to go about this, but just something to think about

If you do that, you get this:
`RuntimeError: Slice does not contain any calibration items. Is this an unprocessed force channel?`

So that case should already be covered?

rpauszek:
> > one thing I realized after reviewing, if it's feasible it might be a good idea to add a check that the slice is actually force data. Not exactly sure of the best way to go about this, but just something to think about
> 
> If you do that, you get this: `RuntimeError: Slice does not contain any calibration items. Is this an unprocessed force channel?`
> 
> So that case should already be covered?

ok yeah, seems ok then

PR #705, opened 2024-11-12
calibration: make sure offset is accessible

JoepVanlier:
**Why this PR?**
The key used to fetch the offset was incorrect and missed in testing (likely an oversight).

Note that this hasn't been in any release yet, so no changelog required.


tobiasjj (approved):
LGTM! 🚴

PR #708, opened 2024-11-22
tracks: expose `load_tracks` to public API

JoepVanlier:
**Why this PR?**
Users would like a way to import tracks they tracked in `pylake` and `lakeview` back into a notebook for further processing.

We already had the private API `import_kymotrackgroup_from_csv`. I have changed the error handling slightly to be more informative, and added a small method on the widget so that you can also load tracks there.

I think `load_tracks` would be a more discoverable name than `import_kymotrackgroup_from_csv`, but I'm open to other suggestions.


lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -195,22 +199,28 @@ def _check_summing_mismatch(track, sampling_width):
...
+        the kymograph you supply to this function in the same way.
+
     Parameters
     ----------
     filename : str | os.PathLike
         filename to import from.
     kymo : Kymo
-        kymograph instance that the CSV data was tracked from.
+        Kymograph instance that the CSV data was tracked from. This kymograph has to be the same
+        as the one used to create the tracks.
```
tiagobonetti, line 223: I think `instance` might not be the best word here since more literal-minded users might interpret it as having to be the same object (think identity, not equality). I think the new phrase is clearer and mostly obsoletes the previous one. I would also mention stuff like ROI also matters and refer users to read the `note` above.

lumicks/pylake/nb_widgets/kymotracker_widgets.py
```diff
@@ -361,14 +361,30 @@ def save_tracks(self, filename, delimiter=";", sampling_width=None):
...
+        except (RuntimeError, ValueError, IOError) as exception:
             self._set_label("status", str(exception))
 
+    def load_tracks(self, filename, delimiter=";"):
+        """Loads tracks from a csv file into the widget
+
+        .. note::
+
+            If you cropped the kymograph prior to tracking, you will have to make sure that you crop
+            the kymograph you supply to this function in the same way.
```
tiagobonetti, line 376: I also mention in this docs that the restrictions are the same from `load_tracks` and recommend users to read that if they want more details.

tiagobonetti (approved):
Looks pretty good to me.
I added some suggestions but nothing critical.

`honk`

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -195,22 +199,28 @@ def _check_summing_mismatch(track, sampling_width):
             else:
                 return RuntimeWarning(wrong_kymo_warning)
     except IndexError:
-        return RuntimeWarning(wrong_kymo_warning)
+        raise ValueError(wrong_kymo_warning)
```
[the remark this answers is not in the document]
JoepVanlier: Yeah, it's a good point. You do have more information in the case of the exception (that's why the second one is an exception and the first just a warning).

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -195,22 +199,28 @@ def _check_summing_mismatch(track, sampling_width):
...
+    """Loads a :class:`~lumicks.pylake.kymotracker.kymotrack.KymoTrackGroup` from a csv file.
 
     The file format contains a series of columns as follows:
     track index, time (pixels), coordinate (pixels), time (optional), coordinate (optional),
     sampled_counts (optional), minimum length
 
+    .. note::
+
+        If you cropped the kymograph prior to tracking, you will have to make sure that you crop
+        the kymograph you supply to this function in the same way.
```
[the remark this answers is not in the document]
JoepVanlier: It is indeed. The list is not exhaustive. I will phrase it as processing and make clear that the list may not be exhaustive (because we may forget to update the docstring when additional processing is added).

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Good point! I rephrased it to:

```
Kymograph that the CSV data was tracked from. This kymograph has to be the one used to
create the tracks (this includes any processing). See the note above for more information.
```

lumicks/pylake/nb_widgets/kymotracker_widgets.py
JoepVanlier: Added!

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -167,50 +171,61 @@ def store_column(column_title, format_string, new_data):
...
+                "used summing window was very small, there may be a bias in the counts."
+                "To recreate these counts without bias invoke:"
+                f"`track.sample_from_image({(sampling_width - 1) // 2}, correct_origin=True)`"
+            )
+        else:
+            return RuntimeWarning(
+                "Photon counts do not match the photon counts found in the file. It is "
+                "possible that the loaded kymo or channel doesn't match the one used to "
+                "create this file."
+            )
```
aafkevandenberg, line 201: I see you already mention the processing in other comments. Also here it applies that another way to get a mismatch of kymo and tracks is when tracking was performed using `crop_by_distance`, but when loading the full kymo is given. Then the kymo matches the file but is not processed the same way.

lumicks/pylake/kymotracker/kymotrack.py
```diff
@@ -223,16 +238,32 @@ def import_kymotrackgroup_from_csv(filename, kymo, channel, delimiter=";"):
...
+
+        import lumicks.pylake as lk
+
+        file = lk.File("test_data/kymo.h5")
+        kymo = file.kymos["16"]  # Extract the kymo named 16
+
+        tracks = lk.track_greedy(kymo, channel="red", pixel_threshold=5)
+        tracks.save("tracks.csv")
+
+        loaded_tracks = lk.load_tracks("tracks.csv", kymo, "red")
```
aafkevandenberg, line 256: Maybe nice to add, as this would be the first thing I'd do when trying this function for the first time:

```
# Plot kymo and overlay imported tracks for comparison:
plt.figure()
kymo1.crop_by_distance(4.9,13.8).plot()
tracks.plot()
```

aafkevandenberg (commented):
The functions is very logical to use.
The pre-processing of the kymo will be something important to make clear to the user, but you already added this in various places.

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Yeah, that's a good point. Considering we're tracking red, what do you think about plotting the red channel with the red tracks?

Something like this as example then:
```python
        import lumicks.pylake as lk

        file = lk.File("test_data/kymo.h5")
        kymo = file.kymos["16"]  # Extract the kymo named 16

        kymo_cropped = kymo.crop_by_distance(10, 25)
        tracks = lk.track_greedy(kymo_cropped, channel="red", pixel_threshold=5)
        tracks.save("tracks.csv")

        loaded_tracks = lk.load_tracks("tracks.csv", kymo_cropped, "red")

        # Plot the red channel of the kymograph and tracks together
        kymo_cropped.plot("red", adjustment=lk.ColorAdjustment(5, 95, "percentile"))
        loaded_tracks.plot()
```

lumicks/pylake/kymotracker/kymotrack.py
JoepVanlier: Ah yes, good point! Probably the most important place to mention it even as this might be the first thing users end up getting if they forget to do it. I changed it to:

```
                "Photon counts do not match the photon counts found in the file. It is "
                "possible that the loaded kymo or channel doesn't match the one used to "
                "create this file. Note that if you processed (e.g. sliced, cropped, flipped or "
                "downsampled) the kymograph prior to tracking, you will have to make sure that you "
                "crop the kymograph you supply to this function in the same way."
```

for the warning. The exception is the same except for one detail, namely that flipping cannot cause that indexing error and is omitted.

docs/tutorial/kymotracking.rst
```diff
@@ -484,6 +484,15 @@ by passing a width in pixels to sum counts over::
...
     tracks.save("tracks_signal.csv", sampling_width=3, correct_origin=True)
 
+We can re-import these tracks into Pylake at a later time using the :func:`~lumicks.pylake.load_tracks()` function::
+
+    tracks = lk.load_tracks("tracks.csv", kymo40, "green")
+
+.. note::
+
+    The `kymo` and `channel` arguments provided must be the same as the ones used to generate the tracks.
+    This includes any processing (e.g. cropping, downsampling, flipping) that was performed on the kymograph.
```
rpauszek, line 494: I think Tiago pointed this out before, but I think we can be explicit that this is not an exhaustive list
```suggestion
    This includes any processing (e.g. cropping, downsampling, flipping, etc.) that was performed on the kymograph.
```

rpauszek (approved):
nice to finally have this public, and I very much appreciate the shorter function name 😃

PR #712, opened 2024-12-09
examples: plot tracks in binding lifetime example

aafkevandenberg:
I added a plot of the tracks, to make more clear what we are actually analyzing in the rest of the example.

### DIFF
--- a/docs/examples/binding_lifetime/binding_lifetime.rst
+++ b/docs/examples/binding_lifetime/binding_lifetime.rst
@@ -91,6 +91,18 @@ Load the tracks into Pylake::
 
 Use the same approach as above to load the tracks from Lakeview, except that the part :func:`Kymo.crop_by_distance() <lumicks.pylake.kymo.Kymo.crop_by_distance>` has to be removed.
 
+The (loaded) tracks can be plotted on top of the original kymograph to visualize the result of the tracking::
+
+    plt.figure()
+    kymo1_selection.plot("g", aspect=5, adjustment=lk.ColorAdjustment(0, 5))
+    tracks1.plot()
+
+.. image:: tracks1.png
+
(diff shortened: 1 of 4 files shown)


docs/examples/binding_lifetime/binding_lifetime.rst
```diff
@@ -91,6 +91,18 @@ Load the tracks into Pylake::
 
 Use the same approach as above to load the tracks from Lakeview, except that the part :func:`Kymo.crop_by_distance() <lumicks.pylake.kymo.Kymo.crop_by_distance>` has to be removed.
 
+The (loaded) tracks can be plotted on top of the original kymograph to visualize the result of the tracking::
+
+    plt.figure()
+    kymo1_selection.plot("g", aspect = 5, adjustment=lk.ColorAdjustment([0], [5]))
```
JoepVanlier, line 97: Minor, but you don't need to make the adjustments lists if you specify only a single value.

```suggestion
    kymo1_selection.plot("g", aspect=5, adjustment=lk.ColorAdjustment(0, 5))
```

docs/examples/binding_lifetime/binding_lifetime.rst
```diff
@@ -91,6 +91,18 @@ Load the tracks into Pylake::
...
+    tracks1.plot()
+
+.. image:: tracks1.png
```
JoepVanlier, line 100: Minor, but if you save the image with `plt.savefig("tracks1.png", bbox_inches="tight")`, you'll get a figure with less whitespace. Might look slightly better in the rendered docs.

JoepVanlier (approved):
Looks good to me! Two small suggestions, but nothing blocking.

docs/examples/binding_lifetime/binding_lifetime.rst
aafkevandenberg: Indeed, that looks much better. I'll also update the two images with kymographs, as they have a lot of white space.

PR #714, opened 2024-12-16
release procedure fixes

JoepVanlier:
**Why this PR?**
I had hoped this would be the release PR already, but it turns out we need a few more small fixes before we can release.

1. It seems we haven't run the notebooks in a while, and some of the figures/numbers went out of date. The figures in the theory section were updated since they didn't take into account this [PR](https://github.com/lumicks/pylake/pull/613/files) yet. Some power spectral plots were updated, since they didn't take into account that we explicitly show the cuts now. Finally, [when it was first introduced](https://github.com/lumicks/pylake/pull/541/files#diff-2c94bbb8cdc4e7320db76b6434fff877418be37b52edc9c9b14a09fd86d7c5e7), the discrete dwell time plot did not enforce specifying a minimum observation time. Now it does, (as it should) by default, but the plot was never updated! The other things are tiny fixups that hopefully speak for themselves.

2. One of the figures in the binding lifetime notebook was incorrect (namely the one for the double exponential fit).

3. I noticed that our example notebook shows a bunch of warnings when doing the bootstrap. These came from 0/0s in the numerical gradient. We can rephrase this gradient to avoid this problem.

4. I noticed that in some pathological cases, the numerical derivative of the Jacobian that we calculate for the standard deviation could fail. Reason being that during numerical differentiation we could exceed 1 for an amplitude meaning that the sum constraint is by definition violated. I have also loosened the tolerances on the tests for clearly overparametrized models (as their uncertainties are not as well defined and depend on numerics). I deliberately added such a case to the tests.

Finally, I have added a what's new entry to highlight some of the nice examples @aafkevandenberg has been contributing!

Highly recommend going commit by commit. Plan is to squash commit 2 into 3 (right now it is separate in case you wish to test that the test indeed hits that corner case).

Note tests are failing because of brittle test (fix in [718](https://github.com/lumicks/pylake/pull/718)).

Docs build [here](https://lumicks-pylake.readthedocs.io/en/final_prep/whatsnew/1.6.0/1_6_0.html).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -33,6 +33,7 @@
 * Fixed a bug where bead edge determination could fail with an unhandled exception during background estimation. This raised a `np.linalg.LinAlgError` when determining the background failed rather than the expected `RuntimeError`. In this case, a simple median is used as a fallback option.
 * Fix a bug to ensure that [`lk.GaussianMixtureModel`](https://lumicks-pylake.readthedocs.io/en/latest/_api/lumicks.pylake.GaussianMixtureModel.html) can also be used with a single state.
 * Fixed bug that prevented opening the force distance widgets when using them with the `widget` backend on `matplotlib >= 3.9.0`.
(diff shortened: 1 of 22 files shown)


rpauszek (approved):
looks good, nice cleanup for release

rpauszek (approved):

PR #717, opened 2025-01-02
kymograph tutorial: add section on photon counts

aafkevandenberg:
I frequently get questions about this workflow, so I added an example.
I decided to add the example in the 'Tutorial', but it would also fit under 'Examples'
On the one hand, there is no new syntax introduced, which makes it fit under 'Examples'.
On the other hand, the example is just very short and does not represent a full workflow, which makes it fit in the 'Tutorial'. Let me know what you think.

### DIFF
--- a/docs/tutorial/kymographs.rst
+++ b/docs/tutorial/kymographs.rst
@@ -140,6 +140,50 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
 
 .. image:: figures/kymographs/kymo_flipped.png
 
+Extracting photon counts
+------------------------
+
+Photon counts can be extracted from the raw image, for example for stoichiometry analysis.
+First, select a region for which the photon counts have to be determined::
+
+    kymo_selection = kymo["88.5s":"93s"].crop_by_distance(21.4, 22.3)
+
+    plt.figure()
+    kymo_selection.plot("rgb", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+
+Get the raw data (photon counts for each pixel) for the selected region::
+
+    selection_green = kymo_selection.get_image(channel="green")
+
+.. image:: figures/kymographs/kymograph_selection.png
+
+If the background has to be subtracted, choose a region without binding events as background::
+
+    background =  kymo["92.1s":"100s"].crop_by_distance(21.4, 22.3)
+    
+    plt.figure()
+    background.plot("rgb", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+
+.. image:: figures/kymographs/background.png
+
+Sum over the vertical axis (along the width of the binding event) to obtain the total photon counts over the selected regions.
+Since every pixel has an average background signal of `background_mean`, we need to sum the background over the same region to obtain the total background::
+
+    summed_photon_counts = np.sum(selection_green, axis=0)
+    background_mean = np.mean(background.get_image(channel="green"))
+    summed_background = background_mean * selection_green.shape[0]
+    time = np.arange(len(summed_photon_counts)) * kymo.line_time_seconds
+
+    plt.figure()
+    plt.plot(time, summed_photon_counts, label = "Binding event")
+    plt.axhline(summed_background, color='k', linestyle='--', label = "background")
+    plt.xlabel("time (s)")
+    plt.ylabel("Photon counts")
+    plt.title("Photon counts along binding event")
+    plt.legend()
+
+.. image:: figures/kymographs/photon_counts.png
+
 Calibrating to base pairs
 -------------------------
 
--- a/docs/tutorial/figures/kymographs/background.png
+++ b/docs/tutorial/figures/kymographs/background.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:e24221d1969438ac81f50ae67363399811f05b71da4eacd4ad308f1186f856f1
+size 11316
--- a/docs/tutorial/figures/kymographs/kymograph_selection.png
+++ b/docs/tutorial/figures/kymographs/kymograph_selection.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:9f56509037b53fa3aa5b5ba390cc0eda3921209013f57854ce6fcc6c4937b660
+size 13137
--- a/docs/tutorial/figures/kymographs/photon_counts.png
+++ b/docs/tutorial/figures/kymographs/photon_counts.png
@@ -1,0 +1,3 @@
+version https://git-lfs.github.com/spec/v1
+oid sha256:979e81baee4f80b54e50e9993a38ed19ab16009c3a2c6bf16f63fd0f43346abb
+size 45412


docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,45 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
...
+If the background has to be subtracted, choose a region without binding events as background::
+
+    background =  kymo["92.1s":"100s"].crop_by_distance(21.4,22.3)
+    
+    plt.figure()
+    background.plot("rgb", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+
+.. image:: figures/kymographs/background.png
+
+Subtract the background and sum over the vertical axis (along the width of the binding event) to obtain the photon counts over time::
```
JoepVanlier, line 168:
```suggestion
Subtract the background and sum over the vertical axis (along the width of the binding event) to obtain a signal intensity over time::
```

While I think it's fine for plotting, depending on how many counts you have, subtracting the estimated offset can make subsequent analysis harder in some cases.

docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,45 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
 
 .. image:: figures/kymographs/kymo_flipped.png
 
+Extracting photon counts
+------------------------
+Photon counts can be extracted from the raw image, for example for stoichiometry analysis.
```
JoepVanlier, line 145:
```suggestion
Extracting photon counts
------------------------

Photon counts can be extracted from the raw image, for example for stoichiometry analysis.
```

JoepVanlier (changes requested):
> I frequently get questions about this workflow, so I added an example.

Nice! What questions do you typically get? I'm wondering whether people ask about this mostly for plotting or analyzing the counts further.

The reason I'm asking is because subtracting the offset changes the distribution. Especially when counts are low, it can make downstream analysis messier than if one were to keep the counts as is, but include a constant background count in the model used to analyze the data.

aafkevandenberg:
> > I frequently get questions about this workflow, so I added an example.
> 
> Nice! What questions do you typically get? I'm wondering whether people ask about this mostly for plotting or analyzing the counts further.
> 
> The reason I'm asking is because subtracting the offset changes the distribution. Especially when counts are low, it can make downstream analysis messier than if one were to keep the counts as is, but include a constant background count in the model used to analyze the data.

Hi Joep,

I often get request such as 'I would like to plot the photon intensity of a specific region and subtract background'. A possible application could be 

- determine the stoichiometry (ie look at step sizes of the signal intensity)
- Sum the intensity over a certain region of DNA and observe how the summed intensity decreases over time, due to bleaching or unbinding.

Could you give a specific example of when it is better to keep the background as a separate parameter?

JoepVanlier:
> I often get request such as 'I would like to plot the photon intensity of a specific region and subtract background'. A possible application could be

For this what you have in this PR is fine.

> determine the stoichiometry (ie look at step sizes in the signal intensity)

Given that counts are high and the steps are very clear, this should also be fine.

> Sum the intensity over a certain region of DNA and observe how the summed intensity decreases over time, due to bleaching or unbinding. For example, because users are interested in facilitated dissociation, or in the exchange rate of molecules in a biocondensate.

> Often users would like to translate the signal intensity to the number of dyes bound to DNA, and therefore subtract the background. Could you give a specific example of when it is better to keep the background as a separate parameter?

It depends. When Poisson-noise dominates, the distributions are very tailed and you may find that subtracting offsets doesn't allow you to use a Poisson model for fitting anymore. Reason being that the count model depends on the total intensity (and that includes the background). In that case, it might be better to include the background (you could still pre-estimate it on a different section of data and make it a fixed parameter) in the model and fit the raw data as is.

> Keeping the background separate is anyway a good idea. I have rewritten the instructions, such that `photon_counts` and `background` remain separate parameters.

👍

docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,48 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
...
+
+    background =  kymo["92.1s":"100s"].crop_by_distance(21.4, 22.3)
+    
+    plt.figure()
+    background.plot("rgb", adjustment=lk.ColorAdjustment(0, 98, mode="percentile"))
+
+.. image:: figures/kymographs/background.png
+
+Sum over the vertical axis (along the width of the binding event) to obtain the total photon counts over the selected regions.
+Since every pixel has an average background signal of `background_mean`, we need to sum the background over the same region to obtain the total background.
```
JoepVanlier, line 170:
```suggestion
Since every pixel has an average background signal of `background_mean`, we need to sum the background over the same region to obtain the total background::
```

docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,48 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
...
+
+    summed_photon_counts = np.sum(selection_green, axis=0)
+    background_mean = np.mean(background.get_image(channel="green"))
+    summed_background = background_mean * selection_green.shape[0]
+    time = np.arange(len(photon_counts)) * kymo.line_time_seconds
```
JoepVanlier, line 175:
```suggestion
    time = np.arange(len(summed_photon_counts)) * kymo.line_time_seconds
```

docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,48 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
...
+
+    plt.figure()
+    plt.plot(time, summed_photon_counts - summed_background)
```
JoepVanlier, line 178: I don't feel super strongly about this, but another option would be to indicate the background level and leave the data as-is.

```suggestion
    plt.plot(time, summed_photon_counts)
    plt.axhline(summed_background, color='k', linestyle='--')
```

Then you could still call them counts. Anyways, your call on that one.

It would be nice if we had some data with clear bleaching steps and just make a fully worked example for stoichiometry analysis.

JoepVanlier (changes requested):
Found a few small errors that prevent the notebook from being rendered/run correctly.

docs/tutorial/kymographs.rst
```diff
@@ -140,6 +140,49 @@ This returns a new (but flipped) :class:`~lumicks.pylake.kymo.Kymo`::
...
+    background_mean = np.mean(background.get_image(channel="green"))
+    summed_background = background_mean * selection_green.shape[0]
+    time = np.arange(len(summed_photon_counts)) * kymo.line_time_seconds
+
+    plt.figure()
+    plt.plot(time, summed_photon_counts, label = "Binding event")
+    plt.axhline(summed_background, color='k', linestyle='--', label = "background")
+    plt.xlabel("time (s)")
+    plt.ylabel("Photon counts")
+    plt.title("Photon counts along binding event")
```
JoepVanlier, line 182: Extremely minor, but you have labels specified in the `plt.plot()` calls, but don't actually show them by invoking `plt.legend()`.

JoepVanlier (approved):
LGTM! :shipit:

docs/tutorial/kymographs.rst
aafkevandenberg: good catch! I updated it.

PR #718, opened 2025-01-07
touchdown: fix brittle test

JoepVanlier:
**Why this PR?**
This test is brittle. In some cases, it ends up with no datapoints due to being right on the edge when masking.

Better to have specific tests for both cases (and gracefully handle the no-data case as well). No changelog entry since it is private API.


mikhas:
You're saying the test already fails, but for the wrong reasons? Not sure how the new data points avoid that.

JoepVanlier:
> You're saying the test already fails, but for the wrong reasons? Not sure how the new data points avoid that.

The test currently fails while it shouldn't. What happens now is that we end up with an empty array internally. Initially this case was entirely unhandled so we get an exception from inside scipy.

The reason is that we end up with an empty array is because we generated data right on the edge of what gets included.

Now both cases are tested explicitly (insufficient data to do anything and insufficient data to do it reliably) and for the latter case we generate a few extra data points.

lumicks/pylake/force_calibration/tests/test_touchdown.py
```diff
@@ -148,14 +148,20 @@ def test_exp_sine_fits(decay, amplitude, frequency, phase_shift):
 @pytest.mark.filterwarnings("ignore:Covariance of the parameters could not be estimated")
 @pytest.mark.filterwarnings("ignore:Surface detection failed")
 def test_insufficient_data(mack_parameters):
-    stage_positions, simulation = simulate_touchdown(102.5, 103.5, 0.01, mack_parameters)
+    stage_positions, simulation = simulate_touchdown(102.5, 104, 0.01, mack_parameters)
```
mikhas, line 151: Is it really called `mack_parameters`?

JoepVanlier: [Yes](https://pmc.ncbi.nlm.nih.gov/articles/PMC3482253/). :D

mikhas: Gotcha, thanks for the `vanlier_comment`.

PR #722, opened 2025-01-10
ci: verify before uploading, unpin setuptools

JoepVanlier:
**Why this PR**
- The strict `twine` check verifies the build before it goes out. Previously, I missed that the `README.md` wouldn't be picked up on a case-sensitive OS such as the one we use to deploy.
- We should also unpin `setuptools` to not stick with the deprecated source distribution name (`lumicks.pylake` -> `lumicks_pylake`).

### DIFF
--- a/.github/actions/check-python-packaging/action.yml
+++ b/.github/actions/check-python-packaging/action.yml
@@ -1,0 +1,22 @@
+name: "Check Python Packaging"
+description: ""
+runs:
+  using: "composite"
+  steps:
+    - name: Install packaging dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install "setuptools>=69.3" build wheel twine
+      shell: bash
+    - name: Package pylake
+      run: |
+        python -m build
+        python -m twine check --strict dist/*
+      shell: bash
+    - name: Unpack the generated package
+      run: tar -xvzf dist/lumicks_pylake*.tar.gz
+      shell: bash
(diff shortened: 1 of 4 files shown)


rpauszek (approved):

pyproject.toml
```diff
@@ -1,5 +1,5 @@
 [build-system]
-requires = ["setuptools<69.3", "wheel"]
+requires = ["setuptools", "wheel"]
```
alessiamarcolini, line 2: it would probably be wise to set a lower bound then, since they introduce a backward incompatible change

and related: does pylake not have a CI workflow to check if the project is packaged correctly (before attempting a release)?

alessiamarcolini (approved):
looks good! just remember to change the wording in the first commit message (we are not unpinning anymore)

.github/actions/check-python-packaging/action.yml
```diff
@@ -0,0 +1,22 @@
+name: "Check Python Packaging"
+description: ""
+runs:
+  using: "composite"
+  steps:
+    - name: Install packaging dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install setuptools build wheel twine
```
alessiamarcolini, line 9: we probably need to set the lower bound of setuptools here as well

JoepVanlier: good catch!

PR #724, opened 2025-01-23
calibration: add APIs for accessing raw data from calibration item

JoepVanlier:
**Why this PR?**
It's been a longstanding request to store the raw calibration data with the calibration items. This exposes APIs to access this data with `pylake`.

You can get the voltage data using:
```python
f = lk.File("test_data/raw_data_in_item.h5")
voltage_slice = f.force1x.calibration[0].voltage
voltage_slice.plot()
```

And the driving data using:
```python
f = lk.File("test_data/raw_data_in_item.h5")
driving_data = f.force1x.calibration[0].driving
driving_data.plot()
```

I also added a few convenience APIs which make calibration with different parameters pretty trivial now.

```python
f = lk.File("test_data/raw_data_in_item.h5")
calibration = f.force1x.calibration[0]
recalibrated = calibration.recalibrate_with(hydrodynamically_correct=True)
recalibrated_force1x = f.force1x.recalibrate_force(recalibrated)
```
Note that recalibrate force is introduced in https://github.com/lumicks/pylake/pull/704, so I would suggest reviewing that one first.

You can also just plot it now:
```python
f = lk.File("test_data/raw_data_in_item.h5")
f.force1x.calibration[0].plot()
```

Contrast that with what you used to have to do:
```python
f = lk.File("test_data/passive_calibration.h5")
force1x_slice = f.force1x[f.force1x.calibration[1]]
old_calibration = force1x_slice.calibration[0]
volts1x_slice = force1x_slice / old_calibration.force_sensitivity
calibration_params = f.force1x.calibration[1].calibration_params()
calibration_params["hydrodynamically_correct"] = True
recalibrated_hyco = lk.calibrate_force(volts1x_slice.data, **calibration_params)
correction_factor = recalibrated_hyco.force_sensitivity / old_calibration.force_sensitivity
recalibrated_force1x = force1x_slice * correction_factor
```

I think it should also not be too hard to implement a similar API improvement for calibrations where the raw data is not in the item, but is still present in the file, but I will leave that for a future PR.


docs/tutorial/force_calibration/calibration_items.rst
```diff
@@ -69,11 +69,78 @@ These parameters are properties and can be extracted as such::
...
+Note that any of the calibration parameters can easily be changed this way.
+To investigate the effect of the hydrodynamically correct model for example, we can try turning it off::
+
+    recalibrated_no_hyco = calibration.recalibrate_with(hydrodynamically_correct=False)
+    recalibrated_no_hyco.plot()
+
+.. image:: figures/in_item_no_hyco.png
+
+This clearly fits the data poorly.
+To see what it would do to our timeline data, we can simply recalibrate that by applying the calibration item to :meth:`~lumicks.pylake.channel.Slice.recalibrate_force`::
```
rpauszek, line 119:
```suggestion
To see what it would do to our timeline data, we can simply recalibrate that by applying the calibration item to
:meth:`~lumicks.pylake.channel.Slice.recalibrate_force`::
```

lumicks/pylake/force_calibration/calibration_item.py
```diff
@@ -3,13 +3,43 @@
...
+    ):
+        super().__init__(dictionary)
+        self._voltage = voltage
+        self._sum_voltage = sum_voltage
+        self._driving = driving
+
+    @property
+    def voltage(self):
+        """Uncalibrated voltage reading on the detector"""
+        return self._voltage if self._voltage else empty_slice
```
rpauszek, line 31: generally  for things like this I test explicitly `if x is not None`. I don't expect it to happen but if somehow `self._voltage` becomes something falsey it might be harder to track down because you'll get an empty slice which might look ok. Again, I don't know if this'll be a problem, just my personal preference

lumicks/pylake/force_calibration/calibration_item.py
```diff
@@ -130,6 +160,43 @@ def __repr__(self):
         )
         return f"{self.__class__.__name__}({properties})"
 
+    def _check_has_data(self):
+        if not self.voltage:
```
rpauszek, line 164: again here might be safer to do `if self.voltage is None`

lumicks/pylake/force_calibration/calibration_item.py
```diff
@@ -130,6 +160,43 @@ def __repr__(self):
...
+        This diagnostic plot can be used to determine how well the spectrum fits the data. While
+        it cannot be used to diagnose over-fitting (being unable to reliably estimate parameters
+        due to insufficient information in the data), it can be used to diagnose under-fitting (the
+        model not fitting the data adequately).
+
+        In an ideal situation, the residual plot should show a noise band around 1 without any
+        systematic deviations.
+        """
+        self._check_has_data()
+        self.recalibrate_with().plot_spectrum_residual()
```
rpauszek, line 187: why do you need an empty recalibrate here?

rpauszek (approved):
LGTM, just some nit pick suggestions

lumicks/pylake/force_calibration/calibration_item.py
JoepVanlier: Yeah, I'm aware of the distinction. In this particular case, I intended falsey. We would get either `None` or a Slice with no data (an empty slice).

I'll make it a little more explicit and check upstream for the emptiness and then only check `None` case here. I agree that that is probably a little safer.

lumicks/pylake/force_calibration/calibration_item.py
JoepVanlier: Because I need to do the calibration. There is no calibration stored, only raw data, so I need to actually perform the calibration with all the raw data in order to show the spectra and fit.

lumicks/pylake/force_calibration/calibration_item.py
JoepVanlier: Here the intent was falsey. By default construction, this would be an `empty_slice`. If you think it is clearer, I can change it to `not len(voltage)`.

PR #728, opened 2025-01-28
Bugfix: `load_tracks` doesn't handle basepairs well

rpauszek:
**Why this PR?**
Noticed this while working on #726, `load_tracks` uses `pixelsize_um` always when loading tracks from a CSV if photon counts are present.  If there are no photon counts, it works fine. 

A better solution to this and more robust test is included in #726 but it's good to have this in the next patch release


JoepVanlier (approved):
Good catch!

PR #732, opened 2025-01-30
calibrate tether upside down

rpauszek:
**Why this PR?**
Sometimes you want to calibrate your tether with a known orientation but without flipping the kymo. Followup to #726

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -334,7 +334,7 @@ class Kymo(ConfocalImage):
     def pixelsize(self):
         """Returns a `List` of axes dimensions in calibrated units. The length of the
         list corresponds to the number of scan axes."""
-        return [self._calibration.value]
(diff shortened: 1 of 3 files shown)


lumicks/pylake/kymo.py
```diff
@@ -1123,7 +1122,7 @@ def label(self):
 @dataclass(frozen=True)
 class PositionCalibration:
     unit: PositionUnit = PositionUnit.pixel
-    value: float = 1.0
+    factor: float = 1.0
```
JoepVanlier, line 1125: I have a mild preference for `scale` given that one often talks about `scale` and `offset` in the context of transforms, but I don't feel very strongly about it.

JoepVanlier (approved):
LGTM. Small optional suggestion regarding naming.

Other than that, don't forget to squash the fixup.

PR #733, opened 2025-01-30
cleanup position unit

rpauszek:
**Why this PR?**
Now with `PositionUnit` I thought it would be nice to cleanup and centralize everything having to do with position units for kymos, tracking and msd estimation. I think especially useful is the string formatting for diffusion constants handled as a method on the enum so we can guarantee consistency.

### DIFF
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -1107,6 +1107,7 @@ class PositionUnit(Enum):
     um = UnitInfo(name="um", label=r"μm")
     kbp = UnitInfo(name="kbp", label="kbp")
     pixel = UnitInfo(name="pixel", label="pixels")
+    au = UnitInfo(name="au", label="au")
 
     def __str__(self):
         return self.value.name
@@ -1117,6 +1118,18 @@ class PositionUnit(Enum):
     @property
     def label(self):
         return self.value.label
+
+    def as_diffusion(self):
+        return {
+            "unit": f"{self}^2 / s",
+            "_unit_label": f"{self.label}²/s",
+        }
+
+    def as_squared(self):
+        return {
+            "unit": f"{self}^2",
+            "_unit_label": f"{self.label}²",
+        }
 
 
 @dataclass(frozen=True)
--- a/lumicks/pylake/kymo.py
+++ b/lumicks/pylake/kymo.py
@@ -382,7 +382,7 @@ class Kymo(ConfocalImage):
             axes.set_axis_off()
 
         if scale_bar and not image_handle:
-            scale_bar._attach_scale_bar(axes, 60.0, 1.0, "s", self._calibration.unit_label)
+            scale_bar._attach_scale_bar(axes, 60.0, 1.0, "s", self._calibration.unit.label)
 
         image = self._get_plot_data(channel, adjustment)
 
@@ -410,7 +410,7 @@ class Kymo(ConfocalImage):
             **{**default_kwargs, **kwargs},
         )
         axes.set_xlabel("time (s)")
-        axes.set_ylabel(f"position ({self._calibration.unit_label})")
+        axes.set_ylabel(f"position ({self._calibration.unit.label})")
         if show_title:
             axes.set_title(self.name)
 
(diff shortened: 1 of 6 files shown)


lumicks/pylake/kymo.py
```diff
@@ -1118,6 +1119,18 @@ def __hash__(self):
...
+        return {
+            "unit": f"{self}^2 / s",
+            "_unit_label": f"{self.label}²/s",
+        }
+
+    def as_squared(self):
+        return {
+            "unit": f"{self}^2",
+            "_unit_label": f"{self.label}²",
+        }
```
JoepVanlier, line 1132: I feel a bit conflicted about these helper functions. I do like that it is in one place now. It does feel a little odd to put this in the `PositionUnit` itself. For now this is probably OK and an improvement from what we had before, but if we get more of these in the future, we might want to consider making a unit formatter or something like that, to separate the logic and not pollute the unit class too much with implementations for rather specific situations. In this case, the output feels very specifically tailored to those dataclasses we have in the diffusion module.

A second thing is the naming. The `as_` would imply to me that I'd be getting a new representation of this unit, rather than labels for a different unit alltogether. I'd sooner call it `get_*_labels` or something along those lines to make that a little more clear.

I think it would also make sense to add a type annotation for the return type in this case (as it is not immediately clear to me from the name).

Sorry for all the nitpicking. 😅

JoepVanlier (changes requested):
Seems like an improvement over what we had. I like that it is in one place now, despite the fact that I feel a bit uncertain about the place where this functionality should live.

lumicks/pylake/kymo.py
JoepVanlier: I do wonder whether it isn't time to rope in an actual unit library to track units for us in these cases so that we also get composite units. Considering that `_unit_label`_is_ private in all the dataclass input args, one could consider changing these. You could have them take a `UnitInfo | String` directly for the unit argument and unpack to add the `unit` as string and the `_unit_label` in `__post_init__`. If we ensure that it can take both strings and units that shouldn't be breaking. That said, that would make more sense if we actually had composite units, because as it stands right now, you'd still need getters of the type you implemented, so this is probably not something to do already now.

lumicks/pylake/kymotracker/detail/msd_estimation.py
```diff
@@ -1159,14 +1149,13 @@ def ensemble_ols(kymotracks, max_lag):
...
         value=slope * to_time,
         std_err=np.sqrt(var_slope / np.mean(ensemble_msd.effective_sample_size)) * to_time,
         num_lags=optimal_lags,
         num_points=sum(len(t) for t in kymotracks),
         localization_variance=intercept / 2.0,
         method="ensemble ols",
-        unit=f"{src_calibration.unit}^2 / s",
-        _unit_label=f"{src_calibration.unit_label}²/s",
+        unit=f"{kymotracks._calibration.unit}^2 / s",
+        _unit_label=f"{kymotracks._calibration.unit.label}²/s",
```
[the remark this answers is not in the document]
rpauszek: good catch!

lumicks/pylake/kymo.py
rpauszek: Yeah I agree with pretty much all of this and I had tried going down that route initially but it got pretty complicated pretty fast. I don't particularly like these small super-specific methods, but as far as I can see we only need something like this for the diffusion stuff at the moment so figured this way was the biggest payoff for the least amount of effort. I had thought about changing up the `DiffusionEstimate` dataclass too but you can't process constructor arguments with `__post_init__` for frozen dataclasses so that's a complication. I'm sure there's a way to get all of this working, but again for just this one instance so far it seemed wise to go with the path of least resistance.

I do really like the name suggestion though, so I did change that 👍

JoepVanlier: Makes sense yeah. We can always introduce a different abstraction later once we know a few more places where we'd need similar functionality.

JoepVanlier (approved):
Looks good to me.

PR #736, opened 2025-03-17
dwelltime: allow parallel bootstraps

JoepVanlier:
**Why this PR?**
Bootstrapping can become prohibitively expensive for large number of samples. This adds a small API which can be used to speed things up. Considering we may want >= 2 components in models we end up ranking this is low hanging fruit for speeding things up.

### DIFF
--- a/lumicks/pylake/population/tests/test_dwelltimes.py
+++ b/lumicks/pylake/population/tests/test_dwelltimes.py
@@ -158,6 +158,24 @@ def test_bootstrap_multi(min_obs, max_obs, ref_ci, time_step):
     np.testing.assert_allclose(ci, ref_ci, rtol=1e-5)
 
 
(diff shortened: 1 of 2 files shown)


lumicks/pylake/population/tests/test_dwelltimes.py
```diff
@@ -158,6 +158,24 @@ def test_bootstrap_multi(min_obs, max_obs, ref_ci, time_step):
     np.testing.assert_allclose(ci, ref_ci, rtol=1e-5)
 
 
+@pytest.mark.parametrize("num_components", [1, 2])
+def test_bootstrap_parallel(monkeypatch, exponential_data, num_components):
+    import multiprocessing.dummy
```
alessiamarcolini, line 163: TIL

alessiamarcolini (approved):

PR #737, opened 2025-03-21
export: add `bigtiff` option for exporting tiffs

JoepVanlier:
**Why this PR?**
BL by default splits TIF files into 4 GB chunks (since this is what the normal file format supports). Some people prefer to just have bigger files which requires using BigTIFF. This allows them to stitch together their TIFFs with Pylake. The error `tifffile` produces by default is also less than informative, so I decided to catch and rethrow that.

I decided to not change to `bigtiff` by default when the content is bigger than 4 GB because it should be a deliberate choice as the format is not supported by the same set of software.

Note that I did touch a few tests, because while doing this, I noticed they were not even running as their names were duplicates 😱. Thankfully no regressions there. I split these changes out in their own commit which I will squash (but keeping it separate for review makes the actual changes easier to see I hope).


lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -68,23 +80,34 @@ def extratags(timestamp_range):
...
+                        frame,
+                        contiguous=False,  # write tags on each page
+                        extratags=extratags(timestamp_range),
+                        metadata=None,  # suppress tifffile default ImageDescription tag
+                        description=json.dumps(metadata, indent=4),
+                        **self._tiff_writer_kwargs(),
+                    )
+        except Exception as e:
+            if "format requires 0 <= number <= 4294967295" in str(e):
+                raise ValueError(
```
tobiasjj, line 104: I'm wondering if it's a `ValueError` or if a `RuntimError` is better suited as the image to be exported is not passed as an argument the the `export_tiff` method.

tobiasjj (approved):
LGTM. Two minor comments. You may want to add a test that checks for the `ValueError` (`RuntimeError`). 🚴

lumicks/pylake/detail/imaging_mixins.py
JoepVanlier: Yeah, I was doubting the exact same thing. I think it depends on whether you consider the error to be "oh no, the tiff turns out to big", or whether you consider the error to be the fact that `bigtiff` is erroneously set to `False`.

tobiasjj: I like the explanation that the user did not explicitly ask for a `bigtiff`. So no, in this sense `RuntimeError` is not more appropriate, and I would leave `ValueError` as it is, as it also implicitly conveys the message that the user can change the behavior by specifying a different value for `bigtiff`.

lumicks/pylake/detail/imaging_mixins.py
```diff
@@ -68,23 +80,34 @@ def extratags(timestamp_range):
...
+                    tif.write(
+                        frame,
+                        contiguous=False,  # write tags on each page
+                        extratags=extratags(timestamp_range),
+                        metadata=None,  # suppress tifffile default ImageDescription tag
+                        description=json.dumps(metadata, indent=4),
+                        **self._tiff_writer_kwargs(),
+                    )
+        except Exception as e:
+            if "format requires 0 <= number <= 4294967295" in str(e):
```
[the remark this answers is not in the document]
JoepVanlier: Added!

lumicks/pylake/detail/imaging_mixins.py
JoepVanlier: Alright! Added a test, so we'd spot it if this upstream error ever changes! :)

PR #738, opened 2025-03-24
dwelltime: fix profile likelihood bounds

JoepVanlier:
**Why this PR?**
Currently, the confidence interval calculated by running profile likelihood on a dwell time model is overly pessimistic thanks to a missing factor 2. The profile likelihood code was written expecting chi squared values (-2 log L), and we gave it -log L.

To test whether this is correct, we can use 3 methods for inferring confidence intervals.
- Asymptotic intervals (relies on large N_datapoints), relies on the scale of the log likelihood.
- Bootstrapping (relies on large N_bootstrap, N_datapoints), but does not rely on the scale of the log likelihood.
- Profile likelihood (relies on medium to large N_datapoints), relies on the scale of the log likelihood.

image

If we make several realizations of large datasets (10000 dwells), we can see the three agree in the limit of lots of data (as they should).

Below a typical set of profiles / standard error based / bootstrap results:
image

Docs build of the example [here](https://lumicks-pylake.readthedocs.io/en/joep-dwell_fixes/examples/binding_lifetime/binding_lifetime.html).


aafkevandenberg (approved):
Thanks for fixing this bug and directly updating this Example!

I am wondering whether it makes sense to remove the Likelihood profile altogether from this Example, as it does not add new insights as compared to the Bootstrapping distributions.
We can also keep it and let users pick their favorite method. Going through this example did help me understand the different methods.

JoepVanlier:
> Thanks for fixing this bug and directly updating this Example!
> 
> I am wondering whether it makes sense to remove the Likelihood profile altogether from this Example, as it does not add new insights as compared to the Bootstrapping distributions. We can also keep it and let users pick their favorite method. Going through this example did help me understand the different methods.

Yes, you are right. In this case, they don't give different outcomes.

Just to summarize the discussion we just had. Probably a good idea to keep it here for now as it can be helpful to compare the methods as they each respond differently to violation of their assumptions and checking whether they give the same result is a reasonable indication that we aren't dealing with a pathological example where more data may be warranted.

Longer term, it may make sense to add something about differences between the methods in a theory section, but for now, to not block the bug-fix, we keep the example as is.

tobiasjj (approved):
The comparison between the three different methods is convincing in terms of correctly fixing the bug. I only have a nitpick for the changelog. 🚴

PR #739, opened 2025-04-09
ImageStack: allow `io.BytesIO` as input source

JoepVanlier:
**Why this PR?**
It is useful to be able to pass byte streams to `ImageStack`. Since we already implicitly support this, we can do this for relatively little extra work.


rjmoerland (approved):
I'm happy with the explicit support of streams, and the minor bug fix in `WrappedTiffFile`! I will nick some of your pytesting code for AutoFAT as well 😁

changelog.md
```diff
@@ -6,6 +6,7 @@
 
 * Fixed a bug where [`ForceCalibrationItem.force_sensitivity`](https://lumicks-pylake.readthedocs.io/en/v1.6.1/_api/lumicks.pylake.calibration.ForceCalibrationItem.html#lumicks.pylake.calibration.ForceCalibrationItem.force_sensitivity) would return the signed force response rather than the unsigned force sensitivity of a Bluelake calibration. This can cause issues with the force sign when using ratios of sensitivities obtained from Bluelake or Pylake. Now Pylake and Bluelake items both return the unsigned force sensitivity.
 * Fixed a bug where [`lk.load_tracks()`](https://lumicks-pylake.readthedocs.io/en/v1.6.1/_api/lumicks.pylake.load_tracks.html#lumicks.pylake.load_tracks) incorrectly loaded the position coordinates of tracks that had been calibrated to base pairs.
+* Fixed a bug where attempting to load a TIF file with insufficient metadata results in an unrelated `AttributeError` when deleting the `ImageStack`.
```
alessiamarcolini, line 9:
```suggestion
* Fixed a bug where attempting to load a TIF file with insufficient metadata resulted in an unrelated `AttributeError` when deleting the `ImageStack`.
```

lumicks/pylake/detail/widefield.py
```diff
@@ -154,8 +154,9 @@ class WrappedTiffFile:
     or TiffStack itself is because TiffFiles are shared between instances of these."""
 
     def __init__(self, file, *args, **kwargs):
+        self._src = None
         self._src = tifffile.TiffFile(file, *args, **kwargs)
```
alessiamarcolini, line 158: what is going on under the hood that we need to set `self._src` to None first?

rjmoerland: If you feed `WrappedTiffFile` a file name or stream that actually doesn't belong to a TIFF file, `tifffile.TiffFile` throws an exception and `self._src` wouldn't exist. Below on line 169 in `WrappedTIffFile.close()`, that causes an `AttributeError` exception. Therefore, a user who is ready to catch a `TiffFileError` exception still runs into an unhandled `AttributeError` exception. See https://github.com/lumicks/autofat/pull/1090

JoepVanlier: It's a good point though it looks like it wouldn't have a function at first glance, I'll add a little comment that mentions why it's there. Probably a good idea, since otherwise someone might remove it in the future.

lumicks/pylake/detail/utilities.py
```diff
@@ -225,3 +226,9 @@ def convert_to_scalar(value):
 
     if isinstance(value, numbers.Number):
         return value
+
+
+def to_stream(filename):
```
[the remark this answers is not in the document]
JoepVanlier: We usually don't explicitly mention fully private changes in the `changelog` as they are user facing. This is basically a small helper to not have to repeat this all over the tests, but it's in detail, so end users of Pylake shouldn't be directly using it.

lumicks/pylake/detail/widefield.py
tobiasjj: Could you also handle the `AttributeError`in `close()` when trying to access `self._src` instead? The reasoning would be that if `_src` is not set, there is no file to close.

JoepVanlier: Personally, I prefer direct initialization, as I feel it is more explicit. It also immediately warns you that it may be `None`, without having to look at `__del__` (I suspect it is more common to inspect the init than the delete dunder).

If the property `_src` changes, the cleanup code might silently start failing.

alessiamarcolini (approved):

PR #743, opened 2025-05-12
dwelltime: expose asymptotic standard errors

JoepVanlier:
**Why this PR?**
Having the option to quickly calculate asymptotic standard errors is convenient.


docs/tutorial/population_dynamics.rst
```diff
@@ -278,6 +278,43 @@ These information criteria values weigh the log likelihood against the model com
...
 exponential model, but only slightly so it is not strong evidence to choose the more complex model.
 
+.. _pop_std_err:
+
+Standard errors
+^^^^^^^^^^^^^^^
+
+Obtaining lifetime estimates is important, but it is equally important to know how reliable these estimates are.
+One way to gauge the reliability of the estimates is to calculate the asymptotic standard errors of the parameters.
+We can find these in the `err_amplitudes` and `err_lifetimes` properties.
```
aafkevandenberg, line 288: I don’t think you need to explain that knowing the reliability of an estimate is important. How about rephrasing the above 3 lines as:

After the optimal model for the data has been selected using the methods above, the reliability of the parameter estimates can be determined. Pylake provides several methods to determine both the confidence intervals and the standard errors of the parameters. 
For large datasets, the standard error under `err_amplitudes` and `err_lifetimes` can be used. Note that these are only asymptotically correct:

docs/tutorial/population_dynamics.rst
```diff
@@ -278,6 +278,43 @@ These information criteria values weigh the log likelihood against the model com
...
+    >>> dwell_1.err_lifetimes
+    array([0.00039664])
+
+    >>> dwell_2.err_amplitudes
+    array([0.03601527, 0.03601527])
+
+    >>> dwell_2.err_lifetimes
+    array([0.00072262, 0.00063035])
+
+When we have sufficient data, these approximate standard errors may be sufficient, but it should be noted that they are only asymptotically correct (for large amounts of data).
```
aafkevandenberg, line 299: I propose to mention the above code block, otherwise people will not read it.

docs/tutorial/population_dynamics.rst
```diff
@@ -278,6 +278,43 @@ These information criteria values weigh the log likelihood against the model com
...
+We can convert these asymptotic standard errors into confidence intervals (CI) using the :math:`\chi^2` distribution::
+
+    def to_ci(parameter_value, std_err, confidence_level):
+        import scipy
+
+        chi_squared = scipy.stats.chi2.ppf(confidence_level, 1)
+        d = np.sqrt(chi_squared * std_err ** 2)
+        return [parameter_value - d, parameter_value + d]
+
+For example, to get the confidence interval for dwell time `2`, we could invoke:
```
aafkevandenberg, line 311: ‘the 95% confidence interval’

aafkevandenberg (approved):
LGTM!

docs/tutorial/population_dynamics.rst
JoepVanlier: > I don’t think you need to explain that knowing the reliability of an estimate is important.

Haha, yes, fair point! 😅 

I will swap it out for your section!

docs/tutorial/population_dynamics.rst
JoepVanlier: Something like `When we have sufficient data, the approximate standard errors calculated in the code block above may be sufficient, but it should be noted that they are only asymptotically correct (for large amounts of data).`?

aafkevandenberg: Sorry, I mistyped by comment. What I wanted to say, is that you should mention this above the code block. Otherwise people will not read it. In the above comment I already included the assumption about sufficient data, let me know what you think.

JoepVanlier: Ohhh, but then it is indeed already covered by your previous suggestion. I will omit this sentence now then.

lumicks/pylake/population/dwelltime.py
```diff
@@ -221,11 +221,11 @@ def plot(self, alpha=None, **kwargs):
             Significance level. Confidence intervals are calculated as 100 * (1 - alpha)%. The
             default value of `None` results in using the significance level applied when
             profiling (default: 0.05).
+        with_stderr : bool
```
alessiamarcolini, line 224: nit 😬 
```suggestion
        with_stderr : bool, optional
```

JoepVanlier: Nice catch. Thank you!

tobiasjj (approved):
Nice addition! LGTM. 🚴

PR #744, opened 2025-05-23
kymo: fix bug in `Kymo.plot_with_channels()`

JoepVanlier:
**Why this PR?**
It seems that we did not consider the start time of the `kymo` versus the `timestamp` of the first data point when implementing `Kymo.plot_with_channels()`.

Considering that `line_timestamp_ranges()` typically does a lot more than we need (we only need the very first timestamp), I also added a small method to shortcut this calculation. The difference in performance is about 9x for a typical kymograph.

image

```python
f = lk.File("kymo.h5")
kymo = f.kymos["test_kymo"]

kymo.plot_with_channels(
    [f["Photon count"]["Red"].downsampled_by(100),
     f["Photon count"]["Red"].downsampled_over(kymo.line_timestamp_ranges())],
    adjustment=lk.ColorAdjustment([0, 0, 0], [95, 100, 100], "percentile"),
    titles=("Kymograph", "Downsampled 100", "Downsampled kymo")
)

plt.gcf().axes[0].get_images()[0].set_interpolation('none')
for i, ax in enumerate(plt.gcf().axes):
    ax.axvline(13.61, color="w", linewidth=2, alpha=0.5)
    ax.axvline(13.61, color="k", linewidth=1, alpha=0.5)

    ax.axvline(49.6, color="w", linewidth=2, alpha=0.5)
    ax.axvline(49.6, color="k", linewidth=1, alpha=0.5)
    ax.set_aspect("auto")

plt.xlim([49, 50])
plt.tight_layout()
```

Without:
image

With:
image

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## v1.6.3 | t.b.d.
+
+#### Bug fixes
+
+* Fixed a bug in `Kymo.plot_with_channels()` that would not correctly align the timeline channels in some cases. Prior to the fix, it assumed that slicing the channel with the kymograph start and stop time results in the origin being at time zero, but this may not be the case if the channel is sparsely sampled and the first point is much later (such as is the case for diagnostic channels).
+
 ## v1.6.2 | 2025-05-06
 
 #### Bug fixes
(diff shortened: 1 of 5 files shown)


tiagobonetti (approved):
Nice fix and well tested.
I used this work as a reference to fix a similar bug in Lakeview.
I left one small suggestion.

`honk`

PR #745, opened 2025-05-23
force calibration: automated fitting ranges

JoepVanlier:
**Why this PR?**
The model we typically fit for lateral calibration consists of two parts. The physical power spectrum (which is based on the physics of the bead in the trap) and a parasitic filtering effect which is a model of the diode sensor. This latter effect is typically at high frequencies (>8 kHz). The power spectrum of these two is quite similar, which leads to fitting problems at high powers where the corner frequency of the physical power spectrum approaches that of the diode. For this reason, we characterize the diodes in the factory.

There are two issues with this characterization:
- Diode characterization at very low powers is not as accurate as at high powers however. By restricting the fitting range to exclude the diode effect when the corner frequency is very low reduces the sensitivity to such biases considerably while not hampering the precision of the fit considerably.
- When working at very low powers, noise floors can be observed at high frequencies. Considering that we do not model this noise floor, this noise floor biases calibration when fitting part of the power spectrum where it constitutes a substantial part of the signal. Fitting the noise floor can make fitting brittle when this inflection is not clearly visible and/or the diode is free. Rather than estimating the noise online, we omit the part of the spectrum that may be hampered by it.

While the fitting ranges could be set manually, it may be difficult to identify these two failure modes, hence it makes sense to offer the option to do so automatically. One thing to keep in mind is that when we are using automated fitting ranges, we should not fit the diode. The reason is that we clip off the part of the spectrum that contains the information on the diode's correct parameters and therefore fitting the diode parameters becomes very unreliable.

I recommend reviewing commit by commit. Not that much changed, but it looks like more than it is because of the docs / test changes.

Docs build [here](https://lumicks-pylake.readthedocs.io/en/iterate_fc/tutorial/force_calibration/diode_model.html).

image

image

### DIFF
--- a/docs/tutorial/force_calibration/diode_model.rst
+++ b/docs/tutorial/force_calibration/diode_model.rst
@@ -73,7 +73,7 @@ We can see that this updated the fixed diode parameters.
 
     Each sensor has its own diode characteristic.
     If you are calibrating multiple traps with pre-calibrated diodes, you will need to provide
-    the correct diode parameters for each trap.
+    the correct diode parameters for each trap. Each sensor has their own diode calibration values!
 
 We can calibrate with these parameters directly by unpacking this dictionary into the :meth:`~lumicks.pylake.calibrate_force` function::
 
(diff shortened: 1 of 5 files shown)


rjmoerland (approved):
🍌

PR #757, opened 2025-10-23
support bluelake items with adaptive fitting ranges

JoepVanlier:
**Why this PR?**
This PR ensures that we can read calibration items coming from Bluelake that were performed with adaptive fitting ranges enabled.

It also ensures that the properties are uniform between the two.

The following properties exist after this PR:
- `item.fit_range` will contain the fit range that was used in the final fit.
- `item.initial_fit_range` will use the first estimate that the user provided through the UI.
- `item.corner_frequency_factor` will contain the algorithm parameter used in optimizing the frequency range that is incorporated in the fit.

When using `item.recalibrate_with()`, the initial fit range in conjunction with the `corner_frequency_factor` is used, such that the full calibration procedure carried out on the Bluelake side is effectively reproduced.

The reason why this is important is because when a user overrides the model (let's say they want to try and recalibrate some data), changing the model here would result in a different corner frequency (which would warrant a different fitting range).

Docs build here: https://lumicks-pylake.readthedocs.io/en/joep-diode/tutorial/force_calibration/diode_model.html
For more info on what this adaptive stuff actually does, see [this](https://github.com/lumicks/pylake/pull/745), but it shouldn't be necessary to review this PR.


lumicks/pylake/force_calibration/calibration_item.py
```diff
@@ -299,8 +304,19 @@ def _excluded_ranges(self):
         ]
 
     @property
-    def _fit_range(self):
-        """Return the spectral fit range used for the calibration"""
+    def _fit_range(self) -> tuple[float, float] | None:
+        """Return the actual spectral fit range used for the calibration"""
+        if "Adaptive fit range (min.) (Hz)" in self.data:
```
agom: Do you need to check also for `Adaptive fit range (max.) (Hz)` or you can assume it's always there if `min` is present?

lumicks/pylake/force_calibration/calibration_results.py
```diff
@@ -75,10 +75,21 @@ def __repr__(self):
...
     @property
-    def _fit_range(self):
+    def _fit_range(self) -> tuple[float, float]:
         """Fitted range"""
         return self.ps_data._fit_range
 
+    @property
+    def _initial_fit_range(self) -> tuple[float, float] | None:
+        """Initial fit range"""
+        if "Initial fit range (min.)" not in self.params:
```
agom: Same as above.

agom (approved):
Look good!

alessiamarcolini (approved):

lumicks/pylake/force_calibration/calibration_item.py
JoepVanlier: Currently, I'm using this purely as a differentiator between the two cases and would indeed expect that you would always have both. I mean, if we check both, then I would sooner raise if only one of them is present, because that condition doesn't make sense.

lumicks/pylake/force_calibration/calibration_results.py
JoepVanlier: Updated to require both and raise if only one exists.

